PR #4, opened 2021-11-18
OCR monitoring

topliceanu:


core/services/monitoring/client.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+	"sync/atomic"
+
+	"github.com/smartcontractkit/libocr/commontypes"
+	"github.com/smartcontractkit/wsrpc"
+)
+
+type Client interface {
+	commontypes.MonitoringEndpoint
```
krebernisak, line 12: The interface here should use the `TelemetryRequest` as input, not the `commontypes.MonitoringEndpoint` which is just OCRs implementation detail. OCR service should format and send a `TelemetryRequest` via the telemetry client.

That way other system services could also require the telemetry client as a dependency and use it to send their telemetry in the future.

core/services/monitoring/client.go
```diff
@@ -0,0 +1,101 @@
...
+
+	Start()
```
krebernisak, line 14: Please use either Start/Stop or Open/Close for lifecycle functions.

I suggest we use:

- Open/Close for clients: a connection you either open yourself or one gets an already connected client and just uses it
- Start/Stop for services: components/services that are started/stopped by the system (as one of root subsystems) or by other services that manage multiple subservices.

So in this case I'd suggest we design a TelemetryService that we start on system start. That service onStart opens a TelemetryClient connection and exposes it as a port (interface others can inject and use directly).

core/services/ocr2.go
```diff
@@ -159,6 +179,9 @@ func (o *OCR2) Start() {
 	// start contract subscription
 	go o.contract.Start()
 
+	// start client for the monitoring endpoint which pushes telemetry to OTI
+	go o.monitoringEndpoint.Start()
```
krebernisak, line 183: Then here we don't start a new connection, as this will create N connections for N OCR jobs.

We want one durable connection between the Relay & remote OTI service. Then every OCR job (service that runs the job) uses that one connection to report its telemetry.

So OCR service should just require a TelemetryClient as a dependency. As this is a dependency, not a new client we construct ourselves we expect it's already connected by some other system service (telemetry/monitoring).

Then the OCR service directly uses the telemetry port (client) to do things (send telemetry).

topliceanu:
@krebernisak thanks a lot for the review. It offered much needed clarity. I refactored the package. Let me know what you think of the structure and separation of responsibilities. If you're OK with the approach I'll add tests.

core/services/telemetry/client.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"sync/atomic"
+
+	"github.com/smartcontractkit/chainlink/core/logger"
+	"github.com/smartcontractkit/wsrpc"
+)
+
+type Client struct {
+	ctx     context.Context
+	backend TelemetryClient
```
krebernisak, line 13: Now we have two telemetry clients, one to our internal relay telemetry service and the other to the remote telemetry RPC service.

Unfortunate naming collision here, so just thinking how to be more explicit about this. But ok for now.

core/services/telemetry/service.go
```diff
@@ -0,0 +1,56 @@
...
+		s.serverURL.String(),
+		wsrpc.WithTransportCreds(
+			s.clientPrivateKey,
+			s.serverPublicKey,
+		),
+	)
+	if err != nil {
+		return Client{}, err
+	}
+	client := NewClient(s.ctx, NewTelemetryClient(conn), 100, s.log)
```
krebernisak, line 50: This line is slightly cryptic as I don't expect this will actually spin up a new goroutine and start listening to the input channel. What I expect is just a constructor making me a data type.

Consider how something like this is a lot more explicit:

```go
bufferCapacity := 100
client := NewClient(NewTelemetryClient(conn), bufferCapacity, s.log)
go client.Run/Open(ctx)
return client, nil
```

core/services.go
```diff
@@ -70,7 +97,12 @@ func (s *Services) Start(job models.Job) error {
 			}
 		}
 
-		ocr2, err := services.NewOCR2(job, s.db, s.webhook, s.cfg, s.keys, s.blockchain, s.peerWrapper)
+		telemetryClient, err := s.telemetryService.Start()
```
krebernisak, line 100: This will still start a new telemetry per OCR job, as this service is started per OCR job.

Better to extract it here than where it was before, but we'd need another layer, a `relay.App` that is started just once, and where all root level services are started once (telemetry included).

We'll add that, no need to worry now about it.

Just as a suggestion add a TODO comment:

```go
// TODO: telemetry service should be started once (at app start), and client injected into OCR service to be used by every job. Currently, here we start a new telemetry service for every OCR job (N background connections).
telemetryClient, err := s.telemetryService.Start()
```

krebernisak (commented):
Looks great @topliceanu!

Almost there, left a couple of comments. Still some problems (not multiplexing over one connection), but not really because of this PR but underlying Relay SDK structure we need to update.

topliceanu:
Hi @krebernisak

topliceanu:
Hi @krebernisak. Thanks again for the review

Re. https://github.com/smartcontractkit/chainlink-relay/pull/4#discussion_r753185268 Yes I agree. I looked into getting go-protobuf to generate private types but that’s not supported afaictl. So, I moved all the protobuf generated code in a separate `generated` package. Hopefully this makes it clearer. Alternatively I can have protobug generate in an `internal` package. 

Re. https://github.com/smartcontractkit/chainlink-relay/pull/4#discussion_r753192160 and https://github.com/smartcontractkit/chainlink-relay/pull/4#discussion_r753197161 
The service is the one that “maintains” the WSRPC connection. When you “Start” the service you get a Client which you can use in multiple instances of OCR. This does multiplex producers over the same WS connection, but the ordering of the messages is left to the golang goroutine scheduler. Client instance is thread-safe insomuch as the wsrpc.Client is threadsafe.

Re. the relay repo, I understand that for every instance of the ocr protocol we will instantiate a core#Services structure. This should receive the shared telemetry#Client instance instead of telemetry#Service - which wraps the connection. The question is then, where do I instantiate the Service? I can’t find an “entrypoint”, but maybe there doesn’t need to be one atm.

PR #5, opened 2021-11-19
libocr update + dataSource fix

aalu1418:
Updating the libocr version

DataSource changes:
* using `uber/atomic` (safer library then the built in `atomic` module)
* implemented `ctx.Done()` handling
  * noticed once it happen once, but `Observe` can potentially freeze if data is not returned.
  * each time there is a wait for a channel response, added a `select` statement to also allow `ctx.Done()` to stop the function

### DIFF
--- a/core/services/data_source.go
+++ b/core/services/data_source.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"errors"
 	"math/big"
-	"sync/atomic"
 
 	"github.com/smartcontractkit/chainlink-relay/core/server/webhook"
 	"github.com/smartcontractkit/chainlink/core/logger"
 	"github.com/smartcontractkit/libocr/offchainreporting2/reportingplugin/median"
+	"go.uber.org/atomic"
 )
 
 var (
@@ -29,7 +29,7 @@ type dataSourceState struct {
 	RunData *chan *big.Int
 	Log     *logger.Logger
 	Prices  map[string]*big.Int
-	locked  uint32
+	locked  *atomic.Bool
 	Done    chan struct{}
 }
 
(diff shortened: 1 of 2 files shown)


core/services/data_source.go
```diff
@@ -57,26 +59,34 @@ func NewDataSources(id string, trigger *webhook.Trigger, runChannel *chan *big.I
...
+	case dss.Prices[priceFeedParam] = <-*dss.RunData: // use first value
+	case <-ctx.Done():
+		return fmt.Errorf("[%s] Observe (job run) cancelled", dss.ID)
+	}
+	select {
+	case dss.Prices[juelsToXParam] = <-*dss.RunData: // use second value
+	case <-ctx.Done():
+		return fmt.Errorf("[%s] Observe (job run) cancelled", dss.ID)
+	}
+
```
archseer, line 87: Sending back two separate sequential values and treating them differently feels kind of weird. What if they arrive out of order? If the same routine is sending them both back, just use a `struct x { priceFeed, juelsToUnit }`

archseer: (Also, how does `core` handle this?)

aalu1418: this probably not be relevant anymore because of switching over to core but i'll answer them 😄 
* the order should be fine since there's a part that deconstructs a struct and returns two separate values: https://github.com/smartcontractkit/chainlink-relay/blob/7839e322d0e1e6022f266c6b3b9425556a97404f/core/services/ocr2.go#L186 (and yes, if we continue to use this i should just return a struct rather than values twice)
* it's just mocked in core right now 😆

PR #7, opened 2021-12-02
refactor ops core for latest architecture

aalu1418:
- [x] Remove relay functionality, replaced with core nodes
  - [x] capability to build from local files or pull from ECR
- [x] Fetch public keys from core node
  - [x] need keys finalized in core 
  - [x] potentially need update to integrations framework for new keys?
- [x] New job spec sent to nodes
  - [x] need job spec  finalized in core
  - [x] potentially need to update the integrations framework for new job type
- [x] Final clean up (remove old code, etc)


krebernisak (approved):
Let's merge what we have here 🚀

PR #12, opened 2022-01-25
feature (ops): configuring migrating nodes to new version

aalu1418:


krebernisak (approved):
A quick sanity check! Looks good.

We're only missing some docs about those options, so we know what's available and have a better understanding of the local env and how to re/structure it.

aalu1418:
👍🏼 good point, I made an issue to revisit #14

PR #25, opened 2022-02-04
Introduce JuelsPerFeeCoing, AggregatorRoundID and TxResultsSucceeded/Failed metrics

topliceanu:


.github/workflows/monitoring.yml
```diff
@@ -26,5 +26,5 @@ jobs:
           go-version: 1.17
       - name: Test
         run: go test -count=1 -race -v ./pkg/monitoring/...
-      - name: Bench
-        run: go test -count=1 -bench=. ./pkg/monitoring/...
+      #- name: Bench
```
pedrovsg, line 29: is this on propuse?

pedrovsg (commented):
can't understand where we check if is failed or not

.github/workflows/monitoring.yml
topliceanu: Yeah, they crash on CI
I'll fix this in a separate PR!

pedrovsg (approved):

PR #48, opened 2022-05-13
pkg/monitoring: copy single hex Encode func instead of importing go-ethereum

jmank88:


topliceanu (dismissed):
That's insane! Thanks for fixing this!

topliceanu (approved):

PR #49, opened 2022-05-13
pkg/logger: add basic zap test and prod logger

jmank88:
https://app.shortcut.com/chainlinklabs/story/40148/chainlink-relay-create-logger-impl-for-testing-and-scripts


topliceanu (approved):

pkg/logger/logger.go
```diff
@@ -0,0 +1,87 @@
...
+	Panicw(msg string, keysAndValues ...interface{})
+	Fatalw(msg string, keysAndValues ...interface{})
+
+	Sync() error
+}
+
+// New returns a new Logger.
+func New() (Logger, error) {
+	var lvl zapcore.Level
+	err := lvl.Set(os.Getenv("LOG_LEVEL"))
```
topliceanu, line 45: I would prefer if log level was passed as a parameter to the constructor. 
I don't like env vars peppered around the code and the dependencies!
_my 2c_

pkg/logger/logger.go
```diff
@@ -0,0 +1,87 @@
...
+	Fatalf(format string, values ...interface{})
+
+	Debugw(msg string, keysAndValues ...interface{})
+	Infow(msg string, keysAndValues ...interface{})
+	Warnw(msg string, keysAndValues ...interface{})
+	Errorw(msg string, keysAndValues ...interface{})
+	Panicw(msg string, keysAndValues ...interface{})
+	Fatalw(msg string, keysAndValues ...interface{})
+
+	Sync() error
```
topliceanu, line 39: Can you add some documentation about when to use `Sync()` and what it does?

pkg/logger/logger.go
jmank88: Will do! Almost did it before but was trying to capture the API exactly at first.

pkg/logger/logger.go
jmank88: Is pointing to core `logger.Logger` and `zap.SugaredLogger` from the type level comment sufficient?

pkg/logger/logger.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+// Logger is a minimal subset of smartcontractkit/chainlink/core/logger.Logger implemented by go.uber.org/zap.SugaredLogger
+type Logger interface {
+	With(args ...interface{}) Logger
```
jmank88, line 14: This self reference is a problem. Although core/logger.Logger also contains a `With(args ...interface) Logger`, the method signatures don't match. I don't think there is any way to express this with generics, but a transitional helper func might be possible, like `logger.With(Logger, args ...interface{})` :thinking:

jmank88: Added `func With`, which is not pretty, but fixes the problem. Open to better ideas :shrug:

pkg/logger/logger.go
```diff
@@ -0,0 +1,114 @@
...
+	*zap.SugaredLogger
+}
+
+func (l *logger) with(args ...interface{}) Logger {
+	return &logger{l.SugaredLogger.With(args...)}
+}
+
+var typeOfLogger = reflect.TypeOf((Logger)(nil))
+
+func With(l Logger, args ...interface{}) Logger {
```
jmank88, line 88: - [ ] doc
- [ ] test

...or replace with some alternative?

pkg/logger/logger.go
topliceanu: Yeah!

pkg/logger/logger.go
topliceanu: I know this pain. I've had to do this https://github.com/smartcontractkit/chainlink-terra/blob/develop/cmd/monitoring/main.go#L74-L82 
I think a separate function `With(args ...interface) Logger` is fine!

pkg/logger/logger.go
topliceanu: I think this works. I pick doc, because YOLO!

topliceanu (approved):

topliceanu:
`core/logger` is the only dependency `chainlink-relay/pkg/monitoring` has on `core`. I wanted to remove it and replace it with zap directly for months now! I appreciate you adding this!

pkg/logger/logger.go
jmank88: Done, and it needed a bit of fixing too.

topliceanu (approved):
lgtm

PR #65, opened 2022-07-25
Add counters for failed/succeded transactions broken down by sender oracle

topliceanu:


topliceanu:
This doesn't work and is not needed anymore!

PR #70, opened 2022-09-13
Add Checks For External Products So We Know We Need To Update

tateexon:
Add workflow to check products that rely on the relay for compatibility so we can in the future more quickly file tickets for those products to be updated.


aalu1418 (changes requested):
could we also add `chainlink-terra` to this? it's not actively maintained but it is affected if we have an interface change

aalu1418 (approved):
✅

PR #79, opened 2023-02-09
Add mercury interfaces and implementation

samsondav:
- Add Mercury


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/confluentinc/confluent-kafka-go v1.8.2
+	github.com/ethereum/go-ethereum v1.10.11
```
jmank88, line 7: Why is ethereum leaking in here?

jmank88 (changes requested):

pkg/plugins/mercury/transmitter.go
```diff
@@ -0,0 +1,11 @@
+package mercury
+
+import (
+	"context"
+
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2/types"
+)
+
+type Transmitter interface {
+	Transmit(ctx context.Context, reportCtx ocrtypes.ReportContext, report ocrtypes.Report, signatures []ocrtypes.AttributedOnchainSignature) error
```
jmank88, line 10: Why does it need a custom transmitter? Isn't this the same signature as the `ContractTransmitter`?

go.mod
samsondav: It's used for ABI encoding to create the report

pkg/plugins/mercury/transmitter.go
samsondav: Correct, fixed

go.mod
jmank88: This repo and these interfaces are meant to remain chain-agnostic. If this version is still EVM specific then it is not ready to move out of the core repo yet.

jmank88: If the dependency is limited to the _implementations_, that is fine, but those pieces aren't supposed to be in this repo.

.tool-versions
```diff
@@ -1 +1 @@
-golang 1.19.2
+golang 1.19.4
```
jmank88, line 1:
```suggestion
golang 1.19.5
```

go.mod
samsondav: Mercury works a bit differently, since this actually _is_ chain agnostic on the transmitter side. The eventual implementation will encode multiple different reports for different types of chain.

jmank88: Median is not an example to be followed in this case. Chain specific implementations of the various components need to be in chain specific repos (or core in the case of EVM).

samsondav: Moved it into core

aalu1418: I would second what jordan said - ideally we don't bring geth into this repo. iirc, in the past it has caused a lot of dependency bloat and also some issues with keeping it in sync across testing/core/integrations repos 😬 

I think in the past we've duplicated some simple logic in this repo - and it seems like the only thing used right now is the hash type which is a `[length]byte` implementation + a utility to convert `[]byte` to `common.Hash`

samsondav: Go-ethereum is now completely expunged

pkg/plugins/mercury/mercury.go
```diff
@@ -0,0 +1,516 @@
...
+	}
+
+	rp.logger.Debug("ShouldAcceptFinalizedReport() = result", commontypes.LogFields{
+		// "contractEpochRound":       contractEpochRound,
+		// "reportEpochRound":         reportEpochRound,
+		"latestAcceptedEpochRound": rp.latestAcceptedEpochRound,
+		// "result":                   result,
+	})
+
+	// rp.latestAcceptedEpochRound = reportEpochRound
```
krehermann, line 504: remove?

pkg/plugins/mercury/mercury.go
```diff
@@ -0,0 +1,516 @@
...
+}
+
+func (rp *reportingPlugin) ShouldAcceptFinalizedReport(ctx context.Context, repts types.ReportTimestamp, report types.Report) (bool, error) {
+	//     Set maxFinalizedBlocknumber to the endBlockNumber from the report.
+	// return true
+	// TODO
+
+	if !(len(report) <= rp.maxReportLength) {
+		rp.logger.Warn("report violates MaxReportLength limit set by ReportCodec", commontypes.LogFields{
+			// "reportEpochRound": reportEpochRound,
```
krehermann, line 490: remove?

jmank88 (approved):

pkg/plugins/mercury/value.go
```diff
@@ -0,0 +1,41 @@
...
+// Bounds on an int192
+const byteWidth = 24
+const bitWidth = byteWidth * 8
+
+var one *big.Int = big.NewInt(1)
+
+// 2**191-1
+func MaxValue() *big.Int {
+	result := MinValue()
+	result.Abs(result)
```
reductionista, line 18: Why not just `result.Lsh(one, bitWidth-1)` instead of `Abs(Neg(...))` of that?

jmank88 (dismissed):

pkg/plugins/mercury/value.go
samsondav: I copied this from libocr :D

samsondav: Its not even used. Removed.

jmank88 (dismissed):

pkg/plugins/mercury/mercury.go
```diff
@@ -0,0 +1,516 @@
...
+
+	obs, err := rp.dataSource.Observe(ctx)
+	if err != nil {
+		return nil, errors.Errorf("DataSource.Observe returned an error: %s", err)
+	}
+
+	benchmarkPrice, err2 := EncodeValue(obs.BenchmarkPrice)
+	err = multierr.Append(err, err2)
+	bid, err := EncodeValue(obs.Bid)
+	err = multierr.Append(err, err2)
```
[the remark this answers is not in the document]
samsondav: Good catch, it was wrong before, but I think it's correct now. Mind checking again?

pkg/plugins/mercury/mercury_observation.proto
```diff
@@ -0,0 +1,15 @@
+syntax="proto3";
+
+package mercury;
+option go_package = ".;mercury";
+
+// TODO: what about different report formats for different clients?
+message MercuryObservationProto {
+    uint32 timestamp = 1;
```
krehermann, line 8: did you consider using the  built in proto3 timestamp type? if so, why not choose it? 

using uint32 risks weird timezone issues. 

in case you haven't seen it, there is a nice golang package for converting back and forth btw time.Time and protobuf.Timestamp https://pkg.go.dev/google.golang.org/protobuf/types/known/timestamppb

.github/workflows/pkg.yml
```diff
@@ -7,11 +7,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.19
```
jmank88, line 12:
```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/pkg.yml
```diff
@@ -36,7 +39,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: 1.18
+          go-version: 1.19
```
jmank88, line 42:
```suggestion
          go-version-file: 'go.mod'
```

jmank88: I noticed that below here we are running `build` and `test` with verbose logging - might get some speedup from dropping the `-v`.

pkg/plugins/mercury/mercury_observation.proto
samsondav: I did not consider it, mostly because I this is copied from the median plugin. This is always assumed to be a unix timestamp which is always UTC/GMT and does not have timezone issues.

jmank88 (approved):

pkg/plugins/mercury/mercury.go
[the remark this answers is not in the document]
krehermann: looks good now.

PR #80, opened 2023-02-09
pkg/loop: loop plugin grpc services

jmank88:
This PR adds GRPC protobufs and client/service implementations for `ReportingPlugin` and the common component service methods, as well as a client `ChainPluginService` to launch and keep a `ChainRelayerPlugin` alive.

Supports the core follow-up: https://github.com/smartcontractkit/chainlink/pull/8517

https://smartcontract-it.atlassian.net/browse/BCF-2052

### Exported API
image
This grew a lot, but will be trimmed down in the follow-up https://smartcontract-it.atlassian.net/browse/BCF-2112

### TODO
- [x] protocol buffers
- [x] grpc clients/servers
- [x] Plugin _service_, with keep-alive
- [x] strict length/overflow checks when converting to/from protobufs
- [x] wrap errors
- [x] handle `Serve()` errors

Graceful shutdown of GRPC servers and `context.TODO()`s to be addressed in follow-up https://smartcontract-it.atlassian.net/browse/BCF-2112


jmank88:
@krebernisak This still needs work before formal review/merge, but what do you think about the location, `package pkg/loop`?
And is there any reason to consider a separate _module_ instead? :thinking:

pkg/types/types.go
```diff
@@ -33,7 +33,6 @@ type RelayArgs struct {
 }
 
 type Relayer interface {
-	Service
```
jmank88, line 36: None of the Relayer implementations actually use the Service methods for anything interesting, so this is just requiring unnecessary no-op boilerplate.

aalu1418 (commented):
this looks really great!

just had some high-level questions for now :)

pkg/loop/config.go
```diff
@@ -0,0 +1,224 @@
...
+	if l := len(reply.ConfigDigest); l != 32 {
+		err = fmt.Errorf("wrong digest length: expect 32 but got %d", l) //TODO necessary? then add more places?
+		return
+	}
+	copy(digest[:], reply.ConfigDigest)
+	return
+}
+
+func (o *offchainConfigDigesterClient) ConfigDigestPrefix() libocr.ConfigDigestPrefix {
+	reply, err := o.grpc.ConfigDigestPrefix(context.TODO(), &pb.ConfigDigestPrefixRequest{})
```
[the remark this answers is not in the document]
jmank88: What did you have in mind? I don't think there are any real workarounds unfortunately. Context is not such a big loss, but without anywhere to send the errors we either have to panic or return a phony value. I'm hoping to follow up with https://smartcontract-it.atlassian.net/browse/BCF-2073 right away (or even in parallel if we want to hold on merging this :shrug:).

pkg/loop/config.go
```diff
@@ -0,0 +1,224 @@
...
+	return &pb.ConfigDigestPrefixReply{ConfigDigestPrefix: uint32(p)}, nil
+}
+
+var _ libocr.ContractConfigTracker = (*contractConfigTrackerClient)(nil)
+
+type contractConfigTrackerClient struct {
+	grpc pb.ContractConfigTrackerClient
+}
+
+func (c *contractConfigTrackerClient) Notify() <-chan struct{} { return nil }
```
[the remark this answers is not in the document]
jmank88: The libocr doc says this is completely optional, so I think it's NBD to leave out. But regardless, we can use GRPC streams to implement it if we need to.

pkg/loop/config.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	cfg.F = uint8(reply.ContractConfig.F)
+	cfg.OnchainConfig = reply.ContractConfig.OnchainConfig
+	cfg.OffchainConfigVersion = reply.ContractConfig.OffchainConfigVersion
+	cfg.OffchainConfig = reply.ContractConfig.OffchainConfig
+
+	return cfg, nil
+}
+
+func (c *contractConfigTrackerClient) LatestBlockHeight(ctx context.Context) (blockHeight uint64, err error) {
```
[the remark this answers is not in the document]
jmank88: Yeah, client and server, one on each side, but it's bi-directional so based on where the impl lives, not just core vs plugin.
The clients implement the Go interface and translate the calls to GRPC and back. Servers implement the GRPC interface and translate the calls from GRPC to regular Go for an internal Go impl.

pkg/loop/internal/pb/loop.proto
```diff
@@ -0,0 +1,246 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
jmank88: It's not quite 1-1 with Go types, but a generator could be written in theory. Just have to account for translating the types in to more general ones, e.g. protobuf only has 32/64 bit ints, so smaller go types have to convert back and forth.

pkg/loop/config.go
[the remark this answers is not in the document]
aalu1418: i was just thinking you could handle a context with a `select` and return an error if context was cancelled early

and then for panics - i vaguely recall panic handling to be a thing but not sure how it technically works? if we can "catch" a panic - then we can return the panic error upstream

pkg/loop/internal/pb/loop.proto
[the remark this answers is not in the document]
jmank88: Might not be worth it though, since this should be a pretty stable interface.

pkg/loop/config.go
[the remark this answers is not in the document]
aalu1418: makes sense! a small improvement might be to just add a comment that this belongs on the requester vs responder sides - it's a tad hard to tell without having some previous context, but I could sort of figure it out based on the func IO

pkg/loop/internal/pb/loop.proto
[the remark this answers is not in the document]
aalu1418: yeah more of something to think about for the future if we have various providers, manually writing out protobuf schemas feels tedious -  and then the usual: a test case fails cuz things didn't match, etc. so maybe this is more about a CI check to make sure schemas match functions 🤷🏼

pkg/loop/config.go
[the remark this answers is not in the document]
jmank88: > i was just thinking you could handle a context with a select and return an error if context was cancelled early

Oh yeah that would be possible, although I'm not sure that those cases actually block, since the client side grpc caller could just move on after timeout anyways.

Yeah we could try to panic and recover, but the stack goes through libocr in complicated ways (e.g. `ReportPlugin`s are highly concurrent), so it may not be possible to do without refactoring those internals is a way that is even more disruptive than regular err/ctx.

pkg/loop/Makefile
```diff
@@ -0,0 +1,5 @@
+.PHONY: proto
```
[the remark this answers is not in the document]
jmank88: Yeah maybe. Although, what about a go generate statement? I think that is more typical in this kind of case anyways :thinking:

pkg/loop/config.go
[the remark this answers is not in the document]
jmank88: Yeah I can add more comments. Fortunately all this boilerplate is hidden is hidden and the exported API surface is minimal.

pkg/loop/internal/pb/loop.proto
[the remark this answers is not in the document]
jmank88: I'd have to walk through what an addition would look like, but I _think_ that compile time errors would mostly guide what needs to be filled in, step by step... :thinking:

pkg/loop/service.go
```diff
@@ -0,0 +1,154 @@
...
+var ErrPluginUnavailable = errors.New("plugin unavailable")
+
+const keepAliveTickDuration = 5 * time.Second //TODO from config
+
+var _ types.Service = (*ChainPluginService)(nil)
+var _ ChainPlugin = (*ChainPluginService)(nil)
+
+// ChainPluginService is a [types.Service] that implements [ChainPlugin], and maintains an internal client instance by
+// launching and re-launching as necessary using the provided [*plugin.ClientConfig].
+type ChainPluginService struct {
```
jmank88, line 25: This could live in core, but is useful as a testing tool for this repo and the chain specific ones.

hans-kessock: This is one of the reasons I *think* (but am not sure) we should have a single 'extensibility/plugins' repo where this is part of a 'chainplugin sdk', there will surely be other plugin types there in the future as well  - core can access it without needing relayers, relayers can access without needing core for testing purposes/tooling.

If we hate this idea, let's revisit post POC

pkg/loop/plugin.go
```diff
@@ -0,0 +1,168 @@
...
+var _ pb.KeystoreServer = (*keystoreServer)(nil)
+
+type keystoreServer struct {
+	pb.UnimplementedKeystoreServer
+
+	impl Keystore
+}
+
+func (k *keystoreServer) Sign(ctx context.Context, request *pb.SignRequest) (*pb.SignReply, error) {
+	signed, err := k.impl.Sign(ctx, request.Account, request.Data)
```
hans-kessock, line 163: Is the keystoreServer intended to be implemented in the client (plugin host), e.g. the core node?  Or, is this something that would be called by OCR (for example) and is chain specific?

hans-kessock (commented):
Really valuable learning tool for OCR->Chain!  Can't wait to see it implementing for Solana

pkg/loop/plugin.go
jmank88: The `keystoreServer` runs on the host side, and packages up the existing two-step keystore interface of `Get(id)`ing a `Key` and then calling `Sign([]byte)` that the tx manager currently uses. Solana happens to be quite generic, but our keystore APIs are technically still chain-specific, so it's still an open question whether this pure byte signing API could work for every chain. The generic tx manager work is addressing this more directly.
```go
type SolanaKeystore interface {
    Get(id string) (Key, error)
}
type Key interface {
    Sign(msg []byte) ([]byte, error)
}
```

hans-kessock: That's because the keystore does more than store/get keys, it handles signing as well, correct?  Making it chain specific (or, at the very least, chains often do this differently)

pkg/loop/plugin_median.go
```diff
@@ -0,0 +1,484 @@
...
+
+type ErrorLog interface {
+	SaveError(ctx context.Context, msg string) error
+}
+
+func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
+	return plugin.HandshakeConfig{
+		ProtocolVersion:  0,
+		MagicCookieKey:   "magic-key-median-todo",
+		MagicCookieValue: "magic-value-median-todo",
```
jmank88, line 36: Will follow up on these soon here https://smartcontract-it.atlassian.net/browse/BCF-2198

pkg/loop/plugin_relayer.go
```diff
@@ -0,0 +1,212 @@
...
+
+type PluginRelayer interface {
+	NewRelayer(ctx context.Context, config string, keystore Keystore) (Relayer, error)
+}
+
+func PluginRelayerHandshakeConfig() plugin.HandshakeConfig {
+	return plugin.HandshakeConfig{
+		ProtocolVersion:  0,
+		MagicCookieKey:   "magic-key-relayer-todo",
+		MagicCookieValue: "magic-value-relayer-todo",
```
jmank88, line 28: Will follow up on these soon here https://smartcontract-it.atlassian.net/browse/BCF-2198

pkg/loop/plugin_median.go
```diff
@@ -0,0 +1,484 @@
...
+	id := reply.ReportingPluginFactoryID
+	cc, err := m.broker.Dial(id)
+	if err != nil {
+		dsSrv.Stop()
+		juelsSrv.Stop()
+		providerSrv.Stop()
+		errorLogSrv.Stop()
+		return nil, ErrConnDial{Name: "MedianPluginFactory", ID: id, Err: err}
+	}
+	//TODO client should close everything
```
jmank88, line 148: Fixed in upcoming https://smartcontract-it.atlassian.net/browse/BCF-2112

krehermann (commented):
test

pkg/loop/config.go
```diff
@@ -0,0 +1,218 @@
...
+var _ libocr.OffchainConfigDigester = (*offchainConfigDigesterClient)(nil)
+
+type offchainConfigDigesterClient struct {
+	grpc pb.OffchainConfigDigesterClient
+}
+
+func (o *offchainConfigDigesterClient) ConfigDigest(config libocr.ContractConfig) (digest libocr.ConfigDigest, err error) {
+	pbCfg := pbContractConfig(config)
+	var reply *pb.ConfigDigestReply
+	reply, err = o.grpc.ConfigDigest(context.TODO(), &pb.ConfigDigestRequest{
```
aalu1418, line 47: are we planning to propagate the context into libocr? (if yes, a ticket link would be nice)

pkg/loop/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+type staticConfigProvider struct{}
+
+func (s staticConfigProvider) Start(ctx context.Context) error { return nil }
+
+func (s staticConfigProvider) Close() error { return nil }
+
+func (s staticConfigProvider) Ready() error { panic("unimplemented") }
+
+func (s staticConfigProvider) Name() string { panic("implement me") }
```
aalu1418, line 20: nit: could we use the same panic message for unimplemented methods?

pkg/loop/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return nil
+}
+
+func (s staticContractTransmitter) LatestConfigDigestAndEpoch(ctx context.Context) (libocr.ConfigDigest, uint32, error) {
+	return configDigest, epoch, nil
+}
+
+func (s staticContractTransmitter) FromAccount() libocr.Account {
+	return account
```
aalu1418, line 84: just curious, where are these defined? or are they just placeholders right now (I feel like i'm missing something)

pkg/loop/errors.go
aalu1418, line 1: interesting that we only have a few predefined error types. I wonder if libocr are also a standard set of errors too?

pkg/loop/config_test.go
aalu1418: oh nevermind, found them: pkg/loop/plugin_relayer_test.go
github minimized it so it wasn't searchable

pkg/loop/plugin_relayer_test.go
```diff
@@ -0,0 +1,484 @@
...
+	limit            = 42
+	lookbackDuration = time.Minute + 4*time.Second
+	max              = 101
+	n                = 12
+	offset           = 11
+	round            = uint8(74)
+	to               = "0x1234"
+)
+
+var (
```
aalu1418, line 46: nit: might be nice to have these const + var in a separate file like constants_test or even in helpers_test, just thinking these are used in more than just the plugin relayer tests

(and I struggled to find them earlier haha)

aalu1418 (commented):
looking really good! just a few Qs and nits :)

krehermann (commented):
only reading the proto for now

pkg/types/chain.go
```diff
@@ -0,0 +1,38 @@
...
+	ID      string
+	Enabled bool
+	Config  string // TOML
+}
+
+type NodeStatus struct {
+	ChainID string
+	Name    string
+	Config  string // TOML
+	State   string
```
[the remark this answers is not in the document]
jmank88: I didn't want to impose anything since this is just for display :shrug:
We actually only use a string in core for EVM now anyways, so we're not giving anything up.

pkg/loop/config.go
jmank88: Planning to follow up here: https://smartcontract-it.atlassian.net/browse/BCF-2112
They will at least have shutdown triggers wired in. It would simplify things to have context and error wired through normally, but it may not be addressed in the initial version. We can sort of work around for now, provided that we also want to fallback to restarting plugins when they crash.

pkg/loop/internal/pb/median.proto
```diff
@@ -0,0 +1,114 @@
...
+import "google/protobuf/empty.proto";
+
+service PluginMedian {
+  rpc NewMedianPluginFactory (NewMedianPluginFactoryRequest) returns (NewMedianPluginFactoryReply) {}
+}
+
+message NewMedianPluginFactoryRequest {
+  uint32 medianProviderID = 1;
+  uint32 dataSourceID = 2;
+  uint32 juelsPerFeeCoinDataSourceID = 3;
```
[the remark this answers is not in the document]
jmank88: The corresponding Go vars are named:
```go
dataSource, juelsPerFeeCoin median.DataSource
```
Any suggestions?

pkg/loop/internal/pb/median.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message LatestTransmissionDetailsRequest {}
+
+message LatestTransmissionDetailsReply {
+  bytes configDigest = 1; // [32]byte
+  uint32 epoch = 2;
+  uint32 round = 3; // uint8
+  bytes latestAnswer = 4;
+  int64 latestTimestamp = 5; // unix milliseconds
```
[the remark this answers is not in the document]
jmank88: Hmmm... Yeah, I don't recall why I did this. We have this precision limitation in the EVM implementation, but I should have matched the interface.

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+  repeated string accounts = 1;
+}
+
+message SignRequest {
+  string account = 1;
+  bytes data = 2;
+}
+
+message SignReply {
+  bytes signed = 1;
```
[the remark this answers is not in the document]
jmank88: Is `signedData` more clear?
```suggestion
  bytes signedData = 1;
```

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+  rpc Keys (google.protobuf.Empty) returns (KeysReply) {}
+  rpc Sign (SignRequest) returns (SignReply) {}
+}
+
+message KeysReply {
+  repeated string accounts = 1;
+}
+
+message SignRequest {
+  string account = 1;
```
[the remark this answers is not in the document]
jmank88: No restrictions imposed by this API. This is a bit of a weird case where the chain-specific relayer uses a chain-specific key, so this can vary by chain family. I expect it to continue to evolve as we implement the others.

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+
+  rpc NodeStatuses (NodeStatusesRequest) returns (NodeStatusesReply) {}
+
+  rpc SendTx (SendTxRequest) returns (google.protobuf.Empty) {}
+}
+
+message RelayArgs {
+  bytes externalJobID = 1; // [32]byte
+  int32 jobID = 2;
+  string contractID = 3;
```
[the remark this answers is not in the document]
jmank88: Matching this type:
https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/types/types.go#L33

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+  bool new = 4;
+  bytes relayConfig = 5;
+}
+
+message PluginArgs {
+  string transmitterID = 1;
+  bytes pluginConfig = 2;
```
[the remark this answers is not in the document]
jmank88: Just matching this:
https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/types/types.go#L27
Nothing imposed by this API since they are defined per plugin (as in OCR2 `ReportingPlugin` - not LOOPP).

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+}
+
+message ChainStatusesRequest {
+  int32 offset = 1;
+  int32 limit = 2;
+}
+
+message ChainStatusesReply {
+  repeated ChainStatus chains = 1;
+  int32 count = 2;
```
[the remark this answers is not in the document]
jmank88: Nearly all of this API is acting as a middle-ware for existing types, so I just started with the simplest version. I'm not opposed to changing these, but it would have cascading effects in legacy core code, so I'd rather do it separately.

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+
+message NodeStatusesRequest {
+  int32 offset = 1;
+  int32 limit = 2;
+  repeated string chainIDs = 3;
+}
+
+message NodeStatusesReply {
+  repeated NodeStatus nodes = 1;
+  int32 count = 2;
```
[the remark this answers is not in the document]
jmank88: It's a total count for paged results. Maybe `total` is better? This is used all over core though too... :thinking:

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+service OffchainConfigDigester {
+  rpc ConfigDigest (ConfigDigestRequest) returns (ConfigDigestReply) {}
+  rpc ConfigDigestPrefix (ConfigDigestPrefixRequest) returns (ConfigDigestPrefixReply) {}
+}
+
+message ContractConfig {
+  bytes configDigest = 1; // [32]byte
+  uint64 configCount = 2;
+  repeated bytes signers = 3; // []OnchainPublicKey
+  repeated string transmitters = 4; // []Account
```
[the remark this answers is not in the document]
jmank88: Is it possible to alias them without paying the extra cost of having full messages?

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+  uint32 F = 5; // uint8
```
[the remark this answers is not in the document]
jmank88: https://github.com/smartcontractkit/libocr/blob/master/offchainreporting2/types/types.go#L103-L104

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,262 @@
...
+}
+
+message ReportContext {
+    ReportTimestamp reportTimestamp = 1;
+    bytes extraHash = 2; // [32]byte
+}
+
+message AttributedOnchainSignature {
+  bytes signature = 1;
+  uint32 signer = 2; // [32]byte
```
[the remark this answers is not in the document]
jmank88: I just typo-ed this one - it is meant for the previous line.

> maybe we should define a named type to represent a 32 byte slice

Can we do that without having an separate message?

pkg/loop/internal/pb/reporting.proto
```diff
@@ -0,0 +1,108 @@
...
+  rpc ShouldTransmitAcceptedReport (ShouldTransmitAcceptedReportRequest) returns (ShouldTransmitAcceptedReportReply) {}
+  rpc Close (google.protobuf.Empty) returns (google.protobuf.Empty) {}
+}
+
+message QueryRequest {
+  ReportTimestamp reportTimestamp = 1;
+}
+
+message QueryReply {
+  bytes query = 1;
```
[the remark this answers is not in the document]
jmank88: Just matching libocr, but it looks like Query creates a query rather than executes it:
https://github.com/smartcontractkit/libocr/blob/master/offchainreporting2/types/types.go#L209-L221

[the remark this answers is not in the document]
jmank88: The Query is passed to the DataSource to obtain an Observation:
https://github.com/smartcontractkit/libocr/blob/master/offchainreporting2/types/types.go#L229

pkg/loop/internal/pb/reporting.proto
```diff
@@ -0,0 +1,108 @@
...
+  bytes query = 2;
+  repeated AttributedObservation observations = 3;
+}
+
+message ReportReply {
+  bool shouldReport = 1;
+  bytes report = 2;
+}
+
+message ShouldAcceptFinalizedReportRequest {
```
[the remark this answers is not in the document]
jmank88: Matching libocr:
https://github.com/smartcontractkit/libocr/blob/master/offchainreporting2/types/types.go#L244

pkg/loop/internal/pb/median.proto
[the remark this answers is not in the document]
krehermann: i meant specifically this one ` juelsPerFeeCoinDataSourceID` is that an accidental concat ?

what does per fee-coin-datasourceid mean?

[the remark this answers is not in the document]
jmank88: It was intentional b/c the type `DataSource` is not included here for context like on the Go side, but there is an existing asymmetry where two variables of the same type have are named as _only_ the type and then without the type at all, so it got weird.
The broader context is that `ReportingPlugin`s expect to have one `DataSource`, and median uses an additional `juelsPerFeeCoin` `DataSource`.

pkg/loop/internal/pb/relayer.proto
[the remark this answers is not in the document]
jmank88: Oh, actually this is a `uint8`
```suggestion
  uint32 signer = 2; // uint8
```

pkg/loop/internal/pb/median.proto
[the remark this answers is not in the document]
jmank88: This actually broke the tests due to timezone differences. The resulting times are no longer `==`, only `Equal()`. I think that should be fine but it was surprising... :thinking:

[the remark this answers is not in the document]
jmank88: This actually broke the tests due to timezone differences. The resulting times are no longer `==`, only `Equal()`. I think that should be fine but it was surprising... :thinking:

pkg/loop/plugin_relayer_test.go
jmank88: `helpers` is not a `package *_test` in this case, but moved consts and vars to `staticdata_test.go` :shrug:

jmank88: `helpers` is not a `package *_test` in this case, but moved consts and vars to `staticdata_test.go` :shrug:

pkg/loop/context.go
```diff
@@ -0,0 +1,48 @@
...
+type ctxKey int
+
+const (
+	ctxKeyJobID ctxKey = iota
+	ctxKeyJobName
+	ctxKeyContractID
+	ctxKeyFeedID
+)
+
+// ContextValues is a helper for passing values via a [context.Context].
```
krehermann, line 14: passing values between what? is this across the plugin boundary? is this indirection/obfuscation needed?

pkg/loop/datasource.go
```diff
@@ -0,0 +1,53 @@
...
+	return &dataSourceClient{grpc: pb.NewDataSourceClient(cc)}
+}
+
+func (d *dataSourceClient) Observe(ctx context.Context) (*big.Int, error) {
+	reply, err := d.grpc.Observe(ctx, &emptypb.Empty{})
+	if err != nil {
+		return nil, err
+	}
+	i := new(big.Int)
+	i.SetBytes(reply.Value)
```
krehermann, line 31: should there be error handling or validation here? i'm surprised it's ok to set this to any byte slice

krehermann (commented):
review of the go code

pkg/loop/datasource.go
jmank88: Nah, it just returns itself: https://pkg.go.dev/math/big#Int.SetBytes
I suppose there are no invalid values since it is just interpreted as a big-endian, but now I see that this is actually a lossy conversion that does absolute-value and then assumes uint :thinking: We'll have to fix this somehow - I don't see a drop-in google.golang.org/protobuf

pkg/loop/context.go
jmank88: It is used across the plugin boundary, but not always/only. This was introduced to preserve logger key/vals, since passing a logger as a GRPC service argument doesn't make sense, and the point was just to have the 'request scoped' logger anyways. It is not the only way to achieve this, and it was not strictly necessary, but it keeps things consistent during the transition phase when the feature can be flipped on/off.

pkg/loop/config.go
```diff
@@ -0,0 +1,218 @@
...
+	pb.UnimplementedContractConfigTrackerServer
+	impl libocr.ContractConfigTracker
+}
+
+func (c *contractConfigTrackerServer) LatestConfigDetails(ctx context.Context, request *pb.LatestConfigDetailsRequest) (*pb.LatestConfigDetailsReply, error) {
+	changedInBlock, configDigest, err := c.impl.LatestConfigDetails(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &pb.LatestConfigDetailsReply{ChangedInBlock: changedInBlock, ConfigDigest: configDigest[:]}, nil
```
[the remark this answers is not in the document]
jmank88: There is nothing to check in this direction since we are converting a fixed length array to a variable length slice.

pkg/loop/config.go
```diff
@@ -0,0 +1,218 @@
...
+
+var _ libocr.ContractConfigTracker = (*contractConfigTrackerClient)(nil)
+
+type contractConfigTrackerClient struct {
+	grpc pb.ContractConfigTrackerClient
+}
+
+func (c *contractConfigTrackerClient) Notify() <-chan struct{} { return nil }
+
+func (c *contractConfigTrackerClient) LatestConfigDetails(ctx context.Context) (changedInBlock uint64, configDigest libocr.ConfigDigest, err error) {
```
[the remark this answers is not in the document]
jmank88: In most of these cases it was just more succinct to express this way, rather than e.g. add a line to declare each of these and a bunch of extra characters on the returns since we have value types with long names.

pkg/loop/config.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	copy(cfg.ConfigDigest[:], reply.ContractConfig.ConfigDigest)
+	cfg.ConfigCount = reply.ContractConfig.ConfigCount
+	for _, s := range reply.ContractConfig.Signers {
+		cfg.Signers = append(cfg.Signers, s)
+	}
+	for _, t := range reply.ContractConfig.Transmitters {
+		cfg.Transmitters = append(cfg.Transmitters, libocr.Account(t))
+	}
+	if reply.ContractConfig.F > math.MaxUint8 {
```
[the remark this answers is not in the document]
jmank88: Hmmm I was trying to keep them located next to the assignment though, and these are aligned with the struct field order - just coincidence that only other check is at the top. Maybe doesn't matter - I'll try it.

pkg/loop/internal/pb/relayer.proto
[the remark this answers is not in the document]
krehermann: yes, thank you

pkg/loop/internal/pb/relayer.proto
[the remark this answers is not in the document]
krehermann: can we tie/link this and the other proto definitions to the libocr code? it would be great to indicate in general that this is all or mostly a wrapper to libocr and point the reader/maintainer to that code base for more documentation and context

pkg/loop/internal/pb/reporting.proto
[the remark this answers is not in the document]
krehermann: same thought as my latest comment re:F

pkg/loop/internal/pb/relayer.proto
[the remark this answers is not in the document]
jmank88: Well we'd pay some performance cost that I'd like to avoid, and it would create an asymmetry between the proto and Go types that seems odd to me.

pkg/loop/internal/pb/relayer.proto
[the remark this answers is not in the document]
jmank88: Updated all of the messages to point to the corresponding types/methods.

pkg/loop/internal/pb/median.proto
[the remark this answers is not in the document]
jmank88: I think it might be too much detail out of context at this level, but the message links back to `NewMedianPluginFactory` now.

pkg/loop/big.go
```diff
@@ -0,0 +1,30 @@
...
+// The encoding uses [*big.Int.Bytes] and a leading sign byte.
+func BigToBytes(b *big.Int) []byte {
+	if b == nil {
+		return nil
+	}
+	var neg byte
+	if b.Sign() < 0 {
+		neg = 1
+	}
+	return append([]byte{neg}, b.Bytes()...)
```
jmank88, line 15: I thought I would be able to leverage something pre-existing, but since I didn't, maybe it would be cleaner to just use a proper proto message for these... :thinking:

pkg/loop/internal/pb/relayer.proto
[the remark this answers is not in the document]
krehermann: probably simplest to add a quick comment here that is the the total count of paged results.

makes it clear, doesn't cause ripples/inconsistency elsewhere, and the reader isn't force to look at ocr docs to understand.

pkg/loop/big.go
```diff
@@ -0,0 +1,30 @@
+package loop
+
+import "math/big"
+
+// BigToBytes returns an encoded [*big.Int] or nil when nil.
+// The encoding uses [*big.Int.Bytes] and a leading sign byte.
+func BigToBytes(b *big.Int) []byte {
```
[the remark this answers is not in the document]
jmank88: String encoding is easy but it has performance costs. I think a pb message for what I'm doing here manually would work well:
```proto
message BigInt {
	boolean negative;
	bytes value;
}
```

pkg/loop/internal/pb/big.go
```diff
@@ -0,0 +1,27 @@
...
+	if b == nil {
+		return nil
+	}
+	i := new(big.Int)
+	i.SetBytes(b.Value)
+	if b.Negative {
+		i = i.Neg(i)
+	}
+	return i
+}
```
jmank88, line 27: @krehermann WDYT about this?

pkg/loop/big.go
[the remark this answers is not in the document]
jmank88: Pushed.

pkg/loop/internal/pb/big.go
krehermann: this look good to me

krehermann (approved):

aalu1418 (approved):

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,306 @@
...
+// NodeStatus represents [github.com/smartcontractkit/chainlink-relay/pkg/types.NodeStatus].
+message NodeStatus {
+  string chainID = 1;
+  string name = 2;
+  string config = 3; // TOML
+  string state = 4;
+}
+
+// SendTxRequest has arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.SendTx].
+message SendTxRequest {
```
prashantkumar1982, line 153: Is this modeled on something existing?
Seems incomplete to me.

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,306 @@
...
+// ChainStatusesRequest has arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.ChainStatuses].
+message ChainStatusesRequest {
+  int32 offset = 1;
+  int32 limit = 2;
+}
+
+// ChainStatusesReply has return arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.ChainStatuses].
+message ChainStatusesReply {
+  repeated ChainStatus chains = 1;
+  int32 count = 2; // total count of chains
```
prashantkumar1982, line 121: Doesn't the above repeated field already support a length() argument?
Why also return count?

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// ChainStatus represents [github.com/smartcontractkit/chainlink-relay/pkg/types.ChainStatus].
+message ChainStatus {
+  string id = 1;
+  bool enabled = 2;
+  string config = 3; // TOML
```
prashantkumar1982, line 128: So the whole TOML is represented as a string here?

Could we convert TOML into a protobuff?

Also, which configs will this include?
Chain-specific, and node specific? Or only 1?

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// NodeStatusesRequest has arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.NodeStatuses].
+message NodeStatusesRequest {
+  int32 offset = 1;
```
prashantkumar1982, line 133: offset and limit should both be uintXX.
Or does the proto guidelines prefer all ints be int32?

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,306 @@
...
+
+// NodeStatusesReply has return arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.NodeStatuses].
+message NodeStatusesReply {
+  repeated NodeStatus nodes = 1;
+  int32 count = 2; // total count of nodes
+}
+
+// NodeStatus represents [github.com/smartcontractkit/chainlink-relay/pkg/types.NodeStatus].
+message NodeStatus {
+  string chainID = 1;
```
prashantkumar1982, line 146: Should we think about making chainID a separate proto?

It could then have structure like:

message ChainID {
    string chainFamily = 1;
    string id = 2;   // Unique id within the chain
    (optional) string globalId = 3;   // A globally unique string id for the chain
}

pkg/loop/internal/pb/relayer.proto
```diff
@@ -0,0 +1,306 @@
...
+  string name = 2;
+  string config = 3; // TOML
+  string state = 4;
```
prashantkumar1982, line 149: Although we currently have state as string, but now that we're separating things across GRPC lines, maybe good time to use enums etc. It is a stronger type, that removes possibility of breaking changes.

pkg/loop/internal/pb/median.proto
```diff
@@ -0,0 +1,138 @@
+syntax = "proto3";
```
prashantkumar1982, line 1: Now that we're using protos, do we follow any style/naming guide for proto files?

Will be good to tag onto some existing standards. Otherwise people will keep tripping over personal preferences etc over how to name services, rpcs, fields, etc etc.

pkg/loop/internal/pb/reporting.proto
```diff
@@ -0,0 +1,122 @@
...
+
+// ReportingPluginConfig represents [github.com/smartcontractkit/libocr/offchainreporting2/types.ReportingPluginConfig]
+message ReportingPluginConfig {
+  bytes configDigest = 1; // [32]byte
+  uint32 oracleID = 2; // uint8
+  uint32 n = 3;
+  uint32 f = 4;
+  bytes onchainConfig = 5;
+  bytes offchainConfig = 6;
+  int64 estimatedRoundInterval = 7;
```
prashantkumar1982, line 31: Could we include interval unit in the name?
estimatedRoundIntervalSeconds

pkg/types/chain.go
```diff
@@ -0,0 +1,38 @@
...
+
+	NodeStatuses(ctx context.Context, offset, limit int, chainIDs ...string) (nodes []NodeStatus, count int, err error)
+
+	SendTx(ctx context.Context, chainID, from, to string, amount *big.Int, balanceCheck bool) error
+}
+
+type ChainService interface {
+	Service
+
+	SendTx(ctx context.Context, from, to string, amount *big.Int, balanceCheck bool) error
```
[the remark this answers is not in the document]
jmank88: Yeah, this is already how Solana works. This first version just adapts it and uses generic types to support the existing CLI, but we have a spike to investigate whether we need to support this at all going forward. I hope we can drop it or simplify it to just 'drain this account' since that is the most common (only? :crossed_fingers: ) use. Otherwise we can iterate to make it chain agnostic later on.

pkg/loop/internal/pb/relayer.proto
jmank88: Yeah, this supports the solana CLI command.

pkg/loop/internal/pb/relayer.proto
jmank88: These responses are paged - `chains` may only be a subset based on the `offset` and `length` args that were used.

pkg/loop/internal/pb/relayer.proto
jmank88: > So the whole TOML is represented as a string here?

Just for this chain.

> Could we convert TOML into a protobuff?

For what purpose? It would prevent relayers from owning the schema in the future, and it is not currently necessary since we are just displaying to the user.

> Also, which configs will this include?
> Chain-specific, and node specific? Or only 1?

I'm not sure what you mean. This isn't really changing from the old way much at all, which was a bag of JSON for a particular chain. Now we just have dynamic TOML instead of dynamic JSON.

pkg/loop/internal/pb/relayer.proto
jmank88: This was matching the existing types, but IIRC they also make use of `-1`.

pkg/loop/internal/pb/relayer.proto
jmank88: We don't currently ever mix nodes from different families together, so this distinction is not necessary. I can see how it could be useful in the future, but I would prefer different terminology for the combination of fields than reusing `ChainID`.

pkg/loop/internal/pb/relayer.proto
jmank88: That would impose one set of states across all chain families, and e.g. require core releases to introduce new states. I don't see any benefit to constraining this field given that it is only used for display right now.

pkg/loop/internal/pb/median.proto
jmank88: I'm open to suggestions :shrug: 
I tried to follow the style in the basic guides and GoLand did some formatting for me.

pkg/loop/internal/pb/reporting.proto
jmank88: Hmmm I would like to keep the names aligned with the Go type if possible. Is it more clear if we include that a note that these are just `time.Duration`s?
```suggestion
  int64 estimatedRoundInterval = 7; // time.Duration
```

pkg/loop/internal/pb/relayer.proto
prashantkumar1982: I mean, if offset = 5, limit is 10 in request.
And the node has say 20 chains configured.

So reply will have this?
chains has length 10(chain ids 5 to 14).
count = 20

Is this what the response will be?

pkg/loop/internal/pb/relayer.proto
prashantkumar1982: Ok, relayers owning the schema sounds good.

I don't know the config internals, but I was assuming some config values are not chain scoped, but global to the Node.
Was asking if those will be passed onto every plugin. But it is fine, if for now, this is mirroring the existing setup.

pkg/loop/internal/pb/relayer.proto
prashantkumar1982: So this will be a string representation of plugin's chain family id?
For example for EVM, it will be int to string?

Maybe I'll need to look into code more deeply, but how would the Core Node understand the native chain family's string chainId? I assume it needs to keep an index/mapping somewhere.

pkg/loop/internal/pb/median.proto
prashantkumar1982: I see this: https://protobuf.dev/programming-guides/

Is this the first time we are using protos in Chainlink?
If yes, I'd say lets start with some existing guides.
We could also open a conversation with wider eng group, and get more opinions.
But I feel people will prefer to use some guide.

pkg/loop/internal/pb/relayer.proto
prashantkumar1982: Hmm, let me think more on this, later.
I'm also confused about the Nodes service being in Core Node, and not in the plugins itself.

pkg/loop/internal/pb/relayer.proto
jmank88: The relayer internals do not currently depend on any global/root config fields directly, but we can accommodate that in the future if we need to. Although it is common to have it be overridable per-chain anyways, so we may not have to represent it separately.

pkg/loop/internal/pb/relayer.proto
jmank88: Everything is reduced to `string` _because_ the core node can't/shouldn't understand it. We already have to use string representations for TOML, go maps, etc. so this is was just a simplification.

pkg/loop/internal/pb/relayer.proto
jmank88: What do you mean? The Relayer plugins are the server side of this RPC call. Core just queries to display in the UI/CLI.

pkg/loop/internal/pb/median.proto
jmank88: > Is this the first time we are using protos in Chainlink?

No, they are in use in many places already, e.g. libocr telemetry and mercury.

PR #83, opened 2023-02-17
Relay logger expose qualified names

essamhassan:
This change enables relay loggers to use similar nested naming functionality to core logger
```
l.Name() == "" # default
l1 = l.Named("from")
l1.Name() == "from"

l2 = l1.Named("within")
l2.Name() == "from.within"
```

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -12,6 +12,9 @@ import (
 
 // Logger is a minimal subset of smartcontractkit/chainlink/core/logger.Logger implemented by go.uber.org/zap.SugaredLogger
 type Logger interface {
+	Name() string
+	Named(string) Logger
+
 	Debug(args ...interface{})
 	Info(args ...interface{})
 	Warn(args ...interface{})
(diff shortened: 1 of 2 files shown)


jmank88 (approved):

jmank88:
> // Logger is a minimal subset of smartcontractkit/chainlink/core/logger.Logger implemented by go.uber.org/zap.SugaredLogger

There is a chance that something still depends on zap's `SugaredLogger` implementing this interface, but IIRC it was within this same repo so would see it breaking now if that were still the case... :thinking:

essamhassan:
> > // Logger is a minimal subset of smartcontractkit/chainlink/core/logger.Logger implemented by go.uber.org/zap.SugaredLogger
> 
> There is a chance that something still depends on zap's `SugaredLogger` implementing this interface, but IIRC it was within this same repo so would see it breaking now if that were still the case... thinking

Wouldn't it still work (with bursting into flames) if someone captured this into a SugeredLogger :thinking:

jmank88 (approved):

PR #84, opened 2023-02-17
Fix iface to match core node logger

essamhassan:
Small fix to make logger iface match between core node and relay


aalu1418 (approved):
approved! just curious do we need to change anything in the chain-specific repos? (solana & starknet)

essamhassan:
> approved! just curious do we need to change anything in the chain-specific repos? (solana & starknet)

I have the repos corresponding changes up in draft, will start cleaning and sending them one by one. They need to go after the mercury PRs though.

PR #88, opened 2023-03-07
Modules from core

sdrug:


aalu1418 (changes requested):
Can you copy over the test files for the utils that you copied over too?

Also to fix the failing solana CI check I think you can remove this line: https://github.com/smartcontractkit/chainlink-relay/blob/0298f0e913fe98bcbf2d5a4c95f33b08a968dd59/.github/workflows/build_external.yml#L62

pkg/config/error.go
```diff
@@ -31,3 +31,12 @@ type ErrEmpty struct {
 func (e ErrEmpty) Error() string {
 	return fmt.Sprintf("%s: empty: %s", e.Name, e.Msg)
 }
+
+type KeyNotFoundError struct {
```
jmank88, line 35: How will this be used?

sdrug: It's being used in the txm tests, for example [here](https://github.com/smartcontractkit/chainlink-solana/pull/475/files#diff-0c15ff13416d346cba7814e0b6501c118492a504ee2bb0263e77d6b6c10280aaR52).

jmank88: Hmmm OK - we might have to revisit this with plugins, since error types are not preserved over the wire. Fortunately it seems like we could just return nil instead of a custom error for this particular case.

sdrug:
@aalu1418 
I think the CI is failing because it cannot build chainlink-solana (develop branch?) with the golang version lower than 1.20. [This](https://github.com/smartcontractkit/chainlink-solana/pull/479) should help.

aalu1418 (approved):
lgtm! I will take a look at fixing the CI runs in a separate PR

sdrug:
Merging this part of work for now. The further improvements will follow in Aaron's PR

PR #95, opened 2023-04-18
Block range guarantee - Mercury

vyzaldysanchez:
Goes with: https://github.com/smartcontractkit/chainlink/pull/9022


jmank88:
Bumping libocr and updating loop here to unblock you: https://github.com/smartcontractkit/chainlink-relay/pull/101

.tool-versions
```diff
@@ -1,2 +1,2 @@
-golang 1.20.1
+golang 1.20.3
```
jmank88, line 1:
```suggestion
golang 1.20.4
```

go.mod
```diff
@@ -4,52 +4,53 @@ go 1.20
 
 require (
 	github.com/confluentinc/confluent-kafka-go v1.9.2
-	github.com/golang/protobuf v1.5.2
+	github.com/golang/protobuf v1.5.3 // indirect
```
jmank88, line 7: Is this tidied? `indirect` is out of place here.

jmank88 (changes requested):
We need to slow down and get all the pieces in to a consistent re-viewable state before merging.

jmank88 (changes requested):
LGTM but we need to wait on libocr, yeah?

samsondav (approved):

jmank88 (approved):

PR #96, opened 2023-04-19
pkg/types: s/Plugin/PluginProvider

jmank88:
Rename `Plugin` to `PluginProvider`, but leave behind an alias for compatability.


aalu1418 (dismissed):
lgtm! (assuming the build failures are fixed?)

jmank88:
Updated with a build fix.

PR #98, opened 2023-04-19
add golangci-lint config

aalu1418:


pkg/loop/relayer_service.go
```diff
@@ -158,11 +158,11 @@ func (p *RelayerService) keepAlive() {
...
 			if c != nil && !c.Exited() && cp != nil {
 				// launched
-				if err := cp.Ping(); err == nil {
+				err := cp.Ping()
+				if err == nil {
 					continue // healthy
-				} else {
-					p.lggr.Errorw("Relaunching unhealthy plugin", "err", err)
 				}
+				p.lggr.Errorw("Relaunching unhealthy plugin", "err", err)
```
aalu1418, line 165: logic refactor

aalu1418 (commented):
most changes are goimports, err shadow declaration, typos

commented on the few changes that are logic refactor

ops/utils/gauntlet.go
```diff
@@ -84,7 +84,7 @@ func (g Gauntlet) ExecCommand(arg ...string) error {
 	a := []string{}
 	a = append(a, g.arg...)
 	a = append(a, arg...)
-	// Execute
+	// #nosec - Execute
```
jmank88, line 87: I don't recognize this format. Is there an advantage over `//nolint`? https://golangci-lint.run/usage/false-positives/#nolint-directive
```suggestion
	//nolint:nosec - Execute
```

aalu1418: oh I didn't know golangci-lint had it's own - the #nosec is just directly interacting with the underlying linter. let me give it a shot

aalu1418: gave it a run, but `nolint` doesn't do anything?

jmank88: :shrug:

jmank88: Weird:
https://github.com/golangci/golangci-lint/blob/master/pkg/result/processors/testdata/nolint2.go#L9
https://github.com/golangci/golangci-lint/blob/master/pkg/result/processors/testdata/nolint_bad_names.go#L13

aalu1418: I wonder if `nolint` is for other types of linters? and `#nosec` is specific? https://github.com/golangci/golangci-lint/blob/1948081d84d96254f3da9d409d185e0942f1132b/.golangci.reference.yml#L888

jmank88 (approved):

PR #100, opened 2023-04-20
modify context deadline in dataSourceServer.Observe() to work around pipeline behavior

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2176

Core bump: https://github.com/smartcontractkit/chainlink/pull/9083

Follow up in https://smartcontract-it.atlassian.net/browse/BCF-2209

### DIFF
--- a/pkg/loop/datasource.go
+++ b/pkg/loop/datasource.go
@@ -3,6 +3,8 @@ package loop
 import (
 	"context"
 	"math/big"
+	"os"
+	"time"
 
 	"google.golang.org/grpc"
 
@@ -10,7 +12,22 @@ import (
 	"github.com/smartcontractkit/libocr/offchainreporting2/types"
 
 	pb "github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb"
+	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 )
+
+// github.com/smartcontractkit/libocr/offchainreporting2/internal/protocol.ReportingPluginTimeoutWarningGracePeriod
+var datasourceOvertime = 100 * time.Millisecond
+
+func init() {
+	// undocumented escape hatch
+	// TODO: remove with https://smartcontract-it.atlassian.net/browse/BCF-2209
+	if v := os.Getenv("CL_DATASOURCE_OVERTIME"); v != "" {
+		d, err := time.ParseDuration(v)
+		if err == nil {
+			datasourceOvertime = d
+		}
+	}
+}
 
 var _ median.DataSource = (*dataSourceClient)(nil)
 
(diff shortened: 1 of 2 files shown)


pkg/loop/datasource.go
```diff
@@ -39,6 +41,13 @@ type dataSourceServer struct {
 }
 
 func (d *dataSourceServer) Observe(ctx context.Context, _ *emptypb.Empty) (*pb.ObserveReply, error) {
+	// Pipeline observations may return results after the context is cancelled, so we modify the
+	// deadline to give them time to return before the parent context deadline.
```
krehermann, line 45: can you file a ticket to fix the root cause and link that ticket here in the comment? open to other ideas about how to track it, but somehow we need to indicate that this workaround is intended to be temporary.

pkg/loop/datasource.go
```diff
@@ -39,6 +41,13 @@ type dataSourceServer struct {
...
+	var cancel func()
+	ctx, cancel = utils.ContextWithDeadlineFn(ctx, func(old time.Time) time.Time {
+		return old.Add(-time.Until(old) / 10)
```
krehermann, line 48: does dividing a duration by an integer always work as expected? i have a vague memory of being tripped up by this in the past.

pkg/utils/utils.go
```diff
@@ -34,3 +34,16 @@ func ContextFromChan(chStop <-chan struct{}) (context.Context, context.CancelFun
 	}()
 	return ctx, cancel
 }
+
+// ContextWithDeadlineFn returns a copy of the parent context with the deadline modified by deadlineFn.
+// deadlineFn will only be called if the parent has a deadline.
+// The new deadline must be sooner than the old to have an effect.
+func ContextWithDeadlineFn(ctx context.Context, deadlineFn func(old time.Time) time.Time) (context.Context, context.CancelFunc) {
```
krehermann, line 41: nit: i prefer `original` or `orig` to `old`, it makes it easy for me to understand.

nbd either way

pkg/loop/datasource.go
jmank88: `time.Duration` is nanosecond precision, so I think the point that rounding errors become problematic is already too short of a duration to do actual work anyways.

jmank88, line 48: I'm not certain that scaling by some % is necessarily the best solution here :thinking: 
In median's particular use case, we generally have several concurrent tasks fetching data from EAs, and then a final step calculate the answer. Ideally we want to maximize the time allocated to the EAs, and only reduce the deadline just enough to accommodate the final answer calculation, which is some fixed duration that is independent of the parent context deadline.

krehermann: could we introduce a prom histogram to measure how much extra time this takes and have it measure per job?

or is it easy to fix the root cause?

jmank88: I don't think we can measure it from the outside here. Something like that could be used inside the pipeline but would have to be a bit more general/dynamic IIUC.

krehermann (approved):

krehermann (approved):

PR #102, opened 2023-04-24
pkg/loop: rename Keystore.Keys to Accounts

jmank88:
Supporting https://github.com/smartcontractkit/chainlink/pull/9034


aalu1418 (approved):

pkg/loop/internal/pb/relayer.proto
```diff
@@ -22,11 +22,11 @@ message NewRelayerReply {
 }
 
 service Keystore {
-  rpc Keys (google.protobuf.Empty) returns (KeysReply) {}
+  rpc Accounts (google.protobuf.Empty) returns (AccountsReply) {}
```
krebernisak, line 25: Not sure if this is a good change, as some chains have a separation between a key and an account. For example, Starknet with its account abstraction as a contract implementation.

The account is the address of a deployed account contract, and the key is just a key that has access to that account. Keys can be changed, while the account stays the same.

jmank88: Hmmm I'm not sure how to incorporate that here at this point. I think this better reflects the existing APIs that we are currently wrapping, which were using Key for a type that is capable of signing, and Account/Address for the identifier of the key. Is there better terminology we can use that doesn't come with so many assumptions?

PR #103, opened 2023-04-26
pkg/loop: set HandshakeConfig magic cookies

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2198

### DIFF
--- a/pkg/loop/plugin_median.go
+++ b/pkg/loop/plugin_median.go
@@ -32,9 +32,8 @@ type ErrorLog interface {
 
 func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
 	return plugin.HandshakeConfig{
-		ProtocolVersion:  0,
-		MagicCookieKey:   "magic-key-median-todo",
-		MagicCookieValue: "magic-value-median-todo",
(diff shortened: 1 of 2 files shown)


pkg/loop/plugin_median.go
```diff
@@ -32,9 +32,8 @@ type ErrorLog interface {
 
 func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
 	return plugin.HandshakeConfig{
-		ProtocolVersion:  0,
-		MagicCookieKey:   "magic-key-median-todo",
-		MagicCookieValue: "magic-value-median-todo",
+		MagicCookieKey:   "CL_PLUGIN_MEDIAN_MAGIC_COOKIE",
+		MagicCookieValue: "b12a697e19748cd695dd1690c09745ee7cc03717179958e8eadd5a7ca4646728",
```
jmank88, line 36: https://pkg.go.dev/github.com/hashicorp/go-plugin#HandshakeConfig

- implicit zero version until we need it later
- caps seem to be the standard for the keys
- values are 32 random hex bytes

krehermann (approved):

PR #107, opened 2023-05-01
Adding currentBlockTimestamp field to Mercury report

sdrug:


pkg/reportingplugins/mercury/aggregate_functions.go
```diff
@@ -47,7 +47,7 @@ func GetConsensusAsk(paos []ParsedAttributedObservation) *big.Int {
 
 // GetConsensusCurrentBlock gets the most common (mode) block hash/number.
 // In the event of a tie, use the lowest numerical value
-func GetConsensusCurrentBlock(paos []ParsedAttributedObservation, f int) (hash []byte, num int64, err error) {
+func GetConsensusCurrentBlock(paos []ParsedAttributedObservation, f int) (hash []byte, num int64, ts uint64, err error) {
```
sdrug, line 50: @samsondav Is it possible that consensus block num and consensus hash belong to different blocks? If not, please consider this update. 
I commented out the related tests for now, depending on the review here I will either fix them or bring all back.

pkg/reportingplugins/mercury/mercury.go
```diff
@@ -360,6 +362,12 @@ func (rp *reportingPlugin) Observation(ctx context.Context, repts ocrtypes.Repor
 		p.CurrentBlockHashValid = true
 	}
 
+	if obs.CurrentBlockTimestamp.Err != nil {
+		obsErrors = append(obsErrors, pkgerrors.Wrap(obs.CurrentBlockTimestamp.Err, "failed to observe CurrentBlockTimestamp"))
+	} else {
+		p.CurrentBlockTimestamp = obs.CurrentBlockTimestamp.Val
```
sdrug, line 368: do we need `p.CurrentBlockTimestamapValid` field here similar to other block values?

pkg/reportingplugins/mercury/aggregate_functions_test.go
```diff
@@ -42,24 +42,25 @@ func Test_AggregateFunctions(t *testing.T) {
...
-			assert.EqualError(t, err, "no block number matching hash 0x40044147503a81e9f2a225f4717bf5faf5dc574f69943bdcd305d5ed97504a7e with at least f+1 votes")
-		})
+		// t.Run("if there are not at least f+1 in consensus about number", func(t *testing.T) {
+		// 	badPaos := NewParsedAttributedObservations()
+		// 	for i := range badPaos {
+		// 		badPaos[i].CurrentBlockNum = int64(i)
+		// 	}
+		// 	_, _, err := GetConsensusCurrentBlock(badPaos, f)
+		// 	assert.EqualError(t, err, "no block number matching hash 0x40044147503a81e9f2a225f4717bf5faf5dc574f69943bdcd305d5ed97504a7e with at least f+1 votes")
+		// })
```
samsondav, line 63: ??

pkg/reportingplugins/mercury/mercury.go
samsondav: Yes it can be missing

samsondav (changes requested):

pkg/reportingplugins/mercury/aggregate_functions.go
sdrug: Ok. Does it have to be done the same way for the timestamp as well? 
Alternatively, the timestamp can be taken from the observation chosen by hash or by block number.

sdrug: And I'd appreciate if you could explain why block number can be different for the same block hash.

samsondav: Yes it should be done the same way for the timestamp. It could potentially be different e.g. on chains with high probability of re-orgs where you can have two nodes that observe a block with the same number but different hash.

samsondav (changes requested):

pkg/reportingplugins/mercury/aggregate_functions.go
```diff
@@ -78,77 +128,83 @@ func GetConsensusCurrentBlock(paos []ParsedAttributedObservation, f int) (hash [
...
+	return
 }
 
-// GetConsensusValidFromBlock gets the most common (mode) ValidFromBlockNum
-// In the event of a tie, the lower number is chosen
-func GetConsensusValidFromBlock(paos []ParsedAttributedObservation, f int) (int64, error) {
-	// pick the most common block number with at least f+1 votes
-	m := map[int64]int{}
-	maxCnt := 0
+func getConsensusCurrentBlockTimestamp(paos []ParsedAttributedObservation, blockHash string, threshold int) (ts uint64, err error) {
```
samsondav, line 174: This should probably only match blocks with the same hash AND number.

sdrug: Added 👌 
This behaviour isn't specified in the ticket. Please let me know if I should further consult someone on this.

samsondav (dismissed):

samsondav (approved):

PR #108, opened 2023-05-05
pkg/loop: auto-recovery

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2112

This PR introduces automatic recovery of plugins. This is primarily achieved with `*clientConn`, a custom implementation of `grpc.ClientConnInterface` that contains a swappable `*grpc.ClientConn`, allowing us to recreate and reconnect silently whenever necessary.

TODO:
- [x] Median still has some logic in core that should be aligned with Relayer in this repo and better tested


pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+		if err != nil {
+			err = ErrConnDial{Name: c.name, ID: id, Err: err}
+			lggr.Errorw("Client dial failed", "err", err)
+			c.closeAll(c.deps...)
+			return false
+		}
+		return true
+	}
+
+	const maxWait = 5 * time.Second
```
krehermann, line 128: why? where did this come from

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	for wait := 100 * time.Millisecond; !try(); wait *= 2 {
```
krehermann, line 129: can we use an exponential backoff here? there are a number of package to do that

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	return c.cc
+}
+
+func isErrTerminal(err error) bool {
+	switch status.Code(err) {
+	case codes.Unavailable, codes.Canceled:
+		return true
+	case codes.OK, codes.Unknown, codes.InvalidArgument, codes.DeadlineExceeded, codes.NotFound, codes.AlreadyExists,
```
krehermann, line 148: why bother making these explicit? would be easier to read if only the small set of terminal errs is explicit

krehermann: what are the rules for judging whether an err code is terminal? what happens if we wrongly judge one?

pkg/loop/helper_test.go
```diff
@@ -1,24 +1,39 @@
...
 
-func (p *RelayerService) Kill() {
-	p.mu.RLock()
-	defer p.mu.RUnlock()
-	if p.client != nil {
-		p.client.Kill()
-	}
+// TestHook returns a TestRelayerService.
+// It must only be called once, and before Start.
+func (r *pluginService[P, S]) TestHook() TestRelayerService[P, S] {
```
krehermann, line 9: what are P and S? can we colocate with other *pluginService funcs since we are in the same package?

pkg/loop/lggr_broker.go
```diff
@@ -51,3 +105,24 @@ type resource struct {
 	io.Closer
 	name string
 }
+
+type resources []resource
```
krehermann, line 109: where is resource defn? can we put this and the following funcs next to it?

pkg/loop/median_service.go
```diff
@@ -0,0 +1,42 @@
+package loop
```
krehermann, line 1: is this being moved out the the `chainlink` repo? if so, can we do that in separate change?

krehermann (commented):
reviww of first half

krehermann (commented):
reviww of first half

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	*lggrBroker
+	newClient newClientFn
+	name      string
+
+	mu   sync.RWMutex
+	deps resources
+	cc   *grpc.ClientConn
+}
+
+func (lb *lggrBroker) newClientConn(name string, newClient newClientFn) *clientConn {
```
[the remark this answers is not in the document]
jmank88: Yeah the internal broker is swapped out, so it has to be propagated and this is how the API settled after refactoring a regular constructor. Will rethink or relocate.

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	return &clientConn{
+		lggrBroker: lb.named(name),
+		newClient:  newClient,
+		name:       name,
+	}
+}
+
+func (c *clientConn) Invoke(ctx context.Context, method string, args interface{}, reply interface{}, opts ...grpc.CallOption) error {
```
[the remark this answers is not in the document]
jmank88: Is the doc on `clientConn` sufficient? I usually try to avoid spamming around method implementation comments that don't have anything to add.

pkg/loop/client.go
jmank88: The linter insists on exhaustively listing types like this.

jmank88: If we are wrong then we return it to the caller instead of refreshing the connection.

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.cc != orig {
+		return c.cc
+	}
+	if c.cc != nil {
+		if err := c.cc.Close(); err != nil {
+			c.lggr.Errorw("Client close failed", "err", err)
+		}
+		c.closeAll(c.deps...)
```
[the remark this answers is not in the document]
jmank88: Closing deps first leaves a period where the connection is still open and possibly trying to use those deps.

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	try := func() bool {
+		c.lggr.Debug("Client refresh")
+		id, deps, err := c.newClient(ctx)
+		if err != nil {
+			c.lggr.Errorw("Client refresh failed", "err", err)
+			c.closeAll(deps...)
```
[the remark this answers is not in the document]
jmank88: Yeah, it's just a helper func.

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+	c.mu.RLock()
+	cc := c.cc
+	c.mu.RUnlock()
+
+	if cc == nil {
+		cc = c.refresh(ctx, nil)
+	}
+	for cc != nil {
+		err := cc.Invoke(ctx, method, args, reply, opts...)
+		if err == nil || !isErrTerminal(err) {
```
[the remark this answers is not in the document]
jmank88: Yeah this is returning in the normal case of no error, or non-terminal error.
I'll add a comment and tinker, but I think the opposite will end up more confusing due to not returing early: https://danp.net/posts/reducing-go-nesting/

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+			return err
+		}
+		cc = c.refresh(ctx, cc)
```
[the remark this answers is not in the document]
jmank88: Because each iteration represents a terminal error that requires reconnecting. It can only go as fast as we can relaunch and initialize everything, so it is indirectly throttled. We currently get stuck, by design, to support methods w/o error signatures and to not impose upstream refactoring - but we may want to relax this in the future. (FWIW context.Context is still respected here)

pkg/loop/client.go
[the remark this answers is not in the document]
jmank88: Eh, not so bad with a `continue`.

pkg/loop/helper_test.go
jmank88: These are isolated in a `_test.go` file to exclude them from the real exported package. We have a habit of using `helpers_test.go` as a catch all - would `plugin_service_test.go` help?

pkg/loop/lggr_broker.go
jmank88: It is right there ^ if you expand.

pkg/loop/median_service.go
jmank88: What do you mean? Separate how?
Actually using this in core is already separate, what difference does it make if this type is included here now?

pkg/loop/client.go
krehermann: i don't follow spam vs silence. i thought maxWait was a timeout. re-reading, does this loop forever if try() always returns false?

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+var _ grpc.ClientConnInterface = (*clientConn)(nil)
+
+// newClientFn returns a new client connection id to dial, and a set of resource dependencies to close.
+type newClientFn func(context.Context) (id uint32, deps resources, err error)
+
+// clientConn is a [grpc.ClientConnInterface] backed by a [*grpc.ClientConn] which can be recreated and swapped out
+// via the provided [newClientFn].
+type clientConn struct {
+	*lggrBroker
+	newClient newClientFn
```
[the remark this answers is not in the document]
krehermann: this was note to self. ignore

pkg/loop/client.go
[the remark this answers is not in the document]
krehermann: yes, i missed it, sorry.

pkg/loop/client.go
```diff
@@ -0,0 +1,154 @@
...
+func (c *clientConn) NewStream(ctx context.Context, desc *grpc.StreamDesc, method string, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	c.mu.RLock()
+	cc := c.cc
+	c.mu.RUnlock()
+
+	if cc == nil {
+		cc = c.refresh(ctx, nil)
+	}
+	for cc != nil {
+		s, err := cc.NewStream(ctx, desc, method, opts...)
```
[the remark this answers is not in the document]
krehermann: done

pkg/loop/client.go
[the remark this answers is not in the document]
jmank88: I mean we keep retrying until we can return a 'real' response, or the system is shutting down.
If we give up eventually that assumes the conditions can't/won't change. Even with a static docker image, we don't necessarily know that is true, and we are trying to preserve the non-LOOPP behavior of boot-looping when the node crashes.

pkg/loop/plugin_service.go
```diff
@@ -0,0 +1,206 @@
...
+func (r *pluginService[P, S]) launch() (*plugin.Client, plugin.ClientProtocol, error) {
+	ctx, cancelFn := utils.ContextFromChan(r.stopCh)
+	defer cancelFn()
+
+	r.lggr.Debug("Launching")
+
+	cc := r.grpcPlug.ClientConfig()
+	cc.Cmd = r.cmd()
+	client := plugin.NewClient(cc)
+	cp, err := client.Client()
```
krehermann, line 114: why is a ClientProtoocal returned from client.Client()?

is this our naming or hashicorp?

aside: what is the difference between a Client and a client protocol?

pkg/loop/plugin_service.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		client.Kill()
+		return nil, nil, fmt.Errorf("failed to create plugin Client: %w", err)
```
krehermann, line 117: failed to create plugin client protocol?

pkg/loop/client.go
jmank88: Yeah, loops until connection is established or context expires.

pkg/loop/plugin_service.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	i, err := cp.Dispense(r.pluginName)
+	if err != nil {
+		abort()
+		return nil, nil, fmt.Errorf("failed to Dispense %q plugin: %w", r.pluginName, err)
+	}
+
+	select {
+	case <-r.serviceCh:
+		// already set
```
krehermann, line 133: what is already set?

pkg/loop/plugin_service.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		defer close(r.serviceCh)
+	}
+	return client, cp, nil
+}
+
+func (r *pluginService[P, S]) Start(context.Context) error {
+	return r.StartOnce("PluginService", func() error {
+		r.wg.Add(1)
+		go r.keepAlive()
```
krehermann, line 148: easier to read if the keepAlive accepts the wg as an arg

pkg/loop/client.go
jmank88: > re:linter doesn't a default case with return false work to avoid explicit false case?

Works now (and that's what I wrote originally), but is brittle since it pre-classifies future types I suppose. If a new type is introduce, we would be forced to classify it explicitly.
https://golangci-lint.run/usage/linters/#exhaustive
https://pkg.go.dev/github.com/nishanths/exhaustive

> re: bad judgement - does that imply that we could get stuck with a bad connection indefinitely?

Yes, and the opposite mistake would needlessly recreate clients for non-terminal cases.

pkg/loop/lggr_broker.go
```diff
@@ -1,42 +1,96 @@
...
+
+func (a *atomicBroker) Dial(id uint32) (conn *grpc.ClientConn, err error) {
+	return a.broker.Load().Dial(id)
+}
+
+func (a *atomicBroker) NextId() uint32 {
+	return a.broker.Load().NextId()
+}
+
 type lggrBroker struct {
```
krehermann, line 44: aside: why is the log broker such an important concept throughout the code? it's used all over the place but i don't understand why a broker is needed in the first place.

krehermann (commented):
a few more files

pkg/loop/plugin_service.go
jmank88: Hmmm passing a struct field as a method param seems like an anti-pattern though... :thinking:

pkg/loop/plugin_service.go
jmank88: `s.service`
> serviceCh chan struct{} // closed when service is available

pkg/loop/plugin_service.go
jmank88: It's their naming. It feels like there's an extra type, but I don't fully grok it. Will revisit.

pkg/loop/lggr_broker.go
jmank88: The [`plugin.GRPCBroker`](https://pkg.go.dev/github.com/hashicorp/go-plugin#GRPCBroker) is fundamentally how we connect our grpc clients/servers, both initially and to multiplex as we wire things back and forth.

krehermann: keep helping me - why do we need the logs to be going over grpc? why does the node need to care about the logs from a LOOP implementation?

pkg/loop/plugin_service.go
krehermann: ok, rather than `already set` how about `service is ready`

pkg/loop/client.go
krehermann: i see. it warrants a comment -- i wouldn't expect NewStream to block forever under some circumstance. that is, it's very dangerous for a caller to pass a context with a timeout or deadline

pkg/loop/client.go
krehermann: yeah, both are bad.

getting stuck with a bad connection seems scarier.

why would something like Unimplemented not be terminal?

pkg/loop/lggr_broker.go
krehermann: that's good. but i'm still confused.

why is the node providing a logger? is it solely for signature compatibility?

and how is that even possible in terms of dependencies? the logger in the node is a interfaced defined in the `chainlink` repo while the logger used in this code is defined in this repo. 

are we copy and pasting interface definitions across repos and hashicorp magically un/marshals between them?

pkg/loop/median_service.go
krehermann: i would like to see the atomicity and generic pluginService in one change and then a second change to refactor the existing Median/Relayer code to use them.

is that possible? i get too lost reading the code to disentangle the moving pieces.

pkg/loop/lggr_broker.go
jmank88: AFAIK it is just piping raw lines through (and prefixing them), that's why we have to configure the logger in the plugins with the same options from the host.

jmank88: The Logger interface in this repo being a subset of core is not directly related to how the data is moving around in this case.

pkg/loop/client.go
jmank88: Indeed, but we have no choice currently since we cannot even return errors fully in some cases. If we can get `error`s introduced to a few more libocr methods, then we'll be able to consider returning early from here, but that won't necessarily 'just work' - we have to make sure the code up the stack is prepared to retry on it's own.

pkg/loop/client.go
jmank88: `Unimplemented` is an actual response from a live server - just that it doesn't serve that method.

pkg/loop/median_service.go
jmank88: I don't think that is possible, not without duplicating most things in the mean time.

pkg/loop/helper_test.go
```diff
@@ -1,24 +1,39 @@
 package loop
 
+import "github.com/smartcontractkit/chainlink-relay/pkg/types"
+
 const KeepAliveTickDuration = keepAliveTickDuration
```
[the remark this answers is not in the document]
jmank88: I'd rather not expand the real exported API unnecessarily. This is just a test quirk.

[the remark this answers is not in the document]
krehermann: i overlooked that this is in a `_test.go` file. ignore.

pkg/loop/plugin_median.go
```diff
@@ -17,12 +17,14 @@ import (
 	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
 	pb "github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb"
 	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 )
 
 // PluginMedianName is the name for [PluginMedian]/[NewGRPCPluginMedian].
 const PluginMedianName = "median"
 
 type PluginMedian interface {
```
krehermann, line 26: Is this better as PluginMedianService MedianPluginServicer? i like the latter -- the `er` makes it clear that it's an interface and `MedianPluginServicer` expresses the intention of a service for median plugins

pkg/loop/plugin_median.go
```diff
@@ -17,12 +17,14 @@ import (
...
+	types.Service
 	NewMedianPluginFactory(ctx context.Context, provider types.MedianProvider, dataSource, juelsPerFeeCoin median.DataSource, errorLog ErrorLog) (libocr.ReportingPluginFactory, error)
```
krehermann, line 28: aside -- i believe i'm mildly dyslexic. intermixing names like AB with BA (in this case PluginMedian with NewMedianPlugin) makes it extremely difficult for me to follow.

pkg/loop/lggr_broker.go
krehermann: i'm still confused a simple level.  what is the utility of the lggrBroker?

is a `lggrBroker` a thing that deals out loggers? (that's my first impression from the name) or a GRPC broker impl that happens to log internally?

some comments on the struct definition would help me a lot

pkg/loop/plugin.go
```diff
@@ -0,0 +1,27 @@
...
+
+type pluginClient struct {
+	atomicBroker
+	atomicClient
+	*lggrBroker
+}
+
+func newPluginClient(stopCh <-chan struct{}, lggr logger.Logger, broker *plugin.GRPCBroker, conn *grpc.ClientConn) *pluginClient {
+	var pc pluginClient
+	pc.lggrBroker = &lggrBroker{stopCh: stopCh, lggr: lggr, broker: &pc.atomicBroker}
```
krehermann, line 18: i suspect this line is critical, but for the life of me, i don't know why.

the code reads like you set it and then nothing ever uses it. what's going on?

pkg/loop/plugin_median.go
```diff
@@ -37,116 +39,111 @@ func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
...
-	lggr logger.Logger
-}
+	StopCh <-chan struct{}
+	Logger logger.Logger
 
-func NewGRPCPluginMedian(mp PluginMedian, lggr logger.Logger) plugin.Plugin {
-	return &grpcPluginMedian{impl: mp, lggr: lggr}
+	PluginServer PluginMedian
+
+	pluginClient *pluginMedianClient
```
krehermann, line 50: please use a different name for this field. we use `pluginClient` as a struct name in this package to mean something different that this (right? this is not 1:1 with `pluginClient` defn'd in plugin.go`? it's more and other stuff)

krehermann: right now we have pluginClient := &pluginMedianClient{*pluginClient} which i can't follow

pkg/loop/plugin_median.go
```diff
@@ -37,116 +39,111 @@ func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
...
+		Plugins:          map[string]plugin.Plugin{PluginMedianName: p},
+		AllowedProtocols: []plugin.Protocol{plugin.ProtocolGRPC},
+	}
 }
 
 var _ PluginMedian = (*pluginMedianClient)(nil)
 
 type pluginMedianClient struct {
-	*lggrBroker
+	*pluginClient
```
krehermann, line 83: can we give this field names? when you write m.newClientConn on 96, i can't tell which of these two struct is being used

pkg/loop/reporting.go
```diff
@@ -19,12 +18,15 @@ type reportingPluginFactoryClient struct {
 	grpc pb.ReportingPluginFactoryClient
 }
 
-func newReportingPluginFactoryClient(lb *lggrBroker, cc *grpc.ClientConn) *reportingPluginFactoryClient {
+func newReportingPluginFactoryClient(lb *lggrBroker, cc grpc.ClientConnInterface) *reportingPluginFactoryClient {
```
krehermann, line 21: is a FactoryClient a thing that client to a factory a factory that creates clients?

if the latter, i want it to be renamed to ClientFactory.

krehermann (commented):
more confusion :)

krehermann (commented):
more confusion :)

pkg/loop/client.go
jmank88: libocr defines many of the interfaces that make up this API.

pkg/loop/lggr_broker.go
jmank88: It's just a broker with helper methods that need to log. I'll find a better name.,

pkg/loop/plugin.go
jmank88: Which part are you referring to? All of the methods from the embed types are available and used.
This declaration is just odd because `broker` is an interface and we need to retain a reference to the atomic impl to be able to update it.

I'll update `pluginClient` with a doc.

pkg/loop/plugin_median.go
jmank88: Hmm I think having `Service` and `Servicer` would be confusing... :thinking:

pkg/loop/plugin_median.go
jmank88: What do you mean? The struct name and the field type give the extra context. I even tried to name these just `Server` and `client`, but that clashes with the net rpc  method names. I think saying more would just be redundant no? What do you propose?

There might be a way to simplify this more broadly with generics, but IIRC I already tried that once. e.g. `GRPCPlugin[P ...]`

pkg/loop/plugin_median.go
jmank88: This is by design to inherit methods. More succinct access is generally regarded as a feature, but besides that we save on boilerplate re-implementations of wrapper methods.

jmank88: When there is ambiguity the language forces you to be explicit.

jmank88: (the indirect `serviceClient.lb` method is not inherited in this case)
```go
m.pluginClient.newClientConn()
m.serviceClient.lb.newClientConn()
```

pkg/loop/reporting.go
jmank88: No it's a client for a [`ReportingPluginFactory`](https://github.com/smartcontractkit/libocr/blob/29f534b2de1a05a91a17aab500aef283c5725956/offchainreporting2/types/types.go#L87).

pkg/loop/plugin_median.go
krehermann: i'm all for reduced boilerplate.

however, the code has to be maintainable and readable. i've spent 3 1-2 hr sessions reading the code and i simply can't follow what's happening.

an alternative to naming the fields is to add more comments and figure out clearer, more descriptive names. see next comment for an example of my struggles

pkg/loop/plugin_median.go
```diff
@@ -37,116 +39,111 @@ func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
...
+	median pb.PluginMedianClient
 }
 
-func newMedianPluginClient(lggr logger.Logger, broker *plugin.GRPCBroker, conn *grpc.ClientConn) *pluginMedianClient {
-	lggr = logger.Named(lggr, "MedianReportingClient")
-	return &pluginMedianClient{lggrBroker: &lggrBroker{lggr, broker}, grpc: pb.NewPluginMedianClient(conn)}
+func newPluginMedianClient(stopCh <-chan struct{}, lggr logger.Logger, broker *plugin.GRPCBroker, conn *grpc.ClientConn) *pluginMedianClient {
+	lggr = logger.Named(lggr, "PluginMedianClient")
+	pc := newPluginClient(stopCh, lggr, broker, conn)
+	return &pluginMedianClient{pluginClient: pc, median: pb.NewPluginMedianClient(pc), serviceClient: newServiceClient(pc.lggrBroker, pc)}
```
krehermann, line 92: `&pluginMedianClient{pluginClient: pc, median: pb.NewPluginMedianClient(pc), serviceClient: newServiceClient(pc.lggrBroker, pc)`

this reads like a circle. `pc` is a pluginClient and is the only thing passed to the other fields.

what does it mean to have a pluginMedianClient comprised of a pluginClient and a NewPluginMedianClient? it looks like a circle

and what does it mean to have a serviceClient that is constructed from a *pluginClient? it reads like the service client isn't needed at all.

pkg/loop/median_service.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
+	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+)
+
+var _ PluginMedian = (*MedianService)(nil)
+
+// MedianService is a [types.Service] that maintains an internal [PluginMedian].
+type MedianService struct {
+	*pluginService[*GRPCPluginMedian, PluginMedian]
```
krehermann, line 19: does `Plugin` mean the same thing in these two names? are they both loops or one a loop and the other an ocr?

pkg/loop/median_service.go
```diff
@@ -0,0 +1,42 @@
...
+			return nil, fmt.Errorf("expected PluginMedian but got %T", instance)
+		}
+		return plug, nil
+	}
+	stopCh := make(chan struct{})
+	lggr = logger.Named(lggr, "MedianService")
+	return &MedianService{newPluginService(PluginMedianName, &GRPCPluginMedian{StopCh: stopCh, Logger: lggr}, newService, lggr, cmd, stopCh)}
+}
+
+func (m *MedianService) NewMedianPluginFactory(ctx context.Context, provider types.MedianProvider, dataSource, juelsPerFeeCoin median.DataSource, errorLog ErrorLog) (ocrtypes.ReportingPluginFactory, error) {
```
krehermann, line 37: can we rename this to MedianOCRReportingFactory?

it think it will help with persistent context confusion about the word `plugin`

pkg/loop/median_service.go
jmank88: I'm not sure that I understand the question. The median loop plugin uses the same interfaces defined by libocr. The two types passed here are for the go-plugin type and our own custom type.

pkg/loop/median_service.go
jmank88: I'm sympathetic to the name overloading, but I think that this would introduce a confusing asymmetry. What about one of these?
```suggestion
func (m *MedianService) NewMedianReportingPluginFactory(ctx context.Context, provider types.MedianProvider, dataSource, juelsPerFeeCoin median.DataSource, errorLog ErrorLog) (ocrtypes.ReportingPluginFactory, error) {
```
```suggestion
func (m *MedianService) NewMedianFactory(ctx context.Context, provider types.MedianProvider, dataSource, juelsPerFeeCoin median.DataSource, errorLog ErrorLog) (ocrtypes.ReportingPluginFactory, error) {
```
FWIW the underlying impl is named `NumericalMedianFactory`.

pkg/loop/plugin_median.go
jmank88: Can you be more specific? Giving up language features or polluting type names don't sound like worthwhile tradeoffs. I could update the `newClientConn` call to be explicit about the field name, but then we're undoing language features to dispel incorrect assumptions that a reader might have. Are there more cases?

pkg/loop/plugin_median.go
jmank88: Can you elaborate on this? I'm not following what is circular. We have a custom type which embeds two other types to inherit their methods, and also implements an interface using the `pb` client.

pkg/loop/plugin_service.go
jmank88: It may be the only goroutine, but that may not always be the case, and this is a general, standard pattern used throughout our codebase, and go code in general. We could have `keepAlive` do the spawning, but I'm not sure that is better:
```go
func (r *pluginService[P, S]) keepAlive() {
	r.wg.Add(1)
	go func() {
		defer r.wg.Done()
		...
	}()
}
```
Because then we're hiding the `go` keyword, which seems like a worse version of the original issue.

Is this better?
```suggestion
		go func() {
			defer r.wg.Done()
			r.keepAlive()
		}()
```

pkg/loop/plugin_median.go
```diff
@@ -37,116 +39,111 @@ func PluginMedianHandshakeConfig() plugin.HandshakeConfig {
...
-}
-
-type grpcPluginMedian struct {
+type GRPCPluginMedian struct {
 	plugin.NetRPCUnsupportedPlugin
 
-	impl PluginMedian
-	lggr logger.Logger
-}
+	StopCh <-chan struct{}
```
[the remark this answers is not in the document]
jmank88: Hmm this is just a case of not requiring a constructor, nothing unusual about StopCh. Maybe we should just switch back to constructors?

pkg/loop/plugin_median.go
jmank88: I'll look at these again, but I'm afraid things will unwravel if we switch one or the other - they aren't arbitrary and we don't control all the code that we connect to.

pkg/loop/lggr_broker.go
jmank88: Renamed :shrug: 
```go
// brokerExt extends a broker with various helper methods.
type brokerExt struct {
```

pkg/loop/plugin_median.go
jmank88: I'm actually not sure what you mean here. This is not a service _for_ median plugins, per se - the plugin _is_ a service. Is it more clear with s/`NewMedianPluginFactory`/`NewMedianFactory`?

pkg/loop/median_service.go
jmank88: Updated to `NewMedianFactory`

pkg/loop/median_service.go
krehermann: we use plugin to mean libocr plugin and loop plugin.

it's super hard to keep these different meaning straight. what would it looks to stop using the word Plugin in new dev? could use something like Loop/Loopp/Loopper and OCRReporter

pkg/loop/plugin_median.go
krehermann: yeah better

pkg/loop/median_service.go
jmank88: I don't think that playing with those terms simplifies the problem in this case. We are already stuck with hashicorp/go-plugin and libocr ReportingPlugin which are inseparable. packages should help, but we're always going to have code that wraps and implements on both sides.

krehermann (approved):

PR #112, opened 2023-05-18
pkg/loop: improve docs; align service method signatures

jmank88:
Misc doc improvements, and while looking at the `godoc`s I noticed that we had mixed method receivers due to embedding a pointer. Updated to embed a value, and switched the constructor to an `init` method.
Screenshot from 2023-05-18 08-51-37
Screenshot from 2023-05-18 08-59-18


aalu1418 (approved):
lgtm!

krehermann (approved):

PR #113, opened 2023-05-23
Add integration with Sonar

chudilka1:


.github/workflows/pkg.yml
```diff
@@ -33,15 +42,86 @@ jobs:
...
         with:
           go-version-file: 'go.mod'
-      - name: build
+
+      - name: Build
         run: go build -v ./...
-      - name: test
+
+      - name: Unit Tests
+        run: go test -v ./... -coverpkg=./... -coverprofile=coverage.txt $1 | tee $OUTPUT_FILE
```
jmank88, line 72: Is `-v` required? It spams all of the logs from every test instead of just failures, and requires devs to use tooling to find the actual failures among the noise.
```suggestion
        run: go test ./... -coverpkg=./... -coverprofile=coverage.txt $1 | tee $OUTPUT_FILE
```

.github/workflows/pkg.yml
```diff
@@ -33,15 +42,86 @@ jobs:
...
+      - name: Upload Go test results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: go-test-results
+          path: |
+            ./output.txt
+            ./coverage.txt
+  
+  race-tests:
```
jmank88, line 83: Why is this separate? We don't get any go build caching this way.

.github/workflows/pkg.yml
```diff
@@ -33,15 +42,86 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+      
+      - name: Race tests
         run: go test -count=1 -race ./...
```
aalu1418, line 95: side note: `-count=1` is that what we use for other repos? can we bump this to 10

aalu1418 (commented):
just one nit + I agree with what Jordan said, otherwise it looks good

.github/workflows/pkg.yml
chudilka1: No, it is not required. Thanks, `-v` removed.

.github/workflows/pkg.yml
chudilka1: Moved the race flag under unit tests.

.github/workflows/pkg.yml
chudilka1: Thanks, removed the flag

.github/workflows/pkg.yml
```diff
@@ -33,15 +42,72 @@ jobs:
...
-      - name: build
+
+      - name: Build
         run: go build -v ./...
-      - name: test
-        run: go test -count=1 -race ./...
-        
+
+      - name: Unit Tests
+        run: go test -race ./... -coverpkg=./... -coverprofile=coverage.txt $1 | tee $OUTPUT_FILE
```
jmank88, line 72: @chudilka1 this is not working properly. Failing tests and detected races no longer fail the workflow. For example: https://github.com/smartcontractkit/chainlink-relay/actions/runs/5283496635/jobs/9559819542

PR #119, opened 2023-05-30
Starknet signature adapter logic

krehermann:
Adds a starknet-specific signature wrapper that is intended to be used with the keystore abstraction


pkg/loop/adapters/starknet/signature.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb"
+)
+
+const (
+	maxPointByteLen = 32 // stark curve max is 252 bits
+	signatureLen    = 2 * maxPointByteLen
+)
+
+type Signature struct {
```
jmank88, line 16: Doesn't the `StarknetSignature` message supercede this?

jmank88:
Moved to #121

PR #120, opened 2023-05-30
pkg/loop: grpc prometheus metrics

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2264

Supporting: 
- https://github.com/smartcontractkit/chainlink/pull/9382


pkg/loop/internal/broker.go
```diff
@@ -34,28 +34,39 @@ func (a *atomicBroker) Accept(id uint32) (net.Listener, error) {
...
 }
 
 func (a *atomicBroker) NextId() uint32 {
 	return a.load().NextId()
 }
 
+// Telemetry has grpc client and server options.
+type Telemetry struct {
+	DialOpts  []grpc.DialOption
+	NewServer func([]grpc.ServerOption) *grpc.Server // optional
```
krehermann, line 48: can you add a comment for the use case of the optional NewServer func? why would i want to provide it?

pkg/loop/telem.go
```diff
@@ -0,0 +1,85 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc"
+)
+
+type Telemetry = internal.Telemetry
+
+// SetupTelemetry initializes open telemetry and returns a prometheus [http.Handler], client side [grpc.DialOption]s, and a
+// [*grpc.Server] constructor that includes interceptor options.
+func SetupTelemetry(registerer prometheus.Registerer) Telemetry {
```
krehermann, line 21: Is this more specific than Telemetry? maybe setup GRPCTelemetry? or SetupInterceptor and return a TelemetryInterceptor?

reading the code in the other pr, from the other repo, the scope of the `Telemetry` was not clear.

pkg/loop/internal/broker.go
```diff
@@ -34,28 +34,39 @@ func (a *atomicBroker) Accept(id uint32) (net.Listener, error) {
 	return a.load().Accept(id)
 }
 
-func (a *atomicBroker) Dial(id uint32) (conn *grpc.ClientConn, err error) {
-	return a.load().Dial(id)
+func (a *atomicBroker) DialWithOptions(id uint32, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
```
krehermann, line 37: can you explain to me how the plumb of the options is working here?

i expected you to need an upstream change to hashi corp, but i don't see that

pkg/loop/telem.go
jmank88: I had something GRPC specific, but was trying to set this up in a forward compatible way, so that we don't have to break the API as we add things later.

pkg/loop/internal/broker.go
krehermann: i see it now. i missed the replace on first reading

jmank88: ah shoot it does disrupt the /integration-tests modules too

pkg/loop/telem.go
jmank88: So I think this is odd because we are able to use the same type in two different contexts currently (telem and broker), because the only telemetry things are grpc options right now. So the question is whether in the future if/when we return more telemetry things (e.g. i was returning prometheus stuff from here - note the stale comment :facepalm: ), if those things will also be passed in to the `Broker`, or if we will only pass along the grpc options. :thinking:

jmank88: I think what I really wanted was:
```go
type Telemetry struct {
	GRPCOpts GRPCOpts
}
```
And it seemed weird to have a struct with a single field. It's probably too soon to be fretting over API compatibility anyways - I'll reconsider just `type GRPCOpts`... :thinking:

PR #121, opened 2023-05-30
Starknet Signature Adapter

krehermann:
Adds a starknet-specific signature wrapper that is intended to be used
with the keystore abstraction

NOTE: this is branch that works around the dependency cycle we are currently experiencing.


jmank88 (approved):

jmank88:
Updated for merge, now that we are moving through this transition.

jmank88 (approved):

PR #124, opened 2023-06-09
Add order field to NodeStatus

george-dorin:


pkg/types/chain.go
```diff
@@ -35,4 +35,5 @@ type NodeStatus struct {
 	Name    string
 	Config  string // TOML
 	State   string
+	Order   *int32
```
jmank88, line 38: Does this already come through in the `Config` TOML field?
This is used for all chain families, so we're trying to keep the shared field to a minimum. `Name` was required IIRC and `State` is mutable.

george-dorin: When fetching information on the node for the `cli` we hit the [Index function](https://github.com/smartcontractkit/chainlink/blob/develop/core/web/nodes_controller.go#L41) and then get the information from [NodeStatuses](https://github.com/smartcontractkit/chainlink/blob/develop/core/chains/chain_set.go#L33) so there isn't a lot of flexibility 

I have been looking into alternative ways to do this but I couldn't find anything worthwhile other than extending the [nodesController](https://github.com/smartcontractkit/chainlink/blob/develop/core/web/nodes_controller.go#L20) both of these options seem undesirable tbh. WDYT?

jmank88: What do you mean there isn't flexibility?

I mean that if `Order` is already a part of the TOML data included in `Config`, then we don't need it in another top level field because it would be redundant. Is there some other reason that this needs to be a top level field?

jmank88: (notice we do not include `HTTPURL`, `WSURL`, or `SendOnly` here)

george-dorin: Yeah, I wanted the Order field to be explicit, even when it's not specified in the Config. Unfortunately, the ways in which this can be achieved seem undesirable(specifically because this only affects EVM nodes, and the `NodeStatus` is shared across all node types). I probably will drop this and leave it as is.

PR #126, opened 2023-06-11
pkg/loop: hclog interopability

jmank88:
This PR updates various logging code to support inter-operating with `package hclog`, in order to allow loop logs to be interpreted and forwarded properly by hashicorp/go-plugin.
 - added `loop.NewLogger()` to create a `logger.Logger` which write hclog fields names
 - added `loop.HCLogLogger()` to create an `hclog.Logger` which is backed by one of our own `Logger`s
 - move `NewOCRWrapper` from core
 - add trace and critical helpers to support migrating plugins away from `core/logger.Logger`

Supports:
- https://github.com/smartcontractkit/chainlink/pull/9551

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -49,8 +49,15 @@ func New() (Logger, error) { return defaultConfig.New() }
 
 // New returns a new Logger for Config.
 func (c *Config) New() (Logger, error) {
+	return NewWith(func(cfg *zap.Config) {
+		cfg.Level.SetLevel(c.Level)
+	})
+}
+
+// NewWith returns a new Logger from a modified [zap.Config].
+func NewWith(cfgFn func(*zap.Config)) (Logger, error) {
 	cfg := zap.NewProductionConfig()
-	cfg.Level.SetLevel(c.Level)
+	cfgFn(&cfg)
 	core, err := cfg.Build()
 	if err != nil {
 		return nil, err
@@ -109,7 +116,15 @@ func (l *logger) Name() string {
 	return l.name
 }
 
-// With returns a Logger with keyvals, if l has a method `With(...interface{}) L`, where L implements Logger, otherwise it returns l.
+func (l *logger) helper(skip int) Logger {
+	return &logger{l.sugaredHelper(skip), l.name}
+}
+
+func (l *logger) sugaredHelper(skip int) *zap.SugaredLogger {
+	return l.SugaredLogger.WithOptions(zap.AddCallerSkip(skip))
+}
+
+// With returns a Logger with keyvals, if 'l' has a method `With(...interface{}) L`, where L implements Logger, otherwise it returns l.
 func With(l Logger, keyvals ...interface{}) Logger {
 	switch t := l.(type) {
 	case *logger:
@@ -138,7 +153,7 @@ func With(l Logger, keyvals ...interface{}) Logger {
 	return w
 }
 
-// Named return a logger with name `n“, if l has a method `Named(name string) L`, where L implements Logger, otherwise it returns l.
+// Named returns a logger with name 'n', if 'l' has a method `Named(string) L`, where L implements Logger, otherwise it returns l.
 func Named(l Logger, n string) Logger {
 	switch t := l.(type) {
 	case *logger:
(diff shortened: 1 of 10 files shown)


pkg/logger/logger.go
```diff
@@ -51,6 +51,10 @@ func New() (Logger, error) { return defaultConfig.New() }
 func (c *Config) New() (Logger, error) {
 	cfg := zap.NewProductionConfig()
 	cfg.Level.SetLevel(c.Level)
+	return NewFromZap(cfg)
+}
+
+func NewFromZap(cfg zap.Config) (Logger, error) {
```
jmank88, line 57: Another option is a func to modify the config:
```suggestion
func NewFromZap(cfgFn func(*zap.Config)) (Logger, error) {
```
This way the default can continue to evolve and things will stay in sync without the caller updating on their end... :thinking:

pkg/loop/logger.go
```diff
@@ -0,0 +1,53 @@
...
+
+var _ hclog.SinkAdapter = (*hclSinkAdapter)(nil)
+
+// hclSinkAdapter implements [hclog.SinkAdapter] with a [logger.Logger].
+type hclSinkAdapter struct {
+	l logger.Logger
+}
+
+func (h *hclSinkAdapter) Accept(name string, level hclog.Level, msg string, args ...interface{}) {
+	l := logger.Named(h.l, name) //opt: cache these
```
jmank88, line 30: This is probably worth doing... :thinking:

pkg/logger/logger.go
jmank88: Done

pkg/loop/logger.go
jmank88: Done

pkg/logger/logger.go
```diff
@@ -164,3 +179,81 @@ func Named(l Logger, n string) Logger {
...
+		return t.helper(skip)
+	}
+	v := reflect.ValueOf(l)
+	m := v.MethodByName("Helper")
+	if m == (reflect.Value{}) {
+		// not available
+		return l
+	}
+
+	r := m.Call([]reflect.Value{reflect.ValueOf(skip)})
```
krehermann, line 196: can you add a comment about what this line is doing? what is the meaning of r?

pkg/logger/logger.go
```diff
@@ -164,3 +179,81 @@ func Named(l Logger, n string) Logger {
 	}
 	return ret
 }
+
+// Helper returns a logger 'skip' levels of callers skipped, if 'l' has a method `Helper(int) L`, where L implements Logger, otherwise it returns l.
+func Helper(l Logger, skip int) Logger {
```
krehermann, line 184: is there a more descriptive name for this? I would almost never expect Helper to be a public api.

is Helper zap lingo? What is a Helper method?

pkg/loop/logger.go
```diff
@@ -0,0 +1,95 @@
...
+	hcl.RegisterSink(&hclSinkAdapter{l: l})
+	return hcl
+}
+
+var _ hclog.SinkAdapter = (*hclSinkAdapter)(nil)
+
+// hclSinkAdapter implements [hclog.SinkAdapter] with a [logger.Logger].
+type hclSinkAdapter struct {
+	l logger.Logger
+	m sync.Map // [string]func() l.Logger
```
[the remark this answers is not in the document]
jmank88: That was meant to document the types used

pkg/loop/logger.go
```diff
@@ -0,0 +1,95 @@
...
+	onceVal := onceValue(func() logger.Logger {
+		return logger.Named(h.l, name)
+	})
+	v, _ := h.m.LoadOrStore(name, onceVal)
+	return v.(func() logger.Logger)()
+}
+
+func (h *hclSinkAdapter) Accept(name string, level hclog.Level, msg string, args ...interface{}) {
+	l := h.named(name)
+	switch level {
```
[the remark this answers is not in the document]
jmank88: It should just no-op for unknown levels, but this explicitly lists everything that hclog supports.
What could default do? Seems potentially spammy to reroute to another level.

pkg/logger/logger.go
jmank88: Zap calls it 'caller skip' but it was confusing people so we settled on `Helper` a long time ago, since it's a helper for writing wrapping logger and utility functions :shrug: 
https://github.com/smartcontractkit/chainlink/blob/72ae3be34217d43c413b70c65d17a690994aef9e/core/logger/logger.go#L108-L110
In any case, we are just capturing and adapting the core logger here, and I would like to avoid exporting Helper from the relay logger.

pkg/logger/ocr.go
```diff
@@ -0,0 +1,55 @@
...
+
+type ocrWrapper struct {
+	l         Logger
+	saveError func(string)
+}
+
+// NewOCRWrapper returns a new [ocrtypes.Logger] backed by the given Logger.
+func NewOCRWrapper(l Logger, saveError func(string)) ocrtypes.Logger {
+	return &ocrWrapper{
+		l:         Helper(l, 2),
```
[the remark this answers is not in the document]
jmank88: I expected 1, but this is copied from core. I'll investigate.

pkg/logger/trace.go
```diff
@@ -0,0 +1,29 @@
...
+// Tracef emits trace level logs, which are debug level with a '[trace]' prefix.
+func Tracef(l Logger, format string, values ...interface{}) {
+	t, ok := l.(interface {
+		Tracef(string, ...interface{})
+	})
+	if ok {
+		t.Tracef(msg, keysAndValues...)
+		return
+	}
+	l.Helper(1).Debugf(tracePrefix+format, values...)
```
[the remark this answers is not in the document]
jmank88: 1 is the typical use, since we need to skip one additional level in the stack when using a helper to log a line.

pkg/logger/logger.go
jmank88: zap doesn't have a method for us to add to our interface. It requires calling `WithOptions`:
`l.SugaredLogger.WithOptions(zap.AddCallerSkip(skip))`

We could rename this func to something more intuitive, but I worry about breaking the symmetry.

pkg/logger/logger.go
```diff
@@ -164,3 +179,81 @@ func Named(l Logger, n string) Logger {
...
+
+	// return is not a Logger
+	if !ok {
+		return l
+	}
+	return ret
+}
+
+// Critical emits critical level logs (a remapping of [zap.DPanicLevel]) or falls back to error level with a '[crit]' prefix.
+func Critical(l Logger, args ...interface{}) {
```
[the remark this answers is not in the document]
jmank88: Added some unit tests for each case.

pkg/logger/logger.go
jmank88: Added a link to the zap method at least.

pkg/loop/logger.go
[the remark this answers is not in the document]
krehermann: i don't necessarily think default is a good option here, instead i wanted to confirm that considered all the cases.

PR #127, opened 2023-06-13
pkg/loop: conditional ContextValues

jmank88:


jmank88:
@bolekk WDYT about this?

bolekk:
It makes sense on the surface but is there a risk that some product (Mercury?) depends on those values being set no matter what? I'll try to take a deeper look to understand that.

bolekk (approved):
OK what I wrote earlier doesn't really matter because whoever needs those values will set it, and we can unset for other plugins. LGTM

PR #130, opened 2023-06-14
Fix rare error cases with block number consensus

samsondav:


pkg/reportingplugins/mercury/aggregate_functions.go
```diff
@@ -85,25 +85,47 @@ func GetConsensusCurrentBlock(paos []ParsedAttributedObservation, f int) (hash [
...
+		if blocks[i].num == blocks[j].num && blocks[i].ts == blocks[j].ts {
+			// tie-break on hash, all else being equal
+			return blocks[i].hash < blocks[j].hash
+		} else if blocks[i].num == blocks[j].num {
+			// if block number is equal and timestamps differ, take the latest timestamp
+			return blocks[i].ts > blocks[j].ts
+		} else {
+			// if block number is different, take the higher block number
+			return blocks[i].num > blocks[j].num
+		}
```
jmank88, line 125: This might read a little easier if the comparison is encapsulated in a method on `block`:
```suggestion
		return blocks[i],less(blocks[j])
```
```go
func (b block) less(b2 block) bool {
	if b.num == b2.num && b.ts == b2.ts {
		// tie-break on hash, all else being equal
		return b.hash < b2.hash
	} else ...
}
```

jmank88 (approved):

PR #132, opened 2023-06-20
Change FetchInitialMaxFinalizedBlockNumber interface

samsondav:


pkg/reportingplugins/mercury/mercury.go
```diff
@@ -181,7 +181,7 @@ type ReportCodec interface {
 type Fetcher interface {
 	// FetchInitialMaxFinalizedBlockNumber should fetch the initial max
 	// finalized block number from the mercury server.
-	FetchInitialMaxFinalizedBlockNumber(context.Context) (int64, error)
+	FetchInitialMaxFinalizedBlockNumber(context.Context) (*int64, error)
```
jmank88, line 184: nit/ did you consider using `-1` instead of `nil` to avoid the signature change?

jmank88 (approved):

pkg/reportingplugins/mercury/mercury.go
samsondav: Yes but it's even more confusing because -1 is actually a valid value that is used already

PR #134, opened 2023-06-23
Mercury multischema support

sdrug:


pkg/reportingplugins/mercury/aggregate_functions.go
```diff
@@ -56,11 +56,11 @@ func GetConsensusBid(paos []ParsedAttributedObservation, f int) (*big.Int, error
 }
 
 // GetConsensusAsk gets the median ask
-func GetConsensusAsk(paos []ParsedAttributedObservation, f int) (*big.Int, error) {
+func GetConsensusAsk(paos []ParsedObservation, f int) (*big.Int, error) {
 	var validAsks []*big.Int
 	for _, pao := range paos {
-		if pao.PricesValid {
-			validAsks = append(validAsks, pao.Ask)
+		if pao.GetPricesValid() {
```
samsondav, line 62: Since these are methods now there may be a cleaner way to represent this, e.g.

```
ask, valid := pao.GetAsk()
if valid {
  validAsks = append(validAsks, ask)
}
```

pkg/reportingplugins/mercury/onchain_config.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if !(min.Cmp(max) <= 0) {
+		return OnchainConfig{}, pkgerrors.Errorf("OnchainConfig min (%v) should not be greater than max(%v)", min, max)
+	}
+
+	return OnchainConfig{min, max}, nil
+}
+
+func (StandardOnchainConfigCodec) Encode(c OnchainConfig) ([]byte, error) {
```
samsondav, line 63: Might be nice to get some tests on this, perhaps they could be imported from here?

https://github.com/smartcontractkit/offchain-reporting/blob/master/lib/offchainreporting2/reportingplugin/median/median_fuzz_test.go#L16

pkg/reportingplugins/mercury/types.go
```diff
@@ -0,0 +1,53 @@
...
+	GetBid() *big.Int
+	GetAsk() *big.Int
+	GetPricesValid() bool
+	GetCurrentBlockNum() int64
+	GetCurrentBlockHash() []byte
+	GetCurrentBlockTimestamp() uint64
+	GetCurrentBlockValid() bool
+	GetMaxFinalizedBlockNumber() int64
+	GetMaxFinalizedBlockNumberValid() bool
+	GetMaxFinalizedTimestamp() int64
```
samsondav, line 25: As stated above, an alternative implementation might look like:

```
GetCurrentBlock() (num int64, valid bool)
```

Could reduce the number of methods and make it a bit more readable this way. It also makes it possible to change the underlying schema later.

samsondav (commented):
Looking good

pkg/reportingplugins/mercury/v2/mercury_observation.proto
```diff
@@ -0,0 +1,16 @@
...
+message MercuryObservationProto {
+    uint32 timestamp = 1;
+
+    bytes benchmarkPrice = 2;
+    bytes bid = 3;
+    bytes ask = 4;
+    bool pricesValid = 5;
+
+    int64 maxFinalizedTimestamp = 6;
+    bool maxFinalizedTimestampValid = 7;
```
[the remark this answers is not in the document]
sdrug: When the previous report is unknown for example?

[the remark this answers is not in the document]
sdrug: Although if we fall back to the current timestamp when the previous report is unknown, we don't need it really. Dropping for now

pkg/reportingplugins/mercury/aggregate_functions.go
```diff
@@ -73,104 +72,40 @@ func GetConsensusAsk(paos []ParsedAttributedObservation, f int) (*big.Int, error
...
+	return validLinkFees[len(validLinkFees)/2], nil
 }
 
-// GetConsensusMaxFinalizedBlockNum gets the most common (mode)
-// ConsensusMaxFinalizedBlockNum In the event of a tie, the lower number is
-// chosen
-func GetConsensusMaxFinalizedBlockNum(paos []ParsedAttributedObservation, f int) (int64, error) {
-	var validPaos []ParsedAttributedObservation
+// GetConsensusNativeFee gets the median native fee
+func GetConsensusNativeFee(paos []IParsedAttributedObservation, f int) (*big.Int, error) {
```
samsondav, line 95: Should this be moved to `v1/aggregate_functions.go`?

samsondav (changes requested):

pkg/reportingplugins/mercury/aggregate_functions.go
sdrug: This code will be shared between v1 and v2 so I'd keep it there for now

pkg/reportingplugins/mercury/v0/types.go
```diff
@@ -0,0 +1,41 @@
+package mercury_v0
+
+import (
+	"github.com/smartcontractkit/chainlink-relay/pkg/reportingplugins/mercury"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+)
+
+type IParsedAttributedObservation interface {
```
samsondav, line 9: Just personal preference here, but how about calling this simply `ParsedAttributedObservation` and making the concrete struct private, i.e. `parsedAttributedObservation`?

samsondav (changes requested):

pkg/reportingplugins/mercury/v0/types.go
sdrug: Sure we can do it this way. I wanted to differentiate between struct and interface when splitting the functionality but there is no need for that anymore.

pkg/reportingplugins/mercury/offchain_config.go
```diff
@@ -0,0 +1,20 @@
...
+	"encoding/json"
+)
+
+type OffchainConfig struct {
+	ExpirationWindow uint32 // Integer number of seconds
+	BaseUSDFeeCents  uint32
+}
+
+func DecodeOffchainConfig(b []byte) (o OffchainConfig, err error) {
+	// TODO: consider protobuf for better efficiency
```
samsondav, line 13: I would say let's remove this TODO since we are not planning on changing it for now

pkg/reportingplugins/mercury/v0/mercury.go
```diff
@@ -286,31 +156,39 @@ func (rp *reportingPlugin) Observation(ctx context.Context, repts ocrtypes.Repor
 		}
 	}
 
+	// TODO: common code with v1/v2, can we pull it out?
```
samsondav, line 159: Remove

pkg/reportingplugins/mercury/v1/mercury.go
```diff
@@ -0,0 +1,428 @@
...
+		obsErrors = append(obsErrors, bpErr)
+	} else if benchmarkPrice, err := mercury.EncodeValueInt192(obs.BenchmarkPrice.Val); err != nil {
+		bpErr = pkgerrors.Wrap(err, "failed to observe BenchmarkPrice; encoding failed")
+		obsErrors = append(obsErrors, bpErr)
+	} else {
+		p.BenchmarkPrice = benchmarkPrice
+	}
+
+	if obs.Bid.Err != nil {
+		// TODO: Add tests that its invalid if encoding fails on all
```
samsondav, line 156: Did I add these? May need to be copied from v0

pkg/reportingplugins/mercury/v1/mercury.go
```diff
@@ -0,0 +1,428 @@
...
+	// By assumption, we have at most f malicious oracles, so there should be at least f+1 valid paos
+	if !(rp.f+1 <= len(paos)) {
+		return false, nil, pkgerrors.Errorf("only received %v valid attributed observations, but need at least f+1 (%v)", len(paos), rp.f+1)
+	}
+
+	observationTimestamp := mercury.GetConsensusTimestamp(Convert(paos))
+
+	var validFromTimestamp uint32
+	if previousReport != nil {
+		validFromTimestamp, err = rp.reportCodec.ObservationTimestampFromReport(previousReport)
```
samsondav, line 313: Does it not need 1 second added?

samsondav (changes requested):

pkg/reportingplugins/mercury/v1/mercury.go
sdrug: It's all tested in each specific `mercury_test.go`

pkg/reportingplugins/mercury/v1/mercury.go
sdrug: I don't think so, why? We make sure that `validFromTimestamp` is not greater than current `observationTimestamp`

samsondav (dismissed):

samsondav (approved):

samsondav (approved):

PR #141, opened 2023-07-24
[Functions] Extend Relayer to support Functions

bolekk:
Based on https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/types/types.go#L38

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -40,6 +40,8 @@ type Relayer interface {
 	NewConfigProvider(rargs RelayArgs) (ConfigProvider, error)
 	NewMedianProvider(rargs RelayArgs, pargs PluginArgs) (MedianProvider, error)
 	NewMercuryProvider(rargs RelayArgs, pargs PluginArgs) (MercuryProvider, error)
+	// TODO: enable when all dependencies are ready
+	// NewFunctionsProvider(rargs RelayArgs, pargs PluginArgs) (FunctionsProvider, error)
 }
 
 // The bootstrap jobs only watch config.
@@ -75,4 +77,34 @@ type MercuryProvider interface {
 	ReportCodec() mercury.ReportCodec
 	OnchainConfigCodec() mercury.OnchainConfigCodec
 	ContractTransmitter() mercury.Transmitter
+}
+
+type FunctionsProvider interface {
+	PluginProvider
+	FunctionsEvents() FunctionsEvents
+}
+
+type OracleRequest struct {
+	RequestID           [32]byte
+	RequestingContract  ocrtypes.Account
+	RequestInitiator    ocrtypes.Account
+	SubscriptionId      uint64
+	SubscriptionOwner   ocrtypes.Account
+	Data                []byte
+	DataVersion         uint16
+	Flags               [32]byte
+	CallbackGasLimit    uint64
+	TxHash              []byte
+	CoordinatorContract ocrtypes.Account
+	OnchainMetadata     []byte
+}
+
+type OracleResponse struct {
+	RequestID [32]byte
+}
+
+// An on-chain event source, which understands router proxy contracts.
+type FunctionsEvents interface {
+	Service
+	LatestEvents() ([]OracleRequest, []OracleResponse, error)
 }


bolekk:
@jmank88 Do I need to first implement it in all dependent repos?

jmank88:
> Do I need to first implement it in all dependent repos?

Yes, you will have to add stubs in -solana, -cosmos, and -starknet just like you did for `package loop`.

pkg/types/types.go
```diff
@@ -76,3 +78,36 @@ type MercuryProvider interface {
...
 	ContractTransmitter() mercury.Transmitter
 }
+
+type FunctionsProvider interface {
+	PluginProvider
+	LogPollerWrapper() LogPollerWrapper
+	NodeSignerKey() *ecdsa.PrivateKey
+}
+
+type Address [20]byte
```
jmank88, line 88: Are you able to use the `libocr.Account` type instead? https://github.com/smartcontractkit/libocr/blob/master/offchainreporting2plus/types/types.go#L297-L298

pkg/types/types.go
```diff
@@ -76,3 +78,36 @@ type MercuryProvider interface {
...
+
+type OracleRequest struct {
+	RequestId           [32]byte
```
jmank88, line 91:
```suggestion
	RequestID           [32]byte
```

pkg/types/types.go
```diff
@@ -76,3 +78,36 @@ type MercuryProvider interface {
 	OnchainConfigCodec() mercury.OnchainConfigCodec
 	ContractTransmitter() mercury.Transmitter
 }
+
+type FunctionsProvider interface {
+	PluginProvider
+	LogPollerWrapper() LogPollerWrapper
+	NodeSignerKey() *ecdsa.PrivateKey
```
jmank88, line 85: How is this key used? We have been trying to avoid passing keys across this API, and I'm surprised to see a key owned by the relayer :thinking: - the normal keystore lives in core and we just provide a signing API.

pkg/types/types.go
bolekk: boo, I wanted a fixed size byte array but ok...

pkg/types/types.go
bolekk: Functions needs a signer key for Gateway communication. It has a node address in the config and then finds the corresponding key in keystore.

Can this exist in the delegate? https://github.com/smartcontractkit/chainlink/blob/develop/core/services/ocr2/delegate.go#L80

It's evm-specific so I assumed it needs to go through relay.

jmank88: We pass a keystore handle to the relayers when we construct them - we can do the same with the functions plugin.
Must you have access to the raw key? Or could you work with some kind of signing API? We have this currently for signing txes:
https://github.com/smartcontractkit/chainlink-relay/blob/main/pkg/loop/internal/relayer.go#L18-L22
```go
type Keystore interface {
	Accounts(ctx context.Context) (accounts []string, err error)
	// Sign returns data signed by account.
	// nil data can be used as a no-op to check for account existence.
	Sign(ctx context.Context, account string, data []byte) (signed []byte, err error)
}
```
And we can pack structured data in to the bytes as necessary. (e.g. starknet actually wants some big.Ints to be returned)

jmank88: If you do need the private key, we could also just pass it directly when we construct the plugin (constructor arg or env var?) rather than expand this API that sort of serves a different purpose.

pkg/types/types.go
jmank88: This is sort of a chain agnostic gray area - we must avoid chain-specific types/constraints, but we can have chain-specific data and encodings, just as long as we don't make any assumptions like by trying to parse them. Broadly speaking, we can pass data around from the config and relayer, but can't really interact with it other than simple things like raw comparisons.

pkg/loop/internal/relayer.go
```diff
@@ -168,6 +168,7 @@ type Relayer interface {
 	NewConfigProvider(context.Context, types.RelayArgs) (types.ConfigProvider, error)
 	NewMedianProvider(context.Context, types.RelayArgs, types.PluginArgs) (types.MedianProvider, error)
 	NewMercuryProvider(context.Context, types.RelayArgs, types.PluginArgs) (types.MercuryProvider, error)
+	NewFunctionsProvider(context.Context, types.RelayArgs, types.PluginArgs) (types.FunctionsProvider, error)
```
aalu1418, line 171: 🥳 

+1 to jordan's note on this will need to implemented into all the chain-specific repos otherwise core builds will fail. let me know if you need help with getting those PRs through

pkg/types/types.go
aalu1418: Ideally we use the interface? it feels more future proof than passing the private key around - especially for loop work.

we've had annoying issues with circular imports that were solved once we put things behind a signer interface (won't happen here cuz ecdsa is a 3rd party dependency, but custom key definitions in core imported into relayers have come to sting us a few times 🐝)

jmank88: Excellent!

pkg/types/types.go
```diff
@@ -76,3 +78,34 @@ type MercuryProvider interface {
 	OnchainConfigCodec() mercury.OnchainConfigCodec
 	ContractTransmitter() mercury.Transmitter
 }
+
+type FunctionsProvider interface {
+	PluginProvider
+	LogPollerWrapper() LogPollerWrapper
+	Sign(data ...[]byte) ([]byte, error)
```
jmank88, line 85: We can pass the Keystore served by core in to the Functions ReportingPlugin(Factory) - does that work? Then it can be left out here.

bolekk: That's what we're doing right now. I might've misunderstood your earlier comment. If the plan is to still keep all chain-specific keys in the key store but abstract them behind a chain-agnostic interface then that works too. I'll remove the sign() call from here, thanks.

pkg/types/types.go
```diff
@@ -76,3 +78,33 @@ type MercuryProvider interface {
...
+	CoordinatorContract ocrtypes.Account
+	OnchainMetadata     []byte
+}
+
+type OracleResponse struct {
+	RequestID [32]byte
+}
+
+// A LogPoller wrapper that understands router proxy contracts
+type LogPollerWrapper interface {
```
jmank88, line 107: WDYT about calling this something like `OracleEvents` or `FunctionsContract`?
```suggestion
type OracleEvents interface {
```
```suggestion
type FunctionsContract interface {
```

bolekk: You're right that LogPollerWrapper is a bad name but I also think it might do more things than just supplying Oracle Events in the future, let me thing about it.

bolekk: Maybe FunctionsEvents...

jmank88 (approved):

PR #144, opened 2023-07-31
pkg/loop: grpc proxy MedianProvider connections

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2364

### DIFF
--- a/pkg/loop/internal/broker.go
+++ b/pkg/loop/internal/broker.go
@@ -91,7 +91,18 @@ func (b *brokerExt) dial(id uint32) (conn *grpc.ClientConn, err error) {
 	return b.broker.DialWithOptions(id, b.DialOpts...)
 }
 
-func (b *brokerExt) serve(name string, register func(*grpc.Server), deps ...resource) (uint32, resource, error) {
+func (b *brokerExt) serveNew(name string, register func(*grpc.Server), deps ...resource) (uint32, resource, error) {
+	var server *grpc.Server
+	if b.NewServer == nil {
+		server = grpc.NewServer()
+	} else {
+		server = b.NewServer(nil)
+	}
+	register(server)
+	return b.serve(name, server, deps...)
+}
+
+func (b *brokerExt) serve(name string, server *grpc.Server, deps ...resource) (uint32, resource, error) {
 	id := b.broker.NextId()
 	b.Logger.Debugf("Serving %s on connection %d", name, id)
 	lis, err := b.broker.Accept(id)
@@ -100,13 +111,6 @@ func (b *brokerExt) serve(name string, register func(*grpc.Server), deps ...reso
 		return 0, resource{}, ErrConnAccept{Name: name, ID: id, Err: err}
 	}
 
-	var server *grpc.Server
-	if b.NewServer == nil {
-		server = grpc.NewServer()
-	} else {
-		server = b.NewServer(nil)
-	}
-	register(server)
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
--- a/pkg/loop/internal/median.go
+++ b/pkg/loop/internal/median.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"time"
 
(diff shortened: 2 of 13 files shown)


pkg/loop/internal/broker.go
```diff
@@ -99,13 +110,6 @@ func (b *brokerExt) serve(name string, register func(*grpc.Server), deps ...reso
...
-	var server *grpc.Server
-	if b.NewServer == nil {
-		server = grpc.NewServer()
-	} else {
-		server = b.NewServer(nil)
-	}
-	register(server)
 	go func() {
 		defer b.closeAll(deps...)
 		if err := server.Serve(lis); err != nil {
```
patrickhuie19, line 115: is there a potential logical race here wrt logging after tests are completed (l116)?  `resource` output args doesn't wait for the two created goroutines above to close.

pkg/loop/internal/median.go
```diff
@@ -42,39 +43,53 @@ func NewPluginMedianClient(broker Broker, brokerCfg BrokerConfig, conn *grpc.Cli
...
-		})
+		var (
+			providerID  uint32
+			providerRes resource
+		)
+		if grpcProvider, ok := provider.(HasClientConn); ok { //TODO ensure both paths are tested
+			providerID, providerRes, err = m.serve("MedianProvider", proxy.NewProxy(grpcProvider.ClientConn()))
+		} else {
+			providerID, providerRes, err = m.serveNew("MedianProvider", func(s *grpc.Server) {
+				pb.RegisterServiceServer(s, &serviceServer{srv: provider})
```
patrickhuie19, line 78:
```suggestion
		if grpcProvider, ok := provider.(HasClientConn); ok { //TODO ensure both paths are tested
		        //Provider supports proxying. Communication path is plugin server <-> proxy <-> plugin client
			providerID, providerRes, err = m.serve("MedianProvider", proxy.NewProxy(grpcProvider.ClientConn()))
		} else {
		        // The provider does not support proxying. Register services with the GRPC server directly.
                        // In this case, the communication path is plugin server <-> core client <-> core server <-> plugin client
			providerID, providerRes, err = m.serveNew("MedianProvider", func(s *grpc.Server) {
				pb.RegisterServiceServer(s, &serviceServer{srv: provider})
```

Am I understanding correctly that the branch in logic here chooses the communication path?

pkg/loop/internal/broker.go
patrickhuie19: I think we could get an error if the listener is blocked on accepting a connection, but to your point, fairly unlikely.

pkg/loop/internal/median.go
jmank88: Yes but the alternate path doesn't run quite like that, since it is only activated when the relay is running locally in legacy mode.

jmank88: I updated the README with a mermaid sequence diagram - lmk if that helps

pkg/loop/internal/broker.go
jmank88: I think it's worth patching. Do you want to open a PR?

patrickhuie19: Sure - will do!

pkg/loop/internal/median.go
patrickhuie19: Sweet, the diagram definitely helps.

For line 74, In what case does `NewMedianFactory` receive a provider that needs to be wrapped with a grpc proxy vs not?

jmank88: When the relayer is running in a LOOP Plugin (relayer and product LOOPPs can be enabled/disabled independently).

patrickhuie19: Ah - thanks!

patrickhuie19 (dismissed):
LGTM - nice!

jmank88:
Rebased

patrickhuie19 (approved):

PR #146, opened 2023-08-02
[Functions] Missing methods in the internal relayer

bolekk:


jmank88 (approved):

pkg/loop/internal/relayer.go
```diff
@@ -238,6 +239,10 @@ func (r *relayerClient) NewMercuryProvider(context.Context, types.RelayArgs, typ
 	return nil, errors.New("mercury is not supported")
 }
 
+func (r *relayerClient) NewFunctionsProvider(context.Context, types.RelayArgs, types.PluginArgs) (types.FunctionsProvider, error) {
+	return nil, errors.New("functions are not supported")
```
bolekk, line 243: @jmank88 this is erroring for now because I don't have the gRPC translation ready yet, right? But I don't need it for direct calls from core to EVM at the moment, right?

jmank88: Right, you don't need to implement it until later, just get it to compile.

PR #148, opened 2023-08-03
cleaning a potential logical race in broker

patrickhuie19:
Potential logical race. Initially discussed in #144

### DIFF
--- a/pkg/loop/internal/broker.go
+++ b/pkg/loop/internal/broker.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net"
@@ -10,6 +11,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
+	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 )
 
 // Broker is a subset of the methods exported by *plugin.GRPCBroker.
@@ -65,8 +67,8 @@ type brokerExt struct {
 	BrokerConfig
 }
 
-// withNamedLogger returns a new [*brokerExt] with name added to the logger.
-func (b *brokerExt) withNamedLogger(name string) *brokerExt {
+// withName returns a new [*brokerExt] with name added to the logger.
+func (b *brokerExt) withName(name string) *brokerExt {
 	bn := *b
 	bn.Logger = logger.Named(b.Logger, name)
 	return &bn
@@ -75,10 +77,14 @@ func (b *brokerExt) withNamedLogger(name string) *brokerExt {
 // newClientConn return a new *clientConn backed by this *brokerExt.
 func (b *brokerExt) newClientConn(name string, newClient newClientFn) *clientConn {
 	return &clientConn{
-		brokerExt: b.withNamedLogger(name),
+		brokerExt: b.withName(name),
 		newClient: newClient,
 		name:      name,
 	}
+}
+
+func (b *brokerExt) StopCtx() (context.Context, context.CancelFunc) {
+	return utils.ContextFromChan(b.StopCh)
 }
 
 func (b *brokerExt) dial(id uint32) (conn *grpc.ClientConn, err error) {
--- a/pkg/loop/internal/broker.go
+++ b/pkg/loop/internal/broker.go
@@ -83,7 +83,7 @@ func (b *brokerExt) newClientConn(name string, newClient newClientFn) *clientCon
 	}
 }
 
-func (b *brokerExt) StopCtx() (context.Context, context.CancelFunc) {
+func (b *brokerExt) stopCtx() (context.Context, context.CancelFunc) {
 	return utils.ContextFromChan(b.StopCh)
 }
 
--- a/pkg/loop/internal/broker.go
+++ b/pkg/loop/internal/broker.go
@@ -1,7 +1,6 @@
 package internal
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"net"
@@ -11,7 +10,6 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
-	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 )
 
 // Broker is a subset of the methods exported by *plugin.GRPCBroker.
@@ -65,7 +63,6 @@ type BrokerConfig struct {
 type brokerExt struct {
 	broker Broker
 	BrokerConfig
-	wg *sync.WaitGroup
 }
 
 // withNamedLogger returns a new [*brokerExt] with name added to the logger.
@@ -82,10 +79,6 @@ func (b *brokerExt) newClientConn(name string, newClient newClientFn) *clientCon
 		newClient: newClient,
 		name:      name,
 	}
-}
-
-func (b *brokerExt) ctxAndCancelFromStopCh() (context.Context, context.CancelFunc) {
-	return utils.ContextFromChan(b.StopCh)
 }
 
 func (b *brokerExt) dial(id uint32) (conn *grpc.ClientConn, err error) {
@@ -108,9 +101,10 @@ func (b *brokerExt) serve(name string, register func(*grpc.Server), deps ...reso
 		server = b.NewServer(nil)
 	}
 	register(server)
-	b.wg.Add(1)
+	var wg sync.WaitGroup
+	wg.Add(1)
 	go func() {
-		defer b.wg.Done()
+		defer wg.Done()
 		defer b.closeAll(deps...)
 		if err := server.Serve(lis); err != nil {
 			b.Logger.Errorw(fmt.Sprintf("Failed to serve %s on connection %d", name, id), "err", err)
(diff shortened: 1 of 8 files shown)


pkg/loop/internal/broker.go
```diff
@@ -106,15 +108,19 @@ func (b *brokerExt) serve(name string, register func(*grpc.Server), deps ...reso
 		server = b.NewServer(nil)
 	}
 	register(server)
+	b.wg.Add(1)
```
jmank88, line 111: Sharing the `WaitGroup` between method calls seems problematic (what prevents `Add` from being called after `Wait`?). Can we just use separate instances instead?
```suggestion
	var wg sync.WaitGroup
	b.wg.Add(1)
```

pkg/loop/internal/config.go
```diff
@@ -62,7 +62,7 @@ func (o *offchainConfigDigesterClient) ConfigDigest(config libocr.ContractConfig
 }
 
 func (o *offchainConfigDigesterClient) ConfigDigestPrefix() (libocr.ConfigDigestPrefix, error) {
-	ctx, cancel := o.ctx()
+	ctx, cancel := o.ctxAndCancelFromStopCh()
```
jmank88, line 65: Is this name change just for readability? It was introduced to be a shorthand form, but with a long name like this it doesn't save much over in-lining:
```suggestion
	ctx, cancel := utils.ContextFromChan(b.StopCh)
```
If you think is important info to surface at callsites for readers, then I would favor inlining which seems easier to read.

pkg/loop/internal/broker.go
patrickhuie19: We could use separate instances, but I think the execution order of `serve` guarantees that `Add` is not being called after `Wait`. `Add` is called when each gr is created (but in the same go routine as `serve`), and `Wait` is only called once `serve` returns. 

I can remove it as a field of `brokerExt` though - it's only essential to `serve`.

jmank88: > We could use separate instances, but I think the execution order of serve guarantees that Add is not being called after Wait. Add is called when each gr is created (but in the same go routine as serve), and Wait is only called once serve returns.

But `serve` can be called multiple times. Isn't it possible for an early `serve` call to close before a later `serve` call is made?

pkg/loop/internal/broker.go
```diff
@@ -66,8 +65,8 @@ type brokerExt struct {
 	BrokerConfig
 }
 
-// named returns a new [*brokerExt] with name added to the logger.
-func (b *brokerExt) named(name string) *brokerExt {
+// withNamedLogger returns a new [*brokerExt] with name added to the logger.
+func (b *brokerExt) withNamedLogger(name string) *brokerExt {
```
jmank88, line 69: Was this confusing before? b/c the fact that only the logger is being named is sort of an implementation detail and left out by design. Would `withName` be more clear?
```suggestion
func (b *brokerExt) withName(name string) *brokerExt {
```

pkg/loop/internal/median.go
```diff
@@ -204,7 +205,7 @@ type reportCodecClient struct {
 }
 
 func (r *reportCodecClient) BuildReport(observations []median.ParsedAttributedObservation) (report libocr.Report, err error) {
-	ctx, cancel := r.ctx()
+	ctx, cancel := utils.ContextFromChan(r.StopCh)
```
jmank88, line 208: These still seem very noisy. What exactly was missing from `ctx()`? Would something like `stopCtx` say enough?
```suggestion
	ctx, cancel := r.stopCtx()
```

pkg/loop/internal/broker.go
patrickhuie19: The previous name made me think we were setting a property `name` of broker ext directly, which seemed sensible and I moved on. I ran into the method later on and was surprised to see it do all this:
```
	bn := *b
	bn.Logger = logger.Named(b.Logger, name)
	return &bn
```

I'd  like the method name to reflect what it's actually doing, which is creating a named logger and setting the new brokerExt with it.

pkg/loop/internal/median.go
patrickhuie19: How's `r.CtxFromStopCh()`? 

I'd like to read it and know it's not creating a new ctx from a parent context associated with the receiver. If we don't expect a lot of fresh eyes on this, though, we can go with the shorthand

pkg/loop/internal/broker.go
patrickhuie19: I've refactored it out of brokerExt, so I think this should be taken care of?

pkg/loop/internal/broker.go
jmank88: > I'd like the method name to reflect what it's actually doing, which is creating a named logger and setting the new brokerExt with it.

But from the caller's perspective it is not necessarily doing that. It is returning a named `brokerExt`, and the implementation details are not important and may change.

pkg/loop/internal/broker.go
patrickhuie19: > But serve can be called multiple times. Isn't it possible for an early serve call to close before a later serve call is made?

As long as serve isn't called concurrently! I think you can reuse a waitgroup once `Wait()` is called and the internal counter hits 0.

```
package main

import "sync"
import "fmt"

type Example struct {
	wg sync.WaitGroup
}

func main() {
	ex := &Example{}

	ex.callTwice()
	ex.callTwice()

}

func (e *Example) callTwice() {
	e.wg.Add(1)
	go func() {
		fmt.Println("Hi")
		defer e.wg.Done()
	}()

	e.wg.Wait()
}
```

pkg/loop/internal/median.go
jmank88: > I'd like to read it and know it's not creating a new ctx from a parent context associated with the receiver.

Why? That is an irrelevant implementation detail. The lifetime of the context is the only relevant information, not the mechanism of implementation.

pkg/loop/internal/broker.go
jmank88: It can be used concurrently and I thought our tests exercised that and would trigger a race. Perhaps with more load.

pkg/loop/internal/median.go
patrickhuie19: > The lifetime of the context is the only relevant information, not the mechanism of implementation.

Do you think the mechanism of implementation = lifetime of the context in this case? When `chStop` receives, the context is canceled. The context is passed through to the 3rd party grpc `Invoke` method, so the mechanism of implementation communicates that the calls to `Invoke` won't block even if stop channel receives. 
```
// ContextFromChan creates a context that finishes when the provided channel
// receives or is closed.
// When channel closes, the ctx.Err() will always be context.Canceled
// NOTE: Spins up a goroutine that exits on cancellation.
// REMEMBER TO CALL CANCEL OTHERWISE IT CAN LEAD TO MEMORY LEAKS
func ContextFromChan(chStop <-chan struct{}) (context.Context, context.CancelFunc) {
	ctx, cancel := context.WithCancel(context.Background())
	go func() {
		select {
		case <-chStop:
			cancel()
		case <-ctx.Done():
		}
	}()
	return ctx, cancel
}
```

Outside of renaming, is there a deeper thread to pull on wrt to creating and injecting this type of context into all grpc methods? i.e. 
```
func (s *serviceClient) Close() error {
	ctx, cancel := utils.ContextFromChan(s.b.StopCh)
	defer cancel()

	_, err := s.grpc.Close(ctx, &emptypb.Empty{})
	return err
}

func (s *serviceClient) Ready() error {
	ctx, cancel := utils.ContextFromChan(s.b.StopCh)
	defer cancel()
	ctx, cancel = context.WithTimeout(ctx, time.Second)
	defer cancel()

	_, err := s.grpc.Ready(ctx, &emptypb.Empty{})
	return err
}
```
Would it be feasible to have such a context initialized as field of `serviceClient`?
```
type serviceClient struct {
	b             *brokerExt
	cc           grpc.ClientConnInterface
	grpc       pb.ServiceClient
	stopCtx context.Context
}

func newServiceClient(b *brokerExt, cc grpc.ClientConnInterface) *serviceClient {
	ctx, _ := utils.ContextFromChan(b.StopCh)
	return &serviceClient{b, cc, pb.NewServiceClient(cc), ctx}
}

func (s *serviceClient) Start(ctx context.Context) error {
	return nil // no-op: server side starts automatically
}

func (s *serviceClient) Close() error {
	ctx, cancel := context.WithCancel(s.stopCtx)
	defer cancel()

	_, err := s.grpc.Close(ctx, &emptypb.Empty{})
	return err
}

func (s *serviceClient) Ready() error {
	ctx, cancel := context.WithCancel(s.stopCtx)
	defer cancel()

	ctx, cancel = context.WithTimeout(ctx, time.Second)
	defer cancel()

	_, err := s.grpc.Ready(ctx, &emptypb.Empty{})
	return err
}
```

jmank88: > Do you think the mechanism of implementation = lifetime of the context in this case?

No, because it could be implemented in other ways while exhibiting the exact same behavior for the caller.

> Would it be feasible to have such a context initialized as field of serviceClient?

We do often store context fields on structs instead, but it is an anti-pattern since contexts are "request" scoped.

jmank88: > Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

https://pkg.go.dev/context

PR #154, opened 2023-08-18
[BCI-1792] Moved over TxAttemptBuilder and Tx interfaces

ExtremelySunnyYK:


go.mod
```diff
@@ -62,6 +65,7 @@ require (
 	golang.org/x/sys v0.8.0 // indirect
 	golang.org/x/text v0.9.0 // indirect
 	google.golang.org/genproto v0.0.0-20230306155012-7f2fa6fef1f4 // indirect
+	gopkg.in/guregu/null.v4 v4.0.0
```
jmank88, line 68: Move up to direct block?

There is a second stray `// indirect` block too.

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/riferrei/srclient v0.5.4
 	github.com/shopspring/decimal v1.3.1
 	github.com/smartcontractkit/libocr v0.0.0-20230802221916-2271752fa829
+	github.com/smartcontractkit/sqlx v1.3.5-0.20210805004948-4be295aacbeb
```
aalu1418, line 19: ideally we don't inject sql dependencies - but it seems unavoidable at the current stage until the persistence layer work is done 🤔

aalu1418: correct me if I'm wrong, but introducing having `pg.Opts` requires importing a lot of extra files/types? (like uint32.go or json.go)

ExtremelySunnyYK: @aalu1418 yup that is right. Introducing `pg.Opts` require us to import extra file types. Do you think we should seperate them instead? i.e generalise pg.Opts

ExtremelySunnyYK:
Comments from slack: https://chainlink-core.slack.com/archives/C04Q61M2MFT/p1692585066729749

We should keep the chainlink-relay service interfaces minimal and hide implementation details from core and products.

Previously this wasn't an issue as we used mostly std lib and generics. But now we are pulling in SQL specific libraries that don't belong in the relayer interface.

To start, we can create a chainlink-relay/pkg/common directory matching chainlink/common and move utils there without modification. This minimally disrupts the current structure.

Longer term, we need to abstract the SQL dependencies out of the chainlink-relay package to avoid leaking details. For example, cleaning up the pg.Opts usage in tx store after separating the persistence layer.

For now, it may be best to keep this PR in a branch until we can properly abstract the SQL dependencies. The goal is to keep the relay interface clean and hidden from core and products.

go.mod
ExtremelySunnyYK: agreed with how we should not inject sql dependencies. This PR is better lived in a branch for now

PR #156, opened 2023-08-21
pkg: move loop interfaces to pkg/types to unify type set

jmank88:
This changes moves the loop types in to the pkg/types package (except for Relayer), in order to unify the type set in one place. It also avoids the internal package aliases (but they are deprecated and left behind for compatibility).

This is not a breaking change. We can follow up later to unify the `types.Relayer` and `loop.Relayer` interfaces, and delete the old aliases.


pkg/types/relayer.go
```diff
@@ -0,0 +1,88 @@
...
+	New           bool // Whether this is a first time job add.
+	RelayConfig   []byte
+}
+
+type PluginRelayer interface {
+	NewRelayer(ctx context.Context, config string, keystore Keystore) (RelayerCtx, error)
+}
+
+// Relayer extends [types.Relayer] and includes [context.Context]s.
+type RelayerCtx interface {
```
krehermann, line 31: I dislike the trailing `Ctx` as a long term API signature.

What would it look like to actually remove the depreciated interface `Relayer` and reuse the name here? Any idea how painful that would be for other repos?

how does this type consolidation relate to the other work we need to do -- consolidating the Providers via the proxy and simplifying the ChainStatus, ChainStatuses, NodeStatuses ... and removing ChainStatuses?

that is does the order matter?

jmank88: > What would it look like to actually remove the depreciated interface Relayer and reuse the name here? Any idea how painful that would be for other repos?

I think it would be an easy superficial change, but would have to be coordinated across all of the repos referencing it. 

> that is does the order matter?

I think order is only important for this small `*Ctx` naming issue. For the other stuff, we're already going to be making breaking signature changes in the same module if we don't move them, so I'm not sure it matters :shrug:

jmank88: Updated to drop legacy Relayer to reuse name.

patrickhuie19 (dismissed):

pkg/types/relayer.go
krehermann: thanks. can we hold off on this change until at least tomorrow? i'm trying to organize my thoughts about how to break down the other work. i'd like to minimize the number of cross-repo changes.

jmank88: Yeah no rush on this, and should set up the other related PRs anyways.

pkg/types/relayer.go
```diff
@@ -0,0 +1,79 @@
...
+	SendTx(ctx context.Context, chainID, from, to string, amount *big.Int, balanceCheck bool) error
+}
+
+type ChainStatus struct {
+	ID      string
+	Enabled bool
+	Config  string // TOML
+}
+
+type NodeStatus struct {
```
krehermann, line 53: can you remind me why this is defined here as opposed to in protobuf? are we shadowing definitions? 

or am i thinking about it in an inverted way? is protobuf a hidden implementation detail of this repo?

jmank88: We haven't had any first class protos so far - they are all mirroring some Go type. We could start using them more directly to save on the conversions though I suppose.

jmank88: Actually, we'll have to do this in order to realize compatibility with GRPC services that add new parameters over time. Otherwise importers will have to react to type changes even when they do not interact with the new fields.

jmank88: FWIW if we settle for RelayerCtx in the short term (or just don't move that one yet?), then this is not a breaking change.

jmank88: Pushed again to leave only the Relayer and PluginRelayer interfaces behind in LOOP (as well as aliases for the others). This should maintain compatability, and allow for gradually updating all the repos to use `package types` instead of the now deprecated aliases.

jmank88: Sweet, that fixed some CI.

ilija42 (approved):

cedric-cordenier (approved):

george-dorin (approved):

patrickhuie19 (approved):

krehermann (approved):

PR #161, opened 2023-08-31
Relayer changes

cedric-cordenier:


pkg/loop/internal/pb/median_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.3.0
-// - protoc             v3.21.12
+// - protoc             v4.24.2
```
jmank88, line 4: @krehermann

pkg/loop/internal/pb/relayer.proto
```diff
@@ -65,38 +64,29 @@ message RelayArgs {
...
   PluginArgs pluginArgs = 2;
 }
 
-// NewMedianProviderReply has return arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.NewMedianProvider].
-message NewMedianProviderReply {
-  uint32 medianProviderID = 1;
+// NewPluginProviderReply has return arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.Relayer.NewPluginProvider].
+message NewPluginProviderReply {
+  uint32 pluginProviderID = 1;
+  string providerType = 2;
```
jmank88, line 79: Is a type parameter necessary? I had imagined that the type would be implicit :thinking:

pkg/types/provider.go
```diff
@@ -2,6 +2,13 @@ package types
 
 import ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 
+const (
+	ProviderTypePlugin    = "plugin"
+	ProviderTypeMedian    = "median"
+	ProviderTypeFunctions = "functions"
+	ProviderTypeMercury   = "mercury"
```
jmank88, line 9: Maybe we should relocate this type? https://github.com/smartcontractkit/chainlink/blob/fbae1afbb38b71b6316460cefffd9ce58e94d354/core/services/job/models.go#L316
Used here: https://github.com/smartcontractkit/chainlink/blob/fbae1afbb38b71b6316460cefffd9ce58e94d354/core/services/job/models.go#L354

pkg/loop/internal/pb/relayer.proto
jmank88: I think it would be OK to omit this if it's always going to be the same as `PluginArgs.providerType`.

pkg/types/provider.go
jmank88: Hmmm I think that makes sense.

Related: Why `PluginArgs.ProviderType` rather than `RelayerArgs.ProviderType`?

jmank88: > It feels to me like they're different, except that they happen to match at the moment.

Is there a chance that this distinction goes away completely before the symmetry is broken though? i.e. couldn't everything just configure an agnostic plugin provider in the future?

cedric-cordenier: >Related: Why PluginArgs.ProviderType rather than RelayerArgs.ProviderType?

I thought it should be PluginArgs since it's the plugin that's requiring a specific providerType. That being said, I looked at the doc comments for both types:
```// PluginArgs are the args required to create any OCR2 plugin components.
// It's possible that the plugin config might actually be different
// per relay type, so we pass the config directly through.```
And that makes me wonder if actually RelayArgs is the right place. Wdyt?

>Is there a chance that this distinction goes away completely before the symmetry is broken though? i.e. couldn't everything just configure an agnostic plugin provider in the future?

Yes that is the ultimate aim, but I think that's very far off since it would require removing product-specific providers altogether. Having a generic pluginType is much more proximate (i.e. should be implemented next quarter).

cedric-cordenier:
Closing this PR in favour of one raised from the main repo as I now have access.

PR #163, opened 2023-09-01
Add make targets for consistent code gen

krehermann:


script/install-protoc.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
krebernisak, line 1: We really need to start using Nix ASAP cc @archseer @aalu1418 @willyrgf @andrevmatos 

@krehermann check out what we're doing here: https://github.com/smartcontractkit/chainlink-project-nix-poc

Also take a look here: https://search.nixos.org/packages?channel=23.05&from=0&size=50&sort=relevance&type=packages&query=protoc

This SDK project wouldn't benefit directly from `chainlink-dev` platform CLI which is designed for platform projects like Ethereum, Solana, Cosmos, and Starknet integrations, but it would benefit from using Nix directly so we supercharge our devenvs and avoid the need to write install scripts like these.

jmank88 (dismissed):

script/install-protoc.sh
krehermann: sure.

 i would like to avoid this kind of script. it's pretty gross.
 
 this change is to unblock @jmank88 @cedric-cordenier and i from cobbling each other's code gen right now.
 
 added a ticket to use nix here https://smartcontract-it.atlassian.net/browse/BCF-2617

pkg/monitoring/source_mock.go
jmank88, line 1: It seems like we can drop the statements that generated this and source_factory_mock.go.

pkg/monitoring/exporter_factory_mock.go
jmank88, line 1: Three more unused mocks here.

pkg/monitoring/source_mock.go
krehermann: done

pkg/monitoring/exporter_factory_mock.go
krehermann: done

pkg/monitoring/source_mock.go
jmank88: Oh wait, did I mis-read these? Maybe they were just out of date with only additions, making them all green.

krehermann: maybe. i assumed you had inside knowledge about how the mocks are used.

`go test` worked after i deleted the `generate.go` but i overlooked rm'ing the generated files themselves.

it's back in place now

jmank88 (approved):

PR #164, opened 2023-09-01
[BCF-2365] Combine providers 

cedric-cordenier:
* Remove NewMercuryProvider, NewMedianProvider, NewFunctionsProvider from the Relayer interface. This is to allow transparent proxying of provider calls from reporting plugin loopps to relayer loopps, without the core node needing to be aware of the types.
* From now on, creating a provider should happen via NewPluginProvider, setting RelayArgs.ProviderType as appropriate.
* LOOPP Relayer implementations can still implement NewMercuryProvider/NewMedianProvider/.... The RelayerServer will check whether the underlying implementation contains these methods and will redirect the call to this more specialized function if that is the case.
* For interoperability with legacy code in the core node which needs to be aware of the product providers, the RelayerClient will instantiate a provider of the right type inside NewPluginProvider, thus allowing calling code to typecast to the right underlying provider type.


cedric-cordenier:
Corresponding chainlink PR:
- https://github.com/smartcontractkit/chainlink/pull/10475

jmank88 (approved):

PR #165, opened 2023-09-04
Make MercuryProvider extend PluginProvider

cedric-cordenier:
We want to replace the `New<Product>Provider` methods in the relayer interface with a single, polymorphic `NewPluginProvider` method which extends a common PluginProvider interface exposing the three OCR components required to set up an Oracle.

At the moment the MedianProvider and FunctionsProvider both extend this interface, but Mercury does not since it extends the OCR ContractTransmitter with a few additional methods.

This modifies Mercury's interface by splitting this transmitter into two separate components, once which implements `ocrtypes.ContractTransmitter` and another which encompasses the additional methods (and implements `mercury.MercuryServerFetcher`). This is a largely cosmetic change, since under the hood the transmitter can remain the same struct, just exposed in two different methods.

### DIFF
--- a/pkg/types/provider_mercury.go
+++ b/pkg/types/provider_mercury.go
@@ -10,10 +10,11 @@ import (
 // MercuryProvider provides components needed for a mercury OCR2 plugin.
 // Mercury requires config tracking but does not transmit on-chain.
 type MercuryProvider interface {
-	ConfigProvider
+	PluginProvider
+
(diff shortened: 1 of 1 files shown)


cedric-cordenier:
@jmank88 Sure! You can find it here: https://github.com/smartcontractkit/chainlink/pull/10467

cedric-cordenier:
Closing this in favour of #164 which folds these changes in.

PR #166, opened 2023-09-05
Change PluginMedian Interface

augustbleeds:
Note: CI tests which build external repos will fail until "replace" is removed [because replace is only respected in the top-level go.mod](https://chainlink-core.slack.com/archives/C03Q4FAGZFC/p1694204165593799)

Interface needs to be modified to include the `gasPrice` data source which tells us how much the gas price is on the chain during a report transmission (for chains which do not support onchain methods of retrieving gas price). It's part of the transmission structure similar to juelsPerFee so it's actually a value which is transmitted to the block-chain as a part of the median plugin

[Example of it on Starknet](https://github.com/smartcontractkit/chainlink-starknet/blob/develop/contracts/src/ocr2/aggregator.cairo#L234)

~For now, I've made LOOPP median plugins have a nil value for the gasPrice data source, so it doesn't actually attempt to start a GRPC server. I think this should be ok for now because ocr will not call the method unless it is non-nil. Eventually, either in this PR or another, we will want to start a grpc service for the gas price data source~ *EDIT*: PR has been updated to serve gas price data source server if it exists

Related PRs:
* [libocr: adding median gasPrice to ocr reports](https://github.com/smartcontractkit/offchain-reporting/pull/412/files)
* [chainlink: core node changes](https://github.com/smartcontractkit/chainlink/pull/10489)

- PR Merge Order (from top to bottom in order)
    1. [libocr](https://github.com/smartcontractkit/offchain-reporting/pull/412/files) 
        - NumericalMedianFactory gets changed to have a gas price data source
        - Note: (If someone bumps libocr in core, median factory will need to be updated with nil)
        - Note: (You will not be able to bump libocr in chainlink-relay until the below PR gets merged)
    2. [chainlink-common](https://github.com/smartcontractkit/chainlink-common/pull/166)
        - Currently uses the replace directive, once libocr is merged we remove replace directive and bump the version of libocr
        - The plugin median interface is updated to include gas price data source
        - Begins the grpc service for gas price data source
        - Note: (You cannot bump chainlink-relay in chainlink repo until the PR below bets merged)
    3. [chainlink-feeds](https://github.com/smartcontractkit/chainlink-feeds/pull/28) 
    4. [chainlink](https://github.com/smartcontractkit/chainlink/pull/10489)
        - Uses replace directive for both chainlink-relay and libocr. They should get updated when the respective repos get updated
        - Parses TOML config for (optional) gas price data source and passes it into median factory.


pkg/types/provider_median.go
```diff
@@ -17,7 +17,7 @@ type MedianProvider interface {
 
 type PluginMedian interface {
 	// NewMedianFactory returns a new ReportingPluginFactory. If provider implements GRPCClientConn, it can be forwarded efficiently via proxy.
-	NewMedianFactory(ctx context.Context, provider MedianProvider, dataSource, juelsPerFeeCoin median.DataSource, errorLog ErrorLog) (ReportingPluginFactory, error)
+	NewMedianFactory(ctx context.Context, provider MedianProvider, dataSource, juelsPerFeeCoin, gasPrice median.DataSource, errorLog ErrorLog) (ReportingPluginFactory, error)
```
jmank88, line 20: What is gas price used for? That seems like a chain-specific detail that shouldn't leak out here.

augustbleeds: [Example of it on Starknet](https://github.com/smartcontractkit/chainlink-starknet/blob/develop/contracts/src/ocr2/aggregator.cairo#L234)

augustbleeds: Not all chains require it (like EVM chains), so if it is nil then ocr will not include it in a transmission

jmank88: Oh, is the point that the actual gas price value for each round has to be agreed on?
Is `Gas` general enough? Or is there a more chain-agnostic term for execution units?

cedric-cordenier: >Not all chains require it (like EVM chains), so if it is nil then ocr will not include it in a transmission

🤔 I wonder if we're operating at the wrong level of abstraction. This suggests that whether or not to include it is chain-specific, so any behaviour related to it should happen in the relayer.

Wdyt about always calling the datasource, but only including it in the report if it's solana? That would allow us to use the ReportEncoder (which is part of the relayer) as the abstraction point.

cedric-cordenier: (I might be jumping the gun btw -- I don't see any code here which actually instantiates a gasPrice datasource and passes it into the median plugin.)

augustbleeds: Yeah, exactly. the gas price is another value that has to be agreed upon in OCR (so it's similar to JuelesPerCoinFee in that regard). 

I've heard the use of the word `fee` being used (during the txm generalization) when referring to the total gas cost for a transaction. But, `gasPrice` in this case means the price of a single unit of gas so "fee" wouldn't quite work

augustbleeds: > > Not all chains require it (like EVM chains), so if it is nil then ocr will not include it in a transmission
> 
> 🤔 I wonder if we're operating at the wrong level of abstraction. This suggests that whether or not to include it is chain-specific, so any behaviour related to it should happen in the relayer.
> 
> Wdyt about always calling the datasource, but only including it in the report if it's solana? That would allow us to use the ReportEncoder (which is part of the relayer) as the abstraction point.

It'd be unnecessary to include a data source that isn't used. That means we'd require node ops to include a data source that did nothing for all evm ocr2 pipelines. 

The issue is precisely that whether or not to include a data source of `gasPrice` depends on the chain.

Yeah, there is no code here that instantiates the gasPrice datasource because I haven't gotten to that yet 😅 . Just wanted to get something working first

jmank88: Do other chain families all use the word `gas` though? I can't actually find any other terminology, so maybe nbd :shrug:

augustbleeds: cosmos and ethereum L2s do. I believe solana operates differently and calls them "tx fees"

go.mod
```diff
@@ -65,6 +65,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+replace github.com/smartcontractkit/libocr => github.com/augustbleeds/libocr v0.0.0-20230901193648-78ff9454fecd
```
augustbleeds, line 68: temp change: will get removed and bumped when libocr gets updated

pkg/loop/internal/broker.go
```diff
@@ -141,8 +141,10 @@ func (b *brokerExt) serve(name string, server *grpc.Server, deps ...resource) (u
 
 func (b *brokerExt) closeAll(deps ...resource) {
 	for _, d := range deps {
-		if err := d.Close(); err != nil {
-			b.Logger.Error(fmt.Sprintf("Error closing %s", d.name), "err", err)
+		if d.Closer != nil {
```
augustbleeds, line 144: handle case when the resource is declared but not uninitialized

pkg/loop/internal/median.go
```diff
@@ -53,6 +53,18 @@ func (m *PluginMedianClient) NewMedianFactory(ctx context.Context, provider type
 		}
 		deps.Add(juelsPerFeeCoinDataSourceRes)
 
+		var gasPriceDataSourceID uint32
+		if gasPrice != nil {
```
augustbleeds, line 57: gasPrice may be nil if none is required for the chain

pkg/types/provider_median.go
cedric-cordenier: >It'd be unnecessary to include a data source that isn't used. That means we'd require node ops to include a data source that did nothing for all evm ocr2 pipelines.

>The issue is precisely that whether or not to include a data source of gasPrice depends on the chain.

Can you explain how the gas price data source is going to be instantiated? We are actively working to put all chain-specific details in the relayer, and I want to make sure that we aren't taking a step back here.

pkg/loop/internal/median.go
```diff
@@ -53,6 +53,18 @@ func (m *PluginMedianClient) NewMedianFactory(ctx context.Context, provider type
 		}
 		deps.Add(juelsPerFeeCoinDataSourceRes)
 
+		var gasPriceDataSourceID uint32
```
cedric-cordenier, line 56: Could this route incorrectly? the ID is 0, so we either need to treat this value differently in the client or signal presence/absence in some other way.

pkg/types/provider_median.go
augustbleeds: The data source is another pipeline so it would be instantiated here https://github.com/smartcontractkit/chainlink/pull/10519/files#diff-7358f3249bf57875121f3bcb6211722cf2bee2f9b411dfff89a1974c48c5b67eR109. 

A chain either has one or it doesn't so it's either nil or it's not. There's https://github.com/smartcontractkit/chainlink-relay/pull/166/files#diff-28ea9d6d17fb7499fbbd0f83828a8335ac28eacff629256d75b261bdb2ee123aR59 also when the loopp plugin serves the underlying data source. Would this be a step backward? 

```
This suggests that whether or not to include it is chain-specific, so any behaviour related to it should happen in the relayer.
```

^ If data sources were to be instantiated in the relayer, that means configuration would also occur in the relayer. The infra doesn't exist, but if it were to exist you'd pass in pipeline configs from core to the relay and relay would instantiate them.

archseer: > Why is a pipeline being used to retrieve this data?

We reimburse `transmit` calls on chain. In order to do that correctly a reimbursement has to be calculated. This is easy in EVM where we can access `tx.gasPrice` in Solidity as well as measure `gasleft()` at the start and end of execution. But in basically every non-EVM chain this data isn't sourceable and we need to do two things:
- come up with reasonable constants to approximate gas use based on lots and lots of transactions
- include a medianized gasPrice in the report (we match the `juelsPerFeeCoin` source usage)

We were able to avoid this second one because:
- on Terra the gas prices were fixed and rarely changed based on some external API price list
- on Solana the TX cost was fixed (for now, but this changes with fee markets)

This `gasPrice` source is needed for both Starknet and Cosmos.

pkg/loop/internal/broker.go
cedric-cordenier: 🤔 This doesn't feel like it should be necessary, and I wonder if this will hide incorrect instantiations in the future. How come you had to add this?

cedric-cordenier: Ah, I guess it's due to this: https://github.com/smartcontractkit/chainlink-relay/pull/166/files#diff-28ea9d6d17fb7499fbbd0f83828a8335ac28eacff629256d75b261bdb2ee123aR147?

cedric-cordenier:
@augustbleeds My feedback from last week still stands: it feels like we're having to special case the gasPricePipeline everywhere because it can be nil everywhere and this is causing a ripple effect of changes throughout the data stack.

Would a null object pattern be appropriate here? i.e. we check if a gas price pipeline is configured on the job spec and if it isn't, pass a noop data source into the loop. Adding this won't introduce any additional overhead since we're already having to stand up a GRPC server for the other dependencies.

The noop data source could return an error code that the median plugin can check to see if it should ignore the results.

I think this will simplify the code overall by allowing you to check for the gas price pipeline in just one place, at instantiation time.

augustbleeds:
> @augustbleeds My feedback from last week still stands: it feels like we're having to special case the gasPricePipeline everywhere because it can be nil everywhere and this is causing a ripple effect of changes throughout the data stack.
> 
> Would a null object pattern be appropriate here? i.e. we check if a gas price pipeline is configured on the job spec and if it isn't, pass a noop data source into the loop. Adding this won't introduce any additional overhead since we're already having to stand up a GRPC server for the other dependencies.
> 
> The noop data source could return an error code that the median plugin can check to see if it should ignore the results.
> 
> I think this will simplify the code overall by allowing you to check for the gas price pipeline in just one place, at instantiation time.

Hey! We de-prioritized this last quarter due to CCIP. We'll be tackling this again this quarter -- going to take another look at it and revisit. Thanks

pkg/loop/internal/datasource.go
```diff
@@ -26,6 +27,10 @@ func (d *dataSourceClient) Observe(ctx context.Context, timestamp types.ReportTi
 	reply, err := d.grpc.Observe(ctx, &pb.ObserveRequest{
 		ReportTimestamp: pbReportTimestamp(timestamp),
 	})
+	// return nil value for NOOP (only affects gas price data source)
+	if errors.Is(err, median.ErrNOOPDataSource) {
```
cedric-cordenier, line 31: @augustbleeds This still feels wrong: this `dataSource` abstraction is totally generic, and we're introducing some special error handling of the NOOPDataSource type here.

Could the noop data source just return `nil, nil` instead, therefore removing the need for this?

go.mod
```diff
@@ -66,16 +66,18 @@ require (
...
+	golang.org/x/net v0.18.0 // indirect
+	golang.org/x/sync v0.5.0 // indirect
+	golang.org/x/sys v0.15.0 // indirect
+	golang.org/x/text v0.14.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20230711160842-782d3b101e98 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230711160842-782d3b101e98 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+replace github.com/smartcontractkit/libocr => github.com/augustbleeds/libocr v0.0.0-20240117214314-b678c2c6c62d
```
jmank88, line 79: TODO:
- [ ] update with real version

pkg/loop/median_service_test.go
jmank88, line 1: I am seeing diff here, but the result is the current state. Did a merge get corrupted or something? If so, could squash everything.

pkg/loop/median_service_test.go
```diff
@@ -18,7 +18,39 @@ func TestMedianService(t *testing.T) {
```
jmank88, line 18: Let's use a table of subtests, rather than duplicating the body of the test:
```go
for _, tt := range []struct{
	name string
	gasPriceDataSource median.DataSource
} {
	{"static",  test.StaticGasPriceDataSource()},
	{"noop",  test.NOOPDataSource()},
} {
	t.Run(tt.name, func(t *testing.T) {
```

pkg/loop/internal/datasource.go
augustbleeds: sounds good, I incorporated the changes

pkg/loop/median_service_test.go
augustbleeds: thanks, good point

PR #168, opened 2023-09-08
Merc-1640 fixing Mercury fee calculation

sdrug:


pkg/reportingplugins/mercury/fees.go
```diff
@@ -1,24 +1,41 @@
...
 
 // CalculateFee outputs a fee in wei
-func CalculateFee(tokenPriceInUSD *big.Int, baseUSDFeeCents uint32) (fee *big.Int) {
-	fee = new(big.Int).Mul(big.NewInt(int64(baseUSDFeeCents)), tokenPriceInUSD)
-	fee = fee.Mul(fee, FEE_SCALING_FACTOR)
-	fee = fee.Div(fee, PRICE_SCALING_FACTOR)
-	fee = fee.Div(fee, CENTS_PER_DOLLAR)
-	return
+func CalculateFee(tokenPriceInUSD *big.Int, baseUSDFeeCents uint32) (*big.Int, error) {
+	if tokenPriceInUSD.Cmp(big.NewInt(0)) == 0 || baseUSDFeeCents == 0 {
```
Fletch153, line 22: A 0 token price is technically correct

pkg/reportingplugins/mercury/fees.go
```diff
@@ -1,24 +1,41 @@
...
+		return nil, errors.Errorf("token price and base fee must be non-zero")
+	}
+
+	// big.Float base fee in USD
+	baseFee := new(big.Float).Quo(new(big.Float).SetInt64(int64(baseUSDFeeCents)), CENTS_PER_DOLLAR)
```
Fletch153, line 27: Isn't cents_per_dollar and price_scaling_factor the same thing? it looks like you need to do something like:

baseUSDFeeCents * token.scale / tokenPriceInUsd

pkg/reportingplugins/mercury/fees_test.go
```diff
@@ -3,15 +3,41 @@ package mercury
 import (
 	"math/big"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
+func scaleTokenPrice(usdPrice float64) *big.Int {
+	scaledPrice := new(big.Float).Mul(big.NewFloat(usdPrice), PRICE_SCALING_FACTOR)
```
Fletch153, line 11: ditto

pkg/reportingplugins/mercury/v2/mercury.go
```diff
@@ -188,12 +193,17 @@ func (rp *reportingPlugin) Observation(ctx context.Context, repts ocrtypes.Repor
 	} else if obs.NativePrice.Val.Cmp(MissingPrice) <= 0 {
 		p.NativeFee = mercury.MaxInt192Enc
 	} else {
-		nativeFee := mercury.CalculateFee(obs.NativePrice.Val, rp.offchainConfig.BaseUSDFeeCents)
-		if nativeFeeEncoded, err := mercury.EncodeValueInt192(nativeFee); err != nil {
-			nativeErr = pkgerrors.Wrapf(err, "failed to encode native fee; val=%s", nativeFee)
+		nativeFee, err := mercury.CalculateFee(obs.NativePrice.Val, rp.offchainConfig.BaseUSDFeeCents)
```
Fletch153, line 196: Is it possible for the price to be negative? For assets it's unusual, but if there is a bug where the NativePrice.Val is negative, CalculateFee would return negative, and this would be encoded to a uint192, given the most significant is the sign, this would in turn be an extraordinarily large fee and lead to unverifiable reports.

pkg/reportingplugins/mercury/fees.go
```diff
@@ -1,24 +1,41 @@
...
 // e.g. for a 1e8 multiplier, a LINK/USD value of 7.42 will be represented as 742000000
 // This is what we expect from our data source.
-var PRICE_SCALING_FACTOR = big.NewInt(1e8)
+var PRICE_SCALING_FACTOR = big.NewFloat(1e8)
 
 // FeeScalingFactor indicates the multiplier applied to fees.
 // e.g. for a 1e18 multiplier, a LINK fee of 7.42 will be represented as 7.42e18
 // This is what will be baked into the report for use on-chain.
-var FEE_SCALING_FACTOR = big.NewInt(1e18)
+var FEE_SCALING_FACTOR = big.NewFloat(1e18)
```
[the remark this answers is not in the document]
sdrug: That's a good point. @samsondav how do we want to configure it?

pkg/reportingplugins/mercury/fees.go
sdrug: PRICE_SCALING_FACTOR is the multiplier applied to the price we retrieve from reports. 
CENTS_PER_DOLLAR is just to convert fee in cents that we receive from config into usd.

pkg/reportingplugins/mercury/v2/mercury.go
sdrug: Imo handling negative values doesn't belong to the fee calculation and should happen somewhere upstream. Also having unverifiable report in this case sounds like a good outcome to me. I'll let @samsondav to chime in here as the code owner

pkg/reportingplugins/mercury/fees.go
Fletch153: I believe you only need one, and that is the precision of the token

pkg/reportingplugins/mercury/v2/mercury.go
Fletch153: If we have a bug in the software which leads to int_max fees being generated, there will be a complete outage. I think instead, we should remove the concept of int_max for failures and instead, zero the fee.

This also reduces gas costs quite significantly, as passing 0x00..00 is cheaper than passing 0x_int_max not only because 0 is billed for less, but low entropy data can easily be compressed when rolling up on a L2

pkg/reportingplugins/mercury/fees.go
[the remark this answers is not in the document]
samsondav: For now we can leave it hardcoded since both link and eth use 18 dp

pkg/reportingplugins/mercury/fees.go
```diff
@@ -1,24 +1,38 @@
...
 
-// CalculateFee outputs a fee in wei
-func CalculateFee(tokenPriceInUSD *big.Int, baseUSDFeeCents uint32) (fee *big.Int) {
-	fee = new(big.Int).Mul(big.NewInt(int64(baseUSDFeeCents)), tokenPriceInUSD)
-	fee = fee.Mul(fee, FEE_SCALING_FACTOR)
-	fee = fee.Div(fee, PRICE_SCALING_FACTOR)
-	fee = fee.Div(fee, CENTS_PER_DOLLAR)
-	return
+	// scale baseFee in USD
+	baseFeeScaled := new(big.Float).Mul(new(big.Float).SetInt64(int64(baseUSDFeeCents)), PRICE_SCALING_FACTOR)
```
samsondav, line 25: For prices please do not use float, use decimal for absolute precision

pkg/reportingplugins/mercury/v2/mercury.go
samsondav: Core side will never encode a negative fee. But it doesn't hurt to have extra checks.

samsondav (changes requested):

pkg/reportingplugins/mercury/fees_test.go
```diff
@@ -3,15 +3,46 @@ package mercury
...
+		fee := CalculateFee(tokenPriceInUSD, baseUSDFeeCents)
+		assert.Equal(t, big.NewInt(0), fee)
+	})
+
+	t.Run("with base fee == 0", func(t *testing.T) {
+		tokenPriceInUSD := scalePrice(123)
+		baseUSDFeeCents = 0
+		fee := CalculateFee(tokenPriceInUSD, baseUSDFeeCents)
+		assert.Equal(t, big.NewInt(0), fee)
+	})
```
samsondav, line 47: Let's sanity check handling negative base fee + negative token price as well pls, should return 0 in these cases or error

samsondav (changes requested):

samsondav (approved):

PR #173, opened 2023-09-12
[BCF-2627/BCF-2628/BCF-2629] Add general plugin service, server and clients

cedric-cordenier:
* First, some refactoring:
    * Move the helper process into a separate command
    * Make the PluginService public so it can be used from other packages. This also includes some selected methods, eg. Wait, Init, and attributes such as PluginService.Service.
    * Move TestPlugin into the test package so it can be used from other packages
* Then, add pkg/reporting_plugins, along with the service, client and server implementations needed for the lightweight OCR initiative.


pkg/loop/plugin_generic.go
```diff
@@ -0,0 +1,56 @@
...
+		MagicCookieValue: "2ad981747cd86c4ab3e23170970020fd",
+	}
+}
+
+type GRPCPluginGeneric struct {
+	plugin.NetRPCUnsupportedPlugin
+
+	BrokerConfig
+
+	PluginServer types.PluginGeneric
```
cedric-cordenier, line 28: This should use a different type which converts the connection into a proper provider

pkg/loop/internal/generic_service.go
```diff
@@ -0,0 +1,107 @@
...
+
+	errorLogConn, err := m.dial(request.ErrorLogID)
+	if err != nil {
+		m.closeAll(providerRes)
+		return nil, ErrConnDial{Name: "ErrorLog", ID: request.ErrorLogID, Err: err}
+	}
+	errorLogRes := resource{errorLogConn, "ErrorLog"}
+	errorLog := newErrorLogClient(errorLogConn)
+
+	factory, err := m.impl.NewGenericServiceFactory(ctx, request.PluginConfig, provider.ClientConn(), errorLog)
```
cedric-cordenier, line 92: This is wrong, since this is where we call the implementation: we should do some type coercion here back into the target provider.

pkg/loop/generic_service.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
+	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
+)
+
+var _ ocrtypes.ReportingPluginFactory = (*GenericService)(nil)
+
+// GenericService is a [types.Service] that maintains an internal [types.PluginGeneric].
+type GenericService struct {
```
jmank88, line 19: The modifier `Generic` doesn't feel right to me. Maybe just because the overloading with generic types? Is there another term we could use? `Dynamic`? :thinking:  Or even just drop it in some cases to say only `Plugin`?

cedric-cordenier: Yeah finding a name for this was really tricky. I'll try PluginService.

pkg/types/plugin_service.go
```diff
@@ -0,0 +1,28 @@
...
+	ProviderType string
+	Command      string
+	PluginName   string
+	PluginConfig []byte
+}
+
+// PluginClient is the client interface to a plugin running
+// as a generic job (job type = GenericPlugin) inside the core node.
+type PluginClient interface {
+	NewPluginServiceFactory(ctx context.Context, config PluginServiceConfig, grpcProvider grpc.ClientConnInterface, errorLog ErrorLog) (ReportingPluginFactory, error)
```
jmank88, line 19: Would it make sense to align with the type name here?
```suggestion
	NewReportingPluginFactory(ctx context.Context, config PluginServiceConfig, grpcProvider grpc.ClientConnInterface, errorLog ErrorLog) (ReportingPluginFactory, error)
```
Or even more broadly with other types too?

pkg/loop/internal/pb/plugin_service.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb";
+
+package loop;
+
+service PluginService {
+  rpc NewPluginServiceFactory (NewPluginServiceFactoryRequest) returns (NewPluginServiceFactoryReply) {}
```
krehermann: is this stuttering or is the factory generating a different kind of service than the PluginService itself?

wdyt of `NewFactory`?

pkg/loop/internal/pb/plugin_service.proto
```diff
@@ -0,0 +1,28 @@
...
+// NewPluginServiceFactoryRequest has arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.PluginService.NewPluginServiceFactory].
+message NewPluginServiceFactoryRequest {
+  uint32 providerID = 1;
+  uint32 errorLogID = 2;
+  PluginServiceConfig pluginServiceConfig = 3;
+}
+
+// NewPluginServiceFactoryReply has return arguments for [github.com/smartcontractkit/chainlink-relay/pkg/loop.PluginService.NewPluginServiceFactory].
+message NewPluginServiceFactoryReply {
+  uint32 reportingPluginFactoryID = 1;
```
krehermann, line 27: does the field name need all the extra qualifies? seems like `id` ought to be enough: `NewPluginServiceFactoryReply.ID`

wdyt?

cedric-cordenier: Yes, good point :)

pkg/loop/internal/pb/plugin_service.proto
cedric-cordenier: This was subsequently renamed to NewReportingPluginFactory since that's what we use everywhere else

pkg/loop/internal/plugin_service.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func NewPluginServiceClient(broker Broker, brokerCfg BrokerConfig, conn *grpc.ClientConn) *PluginServiceClient {
+	brokerCfg.Logger = logger.Named(brokerCfg.Logger, "PluginServiceClient")
+	pc := newPluginClient(broker, brokerCfg, conn)
+	return &PluginServiceClient{pluginClient: pc, pluginService: pb.NewPluginServiceClient(pc), serviceClient: newServiceClient(pc.brokerExt, pc)}
+}
+
+func (m *PluginServiceClient) NewReportingPluginFactory(ctx context.Context, config types.PluginServiceConfig, grpcProvider grpc.ClientConnInterface, errorLog types.ErrorLog) (types.ReportingPluginFactory, error) {
+	cc := m.newClientConn("PluginServiceFactory", func(ctx context.Context) (id uint32, deps resources, err error) {
```
krehermann, line 30: should the text in the first arg be updated?

pkg/loop/internal/test/test_plugin.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		GRPCServer: plugin.DefaultGRPCServer,
+		Plugins:    map[string]plugin.Plugin{name: p},
+	})
+
+	// We should get a config
+	var config *plugin.ReattachConfig
+	select {
+	case config = <-ch:
+	case <-time.After(5 * time.Second):
```
krehermann, line 53: is 5 special? are we setting up for flakey tests by using a hardcoded value?

cedric-cordenier: I don't think so: this is just a timeout value. (Note: I just moved this code so would prefer to leave any changes to this out of scope).

pkg/loop/internal/test/generic.go
```diff
@@ -0,0 +1,239 @@
...
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop/internal"
+	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+)
+
+type MockConn struct {
+	grpc.ClientConnInterface
+}
+
+type StaticGenericPluginMedianProvider struct {
```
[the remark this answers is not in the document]
cedric-cordenier: Generic was meant to be a (admittedly awkward) standin for "lightweight OCR". I've renamed this to StaticReportingPluginWithMedianProvider since Jordan recommended using ReportingPlugin throughout, and StaticGenericPluginProvider has been renamed to StaticReportingPluginWithPluginProvider.

Let me know if that makes sense :)

pkg/loop/internal/test/cmd/main.go
jmank88, line 1: Are you confident about this refactor?
This was copied from hashicorp's internal test suite, so it would be simpler not to deviate from that.

pkg/loop/service.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/smartcontractkit/chainlink-relay/pkg/utils"
 )
 
-const keepAliveTickDuration = 5 * time.Second //TODO from config
+const KeepAliveTickDuration = 5 * time.Second //TODO from config
```
jmank88, line 21: Can we avoid exporting this?

pkg/loop/logger.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
 )
 
+const PluginLoggerTestName = "logger-test"
+
+const LoggerTestName = "server-side-logger-name"
```
jmank88, line 17: Why export?

pkg/loop/service.go
```diff
@@ -30,7 +30,7 @@ type grpcPlugin interface {
 
 // pluginService is a [types.Service] wrapper that maintains an internal [types.Service] created from a [grpcPlugin]
 // client instance by launching and re-launching as necessary.
-type pluginService[P grpcPlugin, S services.Service] struct {
+type PluginService[P grpcPlugin, S services.Service] struct {
```
jmank88, line 33: Why does this need to be exported?

pkg/loop/service.go
```diff
@@ -212,11 +212,43 @@ func (s *pluginService[P, S]) closeClient() (err error) {
...
 	case <-s.serviceCh:
 		return nil
+	case <-s.stopCh:
+		return nil
+	}
+}
+
+// XXXTestHook returns a TestPluginService.
+// It must only be called once, and before Start.
+func (s *PluginService[P, S]) XXXTestHook() TestPluginService[P, S] {
```
jmank88, line 228: IMHO we should avoid this at all costs. Have all other options been exhausted?

pkg/loop/service.go
```diff
@@ -212,11 +212,43 @@ func (s *pluginService[P, S]) closeClient() (err error) {
 	return
 }
 
-func (s *pluginService[P, S]) wait(ctx context.Context) error {
+func (s *PluginService[P, S]) Wait(ctx context.Context) error {
```
jmank88, line 215: Why is this exported?

pkg/loop/service.go
```diff
@@ -49,12 +49,12 @@ type pluginService[P grpcPlugin, S services.Service] struct {
...
 	serviceCh chan struct{} // closed when service is available
-	service   S
+	Service   S
 
-	testInterrupt chan func(*pluginService[P, S]) // tests only (via TestHook) to enable access to internals without racing
+	testInterrupt chan func(*PluginService[P, S]) // tests only (via TestHook) to enable access to internals without racing
 }
 
-func (s *pluginService[P, S]) init(pluginName string, p P, newService func(context.Context, any) (S, error), lggr logger.Logger, cmd func() *exec.Cmd, stopCh chan struct{}) {
+func (s *PluginService[P, S]) Init(pluginName string, p P, newService func(context.Context, any) (S, error), lggr logger.Logger, cmd func() *exec.Cmd, stopCh chan struct{}) {
```
jmank88, line 57: Why is this exported?

pkg/loop/service.go
```diff
@@ -212,11 +212,43 @@ func (s *pluginService[P, S]) closeClient() (err error) {
...
+
+// XXXTestHook returns a TestPluginService.
+// It must only be called once, and before Start.
+func (s *PluginService[P, S]) XXXTestHook() TestPluginService[P, S] {
+	s.testInterrupt = make(chan func(*PluginService[P, S]))
+	return s.testInterrupt
+}
+
+// TestPluginService supports Killing & Resetting a running *pluginService.
+type TestPluginService[P grpcPlugin, S services.Service] chan<- func(*PluginService[P, S])
```
jmank88, line 234: Why export a test type?

pkg/loop/service.go
cedric-cordenier: It needs to be exported so it can be used by `pkg/loop/reporting_plugins`

pkg/loop/service.go
cedric-cordenier: It's used by the `reporting_plugins` package where the generic services are located.

pkg/loop/service.go
cedric-cordenier: Same comment :)

pkg/loop/service.go
cedric-cordenier: Yes, I tried moving this to a `_test` package but that won't get exported to other packages.

One alternative is to expose `testInterrupt` instead, but I think that runs into the same problem.

pkg/loop/service.go
cedric-cordenier: It's used by the `reporting_plugins` package.

pkg/loop/internal/test/cmd/main.go
cedric-cordenier: Pretty confident since the tests pass.

My reason for moving it out of a test was mainly since that's really what we're doing here (running a command), and because this needs to be called by various different packages.

jmank88:
I think we need to be _very_ careful with the exported API in these packages. We already have an `internal/test` package - can't we continue to leverage that for most or even all of these cases?

cedric-cordenier:
@jmank88, I've followed up on our offline chat and moved PluginService, LoggerLOOP and KeepAliveTickDuration to internal.

pkg/loop/reporting_plugins/grpc.go
jmank88, line 1: I usually try to avoid underscores in package names. FWIW the google style guide is very strict about this: https://google.github.io/styleguide/go/decisions.html#package-names
> Go package names should not have underscores. If you need to import a package that does have one in its name (usually from generated or third party code), it must be renamed at import time to a name that is suitable for use in Go code.

pkg/loop/process_test.go
```diff
@@ -0,0 +1,11 @@
+package loop_test
+
+import (
+	"os/exec"
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/test"
+)
+
+func HelperProcess(command string, opts test.HelperProcessOptions) *exec.Cmd {
```
jmank88, line 9: Can we hide this one too?

pkg/loop/reporting_plugins/grpc.go
krehermann: +1

pkg/loop/internal/test/logger_loop.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop"
+)
+
+const PluginLoggerTestName = "logger-test"
+
+const LoggerTestName = "server-side-logger-name"
+
+// NOTE: This is part of the main package because it needs to be imported by the test binary at `./internal/test/cmd`
```
krehermann, line 19: should 'part of main package' be updated or removed? this is a `main`

pkg/loop/reporting_plugins/loopp_service_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+	utilstests "github.com/smartcontractkit/chainlink-relay/pkg/utils/tests"
+)
+
+func HelperProcess(command string, opts test.HelperProcessOptions) *exec.Cmd {
+	return test.HelperProcess("../internal/test/cmd/main.go", command, opts)
+}
+
+// pluginGenericMedian is a generic plugin which wants a median provider
+const pluginGenericMedian = "generic-median"
```
krehermann, line 26: the value `generic-median` is used in on of the tests in a switch. should we export this and update the test?

pkg/loop/internal/pb/reporting_plugin_service.proto
```diff
@@ -0,0 +1,28 @@
...
+
+service ReportingPluginService {
+  rpc NewReportingPluginFactory (NewReportingPluginFactoryRequest) returns (NewReportingPluginFactoryReply) {}
+}
+
+message ReportingPluginServiceConfig {
+  string providerType = 1;
+  string command = 2;
+  string pluginName = 3;
+  bytes pluginConfig = 4;
```
jmank88, line 15: So we use `[]byte` in go because that is standard for encoded JSON, but I wonder if we would prefer `string` here? If there are no other tradeoffs, then `string` seems slightly more debuggable for a human.

pkg/loop/internal/service.go
```diff
@@ -11,6 +12,8 @@ import (
 	"github.com/smartcontractkit/chainlink-relay/pkg/services"
 )
 
+var ErrPluginUnavailable = errors.New("plugin unavailable")
```
jmank88, line 15: This seems OK to export from `package loop`, and potentially even useful. I guess we need to be able to reference it from inside here though? Maybe another var to forward in that case? :shrug:

pkg/loop/median_service.go
```diff
@@ -34,15 +34,13 @@ func NewMedianService(lggr logger.Logger, grpcOpts GRPCOpts, cmd func() *exec.Cm
...
-	ms.init(PluginMedianName, &GRPCPluginMedian{BrokerConfig: broker}, newService, lggr, cmd, stopCh)
+	ms.Init(PluginMedianName, &GRPCPluginMedian{BrokerConfig: broker}, newService, lggr, cmd, stopCh)
 	return &ms
 }
 
 func (m *MedianService) NewReportingPlugin(config ocrtypes.ReportingPluginConfig) (ocrtypes.ReportingPlugin, ocrtypes.ReportingPluginInfo, error) {
-	ctx, cancel := utils.ContextFromChan(m.pluginService.stopCh)
-	defer cancel()
-	if err := m.wait(ctx); err != nil {
+	if err := m.Wait(context.Background()); err != nil {
```
jmank88, line 42: Why a background context? Yielding to shutdown is important.

jmank88: I see that `Wait` yields to `stopCh` now, but then can we drop the context?

pkg/loop/median_service_test.go
```diff
@@ -52,7 +52,7 @@ func TestMedianService_recovery(t *testing.T) {
 	t.Parallel()
 	var limit atomic.Int32
 	median := loop.NewMedianService(logger.Test(t), loop.GRPCOpts{}, func() *exec.Cmd {
-		return helperProcess(loop.PluginMedianName, strconv.Itoa(int(limit.Add(1))))
+		return HelperProcess(loop.PluginMedianName, test.HelperProcessOptions{Limit: int(limit.Add(1))})
```
jmank88, line 55: nit/ I am personally a fan of method based constructors, since it is less redundant/verbose for all cases:
```suggestion
		return test.HelperProcessOptions{Limit: int(limit.Add(1))}.New(loop.PluginMedianName)
```
```diff
- return HelperProcess(loop.PluginMedianName, test.HelperProcessOptions{})
+ return NewHelperProcess(loop.PluginMedianName)
```

pkg/loop/reportingplugins/grpc.go
```diff
@@ -0,0 +1,81 @@
...
+	loop.BrokerConfig
+
+	PluginServer ProviderServer[T]
+
+	pluginClient *internal.ReportingPluginServiceClient
+}
+
+type serverAdapter func(context.Context, types.ReportingPluginServiceConfig, grpc.ClientConnInterface, types.ErrorLog) (types.ReportingPluginFactory, error)
+
+func (g serverAdapter) NewReportingPluginFactory(ctx context.Context, config types.ReportingPluginServiceConfig, conn grpc.ClientConnInterface, errorLog types.ErrorLog) (types.ReportingPluginFactory, error) {
```
jmank88, line 44: Why `g` for `serverAdapter`, rather than `s` or `a`?

pkg/loop/reportingplugins/grpc_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop"
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/test"
+	"github.com/smartcontractkit/chainlink-relay/pkg/loop/reportingplugins"
+	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+	"github.com/smartcontractkit/chainlink-relay/pkg/utils/tests"
+)
+
+func newStopCh(t *testing.T) <-chan struct{} {
```
jmank88, line 17: You could `t.CleanUp(func() { close(stopCh)})`, but then I suppose you need to `sync.Once` the close operation so maybe just making a mess :shrug:

pkg/loop/internal/test/cmd/main.go
jmank88: I'm concerned there are subtle changes happening here that could be problematic, since we are no longer invoking a test binary :thinking:
It doesn't seem like we were forwarding any special test flags though, so no obvious changes that I can see :shrug:

pkg/loop/median_service.go
cedric-cordenier: I noticed that `m.wait` was invariably being called after getting the context from the `stopCh`, so rather than exposing that, moved this into the `wait` function.

I think it can still be useful for callers to override the timeout themselves, which is why I left context as part of the API, but can remove it if you prefer.

pkg/loop/reportingplugins/grpc.go
cedric-cordenier: Ugh, this is a holdover from when this was called `grpcServerAdapter`

pkg/loop/median_service.go
jmank88: Yeah I think remove since it is just creating noise? If there are cases with upstream contexts to pass in, then a `waitCtx` variant might actually be cleaner in order to avoid loose `context.Background()`s.

pkg/loop/median_service_test.go
cedric-cordenier: Sounds good :) I've modified this suggestion to take into account the fact that a commandLocation needs to be passed in, but hopefully the API is cleaner now.

pkg/loop/internal/test/test_plugin.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if h.Limit != 0 {
+		cmdArgs = append(cmdArgs, fmt.Sprintf("-limit=%d", h.Limit))
+	}
+	cmd := exec.Command(cmdArgs[0], cmdArgs[1:]...) // #nosec
+	cmd.Env = os.Environ()
+	return cmd
+}
+
+func NewHelperProcess(commandLocation, command string, opts ...func(*HelperProcessCommand)) *exec.Cmd {
```
jmank88, line 41: Hmmm why functional options? I actually think these are quite problematic and not really necessary for an internal API.

pkg/loop/internal/pb/reporting_plugin_service.proto
```diff
@@ -0,0 +1,28 @@
...
+
+service ReportingPluginService {
+  rpc NewReportingPluginFactory (NewReportingPluginFactoryRequest) returns (NewReportingPluginFactoryReply) {}
+}
+
+message ReportingPluginServiceConfig {
+  string providerType = 1;
+  string command = 2;
+  string pluginName = 3;
+  string pluginConfig = 4;
```
krehermann, line 15: i'm reticent to say it, since i see this was just changed do to a comment from @jmank88, but i think bytes is a better type here because it's the most neutral and this is low level wire communication.

pkg/loop/internal/plugin_service.go
```diff
@@ -212,11 +209,47 @@ func (s *pluginService[P, S]) closeClient() (err error) {
...
 	case <-ctx.Done():
 		return context.Cause(ctx)
 	case <-s.serviceCh:
 		return nil
+	case <-s.stopCh:
+		return errors.New("service was stopped while waiting")
+	}
+}
+
+func (s *PluginService[P, S]) Wait() error {
```
krehermann, line 223: can you add a comment above what is being waited for?

when i read the use in NewReportingPlugin() it's not clear why a Wait is needed and in isolation, I would have expected a call to something like Start() or WaitForStart(); `Wait` seems to nebulous without a good comment

krehermann (dismissed):

pkg/loop/internal/test/test_plugin.go
cedric-cordenier: Why do you think they're problematic? I've always found them to be a useful feature for this kind of use case.

The square I'm trying to circle here is making the limit optional while allowing the `commandLocation` to be specified once per package somewhere.

pkg/loop/internal/pb/reporting_plugin_service.proto
jmank88: What do you mean by most neutral? Like we might use config that is not human readable in the future? Do you think we should change the Relayer TOML field from `string` to `bytes`?

jmank88: https://protobuf.dev/programming-guides/proto3/#scalar
IIUC they are the same cost and there is only a superficial difference of limiting strings to utf8.

krehermann: i mean that the plugin config is really a struct or structs. representing that as bytes feels better and more intuitive to me over the wire.

but i'm not going to die on that hill

pkg/loop/internal/plugin_service.go
```diff
@@ -212,11 +209,47 @@ func (s *pluginService[P, S]) closeClient() (err error) {
...
 
-func (s *pluginService[P, S]) wait(ctx context.Context) error {
+func (s *PluginService[P, S]) WaitCtx(ctx context.Context) error {
 	select {
 	case <-ctx.Done():
 		return context.Cause(ctx)
 	case <-s.serviceCh:
 		return nil
+	case <-s.stopCh:
+		return errors.New("service was stopped while waiting")
```
jmank88, line 219: I wonder if it matters that this will no longer satisfy `errors.Is(err, context.Canceled)`. Might want to wrap it :shrug:

pkg/loop/internal/test/test_plugin.go
cedric-cordenier: Hmmm... I'm not really sure how to address this comment in that case, since as I said this was the cleanest way I could think of to allow the limit to be optional while passing in a commandLocation.

Is this a blocking point for you? If so, can you give me a sense of how you would like to fix this?

jmank88: I'm not sure that I understand the problem. Why is the other API insufficient for doing this? Are you trying to prevent `commandLocation` from being set? I think that is still possible to validate in the helper func.

jmank88: Would something like this work in place of each helper func?
```go
type HelperProcessCommand struct {
	test.HelperProcessCommand
}

func (h HelperProcessCommand) New() *exec.Cmd {
	h.CommandLocation = "./internal/test/cmd/main.go"
	return h.HelperProcessCommand.New()
}
```

cedric-cordenier: I've refactored it so that it doesn't use functional options. Let me know what you think.

pkg/loop/internal/plugin_service.go
```diff
@@ -212,11 +209,43 @@ func (s *pluginService[P, S]) closeClient() (err error) {
...
 	case <-s.serviceCh:
 		return nil
+	case <-s.stopCh:
+		return nil
+	}
+}
+
+// XXXTestHook returns a TestPluginService.
+// It must only be called once, and before Start.
+func (s *PluginService[P, S]) XXXTestHook() TestPluginService[P, S] {
```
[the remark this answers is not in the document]
cedric-cordenier: I've spent a bit of time thinking about this, what do you think of adding this hook to a file with a test build tag, and then running our tests with -tags=test? (A default test build tag has been suggested before, but not accepted: https://github.com/golang/go/issues/21360)

Note: this issue is documented here https://github.com/golang/go/issues/39565, and the recommendation from go developers was to move this suggestion into a common testing package and export the shared functionality which is basically what I did.

pkg/loop/internal/test/test_plugin.go
jmank88: LGTM :up:

pkg/loop/process_test.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func NewHelperProcessCommand(command string) *exec.Cmd {
+	h := HelperProcessCommand{
+		HelperProcessCommand: test.HelperProcessCommand{
+			Command: command,
+		},
+	}
+	return h.New()
+}
```
jmank88, line 25: Oh This can be simplified a little further to drop the nested type:
```suggestion
type HelperProcessCommand test.HelperProcessCommand

func (h HelperProcessCommand) New() *exec.Cmd {
	h.CommandLocation = "./internal/test/cmd/main.go"
	return (test.HelperProcessCommand)(h).New()
}

func NewHelperProcessCommand(command string) *exec.Cmd {
	h := HelperProcessCommand{
		Command: command,
	}
	return h.New()
}
```

pkg/loop/internal/test/test_plugin.go
jmank88: Er one small tweak to make the call sites cleaner.

pkg/loop/internal/plugin_service.go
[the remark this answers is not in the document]
jmank88: I don't think it is worth introducing tags for. That was a major hassle in the core repo that we only just moved away from.

> and the recommendation from go developers was to move this suggestion into a common testing package

That is what I was going to suggest. We currently only use an internal package here, with both product and test code. Using internal/test as well (or some new test package) would separate and encapsulate the concerns. However, it might require another shared package below this level, or just moving this type to that other package. For example
- move this type to `internal/plugin.Service`
- add `internal/plugintest.Service` which embeds `Service` and adds these methods
- tests can choose to import `plugintest` in place of plugin to get the extended `Service`

pkg/loop/process_test.go
cedric-cordenier: 😥 I've refactored again, but want to point out gently that the first version (to use embedding) was your suggestion pretty much verbatim.

jmank88: What do you mean? Isn't this strictly an improvement on that version?

pkg/loop/internal/plugin_service.go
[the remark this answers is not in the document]
jmank88: Oh I think that the way that we embed this type makes it more complicated :thinking:

jmank88 (approved):

PR #174, opened 2023-09-12
Use zero values for fee if we cannot come to consensus

samsondav:
It is better to  generate a report that will validate for free,
rather than no report at all, if we cannot come to consensus on a
valid fee.

### DIFF
--- a/pkg/reportingplugins/mercury/aggregate_functions.go
+++ b/pkg/reportingplugins/mercury/aggregate_functions.go
@@ -130,7 +130,7 @@ func GetConsensusLinkFee(paos []PAOLinkFee, f int) (*big.Int, error) {
 	var validLinkFees []*big.Int
 	for _, pao := range paos {
 		fee, valid := pao.GetLinkFee()
-		if valid && fee.Cmp(Zero) >= 0 {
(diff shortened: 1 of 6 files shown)


pkg/reportingplugins/mercury/aggregate_functions.go
```diff
@@ -128,7 +130,7 @@ func GetConsensusLinkFee(paos []PAOLinkFee, f int) (*big.Int, error) {
 	var validLinkFees []*big.Int
 	for _, pao := range paos {
 		fee, valid := pao.GetLinkFee()
-		if valid {
+		if valid && fee.Cmp(Zero) >= 0 {
```
jmank88, line 133: You could use `Sign` for these: https://pkg.go.dev/math/big#Int.Sign
```suggestion
		if valid && fee.Sign() >= 0 {
```

samsondav: Cool didn't know about this one

essamhassan (approved):

Fletch153 (approved):

Fletch153 (approved):

PR #179, opened 2023-09-19
feature/BCF-2404-grpc-tracing: Implementing grpc tracing

patrickhuie19:
Corresponding core PR: https://github.com/smartcontractkit/chainlink/pull/10698


pkg/loop/telem.go
```diff
@@ -30,6 +56,61 @@ func SetupTelemetry(registerer prometheus.Registerer) GRPCOpts {
...
+	if err != nil {
+		log.Errorf("failed to create trace exporter: %w", err)
+		return
+	}
+
+	attributes := []attribute.KeyValue{
+		semconv.ServiceNameKey.String("cl-node"),
+		semconv.ProcessPIDKey.Int(os.Getpid()),
+		// TODO: figure out how to wire node version through?
+		semconv.ServiceVersionKey.String("0.0.1"),
```
jmank88, line 91: Don't we want the LOOPP program version here, rather than the core node?
The runtime should have build info that we can tap in to: https://pkg.go.dev/runtime/debug#ReadBuildInfo

patrickhuie19: I believe we would want either the LOOP program version or the code node version. `ReadBuildInfo` will work great - we can get the module version of the binary through the returned object of type BuildInfo.

jmank88: We might need the full package name for the `ServiceNameKey`, to be able to distinguish between main packages within the same module. (I'm assuming the short program names won't be consistent or reliable?)

pkg/logger/logger.go
```diff
@@ -45,6 +45,7 @@ type Config struct {
 var defaultConfig Config
 
 // New returns a new Logger with the default configuration.
+// TODO: why do API consumers have to worry about err here? Not useful?
```
jmank88, line 48:
```suggestion
```

pkg/loop/telem.go
```diff
@@ -30,6 +55,73 @@ func SetupTelemetry(registerer prometheus.Registerer) GRPCOpts {
...
+	// Enough to shutdown the underlying connection since DialContext is used in blocking mode
+	defer cancel()
+	conn, err := grpc.DialContext(ctx, config.CollectorTarget,
+		// Note the use of insecure transport here. TLS is recommended in production.
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		// safe usage in this case https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#dialing-in-grpc
+		grpc.WithBlock(),
+	)
+	if err != nil {
+		log.Errorf("Connecting to OTEL collector %w failed: %v", config.CollectorTarget, err)
```
jmank88, line 78: Should this return?
```suggestion
		log.Errorf("Connecting to OTEL collector %w failed: %w", config.CollectorTarget, err)
		return
```

.github/workflows/golangci_lint.yml
jmank88, line 1: Can we handle this in a separate PR? There is a standardized CI action https://github.com/smartcontractkit/.github/tree/10d9b22b85c78e6920c00626156365ae035a605d/actions/ci-lint-go

pkg/loop/telem.go
```diff
@@ -2,26 +2,51 @@ package loop
...
 // SetupTelemetry initializes open telemetry and returns GRPCOpts with telemetry interceptors.
-func SetupTelemetry(registerer prometheus.Registerer) GRPCOpts {
-	otel.SetTracerProvider(sdktrace.NewTracerProvider(
-		sdktrace.WithSampler(sdktrace.AlwaysSample()),
-	))
+// It is called from the host and each plugin - intended as there is bidirectional communication
+func SetupTelemetry(registerer prometheus.Registerer, config TracingConfig) GRPCOpts {
+	if config.Enabled {
+		SetupTracing(config)
+	}
```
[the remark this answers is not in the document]
patrickhuie19: Hmm, good point re: separation of concerns. I put them together for locality of behavior, but I agree two separate functions will be more readable and will extend the consumer more control in the future if tracing setup needs grow.

pkg/loop/telem.go
patrickhuie19: Previously, no. With the new API, I think that's the right approach.

pkg/loop/telem.go
```diff
@@ -2,34 +2,118 @@ package loop
...
+
+	resource := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		attributes...,
+	)
+
+	tracerProvider := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(traceExporter),
+		sdktrace.WithResource(resource),
+		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(config.SamplingRatio)),
```
jmank88, line 109: Should we consider `ParentBased`? https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v1.19.0/trace#ParentBased

patrickhuie19: Absolutely, though for right now since there is no active span when the GRPC interceptors are invoked (on the client side of the request), this will have no effect if I pass the `TraceIDRatioBased` sampler in as the root sampler. Or were you imagining something different?

patrickhuie19: Whatever side initiates the request.

jmank88: How can it be the case that we are only tracing on one side? What happens when the server side calls back in to a client with the same context?

patrickhuie19: My point is just that as the code is written right now, there is no parent span being passed into the interceptors. So, when a request is initiated, the client is generating a fresh span and trace that is extracted on the server side. The span and trace being generated means nothing if the server side does not decide to sample that trace. In the code as is, TraceIDRatioBased samples a percentage of traces based on the trace ID. We currently exert 0 control over the trace IDs, so this is effectively random. 

`ParentBased`sampling will honor the sampling decision for the parent span. Including this today will not affect sampling on the server side because there is no parent span.

jmank88: > The span and trace being generated means nothing if the server side does not decide to sample that trace.

Isn't the point of using `ParentBased` to control that decision? https://github.com/open-telemetry/opentelemetry-go/blob/sdk/v1.19.0/sdk/trace/sampling.go#L209

jmank88: > This is the most popular sampler and is the one recommended by the official OpenTelemetry [documentation](https://opentelemetry.io/docs/go/exporting_data/#sampling). 
>
> When a trace begins we make a decision whether or not to sample it. Whatever the decision, it is bubbled down inside the process and in other processes(for example in other services).
>
> There is a notable advantage to this: you always get the complete picture.

https://www.aspecto.io/blog/opentelemetry-sampling-everything-you-need-to-know/#parent-based-sampler

patrickhuie19: There are 4 options here based on whether the local or remote parent is or is not sampled (`[WithLocalParentNotSampled](https://github.com/open-telemetry/opentelemetry-go/blob/sdk/v1.19.0/sdk/trace/sampling.go#L253)` for example) - my understanding is that the current code does not have a parent, so these options will have no impact.

jmank88: When you say 'current code', is that making assumptions outside of this package?

patrickhuie19: Yes - we're currently not doing anything to chain spans across plugin invocations. Once that changes and we track some series of plugins calls (like in a Job), then I believe the Parent based rules will become useful

jmank88: > Yes

We shouldn't. This is a general library at the bottom of the dependency tree. It does not know anything about how it is currently used upstream or will be used in the future. We should aim to be maximally flexible by default. Why leave this out if we need to circle back later?

> we're currently not doing anything to chain spans across plugin invocations.

What would we need to do, besides pass a context?

patrickhuie19: > Why leave this out if we will need to circle back later?

It feels like we're going in circles here, but it's probably my bad. I have no problem with including or not including this. It makes no difference to me. We're going to be the ones showing product teams how to use these traces anyways. 

> What would we need to do, besides pass a context?

Figure out an appropriate place to define a parent span, and then figure out how to pass the corresponding context around.

krehermann: i'm willing to bet something in this code base will have to change once we try to trace on the core side. in that light, i think it's fine make a best effort now and get the code merged

jmank88: > It feels like we're going in circles here, but it's probably my bad. I have no problem with including or not including this. It makes no difference to me. We're going to be the ones showing product teams how to use these traces anyways.

If it's the default suggestion and we know we need it later, then it seems like a no-brainer to include now.

> Figure out an appropriate place to define a parent span, and then figure out how to pass the corresponding context around.

I know that we can enhance things by adding our own ID at some outer level later on, but that's not what I'm worried about here. We already have calls like this, with context wired through:
```mermaid
sequenceDiagram
	core->>product: MethodA
	product->>relayer: MethodB
```
Complicated further by proxies:
```mermaid
sequenceDiagram
	core->>product: MethodA
	product->>core: MethodB(proxy)
	core->>relayer: MethodB(proxy)
```
I thought that whatever is auto-populated by GRPC would be wired through already.

pkg/loop/telem.go
```diff
@@ -2,34 +2,118 @@ package loop
...
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+
+	// Enough to shutdown the underlying connection since DialContext is used in blocking mode
+	defer cancel()
+	conn, err := grpc.DialContext(ctx, config.CollectorTarget,
+		// Note the use of insecure transport here. TLS is recommended in production.
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		// safe usage in this case https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#dialing-in-grpc
+		grpc.WithBlock(),
```
jmank88, line 67: Can you elaborate on why this is safe? I was expecting the opposite. Our policy is to start fast and without external dependencies which block or may be unavailable. We normally connect (and reconnect as necessary) in the background.

jmank88: Even if it were _safe_, why is it preferable? The default behavior of a deferred internal connection is usually a feature - why give that up?

pkg/loop/telem.go
```diff
@@ -2,34 +2,118 @@ package loop
...
+	}
+
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+
+	// Enough to shutdown the underlying connection since DialContext is used in blocking mode
+	defer cancel()
+	conn, err := grpc.DialContext(ctx, config.CollectorTarget,
+		// Note the use of insecure transport here. TLS is recommended in production.
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
jmank88, line 65: This is a pretty big caveat. Any reason not to wire it up to config now? Seems worth noting in the func doc and creating an issue for.

pkg/loop/telem.go
jmank88: Actually, it looks like you don't have to worry about this detail at all, because the trace exporter can handle it for us, if we pass an endpoint and options instead of a connection:
```go
	// Set up a trace exporter
	// Shutting down the traceExporter will not shutdown the underlying connection.
	traceExporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithEndpoint(config.CollectorTarget), otlptracegrpc.WithDialOption(
		// Note the use of insecure transport here. TLS is recommended in production.
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	))
	if err != nil {
		return err
	}
```

patrickhuie19: > The default behavior of a deferred internal connection is usually a feature - why give that up?

When I wrote this, I didn't want `SetupTelemetry` (now `SetupTracing`) to return before a connection could be made to the otel-collector in cases it could take sufficiently long enough to connect to the collector such that early traces would be dropped. 

That is an assumption though, which is that if the otel-collector isn't properly connected to before traces are generated, the traces will be dropped until it is properly connected to.

pkg/loop/telem.go
patrickhuie19: It is - I tabled it for now because the otel collector is always living directly next to the container for node (whether locally or in PRs). Agreed on doc/issue.

pkg/loop/telem.go
jmank88: I don't think it make sense to treat trace collection as critical since it compromises our startup time, and with one blocking attempt we give up in cases where it was temporarily unavailable. Regardless, I would hope that the collection is buffered and not synchronous, so they would only be delayed rather than dropped.

pkg/loop/telem.go
jmank88: Resolved offline: parent sampler apparently not necessary for grpc client-server and nested span associations based on empirical evidence.

patrickhuie19: I included it for now - based on our conversation offline I think it's possible we get more for free and in the worst case it won't sample less than we intend.

jmank88 (approved):

PR #180, opened 2023-09-23
pkg/loop: fix HCLogLogger to preserve logger names

jmank88:
I noticed that we were logging two `logger` keys for the 'name', which is a bug, and not handled well by grafana.
This fixes the `hclSinkAdapter` and adds a test to ensure that the server side log name is observed in the test output.

### DIFF
--- a/pkg/loop/plugin_test.go
+++ b/pkg/loop/plugin_test.go
@@ -124,6 +124,12 @@ func TestHelperProcess(t *testing.T) {
 		}
 	}
 
+	lggr, err := loop.NewLogger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create logger: %s\n", err)
+		os.Exit(2)
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/loop/plugin_test.go
```diff
@@ -124,14 +124,20 @@ func TestHelperProcess(t *testing.T) {
 		}
 	}
 
+	lggr, err := loop.NewLogger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create logger: %s\n", err)
+		os.Exit(2)
```
patrickhuie19, line 130: Why exit code 2?

patrickhuie19 (approved):

pkg/loop/plugin_test.go
jmank88: I had the same thought, but I'm not sure. Just following local style.

PR #187, opened 2023-10-04
BCF-2684: copy types from core to support relayer implementations

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2684

This PR introduces types copied from core to make them available for relayer implementations. These types in particular are required for moving `core/chains/solana` to the chainlink-solana repo.


pkg/services/services.go
```diff
@@ -0,0 +1,239 @@
...
+	return fmt.Sprintf("service is %q, not started", e.state)
+}
+
+var (
+	ErrAlreadyStopped      = errors.New("already stopped")
+	ErrCannotStopUnstarted = errors.New("cannot stop unstarted service")
+)
+
+// StateMachine contains a State integer
+type StateMachine struct {
```
jmank88, line 29: This is only type that isn't a straight copy. It is functionally identical to the latest core version, but with some things renamed and/or unexported.

pkg/utils/start_stop_once.go
```diff
@@ -45,6 +45,7 @@ func (s startStopOnceState) String() string {
 }
 
 // StartStopOnce can be embedded in a struct to help implement types.Service.
+// Deprecated: use services.StateMachine
```
jmank88, line 48: This version was stale.

aalu1418 (dismissed):
overall looks fine to me since this is pretty much just copying from core

linter error about `testing`?

aalu1418 (approved):

PR #189, opened 2023-10-04
pkg/services/srvctest: add helper for testing HealthReport names

jmank88:
Add a `srvctest.AssertHealthReportNames` helper for verifying service names from `HealthReport()` results.

### DIFF
--- a/pkg/services/srvctest/health.go
+++ b/pkg/services/srvctest/health.go
@@ -1,0 +1,17 @@
+package srvctest
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+func AssertHealthReportNames(t *testing.T, hp map[string]error, names ...string) {
+	t.Helper()
+	keys := maps.Keys(hp)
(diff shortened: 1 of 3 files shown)


pkg/services/srvctest/health.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+func AssertHealthReportNames(t *testing.T, hp map[string]error, names ...string) {
```
jmank88, line 11: What are some other ways we will want to use this kind of API?
I imagine we will often want to verify they are all nil, or that some in particular are non-nil or match certain errors, but those things may be better done as a separate funcs :thinking:

jmank88: This might be too simple to be worth it:
```go
func AssertHealthy(t *testing.T, hp map[string]error) {
	t.Helper()
	require.NoError(t, errors.Join(maps.Values(hp)...))
}
```

jmank88:
This will be included in another PR that makes use of it.

PR #190, opened 2023-10-04
BCF-2687: start plugins without delay

jmank88:


pkg/loop/plugin_service.go
```diff
@@ -76,19 +94,7 @@ func (s *pluginService[P, S]) keepAlive() {
 		case <-s.stopCh:
 			return
 		case <-t.C:
```
patrickhuie19, line 96: Unrelated to the scope of this PR, but why have this cron based logic? If a job needs a healthy plugin, shouldn't the check for the healthiness (and corresponding relaunch) happen in the job code?

patrickhuie19 (approved):

pkg/loop/plugin_service.go
jmank88: There are many users of plugins, including the core application itself, so this service sort of operates like a shared client for anything that may need a plugin, with consolidated logic. Relaunching unhealthy plugins happens synchronously if necessary too, but we also need a keep alive loop to ensure it is running in the background even when not actively being used, since systems like the txm expect to be online at all times.

krehermann (approved):

PR #191, opened 2023-10-06
Telemetry server

george-dorin:


pkg/types/telemetry.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/smartcontractkit/libocr/commontypes"
+)
+
+type Telemetry interface {
+	Send(ctx context.Context, contractID string, telemetryType string, network string, chainID string, payload []byte) error
+}
+
+// MonitoringEndpointGenerator almost identical to synchronization.MonitoringEndpointGenerator except for the telemetry type
+type MonitoringEndpointGenerator interface {
+	GenMonitoringEndpoint(contractID string, telemType string, network string, chainID string) commontypes.MonitoringEndpoint
```
jmank88, line 15: nit/ my gut is to order these hierarchically, and it looks like you did that inside of the protobuf message:
```suggestion
	Send(ctx context.Context, network, chainID, contractID string, telemetryType string, payload []byte) error
}

// MonitoringEndpointGenerator almost identical to synchronization.MonitoringEndpointGenerator except for the telemetry type
type MonitoringEndpointGenerator interface {
	GenMonitoringEndpoint(network, chainID, contractID string, telemType string) commontypes.MonitoringEndpoint
```

jmank88:
How are these related to LOOPPs? I expected the types to be connected one way or another, but this seem disconnected so far. Where will it be used?

george-dorin:
> How are these related to LOOPPs? I expected the types to be connected one way or another, but this seem disconnected so far. Where will it be used?

Yes, at the moment they are disconnected, the telem service will be added to `reportingplugins.NewLOOPPService` method

pkg/types/telemetry.go
george-dorin: Updated in [`50f1349` (#191)](https://github.com/smartcontractkit/chainlink-relay/pull/191/commits/50f1349dffba24da1c5a3b03ea520f4ffcede1f5)

jmank88 (approved):

PR #193, opened 2023-10-09
BCF-2685: move types from core to support Solana plugin command extraction

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2685

Supports:
- https://github.com/smartcontractkit/chainlink-solana/pull/550
- https://github.com/smartcontractkit/chainlink/pull/10973

### DIFF
--- a/pkg/loop/adapter.go
+++ b/pkg/loop/adapter.go
@@ -1,0 +1,67 @@
+package loop
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/services"
+	"github.com/smartcontractkit/chainlink-relay/pkg/types"
+)
+
+// RelayerExt is a subset of [loop.Relayer] for adapting [types.Relayer], typically with a Chain. See [RelayerAdapter].
+type RelayerExt interface {
+	types.ChainService
+	ID() string
+}
+
+var _ Relayer = (*RelayerAdapter)(nil)
+
+// RelayerAdapter adapts a [types.Relayer] and [RelayerExt] to implement [Relayer].
+type RelayerAdapter struct {
+	types.Relayer
+	RelayerExt
+}
+
+func (r *RelayerAdapter) NewConfigProvider(ctx context.Context, rargs types.RelayArgs) (types.ConfigProvider, error) {
+	return r.Relayer.NewConfigProvider(rargs)
+}
+
+func (r *RelayerAdapter) NewMedianProvider(ctx context.Context, rargs types.RelayArgs, pargs types.PluginArgs) (types.MedianProvider, error) {
+	return r.Relayer.NewMedianProvider(rargs, pargs)
+}
+
+func (r *RelayerAdapter) NewMercuryProvider(ctx context.Context, rargs types.RelayArgs, pargs types.PluginArgs) (types.MercuryProvider, error) {
+	return r.Relayer.NewMercuryProvider(rargs, pargs)
+}
+
+func (r *RelayerAdapter) NewFunctionsProvider(ctx context.Context, rargs types.RelayArgs, pargs types.PluginArgs) (types.FunctionsProvider, error) {
+	return r.Relayer.NewFunctionsProvider(rargs, pargs)
+}
+
+func (r *RelayerAdapter) NewPluginProvider(ctx context.Context, rargs types.RelayArgs, pargs types.PluginArgs) (types.PluginProvider, error) {
+	return nil, fmt.Errorf("unexpected call to NewPluginProvider: did you forget to wrap RelayerAdapter in a relayerServerAdapter?")
+}
+
+func (r *RelayerAdapter) Start(ctx context.Context) error {
+	var ms services.MultiStart
+	return ms.Start(ctx, r.RelayerExt, r.Relayer)
+}
+
+func (r *RelayerAdapter) Close() error {
+	return services.CloseAll(r.Relayer, r.RelayerExt)
+}
+
+func (r *RelayerAdapter) Name() string {
+	return fmt.Sprintf("%s-%s", r.Relayer.Name(), r.RelayerExt.Name())
+}
+
+func (r *RelayerAdapter) Ready() (err error) {
+	return errors.Join(r.Relayer.Ready(), r.RelayerExt.Ready())
+}
+
+func (r *RelayerAdapter) HealthReport() map[string]error {
+	hr := make(map[string]error)
+	services.CopyHealth(hr, r.Relayer.HealthReport())
+	return hr
+}
--- a/pkg/loop/config.go
+++ b/pkg/loop/config.go
@@ -1,0 +1,116 @@
+package loop
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const (
+	envPromPort               = "CL_PROMETHEUS_PORT"
+	envTracingEnabled         = "CL_TRACING_ENABLED"
+	envTracingCollectorTarget = "CL_TRACING_COLLECTOR_TARGET"
+	envTracingSamplingRatio   = "CL_TRACING_SAMPLING_RATIO"
+	envTracingAttribute       = "CL_TRACING_ATTRIBUTE_"
+)
+
+// EnvConfig is the configuration between the application and the LOOP executable. The values
+// are fully resolved and static and passed via the environment.
+type EnvConfig struct {
+	PrometheusPort int
+
+	TracingEnabled         bool
+	TracingCollectorTarget string
+	TracingSamplingRatio   float64
+	TracingAttributes      map[string]string
+}
+
+// AsCmdEnv returns a slice of environment variable key/value pairs for an exec.Cmd.
+func (e *EnvConfig) AsCmdEnv() (env []string) {
+	injectEnv := map[string]string{
+		envPromPort:               strconv.Itoa(e.PrometheusPort),
+		envTracingEnabled:         strconv.FormatBool(e.TracingEnabled),
+		envTracingCollectorTarget: e.TracingCollectorTarget,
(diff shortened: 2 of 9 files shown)


pkg/loop/config.go
```diff
@@ -0,0 +1,35 @@
+package loop
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+)
+
+var EnvPromPort = "CL_PROMETHEUS_PORT"
```
jmank88, line 9: My instinct was to expose this as mutable so that `package main`s can make their own decisions about what it should be, but maybe this is not necessary :thinking:

pkg/loop/prom.go
```diff
@@ -0,0 +1,109 @@
...
+
+type promServerOpt func(*promServer)
+
+func withHandler(h http.Handler) promServerOpt {
+	return func(s *promServer) {
+		s.handler = h
+	}
+}
+
+// TODO there is one core/web test using this....
```
jmank88, line 35: I wanted to avoid exporting this, since the common helper code can handle it internally, but there is still one test in core using it. We may be able to do the same validation indirectly though :thinking:

pkg/loop/config.go
krehermann: the plugin can't choose the value - it must be guaranteed to be unique and only the core node has enough context to do that

pkg/loop/prom.go
krehermann: The core test is exercising the loop registration and service discovery.

I think we need that test, because is it is only test for the metric plumbing between the two sides.

Part of the contract btw the core node and a LOOPP is that a LOOPP ought to expose a prom server. We could promote that to a real interface somehow and then mock on the core side test for the registry.

or this repo could expose a mock/test implementation that runs a prom server that could be used for that test. 

or we could leave this public and keep the test as is.

i think 1 is the most principled thing to do, but doesn't fit in this PR, so maybe 2 is the best immediate option?

pkg/loop/prom_test.go
```diff
@@ -0,0 +1,56 @@
...
+	require.NoError(t, err)
+	require.Contains(t, string(b), "test_metric")
+	defer resp.Body.Close()
+
+	require.NoError(t, s.Close())
+}
+
+// Port is the resolved port and is only known after Start().
+// returns -1 before it is resolved or if there was an error during resolution.
+func (p *promServer) Port() int {
```
krehermann, line 47: why is this defined in a test file? is that an oversight?

pkg/loop/config.go
jmank88: >  it must be guaranteed to be unique and only the core node has enough context to do that.

That's why I think it's important to be mutable. And I'm trying to avoid making assumptions about the plugin implementations from here. A practical case could be if we want to write a test host for some reason.

pkg/loop/prom_test.go
jmank88: It was only used in tests unless I missed something.

pkg/loop/config.go
jmank88: Oh but we can choose to not inherit the host env with newer hashicorp, so maybe this is nbd.

jmank88: The name of the env var. If there was a collision to avoid, or a different standard to follow than `CL_` prefixing everything, then this would need to be mutable for that to be possible. But we don't want to inherit the host env when launching a plugin anyways, so collisions don't really matter in practice, and the `CL_` standard doesn't even need to apply here. Maybe we should actually considering using something else, e.g. `LOOPP_` :shrug:

pkg/loop/prom_test.go
jmank88: I'll connect the other layers in drafts before merging this to be sure :shrug:

patrickhuie19 (dismissed):

pkg/loop/config.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+const (
+	envLogSQLMigrations = "CL_LOG_SQL_MIGRATIONS"
+	envPromPort         = "CL_PROMETHEUS_PORT"
+)
+
+// EnvConfig is the configuration between the application and the LOOP executable. The values
+// are fully resolved and static and passed via the environment.
+type EnvConfig struct {
```
jmank88, line 17: Something felt off about the old API, and I think it's because this type was only an interface because it was capturing existing config types previous (for log fields), but that is no longer the case. A struct keeps things much simpler and more extensible (we can add fields without breaking users).

jmank88: Also allowed for simplifying the helpers to methods: `(*EnvConfig).SetCmdEnv`

pkg/loop/prom.go
```diff
@@ -0,0 +1,106 @@
...
+
+type PromServerOpts struct {
+	Handler http.Handler
+}
+
+func NewPromServer(port int, lggr logger.Logger) *PromServer {
+	return PromServerOpts{}.New(port, lggr)
+}
+
+func (o PromServerOpts) New(port int, lggr logger.Logger) *PromServer {
```
jmank88, line 35: @krehermann WDYT about dropping the functional options? I couldn't find satisfactory names (old ones weren't qualified enough in the context of this broader package), and there were extra allocations with the default handler. I'm open to other ideas :shrug:

pkg/loop/prom.go
jmank88: Exported again.

pkg/loop/adapter.go
```diff
@@ -0,0 +1,91 @@
...
+	return errors.Join(r.Relayer.Ready(), r.RelayerExt.Ready())
+}
+
+func (r *RelayerAdapter) HealthReport() map[string]error {
+	hr := make(map[string]error)
+	services.CopyHealth(hr, r.Relayer.HealthReport())
+	return hr
+}
+
+// TODO unused?!
```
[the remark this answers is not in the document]
jmank88: Indeed, this was unused :shrug:

pkg/loop/plugin.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"sync"
+
+	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
+	"github.com/smartcontractkit/chainlink-relay/pkg/services"
+)
+
+// Plugin is a base layer for plugins to easily manage sub-[types.Service]s.
+// Useful for implementing PluginRelayer and PluginMedian.
+type Plugin struct {
```
jmank88, line 12: This was called `plugins.Base` before.

pkg/loop/prom.go
jmank88: Oh the allocations is a separate problem.

Still not sure about the naming for functional opts though. `WithHandler` is too general for this package namespace, and even `WithPromHandler` has the same problem. So then is `WithPromServerHandler` the best option? That would mean that callers have to do:
`loop.NewPromServer(port, lggr, loop.WithPromServerHandler(handler))`
With an extra `loop.WithPromServer` per option, compared to a struct+method.

pkg/loop/config.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const (
+	envPromPort = "CL_PROMETHEUS_PORT"
+	//TODO CL_ prefix everything?
```
jmank88, line 13: WDYT? @krehermann @patrickhuie19

patrickhuie19: What's the problem we're solving? Distinguishing between the environment of the host and plugins?

patrickhuie19 (approved):

pkg/loop/config.go
jmank88: Namespacing so we don't have collisions. We do this in core. These aren't user facing so it's not as important, but in theory we could have a conflict with a dependency that has it's own env vars :shrug:

patrickhuie19: Hmm - If we prefix everything with `CL_`, then we're still at risk of collisions.

patrickhuie19 (approved):

pkg/loop/config.go
jmank88: Indeed, but they are less likely :shrug:

patrickhuie19 (approved):

pkg/loop/prom.go
krehermann: sorry, seems like being tagged isn't reach me clear thru github. i didn't realize you had asked my opinion here. aside: how do you handle that? it seems to be lost in a stream of other review comment in an email chain for me.

i find options creating an instance confusing. it's easier for me to understand for the constructor of the thing to take options.

is the only use case for testing right now? 

what about something like 

// WithHandler uses the given HTTP handler. It must be called before Start to have any effect
func (*PromServer) WithHandler(h) *PromServer

and the usage would be ~
s := NewPromServer(port,logger).WithHandler(h)

jmank88: > sorry, seems like being tagged isn't reach me clear thru github. i didn't realize you had asked my opinion here. aside: how do you handle that? it seems to be lost in a stream of other review comment in an email chain for me.

Mentions come through my email stream :shrug: 

> i find options creating an instance confusing. it's easier for me to understand for the constructor of the thing to take options.

Can you elaborate on this? I find the opposite, because functional options come with a ton of extra baggage and behave like 'spooky action at a distance'. They strike me as particularly odd when they are only setting fields rather than doing something complex or dynamic.

A `(*PromServer).WithHandler` method seems cleaner than options to me, but it still doesn't seem like it's in the right place if it comes with a caveat that you can only call it at certain times. A basic func or method constructor doesn't come with any extra baggage or caveats - it just does exactly what we need and reads very plainly.

Do you find `net.ListenConfig` difficult to read? https://pkg.go.dev/net#ListenConfig

jmank88:
Merged to unblock the growing stack of PRs, but I'm happy to circle back on anything.

PR #198, opened 2023-10-14
bump golangci-lint; add make golangci-lint

jmank88:
Also fixing the outstanding issues.


Makefile
```diff
@@ -23,4 +23,8 @@ generate: mockery install-protoc
 # add our installed protoc to the head of the PATH
 # maybe there is a cleaner way to do this
 	 PATH=$$HOME/.local/bin:$$PATH go generate -x ./...
-	  
\ No newline at end of file
+
+.PHONY: golangci-lint
+golangci-lint: ## Run golangci-lint for all issues.
+	[ -d "./golangci-lint" ] || mkdir ./golangci-lint && \
+	docker run --rm -v $(shell pwd):/app -w /app golangci/golangci-lint:v1.54.2 golangci-lint run --max-issues-per-linter 0 --max-same-issues 0 > ./golangci-lint/$(shell date +%Y-%m-%d_%H:%M:%S).txt
```
patrickhuie19, line 30: Nice - had the same idea here: https://github.com/smartcontractkit/chainlink-relay/pull/179/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

My reasoning was trying to bring more parity between local development and workflows in CI. From a local development perspective it might be nice to send the results directly to the shell

patrickhuie19 (approved):

Makefile
jmank88: Oh I just copied the command from core. I'd like to keep the development experience aligned between repos but I guess we don't have the same kind of concerns here without a massive list of outstanding issues :thinking:

jmank88: Oh there is even a separate custom action in core: https://github.com/smartcontractkit/chainlink/blob/develop/.github/actions/golangci-lint/action.yml

jmank88: I agree that is a priority, but we're trying to establish consistency across repos too.

PR #201, opened 2023-10-26
[BCF-2630] Add types for lightweight pipelinerunner service

cedric-cordenier:


cedric-cordenier:
See them used:
* https://github.com/smartcontractkit/chainlink/pull/11091/files

pkg/types/pipeline_runner.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+type Options struct {
+	MaxTaskDuration time.Duration
+}
+
+type TaskResult struct {
+	ID    string
+	Type  string
+	Value interface{}
```
jmank88, line 19: How do we send Vars and Values over the wire? Is there an appropriate interface we could swap in? e.g. `json.(Un)Marshaler`?

jmank88: Well I guess JSON is a bad example since it operates on `interface{}` :shrug:

cedric-cordenier: Vars will be translated to a `structpb.Struct` and Value will be translated to a `structpb.Value`. Those provide extensive support for interface{}

cedric-cordenier: Ah you mean: can we be more structured? Hmmm... let me have a think: I think we could make it more structured but that would require changing the pipeline runner itself (which would be worth it for the type safety).

jmank88: Oh I hadn't seen https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb. I suppose this would simplify some things in chainreader, if we committed to JSON there.

cedric-cordenier:
Closing in favour of a PR that implements the types and the wrapper as well.

PR #202, opened 2023-10-26
[BCF-2630] Add pipeline runner wrapper

cedric-cordenier:


pkg/loop/internal/pb/pipeline_runner.proto
```diff
@@ -0,0 +1,35 @@
...
+
+option go_package = "github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb";
+
+package loop;
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
+
+message Options {
+  google.protobuf.Duration max_task_duration = 2;
```
jmank88, line 11: I am surprised there is special type for `Duration`, since it is just `int64` in Go. Is this for language interoperability?

jmank88 (approved):

george-dorin (approved):

pkg/loop/internal/pb/pipeline_runner.proto
cedric-cordenier: I think so: Go's duration is limited to ~290 years according to the docs, whereas since this proto models seconds and nanoseconds in different int fields (int64 and int32) respectively, it can achieve longer durations.

PR #203, opened 2023-10-29
pkg/utils: deprecate StartStopOnce in favor of StateMachine

jmank88:
This PR refines the `StateMachine` type, moves some tests from core, and advances the deprecation of `StartStopOnce` (alias to `StateMachine` for now).

Supports:
- https://github.com/smartcontractkit/chainlink/pull/11117


pkg/services/state.go
```diff
@@ -145,19 +145,20 @@ func (once *StateMachine) StopOnce(name string, fn func() error) error {
 }
 
 // State retrieves the current state
-func (once *StateMachine) State() State {
-	state := once.state.Load()
-	return State(state)
+func (once *StateMachine) State() string {
```
jmank88, line 148: After some tweaks on the core side, the raw state doesn't need to be exposed anymore, but the string format is still used.

george-dorin (approved):

PR #204, opened 2023-10-29
pkg/services: add services.Config.NewService/Engine helper

jmank88:
This PR introduces `services.Config.NewService/NewServiceEngine` with `services.Engine` to help create `services.Service`s in a standard way that reduces boilerplate and the risk of mistakes that comes with re-implementing logger naming, health reporting, sub-service management, etc.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/13851


pkg/services/service.go
```diff
@@ -0,0 +1,224 @@
...
+// Config is a service configuration.
+type Config struct {
+	// Name is required. It will be logged shorthand on Start and Close, and appended to the logger name.
+	// It must be unique among services sharing the same logger, in order to ensure uniqueness of the fully qualified name.
+	Name string
+	// SubServices is an optional set of dependent Services to Start and Close along with this one.
+	NewSubServices func(logger.Logger) []Service
+	// Start is an optional hook called after starting SubServices.
+	Start func(context.Context) error
+	//TODO Close hook?
```
patrickhuie19, line 114: +1 to close hook

pkg/services/service_example_configured_test.go
```diff
@@ -0,0 +1,102 @@
...
+	if err = s.Start(context.Background()); err != nil {
+		fmt.Println("Failed to start service:", err)
+		return
+	}
+	if err = s.Close(); err != nil {
+		fmt.Println("Failed to close service:", err)
+		return
+	}
+
+	// Output:
```
patrickhuie19, line 87: Is there a txtar like way to test against this output? That might be too brittle since the sub services are closed concurrently.

pkg/internal/example/example.go
```diff
@@ -0,0 +1,34 @@
+// Package example contains helpers for implementing testable examples.
```
patrickhuie19, line 1: Do you have an example in mind? I'm trying to understand how this would be consumed.

pkg/services/service_example_configured_test.go
jmank88: > Is there a txtar like way to test against this output?

It is already built-in. These execute like regular tests.
https://go.dev/blog/examples

> That might be too brittle since the sub services are closed concurrently.

We can probably add a test-only hack to make the close deterministic :thinking:

pkg/internal/example/example.go
jmank88: Oh right I had set them up to be error-less, but it disrupts the test suite, so I switched back to early returns instead.

pkg/services/service.go
```diff
@@ -0,0 +1,224 @@
...
+	w.condsMu.RUnlock()
+	return errors.Join(errs...)
+}
+
+// Config is a service configuration.
+type Config struct {
+	// Name is required. It will be logged shorthand on Start and Close, and appended to the logger name.
+	// It must be unique among services sharing the same logger, in order to ensure uniqueness of the fully qualified name.
+	Name string
+	// SubServices is an optional set of dependent Services to Start and Close along with this one.
```
jmank88, line 110: - [ ] update stale comment

pkg/services/service.go
```diff
@@ -0,0 +1,228 @@
...
+// Do not override this method in your service. Instead, report errors via the Engine.
+func (s *service) Healthy() (err error) {
+	err = s.StateMachine.Healthy()
+	if err == nil {
+		err = s.eng.healthy()
+	}
+	return
+}
+
+func (s *service) HealthReport() map[string]error {
```
jmank88, line 177: As a fail-safe, we could track when this is called, and if it is never called, or not called often enough, then we could periodically emit to logs instead :thinking:

pkg/services/service.go
```diff
@@ -0,0 +1,232 @@
...
+	delete(e.conds, condition)
+}
+
+// Unhealthy causes an unhealthy report, until the returned func() is called.
+// Use this for simple cases where the func() can be kept in scope, and prefer to defer it inline if possible:
+//
+//	defer Unhealthy(fmt.Errorf("foo bar: %i", err))()
+//
+// See SetUnhealthy for an alternative API.
+func (e *Engine) Unhealthy(err error) func() {
```
jmank88, line 96: I am still not satisfied with these names. I think that these four different operations are fundamental, but I'm not sure what to call them that would be more intuitive than this... :thinking:

pkg/services/service.go
```diff
@@ -0,0 +1,232 @@
...
+			if err := s.cfg.Start(ctx); err != nil {
+				return fmt.Errorf("failed to start service %s: %w", s.cfg.Name, err)
+			}
+		}
+		s.eng.Info("Started")
+		return nil
+	})
+}
+
+func (s *service) Close() error {
```
jmank88, line 213: @DeividasK We could prioritize Stop while adapting Close easily here without spamming it in every implementation.
```suggestion
func (s *service) Close() error { return s.Stop() }
func (s *service) Stop() error {
```

pkg/services/service.go
```diff
@@ -0,0 +1,238 @@
...
+	// Start is an optional hook called after starting SubServices.
+	Start func(context.Context) error
+	// Close is an optional hook called before closing SubServices.
+	Close func() error
+}
+
+// NewEngine returns a new Service defined by Config, and an Engine for managing health, goroutines, and logging.
+//   - You *should* embed the Service, in order to inherit the methods.
+//   - You *should not* embed the Engine. Use an unexported field instead.
+func (c Config) NewEngine(lggr logger.Logger) (Service, *Engine) {
```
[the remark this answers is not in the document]
jmank88: Done

pkg/services/service_example_configured_test.go
jmank88: I think there is still some non-determinism in this example

jmank88:
I have added a core side PR with a few example uses https://github.com/smartcontractkit/chainlink/pull/13851

pkg/services/service_example_configured_test.go
jmank88: "Fixed" by commenting out (the output) with a note so it does not run in the test suite. It can still be invoked from the docs site.

samsondav (approved):

pkg/services/service.go
jmank88: I'm still open to this, but as a follow-up after this PR.

Atrax1 (approved):

PR #205, opened 2023-10-30
[BCF-2728] Add telemetry service to LOOPP dependencies

cedric-cordenier:


pkg/loop/internal/reporting_plugin_service.go
```diff
@@ -26,29 +26,44 @@ func NewReportingPluginServiceClient(broker Broker, brokerCfg BrokerConfig, conn
...
 	cc := m.newClientConn("ReportingPluginServiceFactory", func(ctx context.Context) (id uint32, deps resources, err error) {
 		providerID, providerRes, err := m.serve("PluginProvider", proxy.NewProxy(grpcProvider))
 		if err != nil {
 			return 0, nil, err
 		}
 		deps.Add(providerRes)
 
-		errorLogID, errorLogRes, err := m.serveNew("ErrorLog", func(s *grpc.Server) {
-			pb.RegisterErrorLogServer(s, &errorLogServer{impl: errorLog})
+		pipelineRunnerID, pipelineRunnerRes, err := m.serveNew("PipelineRunner", func(s *grpc.Server) {
```
cedric-cordenier, line 44: Sorry for the noisy diff here: I reordered the deps in the order in which they appear at the top of the function.

pkg/loop/internal/telemetry.go
```diff
@@ -3,70 +3,56 @@ package internal
...
 
-	"github.com/smartcontractkit/chainlink-relay/pkg/logger"
 	"github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb"
 	"github.com/smartcontractkit/chainlink-relay/pkg/types"
 )
 
-var _ types.Telemetry = (*telemetryClient)(nil)
-var _ types.MonitoringEndpointGenerator = (*telemetryClient)(nil)
-var _ commontypes.MonitoringEndpoint = (*telemetryEndpoint)(nil)
+var _ types.TelemetryClient = (*telemetryClient)(nil)
```
cedric-cordenier, line 14: Note: I relaxed the interfaces here and allowed the server and client-side interfaces to diverge.

I think the fact that we were validating inputs while having an interface which didn't allow errors to be returned to callers directly was making the API awkward, and opted to relax that.

george-dorin (dismissed):

pkg/types/telemetry.go
```diff
@@ -6,10 +6,19 @@ import (
...
 	Send(ctx context.Context, network string, chainID string, contractID string, telemetryType string, payload []byte) error
 }
 
+type TelemetryClientEndpoint interface {
+	SendLog(ctx context.Context, log []byte) error
+}
+
+type TelemetryClient interface {
+	TelemetryService
+	NewEndpoint(nework string, chainID string, contractID string, telemetryType string) (TelemetryClientEndpoint, error)
```
jmank88, line 19: I think we want to include ctx and error by default on everything new. Any reason not to here?
```suggestion
	NewEndpoint(ctx context.Context, nework string, chainID string, contractID string, telemetryType string) (TelemetryClientEndpoint, error)
```

pkg/loop/internal/test/reporting_plugin.go
```diff
@@ -158,8 +158,16 @@ func (s StaticReportingPluginWithMedianProvider) NewReportingPluginFactory(ctx c
...
-	if err := errorLog.SaveError(ctx, errMsg); err != nil {
-		return nil, fmt.Errorf("failed to save error: %w", err)
+	if err2 := errorLog.SaveError(ctx, errMsg); err2 != nil {
+		return nil, fmt.Errorf("failed to save error: %w", err2)
+	}
+	endpoint, err := telemetry.NewEndpoint(network, chainID, contractID, telemType)
+	if err != nil {
+		return nil, fmt.Errorf("failed to instantiate endpoint: %w", err)
+	}
+	err = endpoint.SendLog(ctx, payload)
```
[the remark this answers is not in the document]
cedric-cordenier: The analogous usages of this in the core node tend to instantiate a single endpoint (eg. in the NewReportingPluginFactory function and then use the generated endpoint.SendLog... to emit Telemetry data.

pkg/types/telemetry.go
cedric-cordenier: No, this was just because I renamed the method from GenMonitoringEndpoint -- definitely agree we want to pass in a context here :)

jmank88 (approved):

PR #206, opened 2023-10-31
pkg/services/srvctest: add helper for testing HealthReport names

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2709


pkg/loop/internal/core/services/capability/standard/standard_capabilities.go
jmank88, line 1: Aligning with other constructors

pkg/loop/internal/core/services/oraclefactory/client.go
jmank88, line 1: Aligning with other constructors. The variable reassignment is deliberate, to prevent misuse of the original variable afterwards.

pkg/loop/plugin_relayer.go
```diff
@@ -64,5 +63,8 @@ func (p *GRPCPluginRelayer) ClientConfig() *plugin.ClientConfig {
 		HandshakeConfig: PluginRelayerHandshakeConfig(),
 		Plugins:         map[string]plugin.Plugin{PluginRelayerName: p},
 	}
-	return ManagedGRPCClientConfig(c, p.BrokerConfig)
```
jmank88, line 67: Passing `p.BrokerConfig` here bypasses the service hierarchy. We need the `p.pluginClient.BrokerConfig`, which wasn't available yet, since it depended on the broker and connect as well. Now it can be constructed without them.

pkg/loop/mercury_service.go
jmank88, line 1: The simplification of having our `newService` funcs do construction made the original cast type inaccessible. However, we need it for health reports. This pattern allows returning both (or just one which satisfies both interfaces), but it feels messy. I'm open to better suggestions.

pkg/loop/internal/core/services/capability/standard/standard_capabilities.go
jmank88, line 1: Aligning with other constructors

pkg/loop/internal/core/services/oraclefactory/client.go
jmank88, line 1: Aligning with other constructors. The variable reassignment is deliberate, to prevent misuse of the original variable afterwards.

pkg/loop/internal/core/services/reportingplugin/ocr2/reporting.go
jmank88, line 1: It is important to do this reassignment first, rather than inline, so that the other uses of b use the reassigned value (with the named logger).

pkg/loop/mercury_service.go
jmank88, line 1: The simplification of having our `newService` funcs do construction made the original cast type inaccessible. However, we need it for health reports. This pattern allows returning both (or just one which satisfies both interfaces), but it feels messy. I'm open to better suggestions.

pkg/loop/internal/goplugin/plugin.go
```diff
@@ -12,10 +12,10 @@ type PluginClient struct {
 	*net.BrokerExt
 }
 
-func NewPluginClient(broker net.Broker, brokerCfg net.BrokerConfig, conn *grpc.ClientConn) *PluginClient {
+// NewPluginClient creates a *PluginClient. Refresh must be called to initialize the net.Broker and *grpc.ClientConn.
+func NewPluginClient(brokerCfg net.BrokerConfig) *PluginClient {
 	var pc PluginClient
 	pc.BrokerExt = &net.BrokerExt{Broker: &pc.AtomicBroker, BrokerConfig: brokerCfg}
-	pc.Refresh(broker, conn)
```
krehermann, line 18: why did this go away? where is it now?

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -43,21 +43,30 @@ type PluginService[P grpcPlugin, S services.Service] struct {
...
 	clientProtocol plugin.ClientProtocol
 
-	newService func(context.Context, any) (S, error)
+	newService NewService[S]
 
-	serviceCh chan struct{} // closed when service is available
-	Service   S
+	serviceCh      chan struct{} // closed when service is available
+	Service        S
+	HealthReporter services.HealthReporter // may be the same as Service
```
krehermann, line 50: please add more commentary about the conditions when the HealthReporter would be the same as Service. 

i have no intuition about it and it would take me a long time read the code to understand if i ought to expect them to the be the same thing or not

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -43,21 +43,30 @@ type PluginService[P grpcPlugin, S services.Service] struct {
...
 
 	testInterrupt chan func(*PluginService[P, S]) // tests only (via TestHook) to enable access to internals without racing
 }
 
-func (s *PluginService[P, S]) Init(pluginName string, p P, newService func(context.Context, any) (S, error),
-	lggr logger.Logger, cmd func() *exec.Cmd, stopCh chan struct{}) {
+type NewService[S any] func(context.Context, any) (S, services.HealthReporter, error)
```
krehermann, line 55: why is S unconstrained? its it really true that S = type `int`, for example, could possible be correct?

pkg/loop/internal/goplugin/plugin_service.go
krehermann: also, is this the primary qualitative change and all/most of the other diffs are knock effects?

pkg/loop/plugin_median.go
```diff
@@ -60,5 +59,8 @@ func (p *GRPCPluginMedian) ClientConfig() *plugin.ClientConfig {
 		HandshakeConfig: PluginMedianHandshakeConfig(),
 		Plugins:         map[string]plugin.Plugin{PluginMedianName: p},
 	}
-	return ManagedGRPCClientConfig(c, p.BrokerConfig)
+	if p.pluginClient == nil {
```
krehermann, line 62: why is this repeated relative to line 49?

seem like that is a new pattern everywhere. what's the fundamental change that provoked the pattern?

krehermann (commented):
first pass, trying to obtain more context

pkg/loop/internal/goplugin/plugin_service.go
jmank88: It is allowed to be, but this code doesn't know or care about why. It only happens when the top level `Plugin*` type is the one returned, rather than constructing another secondary type like we do in most places (`PluginRelayer`->`Relayer`)

pkg/loop/internal/goplugin/plugin.go
jmank88: The callers are responsible for calling `Refresh`:
> Refresh must be called to initialize the net.Broker and *grpc.ClientConn.

pkg/loop/plugin_median.go
jmank88: https://github.com/smartcontractkit/chainlink-common/pull/206#discussion_r1907337647

pkg/loop/internal/goplugin/plugin_service.go
jmank88: Documented on `NewService` now

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -43,21 +43,32 @@ type PluginService[P grpcPlugin, S services.Service] struct {
...
+	Service        S
+	HealthReporter services.HealthReporter // may be the same as Service
 
 	testInterrupt chan func(*PluginService[P, S]) // tests only (via TestHook) to enable access to internals without racing
 }
 
-func (s *PluginService[P, S]) Init(pluginName string, p P, newService func(context.Context, any) (S, error),
-	lggr logger.Logger, cmd func() *exec.Cmd, stopCh chan struct{}) {
+// NewService funcs returns an S and a HeathReporter, which should provide the top level health report for the whole
+// plugin, and which may be the same as S.
```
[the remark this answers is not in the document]
jmank88: Would this be better documented on the PluginService? This type declaration is really just a convenience. It doesn't do anything, and it's use is not restricted in any way.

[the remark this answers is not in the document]
krehermann: yes, sounds like a good place for it

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -172,14 +183,33 @@ func (s *PluginService[P, S]) Ready() error {
...
+	case <-s.stopCh:
+		return map[string]error{s.Name(): fmt.Errorf("service was stoped while waiting: %w", context.Canceled)}
+	case <-ctx.Done():
 		return map[string]error{s.Name(): ErrPluginUnavailable}
+	case <-s.serviceCh:
+	}
+
+	hr := map[string]error{s.Name(): s.Healthy()}
+
+	// wait until service is ready, which also triggers the deferred construction to ensure a complete HealthReport
```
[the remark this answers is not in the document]
krehermann: seems reasonable to me. we either have to pull or push; changing to push would be a lot of work for little gain afaict

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -172,14 +183,33 @@ func (s *PluginService[P, S]) Ready() error {
...
+	ctx, cancel := s.stopCh.CtxWithTimeout(5 * time.Second)
+	defer cancel()
+
 	select {
-	case <-s.serviceCh:
-		hr := map[string]error{s.Name(): s.Healthy()}
-		services.CopyHealth(hr, s.Service.HealthReport())
-		return hr
-	default:
+	case <-s.stopCh:
```
patrickhuie19, line 190: This flow isn't intuitive to me - we're creating a ctx + cancel from `s.stopCh`, but then handling the case where stopCh is closed differently from the context being canceled. What am I missing?

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -36,28 +36,39 @@ type PluginService[P grpcPlugin, S services.Service] struct {
...
 	client         *plugin.Client
 	clientProtocol plugin.ClientProtocol
 
-	newService func(context.Context, any) (S, error)
+	newService NewService[S]
 
-	serviceCh chan struct{} // closed when service is available
-	Service   S
+	serviceCh      chan struct{} // closed when service is available
+	Service        S
```
patrickhuie19, line 49: Do S and HealthReport needed to be synchronously closed when PluginService is closed?

pkg/loop/internal/goplugin/plugin_service.go
jmank88: Ah, right. We want to handle the cases differently, so I guess we need to use a background context.

pkg/loop/internal/goplugin/plugin_service.go
jmank88: It shouldn't really matter :thinking:

patrickhuie19: Or log after a test close

patrickhuie19 (approved):

krehermann (approved):

PR #209, opened 2023-11-01
Creates a codec interface for encoding and decoding data to a chain.

nolag:
Creates a codec interface for encoding and decoding data to a chain.  Implements the relay chain reader to use it and creates interface tests for the chain reader to be run by all implementations

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,12 +4,14 @@ go 1.21
 
 require (
 	github.com/confluentinc/confluent-kafka-go v1.9.2
+	github.com/fxamacker/cbor/v2 v2.5.0
 	github.com/google/uuid v1.3.1
 	github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus v1.0.0
 	github.com/hashicorp/go-hclog v1.5.0
 	github.com/hashicorp/go-plugin v1.5.2
 	github.com/jpillora/backoff v1.0.0
 	github.com/linkedin/goavro/v2 v2.12.0
+	github.com/mitchellh/mapstructure v1.5.0
 	github.com/mwitkow/grpc-proxy v0.0.0-20230212185441-f345521cb9c9
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.17.0
@@ -40,6 +42,7 @@ require (
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/golang/snappy v0.0.5-0.20220116011046-fa5810519dcb // indirect
+	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.0.0-rc.3 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.16.0 // indirect
 	github.com/hashicorp/yamux v0.0.0-20180604194846-3520598351bb // indirect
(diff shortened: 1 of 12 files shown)


pkg/types/interfacetests/codec_interface_tests.go
```diff
@@ -0,0 +1,280 @@
...
+	TestItemArray2Type = "TestItemArray2Type"
+)
+
+// CreateChainReaderInterfaceTests uses TestStruct and TestStructWithSpecialFields
+func CreateChainReaderInterfaceTests(tester ChainReaderInterfaceTester) bool {
+	ctx := context.Background()
+	BeforeEach(tester.Setup)
+	AfterEach(tester.Teardown)
+	return Describe("Codec tests for "+tester.Name(), func() {
+		It("Encodes and decodes a single item", func() {
```
jmank88, line 52: I have personally always avoided this framework because this style of test naming doesn't make any sense to me in go. Is there a clean way to invoke individual sub-tests? Without e.g. replacing all the spaces with underscores?

go.mod
```diff
@@ -4,13 +4,17 @@ go 1.21
...
 	github.com/google/uuid v1.3.1
 	github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus v1.0.0
 	github.com/hashicorp/go-hclog v1.5.0
 	github.com/hashicorp/go-plugin v1.5.2
 	github.com/jpillora/backoff v1.0.0
 	github.com/linkedin/goavro/v2 v2.12.0
+	github.com/mitchellh/mapstructure v1.5.0
 	github.com/mwitkow/grpc-proxy v0.0.0-20230212185441-f345521cb9c9
+	github.com/onsi/ginkgo v1.14.1
+	github.com/onsi/gomega v1.29.0
```
jmank88, line 17: We started moving away from these recently for various reasons. Are you using any particular features of these libraries? If not, pure go tests would be preferred. (especially in this repo, where we are trying to keep dependencies light)

nolag: Ack, I can see if there's another common way to do these kinds of tests.  Just used to doing it that way.

pkg/types/interfacetests/codec_interface_tests.go
nolag: I'll try to remove the library, per your other comment :).  It is convention with these libraries to do this kind of naming.  I'll do subtests instead or something else.

pkg/types/interfacetests/codec_interface_tests.go
```diff
@@ -0,0 +1,281 @@
...
+			codec := tester.GetChainReader()
+			err := codec.Decode(ctx, resp, &items, TestItemSliceType)
+			Expect(err).To(Equal(types.InvalidEncodingError{}))
+		})
+	})
+	return basicTests && strictTests
+}
+
+func Must[T any](t T, err error) T {
+	Expect(err).ToNot(HaveOccurred())
```
jmank88, line 253: How does a failure from this point manifest? Does this framework panic internally?

go.mod
jmank88: Sounds good. I think the high level structure maps to regular subtests well, and we usually lean on testify require/assert.

pkg/types/interfacetests/codec_interface_tests.go
nolag: I think it panics internally, I'm not 100% sure.  Never needed to dig into it.

When initializing it is passed t, it may use something from there.   That code is in a different file.   I'll remove the library anyway.

`RunSpecs(t, "loop internal interface tests")`

jmank88 (approved):

jmank88:
Oh it looks like the linter is complaining

nolag:
> Oh it looks like the linter is complaining

It was actually a good one this time too.  I made a slight change to make a test pass and I shadowed err without noticing.

PR #211, opened 2023-11-02
Add map decoder to be used by EVM and other implementations.

nolag:


pkg/loop/internal/chain_reader_test.go
```diff
@@ -2,17 +2,19 @@ package internal
 
 import (
 	"errors"
-	"github.com/stretchr/testify/assert"
```
nolag, line 5: Lint fixed this in this pr but not the other, not sure why, but I don't feel like it needs to be reverted...

jmank88 (approved):

jmank88 (approved):

PR #212, opened 2023-11-02
Add merge and split fields utilities to be used by encoders

nolag:


pkg/utils/utils_test.go
```diff
@@ -1,12 +1,14 @@
 package utils_test
 
 import (
+	"github.com/stretchr/testify/require"
```
jmank88, line 4: golangci-lint usually flags these for sorting

nolag: thought I did fix, but I might have merged after or something, I'll run again, sorry.

pkg/loop/internal/chain_reader_test.go
```diff
@@ -2,17 +2,19 @@ package internal
...
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/test/bufconn"
+
 	"github.com/stretchr/testify/require"
 
 	"github.com/smartcontractkit/chainlink-relay/pkg/loop/internal/pb"
 
 	"context"
```
jmank88, line 16: Pre-existing, but there's a std lib context down here?

cc @chainchad seems like the linter is failing in this repo too

jmank88 (approved):

PR #214, opened 2023-11-03
migrate common utils from core/internal/testutils/testutils.go

dhaidashenko:


pkg/utils/tests/tests.go
```diff
@@ -33,3 +35,22 @@ func WaitTimeout(t *testing.T) time.Duration {
...
+const TestInterval = 100 * time.Millisecond
+
+// AssertEventually waits for f to return true
+func AssertEventually(t *testing.T, f func() bool) {
+	assert.Eventually(t, f, WaitTimeout(t), TestInterval/2)
+}
+
+// WaitWithTimeout waits for the channel to close (or receive anything) and
+// fatals the test if the default wait timeout is exceeded
+func WaitWithTimeout(t *testing.T, ch <-chan struct{}, failMsg string) {
```
jmank88, line 50: Would it make sense to name this one with the `Require` convention to convey that it does a fatal exit?
```suggestion
func RequireSignal(t *testing.T, ch <-chan struct{}, failMsg string) {
```

pkg/utils/tests/tests.go
```diff
@@ -33,3 +35,22 @@ func WaitTimeout(t *testing.T) time.Duration {
 	}
 	return DefaultWaitTimeout
 }
+
+// TestInterval is just a sensible poll interval that gives fast tests without
+// risk of spamming
+const TestInterval = 100 * time.Millisecond
```
jmank88, line 41: Does this one need to be exported?

pkg/utils/tests/logs.go
```diff
@@ -0,0 +1,56 @@
...
+// WaitForLogMessage waits until at least one log message containing the
+// specified msg is emitted.
+// NOTE: This does not "pop" messages so it cannot be used multiple times to
+// check for new instances of the same msg. See WaitForLogMessageCount instead.
+//
+// Get a *observer.ObservedLogs like so:
+//
+//	observedZapCore, observedLogs := observer.New(zap.DebugLevel)
+//	lggr := logger.TestLogger(t, observedZapCore)
+func WaitForLogMessage(t *testing.T, observedLogs *observer.ObservedLogs, msg string) {
```
jmank88, line 19: Maybe these should be renamed for consistency too?
```suggestion
func AssertLogMessageEventually(t *testing.T, observedLogs *observer.ObservedLogs, msg string) {
```
```go
AssertLogMessageCountEventually
```
Kinda verbose though :shrug:

dhaidashenko: AssertLogEventually or AssertLogMsgEventually?

dhaidashenko: Also we'll have to do the same for WaitForLogMessageCount

jmank88 (approved):

PR #218, opened 2023-11-03
Add get max size to chain reader.

nolag:
I'll need this to fully replace codec calls in median.  Haven't prototyped the how yet, but seems pretty simple given it can assume there's only one level of slice and everything in it will then be fixed size :).


reductionista:
Looks like you need to add something like this to `config.go`:
```
GetMaxSize(ctx context.Context, n int, itemType string) (int, error) {
	return 0, errors.New("not used for these tests")
}
```

pkg/types/codec.go
```diff
@@ -8,6 +8,11 @@ import (
 
 type Encoder interface {
 	Encode(ctx context.Context, item any, itemType string) (types.Report, error)
+	// GetMaxSize returns the max size in bytes if n elements are supplied for all top level dynamically sized elements.
+	// If no elements are dynamically sized, the returned value will be the same for all n.
+	// If there are multiple levels of dynamically sized elements, or itemType cannot be found,
+	// InvalidTypeError will be returned.
+	GetMaxSize(ctx context.Context, n int, itemType string) (int, error)
 }
```
reductionista, line 16: I wonder if we might end up needing something similar for the `Decoder`? ie, would it make sense for a product plugin to ask for a maximum size of data that could be returned, before deciding on what sort of data structure to allocate to hold the retVal?

nolag: Sounds reasonable, I can put it in both.  The current use was when encoding the median report.  My goal was to change the current median codec to use this for encoding and decoding at the same time we changed the use of get the latest value function :)..

PR #219, opened 2023-11-03
Fix Tracing Dial to Otel Collector

patrickhuie19:
Corresponding core PR: https://github.com/smartcontractkit/chainlink/pull/11172

Previously, the trace exporter would silently fail if there was a failure to connect to the otel collector. This fixes dial issues with the otel collector and also ensures there is a log line for failures. 

The default log line when the otel-collector is un-dialable is plain text, which won't show up in grafana. 

Log line:
```
failed to dial: dial tcp: lookup otel-collector on 127.0.0.11:53: no such host
```

We've added `OnDialError` to the `TracingConfig` struct so that consumers of `SetupTracing` can wire a logger through.

### DIFF
--- a/pkg/loop/telem.go
+++ b/pkg/loop/telem.go
@@ -2,7 +2,6 @@ package loop
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"os"
 	"runtime/debug"
@@ -39,6 +38,9 @@ type TracingConfig struct {
 
 	// SamplingRatio is the ratio of traces to sample. 1.0 means sample all traces.
 	SamplingRatio float64
+
+	// OnDialError is called when the dialer fails, providing an opportunity to log.
+	OnDialError func(error)
 }
 
 // NewGRPCOpts initializes open telemetry and returns GRPCOpts with telemetry interceptors.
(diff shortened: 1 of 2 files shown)


pkg/loop/telem.go
```diff
@@ -58,13 +61,24 @@ func SetupTracing(config TracingConfig) error {
...
-	// Shutting down the traceExporter will not shutdown the underlying connection.
-	traceExporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithEndpoint(config.CollectorTarget), otlptracegrpc.WithDialOption(
+	conn, err := grpc.DialContext(ctx, config.CollectorTarget,
 		// Note the use of insecure transport here. TLS is recommended in production.
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
-	))
+		grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
+			conn, err := net.Dial("tcp", s)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "failed to dial: %v", err)
```
jmank88, line 73: Something straightforward seems simplest to me :shrug: 
```suggestion
				config.OnDialError(err)
```
```diff
type TracingConfig struct {
	// NodeAttributes are the attributes to attach to traces.
	NodeAttributes map[string]string
	// Enables tracing; requires a collector to be provided
	Enabled bool
	// Collector is the address of the OTEL collector to send traces to.
	CollectorTarget string
	// SamplingRatio is the ratio of traces to sample. 1.0 means sample all traces.
	SamplingRatio float64
+	// OnDialError is called when the dialer fails, providing an opportunity to log.
+	OnDialError func(error)
}
```
```go
config.OnDialError = func(err error) { lggr.Errorw("Failed to dial", "err", err) }
```

krehermann (approved):

PR #222, opened 2023-11-07
mercury: Add ChainReader to MercuryProvider

brunotm:
This change adds ChainReader() (ChainHead) interface and type to the MercuryProvider relayer to enable the removal of the plugin dependency on the raw evm chain.

### DIFF
--- a/pkg/reportingplugins/mercury/types.go
+++ b/pkg/reportingplugins/mercury/types.go
@@ -44,10 +44,10 @@ type Transmitter interface {
 
 type ChainReader interface {
 	// LatestBlocks returns an ordered list of the latest specified number of blocks
-	LatestBlocks(context.Context, int) ([]Block, error)
+	LatestBlocks(context.Context, int) ([]Head, error)
 }
 
-type Block struct {
+type Head struct {
 	Number    uint64
 	Hash      []byte
 	Timestamp uint64
--- a/pkg/reportingplugins/mercury/types.go
+++ b/pkg/reportingplugins/mercury/types.go
@@ -43,8 +43,8 @@ type Transmitter interface {
 }
 
 type ChainReader interface {
-	// LatestBlocks returns an ordered list of the latest specified number of blocks
-	LatestBlocks(context.Context, int) ([]Head, error)
+	// LatestHeads returns an ordered list of the latest specified number of heads
+	LatestHeads(context.Context, int) ([]Head, error)
 }
 
 type Head struct {
--- a/pkg/reportingplugins/mercury/types.go
+++ b/pkg/reportingplugins/mercury/types.go
@@ -43,8 +43,6 @@ type Transmitter interface {
 }
 
 type ChainReader interface {
-	// Head returns the current chain head
-	Head(context.Context) (Block, error)
 	// LatestBlocks returns an ordered list of the latest specified number of blocks
 	LatestBlocks(context.Context, int) ([]Block, error)
 }
--- a/pkg/reportingplugins/mercury/types.go
+++ b/pkg/reportingplugins/mercury/types.go
@@ -41,3 +41,16 @@ type Transmitter interface {
 	// - FromAccount() should return CSA public key
 	ocrtypes.ContractTransmitter
 }
+
+type ChainReader interface {
+	// Head returns the current chain head
+	Head(context.Context) (Block, error)
+	// LatestBlocks returns an ordered list of the latest specified number of blocks
+	LatestBlocks(context.Context, int) ([]Block, error)
+}
+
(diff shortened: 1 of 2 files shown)


samsondav (dismissed):

pkg/reportingplugins/mercury/types.go
```diff
@@ -41,3 +41,13 @@ type Transmitter interface {
 	// - FromAccount() should return CSA public key
 	ocrtypes.ContractTransmitter
 }
+
+type ChainReader interface {
```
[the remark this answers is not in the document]
brunotm: Here: https://github.com/smartcontractkit/chainlink/pull/11201

krehermann (approved):

pkg/reportingplugins/mercury/types.go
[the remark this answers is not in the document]
brunotm: I had it in draft and still waiting on a pending config for the sigscanner check to pass.

jmank88 (dismissed):

krehermann (dismissed):

samsondav (approved):

pkg/reportingplugins/mercury/types.go
```diff
@@ -41,3 +41,16 @@ type Transmitter interface {
...
+	// Head returns the current chain head
+	Head(context.Context) (Block, error)
```
samsondav, line 47: This method seems redundant.

samsondav (changes requested):

pkg/reportingplugins/mercury/types.go
brunotm: Yes. Calling LatestBlocks(1) is enough. Left it as a placeholder as but it can be removed.

pkg/reportingplugins/mercury/types.go
```diff
@@ -41,3 +41,14 @@ type Transmitter interface {
 	// - FromAccount() should return CSA public key
 	ocrtypes.ContractTransmitter
 }
+
+type ChainReader interface {
+	// LatestBlocks returns an ordered list of the latest specified number of blocks
+	LatestBlocks(context.Context, int) ([]Block, error)
+}
+
+type Block struct {
```
jmank88, line 50: Oh should we consider calling these `Head`s, since they are not full blocks.
```suggestion
type Head struct {
```

brunotm: Was keeping it consistent with the naming in mercury_v1 package, but i agree as they are not full blocks. BlockDescriptor? BlockHeader?

jmank88: There is precedent for some version of `Header` from geth. Not sure about others.

brunotm: Changed to Head. Thanks!

pkg/reportingplugins/mercury/types.go
```diff
@@ -41,3 +41,14 @@ type Transmitter interface {
 	// - FromAccount() should return CSA public key
 	ocrtypes.ContractTransmitter
 }
+
+type ChainReader interface {
+	// LatestBlocks returns an ordered list of the latest specified number of blocks
+	LatestBlocks(context.Context, int) ([]Head, error)
```
jmank88, line 47: Let's align the method name too:
```suggestion
	// LatestHeads returns an ordered list of the latest specified number of heads
	LatestHeads(context.Context, int) ([]Head, error)
```

jmank88 (approved):

pkg/reportingplugins/mercury/types.go
```diff
@@ -41,3 +41,16 @@ type Transmitter interface {
 	// - FromAccount() should return CSA public key
 	ocrtypes.ContractTransmitter
 }
+
+type ChainReader interface {
+	// Head returns the current chain head
+	Head(context.Context) (Block, error)
+	// LatestBlocks returns an ordered list of the latest specified number of blocks
+	LatestBlocks(context.Context, int) ([]Block, error)
+}
```
reductionista, line 48: I'm still looking this over, but wanted to draw attention to this PR where we define `type ChainReader interface` as a part of PluginProvider:
https://github.com/smartcontractkit/chainlink-relay/pull/196

We should coordinate between these to make sure we don't have too many merge conflicts or conflicting definitions.

jmank88: What did you have in mind? Including this on the shared `ChainReader` was ruled out of scope, so we are staying out of the way by working in the encapsulated mercury provider.

samsondav (approved):

PR #226, opened 2023-11-07
Remove the map from inner test struct for the chain reader interface test and show testing with compatible types for encoding.

nolag:


pkg/types/interfacetests/codec_interface_tests.go
```diff
@@ -64,6 +64,61 @@ func RunChainReaderInterfaceTests(t *testing.T, tester ChainReaderInterfaceTeste
...
+			compatibleItem := compatibleTestStruct{
+				Account:        item.Account,
+				Accounts:       item.Accounts,
+				BigField:       item.BigField,
+				DifferentField: item.DifferentField,
+				Field:          item.Field,
+				NestedStruct:   item.NestedStruct,
+				OracleId:       item.OracleId,
+				OracleIds:      item.OracleIds,
+			}
```
reductionista, line 80: Ah, this one is actually a pretty important use case I think!

reductionista (approved):

PR #230, opened 2023-11-10
Add more helper methods in utils, use cases will be split into other PRs

nolag:


jmank88:
I've been trying to move things out of package utils since catch-all packages are an anti pattern in go. 
> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them

https://go.dev/blog/package-names

jmank88:
Are these strictly to support encoding/decoding?
What about a `package codec`? 🤷

pkg/utils/utils.go
```diff
@@ -44,3 +48,92 @@ func ContextWithDeadlineFn(ctx context.Context, deadlineFn func(orig time.Time)
 	}
 	return ctx, cancel
 }
+
+func FitsInNBitsSigned(n int, bi *big.Int) bool {
```
nolag, line 52: This will be used in chain reader implementations.

pkg/utils/utils.go
```diff
@@ -44,3 +48,92 @@ func ContextWithDeadlineFn(ctx context.Context, deadlineFn func(orig time.Time)
...
+	if bi.Sign() < 0 {
+		bi = new(big.Int).Neg(bi)
+		bi.Sub(bi, big.NewInt(1))
+	}
+	return bi.BitLen() <= n-1
+}
+
+func MergeValueFields(valueFields []map[string]any) (map[string]any, error) {
```
nolag, line 60: This function and the one below will be used to create a MapDecoder that can decode to maps.  That can be re-used so that implementations don't need to decode to interfaces directly.  The need for this comes from encoding for functions (or similar) where instead of an object representing fields, you get an array of objects squashed into one object and then encoded to the fields on that object.

nolag:
> Are these strictly to support encoding/decoding? What about a `package codec`? 🤷

happy to, was trying to follow the convention when I put it in utils.

PR #234, opened 2023-11-13
pkt/utils: simpler LazyLoad

jmank88:
Simplify `LazyLoad` by dropping unnecessary synchronization. Mirrors changes from core.

### DIFF
--- a/pkg/utils/lazy.go
+++ b/pkg/utils/lazy.go
@@ -7,8 +7,8 @@ import (
 type LazyLoad[T any] struct {
 	f     func() (T, error)
 	state T
+	ok    bool
 	lock  sync.Mutex
-	once  sync.Once
 }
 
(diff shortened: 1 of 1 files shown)


patrickhuie19 (commented):
Nice catch on the discrepancy. 

 Why have this type in core and in relay?

jmank88:
> Nice catch on the discrepancy.
> 
> Why have this type in core and in relay?

None. Planning to update core to use this.

patrickhuie19 (approved):

patrickhuie19:
> > Nice catch on the discrepancy.
> > Why have this type in core and in relay?
> 
> None. Planning to update core to use this. Edit: or drop it? seems unused... 🤔

I dug around a bit, I agree with dropping it. Makes sense to me why we had it there before and why we don't need it anymore.

PR #237, opened 2023-11-14
Wiring through Tracing.TLSCertPath from TOML config

patrickhuie19:
Pair in core: https://github.com/smartcontractkit/chainlink/pull/11278


pkg/loop/telem.go
```diff
@@ -66,9 +67,16 @@ func SetupTracing(config TracingConfig) error {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
+	var creds credentials.TransportCredentials
+	if config.TLSCertPath != "" {
+		creds, _ = credentials.NewClientTLSFromFile(config.TLSCertPath, "")
+	} else {
+		creds = insecure.NewCredentials()
```
patrickhuie19, line 74: I'm definitely still thinking about this - I disallow setting `Tracing.TLSCertPath = ""` in the TOML configuration in core, so this path occurs when `Tracing.TLSCertPath` is unset. I'd like to be more verbal about this and log at the warn level. There's probably a sensible place to shout about this in core.

pkg/loop/config.go
```diff
@@ -114,3 +123,16 @@ func getTracingSamplingRatio() float64 {
 	}
 	return samplingRatio
 }
+
+// getTLSCertPath parses and validates the CL_TRACING_TLS_CERT_PATH environment variable.
+func getTLSCertPath() (string, error) {
+	// empty string is valid; signals to use insecure credentials
```
krehermann, line 129: i'm confused by this comment and what happens if this is empty.

would LookupEnv be a better fit https://pkg.go.dev/os#LookupEnv  so we can distinguish unset value?

patrickhuie19: Valid TOML:

```
[Tracing]
Enabled = true
CollectorTarget = 'localhost:4317'
TLSCertPath = '/path/to/cert.pem'
```

Valid TOML:

```
[Tracing]
Enabled = true
CollectorTarget = 'localhost:4317'
# Note that TLSCertPath is not specified
```

Invalid TOML:
```
[Tracing]
Enabled = true
CollectorTarget = 'localhost:4317'
TLSCertPath = ''
```
The 2nd option is how I intend users specify insecure connections to the otel-collector are allowed. From a security perspective, this is fine when the nodes and the otel-collector live on the same network.

LookupEnv would be able to tell when the plugin is booted whether or not a value is set, but this doesn't matter because a value of empty string will be caught by the host before it is sent over to the plugin via env var.

pkg/loop/telem.go
```diff
@@ -66,9 +67,16 @@ func SetupTracing(config TracingConfig) error {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
+	var creds credentials.TransportCredentials
+	if config.TLSCertPath != "" {
+		creds, _ = credentials.NewClientTLSFromFile(config.TLSCertPath, "")
```
jmank88, line 72: Unhandled error?

pkg/loop/config.go
```diff
@@ -114,3 +123,16 @@ func getTracingSamplingRatio() float64 {
...
+	tracingTLSCertPath := os.Getenv(envTracingTLSCertPath)
+
+	_, err := credentials.NewClientTLSFromFile(tracingTLSCertPath, "")
```
jmank88, line 132: What is the advantage of doing validation at this level?

pkg/loop/telem.go
patrickhuie19: Yes, in this revision I chose to validate in `parse` in `pkg/loop/config.go`

pkg/loop/config.go
jmank88: Wait, but `SetupTracing` already returns an `error`?

patrickhuie19: I gave this a try. 

It gets messier than I think is worth b/c EnvConfig previously was symmetric on values passed in to serialize to env vars, and values passed out after deserializing from env vars. i can break this symmetry to have parse return a value of type `credentials.TransportCredentials`:

```
type EnvConfig struct {
	PrometheusPort int

	TracingEnabled            bool
	TracingCollectorTarget    string
	TracingSamplingRatio      float64
	TracingTLSCertPath        string
	TracingAttributes         map[string]string
	TracingTLSCreds           credentials.TransportCredentials
}
```

I think I prefer my initial implementation, or your idea to avoid separation of concerns and just return the credentials error in `SetupTracing`

`SetupTracing` does already return an `error`. In my head I had earmarked the error just for dial issues, but it's looking more sensible to include TLS credentials issues as well.

jmank88: I think that using the SetupTracing error makes sense.

pkg/loop/telem.go
patrickhuie19: Resolved by explicit insecure mode in TOML config

pkg/config/error.go
```diff
@@ -40,6 +40,15 @@ func (e ErrEmpty) Error() string {
 	return fmt.Sprintf("%s: empty: %s", e.Name, e.Msg)
 }
 
+type ErrIncluded struct {
```
krehermann, line 43: i don't see this being used. what is happening? it's used in core?

if you need it, can you think of a more descriptive name? the code below KeyNotFoundError is pretty clear what it relates to. ErrIncluded is not intuitive to me

pkg/loop/telem.go
```diff
@@ -66,15 +67,25 @@ func SetupTracing(config TracingConfig) error {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
jmank88, line 67: Is this still reasonable going forward as we continue to add things here? I'm concerned that it is arbitrary and we have no escape hatch. Actually, was it only added when we had a blocking connection? Could we just drop it?
```suggestion
```

jmank88 (approved):

pkg/config/error.go
patrickhuie19: Yep, it's used in core

pkg/loop/telem.go
patrickhuie19: We can drop it, good call

pkg/config/error.go
jmank88: Can you expand the error message so that is clear to the users?
Does this really need a special type? Or could `ErrInvalid` be leveraged instead:
```go
return fmt.Sprintf("%s: invalid value (%v): %s", e.Name, e.Value, e.Msg)
```
There is also the in-between option of having a special constructor for this case, but still returning another type, like we do for duplicates:
https://github.com/smartcontractkit/chainlink-relay/blob/02b76df27fb4369ee017ce5512df1f8041f1afce/pkg/config/error.go#L20-L22

jmank88 (approved):

pkg/config/error.go
patrickhuie19: I didn't want to use ErrInvalid because of `Value`:

If you don't specifiy a value, i.e.
`ErrInvalid{Name: name, Msg: "msg"}`

You get `'<nil>'` in the resulting error string. 

If you specify a value and it's nil, a panic is thrown. 

I suppose I could have hardcoded value to something like 'set', but this felt hacky, and it felt like there could be a category of TOML values that should be expected to be unset.

jmank88: Why don't you want to include the value?
What is the ideal error message for users? We can work backwards from that.

jmank88 (approved):

pkg/config/error.go
patrickhuie19: > Why don't you want to include the value?
Because the value doesn't matter - its presence does

PR #238, opened 2023-11-14
Add ChainsKV

dimriou:


pkg/chains/chain_kv.go
jmank88, line 1: Do you expect to use this with non-evm chains? IIUC it is only leftover to support the legacy EVM uses, so it could live in chainlink-evm instead.

dimriou: I saw it being used by Cosmos here (https://github.com/smartcontractkit/chainlink/blob/develop/core/services/chainlink/relayer_chain_interoperators.go#L370) and I figured it might have a general purpose

jmank88: Hmm but it is not used by the `chainlink-cosmos` repo. Does `chainlink-evm` even need to use it? Or could it stay in core?

dimriou: It's used here(https://github.com/smartcontractkit/chainlink/blob/develop/core/chains/evm/chain.go#L65) for `LegacyChains`. I'm not very familiar with these, so I don't have any strong preference.

jmank88: I don't know much about that, but if it's just a shim then it could stay in core.

cc @krehermann

dimriou: It seems it's also used in `solana_transfer_controler.go`. Again no strong preference here.

krehermann: it should just be a shim in core to support the legacy chains and should be able to remain in core. once we are fully loopified, there will be no use case for this abstraction

dimriou:
Closing this and moving ChainsKV to evm folder

PR #239, opened 2023-11-14
Add currencies

dimriou:
Supports https://github.com/smartcontractkit/chainlink/pull/11287


pkg/utils/utils.go
jmank88, line 1: WDYT about switching to more idiomatic naming? Would that be disruptive? For example:
```go
func HasQuotes([]byte) bool { }
func TrimQuotes([]byte) bool { }
```

jmank88: A case could be made for a bytes-specific package too :thinking: `package byteutil`? Can always do that later though.

dimriou: Since the extraction is a bit time-sensitive, I want us to avoid introducing new changes, so we can speed things up. But perhaps we can open a tech-dept backlog and add all these suggestions there.

jmank88: I am not proposing refactoring - only moving things where they belong. There is no reason to inherit core's mess in to this repo.

dimriou: Oh yeah, I'm okay with creating a separate package. I just wanted to avoid renaming functions.

jmank88: I actually think the naming is more important, but what is the difference?

dimriou: I want us to make the review process as fast as possible so we can keep things going. If we start with renaming then we might get tempted to also refactor some methods, or change some logic. We have to draw a line somewhere and I think the scope of this effort is to simply move things.

jmank88: My claim is that drawing the line between new packages and new identifiers is arbitrary.

Additionally, if you leave behind aliases or forwarders in core anyways, then it will be obvious to reviewers what is going on.

dimriou: Moving things is just that. Renaming involves debating about the names and making the changes back to chainlink repo. It might not seem a lot but given reviews are async, it adds up. Since you'll be the main reviewer I'm ok with it if you're on board and don't mind the additional time it might take.

jmank88: My point is that moving === renaming. Just like the cli `mv` command is used for renaming. I could see how moving a whole package would be a special case, but that is not what is happening here. We are relocating individual types between modules, and any existing similarity between package names is incidental. The relay modules have their own conventions, and the stakes are much higher here than in core, since it is actually used as a library by other modules.

jmank88 (approved):

pkg/assets/currencies.go
```diff
@@ -0,0 +1,191 @@
+package assets
+
```
dhaidashenko, line 2: nit: should we rename this file to link?
Not directly related to this, but we tend to create large files that contains code for multiple types. Is not it easier to navigate packages with large number of small files?

dhaidashenko (approved):
LGTM

pkg/utils/bytes/bytes.go
```diff
@@ -0,0 +1,17 @@
+package bytes
```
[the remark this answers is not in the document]
dimriou: Jordan made the suggestion. The idea is to avoid having a large `utils.go` file with everything cramped together.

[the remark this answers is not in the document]
jmank88: https://go.dev/blog/package-names#bad-package-names
> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.

jmank88 (approved):

pkg/utils/bytes/bytes.go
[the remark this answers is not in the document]
jmank88: IMHO the rule that takes priority here is to not disrupt the local style. This module stands alone, and has much higher standards than core.

dhaidashenko (approved):

PR #240, opened 2023-11-14
pkg/reportingplugins: remove; move mercury to chainlink-data-streams

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2737

Supports:
- https://github.com/smartcontractkit/chainlink-mercury/pull/1
- https://github.com/smartcontractkit/chainlink/pull/11293


pkg/types/provider_mercury.go
```diff
@@ -1,21 +1,97 @@
...
+	BenchmarkPrice     *big.Int
+}
+
+type MercuryFieldsV3 struct {
+	ValidFromTimestamp uint32
+	Timestamp          uint32
+	NativeFee          *big.Int
+	LinkFee            *big.Int
+	ExpiresAt          uint32
+	BenchmarkPrice     *big.Int
```
jmank88, line 78: It looks like this would be a valid simplification too :thinking: 
```suggestion
	MercuryFieldsV2
```

vyzaldysanchez: We could go with that... 🤔

vyzaldysanchez (dismissed):

pkg/types/provider_mercury.go
jmank88: I'm not sure it is really helpful and I'm worried about some kind of edge case failing :shrug:

vyzaldysanchez: That's fair

vyzaldysanchez (dismissed):

vyzaldysanchez (dismissed):

samsondav (dismissed):

samsondav (approved):

PR #242, opened 2023-11-15
Clean up chain reader, removed map decoder, as it will no longer be used

nolag:


jmank88 (approved):

pkg/codec/utils.go
```diff
@@ -11,3 +15,119 @@ func FitsInNBitsSigned(n int, bi *big.Int) bool {
...
+			return data, nil
+		}
+	}
+
+	return data, nil
+}
+
+func SliceToArrayVerifySizeHook(from reflect.Type, to reflect.Type, data any) (any, error) {
+	// By default, if the array is bigger it'll still work. (ie []int{1, 2, 3} => [4]int{} works with 0 at the end
+	// [2]int{} would not work. This seems to lenient, but can be discussed.
```
reductionista, line 124: Yeah, I suspect we'll want to enforce that the length matches exactly, but I guess we should gather feedback from others about use cases. This is fine for now.

reductionista (approved):

PR #244, opened 2023-11-16
[BCF-2793] Support decimal.Decimal in pipeline runner service

cedric-cordenier:


george-dorin (approved):

jmank88:
Why do we need to treat `decimal.Decimal` different than everything else? I thought we already had support for decimal.Decimal<->json.Number in memory, which doesn't require extra type information - why doesn't that work here?

cedric-cordenier:
@jmank88 It's due to decimal.Decimal being JSON serializable but not proto serializable. We could fall back to JSON serialization (and that's an option I considered), but then the problem is that in the LOOPP client you won't be able to distinguish between a string that's actually a string and a string that encodes a decimal.

To me it felt cleaner to handle this as an exception here and properly encode the type.

Note though that I don't this is a good long-term solution, and it's not intended as such. Long-term it feels like we should be making pipeline values more type safe by replacing the existing `interface{}` used to store the value by something like a special value type that knows how to encode itself to a proto.

jmank88:
> but then the problem is that in the LOOPP client you won't be able to distinguish between a string that's actually a string and a string that encodes a decimal.

But that is already the case with the existing code, so why is it a good thing to change the behavior? Shouldn't we be matching behavior by default to minimize risk?

jmank88:
To put this another way, when we encounter a json serialized number that holds a decimal value, we have no way of knowing whether it was a `decimal.Decimal`, `float32/64`, or any other random type. How do we normally decode those? Because they will still show up here. Do we always use a `decimal.Decimal` for non-integer `json.Number`s? If so, why not just do that here? If not, then what happens, because we still need to handle that case in this code.

cedric-cordenier:
>But that is already the case with the existing code, so why is it a good thing to change the behavior? Shouldn't we be matching behavior by default to minimize risk?

Maybe I'm misunderstanding, but I don't think we are changing the behaviour... In a non-loopp world a call to pipelineRunner.ExecuteRun can return a `decimal.Decimal` for some tasks; this ensures that for those tasks we will still return a `decimal.Decimal` even if we crossed a serialization boundary in a loopp world, as opposed to having the API now return a string which the caller then needs to convert back to a decimal to get functional parity.

cedric-cordenier:
>What part of the non-loopp code is preserving the decimal.Decimal type through serialization? Why is this serialization different than the existing?

Genuine question: why do you say "through serialization" here?

Calls to ExecuteRun currently don't serialize the task results, which is what we're getting our values from (it's different from the Run object's PipelineTaskRuns which does serialize the task results, and which may store them in the DB. But that's not relevant to us since ExecuteRun doesn't persist anything.

jmank88:
> Genuine question: why do you say "through serialization" here?
> 
> Calls to ExecuteRun currently don't serialize the task results, which is what we're getting our values from (it's different from the Run object's PipelineTaskRuns which does serialize the task results, and which may store them in the DB. But that's not relevant to us since ExecuteRun doesn't persist anything.

IIUC task results can be serialized mid-pipeline execution (and stored in the DB). Pipeline serialization was a big headache in the past, and I am concerned that we are re-inventing the wheel.

cedric-cordenier:
>IIUC task results can be serialized mid-pipeline execution (and stored in the DB). Pipeline serialization was a big headache in the past, and I am concerned that we are re-inventing the wheel.

I agree with the concern: let me think a bit about how to navigate this...

Basically, task results are serialized and can be stored in the DB if the pipeline is executed via Run/ResumeRun. However, we are only using ExecuteRun here, so nothing is being stored in the database. Also Median specifically is using the non-serialized task values (the values stored in `TaskRunResults`, not the ones stored on the `Run` object which are serialized. This matches the behaviour of the existing median plugin).

That's why I wanted to preserve the existing behaviour as much as possible, i.e.:
- existing: median calls pipelineRunner.ExecuteRun, and fetches the values from the TaskResults -> these are decimal.Decimals.
- new: the same behaviour is preserved since we aren't doing a lossy conversion from decimal.Decimal -> string without the type info

But if I understand your point it is that the existing json serialization handles a bunch of edge cases that we won't have covered here, and that we'll need to duplicate in the worst case. Is that right?

jmank88:
That makes sense, but I was under the impression that a pipeline could include serialization regardless of how it was invoked, and  even without being stored in the DB. If that is not the case, then my only concern is duplicating, as you say.

PR #250, opened 2023-11-20
Add json data type

amit-momin:
Migrated the json data type from `/core/services/pg/datatypes`

### DIFF
--- a/pkg/utils/json/json.go
+++ b/pkg/utils/json/json.go
@@ -1,0 +1,58 @@
+package json
+
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"fmt"
+)
+
+// JSON defined JSON data type, need to implements driver.Valuer, sql.Scanner interface
+type JSON json.RawMessage
+
+// Value return json value, implement driver.Valuer interface
+func (j JSON) Value() (driver.Value, error) {
+	if len(j) == 0 {
+		return nil, nil
+	}
+	bytes, err := json.RawMessage(j).MarshalJSON()
+	return string(bytes), err
+}
+
+// Scan scan value into Jsonb, implements sql.Scanner interface
+func (j *JSON) Scan(value interface{}) error {
+	if value == nil {
+		*j = JSON("null")
+		return nil
+	}
+	var bytes []byte
+	switch v := value.(type) {
+	case []byte:
+		bytes = v
+	case string:
+		bytes = []byte(v)
+	default:
+		return fmt.Errorf("Failed to unmarshal JSONB value: %+v", value)
+	}
+
+	result := json.RawMessage{}
+	err := json.Unmarshal(bytes, &result)
+	*j = JSON(result)
+	return err
+}
+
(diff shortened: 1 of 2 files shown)


jmank88 (commented):
Do we need to call Marshall and unmarshal explicitly on rawmessage? Aren't those no-ops?

jmank88:
> I would test Jordan's comment about marshal & unmarshal methods but otherwise looks ok.

Oh the implementations are basically no-ops anyways - just support for `nil`->`null`:
```go
// MarshalJSON returns m as the JSON encoding of m.
func (m RawMessage) MarshalJSON() ([]byte, error) {
	if m == nil {
		return []byte("null"), nil
	}
	return m, nil
}

// UnmarshalJSON sets *m to a copy of data.
func (m *RawMessage) UnmarshalJSON(data []byte) error {
	if m == nil {
		return errors.New("json.RawMessage: UnmarshalJSON on nil pointer")
	}
	*m = append((*m)[0:0], data...)
	return nil
}
```
https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/encoding/json/stream.go;l=260-275

pkg/utils/json/json.go
jmank88, line 1: I like having a separate package for json related things, but it has two problems:
- if you import std lib also, then you have to alias one of them as something other than `json`
- redundant naming from outside the package: `json.JSON`

What about something like `package jsonutil`? And we could reuse the name `RawMessage`
- `jsonutil.RawMessage`
- 
Or maybe this makes sense in the `sqlutil` package, in which case the `JSON` name is probably fine.
- `sqlutil.JSON`

I'm thinking this feels more SQL related :thinking:

amit-momin: I have this [PR](https://github.com/smartcontractkit/chainlink/pull/11321) open but it still needs the changes to point to the JSON type here in `chainlink-common`. But any place I've replaced `datatypes.JSON` -> `json.RawMessage` is where I'd use this. I think naming the package `jsonutil` makes sense. It doesn't seem completely SQL related since it's used for things like the TxMeta type.

jmank88: You can temporarily use the hash from this PR to show passing tests on the core side.

jmank88: > It doesn't seem completely SQL related since it's used for things like the TxMeta type.

The whole purpose of the type is to extend support for SQL though:
> // JSON defined JSON data type, need to implements driver.Valuer, sql.Scanner interface
> type JSON json.RawMessage

amit-momin: That's true. For the current changes, it's using these methods under the hood so I can also get behind putting this under `sqlutil`. But I could see someone using this type just for `RawMessage` without the extended support being needed. Even though `RawMessage` could be used directly from `encoding/json`, would we want to use this type here going forward for consistency?

amit-momin: Ok in that case I'll move this under the `sqlutil` package

jmank88 (approved):

PR #251, opened 2023-11-21
pkg/services/servicetest: add Run and RunHealthy helpers

jmank88:
Supports:
- https://github.com/smartcontractkit/chainlink/pull/11354

### DIFF
--- a/pkg/services/servicetest/run.go
+++ b/pkg/services/servicetest/run.go
@@ -3,10 +3,8 @@ package servicetest
 import (
 	"context"
 	"errors"
-	"testing"
+	"fmt"
 	"time"
-
-	"golang.org/x/exp/maps"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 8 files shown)


nolag (dismissed):

pkg/services/servicetest/run.go
```diff
@@ -0,0 +1,65 @@
...
+func Run(tb testing.TB, s Runnable) {
+	tb.Helper()
+	require.NoError(tb, s.Start(tests.Context(tb)), "service failed to start")
+	tb.Cleanup(func() { assert.NoError(tb, s.Close(), "error closing service") })
+}
+
+// RunHealthy fails tb if the service fails to start, close, is never ready, or is ever unhealthy (based on periodic checks).
+//   - after starting, readiness will always be checked at least once, before closing
+//   - if ever ready, then health will be checked at least once, before closing
+func RunHealthy(tb testing.TB, s services.Service) {
```
patrickhuie19, line 33: Is there a sensible unit test for `RunHealthy`?

patrickhuie19: Hmm. Probably not. The use in pkg/loop is likely enough. 

Related: should we add this file to sonar exceptions so we don't mess with coverage stats?

jmank88: Added a test.

pkg/services/servicetest/run_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
+
+func TestRunHealthy(t *testing.T) {
+	for _, test := range []struct {
+		name    string
+		s       services.Service
+		expFail bool
```
jmank88, line 18: I tried to validate error messages too, but they contain whole stacks from testify internals.

patrickhuie19 (approved):

PR #246, opened 2023-11-16
fix/lint-action: trying json lint artifact to make parsing file name and errors easier

patrickhuie19:

PR #252, opened 2023-11-22
update golangci-lint rules

jmank88:
This PR enables some new linters and revive rules for things we are already trying to do.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,6 +10,10 @@ linters:
     - misspell
     - rowserrcheck
     - errorlint
+    - unconvert
+    - sqlclosecheck
+    - noctx
+    - whitespace
 linters-settings:
   exhaustive:
     default-signifies-exhaustive: true
@@ -40,9 +44,10 @@ linters-settings:
       - name: error-return
       - name: error-strings
       - name: error-naming
+      - name: exported
       - name: if-return
       - name: increment-decrement
-      # - name: var-naming
+      - name: var-naming
       - name: var-declaration
       - name: package-comments
       - name: range
--- a/pkg/loop/internal/broker.go
+++ b/pkg/loop/internal/broker.go
@@ -39,7 +39,7 @@ func (a *atomicBroker) DialWithOptions(id uint32, opts ...grpc.DialOption) (conn
 	return a.load().DialWithOptions(id, opts...)
 }
 
-func (a *atomicBroker) NextId() uint32 {
+func (a *atomicBroker) NextId() uint32 { //nolint:revive
 	return a.load().NextId()
 }
 
--- a/pkg/monitoring/poller.go
+++ b/pkg/monitoring/poller.go
@@ -75,11 +75,10 @@ func (s *sourcePoller) Run(ctx context.Context) {
 					continue
 				} else if errors.Is(err, context.Canceled) {
 					return
-				} else {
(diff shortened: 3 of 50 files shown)


vyzaldysanchez (dismissed):

jmank88:
I was unable to configure `testifylint` to disable the `require-error` rule, which was producing many false positives, so I have removed `testifylint` for now.

patrickhuie19:
Does this make #246 obsolete? I had a feeling something like this was coming so I didn't merge

.golangci.yml
jmank88, line 1: - https://github.com/mdempsky/unconvert
- https://github.com/ryanrolds/sqlclosecheck
- https://github.com/sonatard/noctx
- https://github.com/ultraware/whitespace

[revive](https://github.com/mgechev/revive):
 - https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#exported
 - https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming

pkg/loop/internal/broker.go
```diff
@@ -39,7 +39,7 @@ func (a *atomicBroker) DialWithOptions(id uint32, opts ...grpc.DialOption) (conn
 	return a.load().DialWithOptions(id, opts...)
 }
 
-func (a *atomicBroker) NextId() uint32 {
+func (a *atomicBroker) NextId() uint32 { //nolint:revive
```
patrickhuie19, line 42: what does this do?

pkg/monitoring/poller.go
```diff
@@ -75,11 +75,10 @@ func (s *sourcePoller) Run(ctx context.Context) {
 					continue
 				} else if errors.Is(err, context.Canceled) {
 					return
-				} else {
-					s.log.Errorw("failed to fetch from source", "error", err)
-					reusedTimer.Reset(s.pollInterval)
-					continue
 				}
+				s.log.Errorw("failed to fetch from source", "error", err)
```
patrickhuie19, line 79: This feels odd to me - the previous formatting was easier to read

jmank88:
> Does this make #246 obsolete? I had a feeling something like this was coming so I didn't merge

No, this doesn't affect CI other than the version bump.

patrickhuie19 (changes requested):
> Does this make #246 obsolete? I had a feeling something like this was coming so I didn't merge

If it does, I think this PR is a good opportunity to update the sonar properties:
```
sonar.projectName=chainlink-common
```

pkg/loop/internal/broker.go
jmank88: Ignores the naming failure on `Id` instead of `ID`, since we are implementing an interface from another package.

patrickhuie19 (approved):

pkg/monitoring/poller.go
jmank88: The idea is to reduce unnecessary indentation.
https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#superfluous-else

jmank88: https://go.dev/doc/effective_go#control-structures
> In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

PR #256, opened 2023-11-27
Utils extraction from chainlink repo

dimriou:
Required for https://github.com/smartcontractkit/chainlink/pull/11425

### DIFF
--- a/pkg/loop/server.go
+++ b/pkg/loop/server.go
@@ -85,7 +85,7 @@ func (s *Server) start() error {
 		return fmt.Errorf("error starting prometheus server: %w", err)
 	}
 
-	s.checker = services.NewChecker()
+	s.checker = services.NewChecker("", "")
 	if err := s.checker.Start(); err != nil {
 		return fmt.Errorf("error starting health checker: %w", err)
 	}
--- a/pkg/services/health.go
+++ b/pkg/services/health.go
@@ -37,6 +37,8 @@ type HealthChecker struct {
 	chStop chan struct{}
 	chDone chan struct{}
 
+	ver, sha string
+
 	servicesMu sync.RWMutex
 	services   map[string]HealthReporter
 
@@ -70,8 +72,23 @@ var (
 	)
 )
 
-func NewChecker() *HealthChecker {
+func NewChecker(ver, sha string) *HealthChecker {
+	if ver == "" || sha == "" {
+		if bi, ok := debug.ReadBuildInfo(); ok {
+			if ver == "" {
+				ver = bi.Main.Version
+			}
+			if sha == "" {
+				sha = bi.Main.Sum
+			}
+		}
+	}
+	if len(sha) > 7 {
+		sha = sha[:7]
+	}
 	return &HealthChecker{
+		ver:      ver,
+		sha:      sha,
 		services: make(map[string]HealthReporter, 10),
 		healthy:  make(map[string]error, 10),
 		ready:    make(map[string]error, 10),
@@ -82,13 +99,7 @@ func NewChecker() *HealthChecker {
 
 func (c *HealthChecker) Start() error {
 	return c.StartOnce("HealthCheck", func() error {
-		if bi, ok := debug.ReadBuildInfo(); ok {
-			hash := bi.Main.Sum
-			if len(hash) > 7 {
-				hash = hash[:7]
-			}
-			version.WithLabelValues(bi.Main.Version, hash).Inc()
-		}
+		version.WithLabelValues(c.ver, c.sha).Inc()
 
 		// update immediately
 		c.update()


pkg/services/state.go
```diff
@@ -236,3 +237,13 @@ func (eb *ErrorBuffer) SetCap(cap int) {
 	}
 	eb.cap = cap
 }
+
+// EnsureClosed closes the io.Closer, returning nil if it was already
+// closed or not started yet
+func EnsureClosed(c io.Closer) error {
+	err := c.Close()
+	if errors.Is(err, ErrAlreadyStopped) || errors.Is(err, ErrCannotStopUnstarted) {
```
jmank88, line 245: Was this being used in production?
I thought it was only a hacky workaround for old problems which no longer exist, and so dropping it would be ideal, but limiting to tests would be an improvement.

pkg/utils/utils.go
jmank88, line 1: Is this going to be used by relay/common code? Or just EVM?

pkg/services/state.go
dimriou: As far as I can tell it's currently used by the [txm](https://github.com/smartcontractkit/chainlink/blob/develop/common/txmgr/txmgr.go#L393), nowhere else. I wasn't aware of the problem we faced in the past.

pkg/utils/utils.go
dimriou: That's what I thought, it looks like a simple util function that can be used on different occasions.

pkg/services/state.go
jmank88: The main problem in the past had to do with test helpers, so I was surprised to see this outside of tests.

pkg/utils/utils.go
jmank88: It would be, but unfortunately I think we do not want to encourage this pattern.

jmank88: True, but does it make sense to align across modules/repos? It is a trivial amount of boiler plate, so the main reason for having the helper at all is to align on the values and declare them in just one place, as a set of defaults. Does it make sense for other repos to use the same defaults as common? Or might each repo want to declare their own version of this? We need to consider a LOOPP based future with different kinds of plugins and third party repos too.

pkg/services/state.go
dimriou: I've added a fix on the main repo

pkg/utils/utils.go
dimriou: Yeah, I agree with that, it's just hard to find a good place to put this. I don't see a utility file inside `/common` on chainlink repo, I guess this is by design to keep everything here. At the same time, I see `NewRedialBackoff` being used by `/common` , `core/chains/evm`, and `core/services`, so it needs to be a place where all of them have access. Perhaps I can change it to receive params, but it will add boiler plate code, which we were trying to avoid in the first place.

dimriou: Yeah, I'm not opposed to that, let me explain what I'm saying. There are three main packages where this is used:

- `/core/services` : we can keep it as it is along with the core-specific implementation of that code
- `/evm/client` : we can add another evm/client implementation there to help us remove the dependency with core and extract the code, but the client will be deprecated anyway once we migrate to MultiNode so it doesn't really matter
- `/common` : for this, we can create a utils file under `/common` and add the method there, but since we don't have one, and the original plan was to move the /common to chainlink-common eventually, it made more sense to keep the method here. I'm not opposed to having a separate implementation, I just want to find a good place to put this for `/common` to use it.

jmank88: We could start a `common/internal` tree to collect things to share among the common/ package tree only.

pkg/utils/mailbox/mailbox.go
dimriou, line 1: @jmank88 I know you're not a big fan of `mailbox.NewMailbox` patter. Should we refactor the methods to `mailbox.New` `mailbox.NewSingle` , etc?

pkg/utils/mailbox/mailbox_prom.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/utils"
+)
+
+var mailboxLoad = promauto.NewGaugeVec(prometheus.GaugeOpts{
+	Name: "mailbox_load_percent",
+	Help: "Percent of mailbox capacity used",
+},
+	[]string{"appID", "name", "capacity"},
+)
```
jmank88, line 22: Is this our first case of global prom metrics in this module? We might have to be careful with these.

pkg/utils/mailbox/mailbox.go
jmank88: Hmmm this is a tough one since the monitor is in this package too.

jmank88: Are there any other similar data structures that it would make sense to group together into a less specific package? e.g. `package queues`?

pkg/utils/mailbox/mailbox_prom.go
dimriou: Looks like we have some on the monitoring package: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/monitoring/metrics.go

jmank88: Oh I wasn't expecting that :thinking: 
I'm think that we'll have to undo all these global prom metrics at some point, but I guess it is out of scope for now if it is not in the way.

pkg/utils/utils.go
```diff
@@ -44,3 +45,16 @@ func ContextWithDeadlineFn(ctx context.Context, deadlineFn func(orig time.Time)
...
 }
+
+func IsZero[C comparable](val C) bool {
+	var zero C
+	return zero == val
+}
+
+// AddHexPrefix adds the prefix (0x) to a given hex string.
+func AddHexPrefix(str string) string {
+	if len(str) < 2 || len(str) > 1 && strings.ToLower(str[0:2]) != "0x" {
```
jmank88, line 56: How about std lib for readability? (should still be plenty fast too)
```suggestion
	if !strings.HasPrefix(s, "0x") {
```

pkg/utils/utils.go
```diff
@@ -44,3 +45,16 @@ func ContextWithDeadlineFn(ctx context.Context, deadlineFn func(orig time.Time)
...
 	return ctx, cancel
 }
+
+func IsZero[C comparable](val C) bool {
+	var zero C
+	return zero == val
+}
+
+// AddHexPrefix adds the prefix (0x) to a given hex string.
+func AddHexPrefix(str string) string {
```
jmank88, line 55: How about `Ensure` since it doesn't add when it's already there.
```suggestion
// EnsureHexPrefix adds the prefix (0x) to a given hex string.
func EnsureHexPrefix(str string) string {
```

jmank88 (approved):

jmank88 (approved):

PR #261, opened 2023-11-28
Add a renamer mod

nolag:


pkg/codec/renamer.go
```diff
@@ -0,0 +1,76 @@
...
+		modifierBase: modifierBase[string]{
+			fields:           fields,
+			onToOffChainType: map[reflect.Type]reflect.Type{},
+			offToOnChainType: map[reflect.Type]reflect.Type{},
+		},
+	}
+	m.modifyFieldForInput = func(field *reflect.StructField, _, newName string) error {
+		field.Name = newName
+		if unicode.IsLower(rune(field.Name[0])) {
+			field.PkgPath = pkgPath
```
ilija42, line 23: What exactly happens here? Field that is unexported gets its PkgPath set to code pckg path. Why would an unexported field need PkgPath and why would we set it to codec pckg path?

pkg/codec/renamer.go
```diff
@@ -0,0 +1,76 @@
...
+		}
```
ilija42, line 24: IIUC pkgpath is a unique identifier for unexported fields as exported fields are identified just by their name? Can there be some runtime issue if two fields with same name get assigned same PkgPath (in this case codec package path)

pkg/codec/renamer.go
reductionista, line 23: I'm not sure I understand the pkgPath stuff.  Does this mean that if we rename a struct in some other package to something new, it will automatically get moved inside the `codec` package?

nolag: pkgpath is the name of the package that the field is defined in. Since nothing else in this package is defining fields on the modified structs, there won't be another name that's the same.

The intended use will be in [config in this PR](https://github.com/smartcontractkit/chainlink-common/pull/264/files#diff-ede4abc6023e679b5795c220bdd7a8fc49acc42011979f467b5f77b2d1ecf654R87) that drops fields by renaming them to private names.

nolag: going to merge answers in [ilija42](https://github.com/ilija42)'s thread.

pkg/codec/renamer.go
ilija42: But in a scenario where we have 2 jobs running that drop same fields there would be two filed with same name and package path right?

nolag: I can be extra safe and make the type use the PkgPath of the type that's being modified.  Then if there are other private fields it won't break.  I didn't see a use case for the codec generating private fields though.

ilija42: I am just worried that if that if this somehow happens it would be a nightmare to track down

pkg/codec/renamer.go
```diff
@@ -0,0 +1,76 @@
...
+	return rOutput.Interface(), nil
+}
+
+func (r *renamer) TransformForOnChain(input any) (any, error) {
+	rOutput, err := renameTransform(r.offToOnChainType, reflect.ValueOf(input))
+	if err != nil {
+		return nil, err
+	}
+	return rOutput.Interface(), nil
+}
```
reductionista, line 48: Does the implementation of `TransformOffChain` and `TransformOnChain` just happen to be identical for the renamer, or is this always going to be the case?  Wondering if we really need two different transformer methods or if we could get by with just one?

nolag: It's not identical, one uses  r.onToOffChainType the other uses r.offToOnChainType.  They just happen to be similar for renamed because in both cases it just **safely** _unsafe_ casts to the other type :)

ilija42 (approved):

pkg/codec/renamer.go
reductionista: Nevermind, I think it's probably better as-is I guess even if they were combined we'd still at least have to pass a boolean for which direction which is a more awkward API, and we can just implement it under the hood with a single method as you've done here.

nolag: Most make helper calls that take different parameters depending on which way they go.  The chain implementation that combines multiple modifiers reverses the order if you call TransformOnChain :).

reductionista: Yeah, it makes sense

pkg/codec/renamer.go
reductionista: > I can be extra safe and make the type use the PkgPath of the type that's being modified.

Yeah, this is how I was thinking it should work... makes more sense to me this way, thanks.

PR #263, opened 2023-11-28
Add a mod that runs multiple mods

nolag:


pkg/codec/chain_modifier.go
```diff
@@ -0,0 +1,38 @@
+package codec
+
+import (
+	"reflect"
+)
+
+type ChainModifier []Modifier
```
ilija42, line 7: does modifierChainer name make more sense in our codebase? Naming anything with chain adds a subconscious expectation 😄

nolag: still has chain in chainer.  How about multiModifier?

ilija42: sure: 👍🏼

PR #265, opened 2023-11-30
pkg/services: add ver & sha params to HealthChecker

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2824

Accept ver & sha params in NewHealthChecker. Only fall back to build info if empty.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/11432

### DIFF
--- a/pkg/loop/server.go
+++ b/pkg/loop/server.go
@@ -85,7 +85,7 @@ func (s *Server) start() error {
 		return fmt.Errorf("error starting prometheus server: %w", err)
 	}
 
-	s.checker = services.NewChecker()
+	s.checker = services.NewChecker("", "")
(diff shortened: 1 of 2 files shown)


pkg/loop/server.go
```diff
@@ -85,7 +85,7 @@ func (s *Server) start() error {
 		return fmt.Errorf("error starting prometheus server: %w", err)
 	}
 
-	s.checker = services.NewChecker()
+	s.checker = services.NewChecker("", "")
```
patrickhuie19, line 88: Where would we expect `NewChecker` to be called with non-empty string values for `ver` and `sha`?

patrickhuie19 (approved):

PR #266, opened 2023-11-30
Adjust price scaling factor from 1e8 => 1e18

samsondav:


Fletch153 (approved):

Fletch153:
@samsondav are we able to derive the precision from the onchain configs by any chance?

jmank88 (approved):

PR #267, opened 2023-12-01
Add mathutil

dimriou:


pkg/utils/mathutil/mathutil.go
jmank88, line 1: Are these still necessary given the new built-ins with go 1.21?
https://tip.golang.org/ref/spec#Min_and_max

jmank88: Ah ok. That's is unfortunate

prashantkumar1982 (approved):

PR #268, opened 2023-12-01
Add a wrapper codec that applies mods

nolag:


pkg/codec/modifier_codec.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (m *modifierCodec) Encode(ctx context.Context, item any, itemType string) ([]byte, error) {
+	tmpOnChain, err := m.typeProvider.CreateType(itemType, true)
+	if err != nil {
+		return nil, err
+	}
+
+	offChainType, err := m.modifier.RetypeForOffChain(reflect.TypeOf(tmpOnChain))
+	if err != nil {
```
ilija42, line 55: Just to make sure I understood. `RetypeForOffChain` returns a type that should be a modified version of `item` for eg. adding a field?
What should `TransformForOffChain` do, rename fields, hard code values?

I feel like some of this code could use a bit more comments even if tests are helpful for understanding it, it can get a bit confusing unless you used reflect package recently.

pkg/codec/modifier_codec.go
```diff
@@ -0,0 +1,185 @@
...
+	return reflect.Zero(nt).Interface(), nil
+}
+
+func (m *modifierCodec) Encode(ctx context.Context, item any, itemType string) ([]byte, error) {
+	tmpOnChain, err := m.typeProvider.CreateType(itemType, true)
+	if err != nil {
+		return nil, err
+	}
+
+	offChainType, err := m.modifier.RetypeForOffChain(reflect.TypeOf(tmpOnChain))
```
ilija42, line 54: `CreateType` already returns a type that went through `RetypeForOffChain`. Is it necessary to call it again? Is this done just so that we don't have to cast?

pkg/codec/modifier_codec.go
nolag: In the simplest terms, with respect to relay, `RetypeForOffChain` will change the type used on-chain to something that can be parsed from the input json.  `TransformForOffChain` does the transformation of the on-chain to the off-chain object.

pkg/codec/modifier_codec.go
ilija42:
```suggestion
	ot, err := m.CreateType(itemType, true)
	if err != nil {
		return nil, err
	}
	offChainType := reflect.TypeOf(ot)
```
Can't we do this instead? Same thing, but less err handling

pkg/codec/modifier_codec.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
nolag: Simplified, I can remove more with that logic too :D.  Good catch

PR #270, opened 2023-12-04
Add a modifier by type and remove an unused struct

nolag:


ilija42 (approved):
looks simple enough, lgtm

PR #275, opened 2023-12-06
merge main to branch since BCF-2612-ChainReader was merged.

nolag:


pkg/loop/internal/client.go
```diff
@@ -162,3 +163,25 @@ func isErrTerminal(err error) bool {
 	}
 	return false
 }
+
+func wrapRpcErr(err error) error {
```
nolag, line 167: This function is net-new.  Need it to allow the clients to pass tests that verify errors act the same regardless of running locally or in LOOP.

PR #280, opened 2023-12-11
Support contract mapping

nolag:


pkg/codec/modifier_codec_test.go
```diff
@@ -140,7 +140,7 @@ func TestModifierCodec(t *testing.T) {
 	})
 
 	var actual any
-	t.Run("CreateType returns modified type", func(t *testing.T) {
+	t.Run("CreateContractType returns modified type", func(t *testing.T) {
```
ilija42, line 143: nit, why was this renamed, it stil calls CreateType

ilija42 (approved):
approved with a small question/nit

pkg/codec/modifier_codec_test.go
nolag: For tests in common, it does.  We need the Contract and the TypeName  for the real implementation when it's chain reader.

PR #283, opened 2023-12-11
Check if the codec and chain reader are implemented before setting them

nolag:


pkg/loop/internal/codec.go
jmank88, line 1: Why are the explicit nil checks necessary? Could we nil check sooner and avoid creating the server all together?

pkg/loop/internal/median.go
```diff
@@ -186,8 +194,20 @@ func newMedianProviderClient(b *brokerExt, cc grpc.ClientConnInterface) *medianP
...
+	if status.Convert(err).Code() != codes.Unknown {
+		m.chainReader = maybeCr
+	}
+
+	maybeCodec := &codecClient{b, pb.NewCodecClient(m.cc)}
+	err = maybeCodec.Decode(context.Background(), []byte{}, &anyRetVal, "")
+	if status.Convert(err).Code() != codes.Unknown {
+		m.codec = maybeCodec
+	}
+
```
jmank88, line 210: Is calling a 'real' service the only option here? Or is there some way to interact with the `pb.NewCodecClient` directly instead?

nolag: I couldn't find any better way, it's not supported out of the box for grpc.  I asked bing chat and it suggested doing this too :(.

pkg/loop/internal/chain_reader_test.go
```diff
@@ -59,37 +60,40 @@ func TestVersionedBytesFunctions(t *testing.T) {
 }
 
 func TestChainReaderClient(t *testing.T) {
-	RunChainReaderInterfaceTests(t, &chainReaderInterfaceTester{chainReader: &fakeChainReader{lock: &sync.Mutex{}}})
+	fake := &fakeChainReader{lock: &sync.Mutex{}}
```
jmank88, line 63: nit/ if you embed the plain `sync.Mutex` type instead of a pointer, then you can just omit it and use the zero value.
```suggestion
	fake := &fakeChainReader{}
```

jmank88 (approved):

jmank88 (approved):

PR #284, opened 2023-12-11
Extract pipeline from EVM

DylanTinianov:


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.21
 
 require (
 	github.com/confluentinc/confluent-kafka-go/v2 v2.3.0
+	github.com/ethereum/go-ethereum v1.10.26
```
jmank88, line 7: This repo needs to remain chain-agnostic. What is geth being used for?

DylanTinianov: I moved `utils.big` which is using `goeth/hexutils`. Do you think this is an issue?

jmank88: We'll have to swap it out, but hex is easy - most of that package is just wrapping std lib, and we can copy whatever we need to.

pkg/utils/interval.go
jmank88, line 1: Let's move this one to `package sqlutil`.

pkg/utils/big.go
jmank88, line 1: @dimriou do we already have another big util package? It seems like this file and `package big_math` belong together in something like `package bigutil`. And then we could name these types normally, as `bigutil.Int` and `bigutil.Float`.

pkg/pipeline/models.go
jmank88, line 1: Where is this used? Pipelines are a core node concept that should not appear in this repo.

pkg/loop/internal/chain_reader_test.go
```diff
@@ -18,7 +18,7 @@ import (
 
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
-	. "github.com/smartcontractkit/chainlink-common/pkg/types/interfacetests"
+	test "github.com/smartcontractkit/chainlink-common/pkg/types/interfacetests"
```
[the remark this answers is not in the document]
DylanTinianov: linter was failing. I can change it back though, it doesn't bother me.

pkg/pipeline/models.go
DylanTinianov: Pipeline.Spec{} is being used in evm code for running tests. We need to extract this from core/services/pipeline to remove the dependency on evm. I agree it might not make sense in this repo, I'm mostly just trying to see how this extraction can be done.

jmank88: EVM depending on core (even for tests) is an inversion of dependencies. The EVM tests needs to be rewritten to avoid using the pipeline, or they need to remain in core. Those tests may be legitimate and useful, but they do not belong embedded in the EVM code - they are integration tests of the node with EVM code.

pkg/loop/internal/chain_reader_test.go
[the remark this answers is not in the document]
jmank88: nbd just curious since the linter didn't stop it from getting in :thinking:

pkg/pipeline/models.go
DylanTinianov: Going to close this PR then. The tests will need to be moved/ re-written instead.

PR #288, opened 2023-12-12
Add chain reader and codec to relay

nolag:
core ref: 3d2875640754ca83c8e1e66e57069f587da306ca
solana ref: 946ba098fdce3bd1dc5a628c4eebdd5acf65ddec
starknet ref: 1f32389044ea91447d4fbf9bd46e05444264970c


pkg/types/provider.go
```diff
@@ -70,6 +71,34 @@ func (e UnimplementedError) Is(target error) bool {
...
+func (e InternalError) GRPCStatus() *status.Status {
+	return status.New(codes.Internal, e.Error())
+}
+
+func (e InternalError) Is(target error) bool {
+	if e == target {
+		return true
+	}
+
+	return grpcErrorHasTypeAndMessage(target, string(e), codes.Unimplemented)
```
jmank88, line 89: An internal error is an unimplemented error? Is that right?

pkg/loop/internal/median.go
```diff
@@ -185,7 +195,20 @@ func newMedianProviderClient(b *brokerExt, cc grpc.ClientConnInterface) *medianP
 	m.reportCodec = &reportCodecClient{b, pb.NewReportCodecClient(m.cc)}
 	m.medianContract = &medianContractClient{pb.NewMedianContractClient(m.cc)}
 	m.onchainConfigCodec = &onchainConfigCodecClient{b, pb.NewOnchainConfigCodecClient(m.cc)}
-	m.chainReader = &chainReaderClient{b, pb.NewChainReaderClient(m.cc)}
+	anyRetVal := 10
+
+	maybeCr := &chainReaderClient{b, pb.NewChainReaderClient(m.cc)}
```
jmank88, line 200:
```suggestion

	maybeCr := &chainReaderClient{b, pb.NewChainReaderClient(m.cc)}
	var anyRetVal int
```

pkg/loop/internal/chain_reader_test.go
jmank88, line 1: Imports could use a sort

pkg/types/provider.go
nolag: No, internal error is for if you have an error that isn't something you expected.  Eg: ChainReader wraps with one if log poller gives it an error.  I think of it like a 5xx, you shouldn't get one.

Unimplemented will be returned when the ChainReader (or codec) isn't initilzide

pkg/loop/internal/chain_reader_test.go
nolag: Not sure why, but `golangci-lint run --fix pkg/loop/internal` didn't change it.  I got rid of all the spacing and goland did...

pkg/types/provider.go
jmank88: But should this be `codes.Internal` instead of `codes.Unimplemented`?
```suggestion
	return grpcErrorHasTypeAndMessage(target, string(e), codes.Internal)
```

pkg/types/provider.go
```diff
@@ -70,6 +71,34 @@ func (e UnimplementedError) Is(target error) bool {
...
+}
+
+type NotFoundError string
```
jmank88, line 92: Should this one implement `Is(error) bool`?

pkg/codec/by_item_type_modifier_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/codec"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+func TestByTypeModifier(t *testing.T) {
+	anyitemType := "anything"
```
jmank88, line 16: nit/ prefer const if possible?
```suggestion
	const anyitemType = "anything"
```

pkg/codec/element_extractor.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type ElementExtractorLocation string
+
+const (
+	FirstElementLocation  ElementExtractorLocation = "first element"
+	MiddleElementLocation ElementExtractorLocation = "middle element"
+	LastElementLocation   ElementExtractorLocation = "last element"
```
jmank88, line 15: Is this type doing double duty as a string message describing itself? Would it make sense to use an int enum instead? with a `String()` method?
```suggestion
	FirstElementLocation  ElementExtractorLocation = iota
	MiddleElementLocation ElementExtractorLocation
	LastElementLocation   ElementExtractorLocation
```
```go
func (l ElementExtractorLocation) String() {
	...
}
```
Or just drop the common suffix?
```suggestion
	FirstElementLocation  ElementExtractorLocation = "first"
	MiddleElementLocation ElementExtractorLocation = "middle"
	LastElementLocation   ElementExtractorLocation = "last"
```

jmank88: Oh and I think that idiomatic naming here would be `ElementExtractorLocation*` instead of `*ElementLocation` - e.g. `ElementExtractorLocationFirst`, like `http.Status*` https://pkg.go.dev/net/http#pkg-constants

pkg/codec/modifier.go
```diff
@@ -0,0 +1,17 @@
+package codec
+
+import (
+	"reflect"
+)
+
+type Modifier interface {
```
jmank88, line 7: Worth a comment?
FWIW you can run `make godoc` to run a local godoc server and see the generated HTML:
```
$ make godoc 
go install golang.org/x/tools/cmd/godoc@latest
# http://localhost:6060/pkg/github.com/smartcontractkit/chainlink-common/
godoc -http=:6060
```

pkg/codec/multi_modifier_test.go
```diff
@@ -0,0 +1,59 @@
...
+		require.NoError(t, err)
+
+		expected := chainModifierTestStruct{A: 100}
+		assert.Equal(t, expected, actual)
+	})
+}
+
+type chainModifierTestStruct struct {
+	A int
+}
```
jmank88, line 59: It should be possible to declare this type inside of the body of `TestMultiModifier`, to reduce the scope and to allow for a simpler name like just `testStruct` :shrug:

pkg/codec/utils.go
jmank88, line 1: These exported funcs could use comments.

pkg/types/chain_reader.go
```diff
@@ -5,10 +5,14 @@ import (
 	"time"
 )
 
+// Errors exposed to product plugins
+
```
jmank88, line 9:
```suggestion
// Errors exposed to product plugins
```

pkg/types/codec.go
```diff
@@ -4,6 +4,13 @@ import (
 	"context"
 )
 
+const (
+	ErrFieldNotFound         = InvalidArgumentError("field not found")
+	ErrInvalidEncoding       = InvalidArgumentError("invalid encoding")
+	ErrWrongNumberOfElements = InvalidArgumentError("wrong number of elements in slice")
```
jmank88, line 10: Would it be correct/equivalent to say 'length' instead of 'number of elements':
```suggestion
	ErrWrongLength = InvalidArgumentError("slice is wrong length")
```
If so, I think some version of this should be preferred to avoid introducing new language.

(and should name contain 'slice'? e.g. `ErrSliceWrongLength`)

pkg/codec/modifier.go
nolag: locally, I can't get some links to work, but I put what I thought I needed in the comments.  lmk if it's good enough or if I need to do something else.

pkg/codec/utils.go
```diff
@@ -0,0 +1,225 @@
+package codec
+
+// [math/big.Int]: https://pkg.go.dev/math/big#Int
+// [time.Time]: https://pkg.go.dev/time#Time
+// [time.Unix]: https://pkg.go.dev/time#Unix
+// [time.Time.Unix]: https://pkg.go.dev/time#Time.Unix
```
jmank88, line 6: These should not be necessary. Do the bracketed types not work as-is? `[math/big.Int]` is the only one that looks unusual. It may need to be `[big.Int]`, but the others should work.

pkg/types/chain_reader.go
```diff
@@ -30,7 +33,11 @@ type ChainReader interface {
 	// 		err = cr.GetLatestValue(ctx, bc, "method", ProductParams{Arg:1}, &resp)
 	// 		return
 	//  }
-	GetLatestValue(ctx context.Context, bc BoundContract, method string, params, returnVal any) error
+	GetLatestValue(ctx context.Context, contractName string, method string, params, returnVal any) error
```
jmank88, line 36: Is codec now the only option? Or is JSON still available?
Either way, these comments need to be updated.

pkg/loop/internal/chain_reader.go
```diff
@@ -106,19 +115,21 @@ type chainReaderServer struct {
...
-	retVal := &map[string]any{}
-	err := c.impl.GetLatestValue(ctx, bc, request.Method, params, retVal)
+	retVal, err := getContractEncodedType(contractName, request.Method, c.impl, false)
+	if err != nil {
+		return nil, err
+	}
+	err = c.impl.GetLatestValue(ctx, contractName, request.Method, params, retVal)
 	if err != nil {
 		return nil, err
 	}
```
jmank88, line 135: I thought we discussed this but I'm not sure: Why do we respond with the `CurrentEncodingVersion` below on 126 instead of the version that they used to initiate this call, `request.Params.Version`?

nolag: I'm not sure, I didn't change that from what was there before.  It makes sense to use the params version.  Should we have the Decoding ask for a version that you also want back?  It's currently sent raw bytes so it doesn't need an encoding for the object in the request.

nolag: Done, and added a variable in `Codec` since it doesn't receive `VersionedBytes`

pkg/codec/by_item_type_modifier.go
```diff
@@ -0,0 +1,52 @@
...
+func (b *byItemTypeModifier) RetypeForOffChain(onChainType reflect.Type, itemType string) (reflect.Type, error) {
+	mod, ok := b.modByitemType[itemType]
+	if !ok {
+		return nil, fmt.Errorf("%w: cannot find modifier for %s", types.ErrInvalidType, itemType)
+	}
+
+	return mod.RetypeForOffChain(onChainType, itemType)
+}
+
+func (b *byItemTypeModifier) TransformForOnChain(offChainValue any, itemType string) (any, error) {
```
jmank88, line 34: Would these methods read a little more clearly if they used `To` instead of `For`?
```suggestion
func (b *byItemTypeModifier) TransformToOnChain(offChainValue any, itemType string) (any, error) {
```

nolag: Sure, I'll rename Retype with To also then :).

jmank88 (approved):

jmank88:
Not blocking, but there is still a lot of exported API without any documentation. I think it will be important to have high quality go docs for the packages we expect other teams to interact with, and we may even start enforcing comments on exported API via golangci-lint soon.

jmank88 (approved):

ilija42 (approved):

PR #291, opened 2023-12-14
Add hex package

dimriou:
Adding `hex` package with common functions from `chainlink` repo to support EVM extraction. Required for: https://github.com/smartcontractkit/chainlink/pull/11584

### DIFF
--- a/pkg/utils/hex/hex.go
+++ b/pkg/utils/hex/hex.go
@@ -16,17 +16,17 @@ func EnsurePrefix(str string) string {
 	return str
 }
 
-// ToBig parses the given hex string or panics if it is invalid.
-func ToBig(s string) *big.Int {
+// ToBig parses the given hex string and returns error if it is invalid.
+func ToBig(s string) (*big.Int, error) {
 	n, ok := new(big.Int).SetString(s, 16)
 	if !ok {
-		panic(fmt.Errorf(`failed to convert "%s" as hex to big.Int`, s))
+		return nil, fmt.Errorf(`failed to convert "%s" as hex to big.Int`, s)
 	}
-	return n
+	return n, nil
 }
 
-// RemovePrefix removes the prefix (0x) of a given hex string.
-func RemovePrefix(str string) string {
+// TrimPrefix removes the prefix (0x) of a given hex string.
+func TrimPrefix(str string) string {
 	if HasPrefix(str) {
 		return str[2:]
 	}
@@ -41,7 +41,7 @@ func HasPrefix(str string) bool {
 // TryParse parses the given hex string to bytes,
 // it can return error if the hex string is invalid.
 // Follows the semantic of ethereum's FromHex.
-func TryParse(s string) (b []byte, err error) {
+func DecodeString(s string) (b []byte, err error) {
 	if !HasPrefix(s) {
 		err = errors.New("hex string must have 0x prefix")
 	} else {
--- a/pkg/utils/hex/hex.go
+++ b/pkg/utils/hex/hex.go
@@ -1,0 +1,55 @@
+package hex
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+// EnsurePrefix adds the prefix (0x) to a given hex string.
+func EnsurePrefix(str string) string {
+	if !strings.HasPrefix(str, "0x") {
+		str = "0x" + str
+	}
+	return str
+}
+
+// ToBig parses the given hex string or panics if it is invalid.
+func ToBig(s string) *big.Int {
+	n, ok := new(big.Int).SetString(s, 16)
+	if !ok {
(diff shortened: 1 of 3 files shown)


pkg/utils/hex/hex.go
```diff
@@ -0,0 +1,55 @@
...
+func ToBig(s string) *big.Int {
+	n, ok := new(big.Int).SetString(s, 16)
+	if !ok {
+		panic(fmt.Errorf(`failed to convert "%s" as hex to big.Int`, s))
+	}
+	return n
+}
+
+// RemovePrefix removes the prefix (0x) of a given hex string.
+func RemovePrefix(str string) string {
```
jmank88, line 29: We could align with the std lib naming here. (e.g. `strings.TrimPrefix`)
```suggestion
// TrimPrefix removes the prefix (0x) of a given hex string.
func TrimPrefix(str string) string {
```

pkg/utils/hex/hex.go
```diff
@@ -0,0 +1,55 @@
...
+		str = "0x" + str
+	}
+	return str
+}
+
+// ToBig parses the given hex string or panics if it is invalid.
+func ToBig(s string) *big.Int {
+	n, ok := new(big.Int).SetString(s, 16)
+	if !ok {
+		panic(fmt.Errorf(`failed to convert "%s" as hex to big.Int`, s))
```
jmank88, line 23: Where is this func used? Ideally we shouldn't be panicing in production code (especially a library) so I would hope that this can move to a test package, or be modified to return the error instead.

pkg/utils/hex/hex.go
```diff
@@ -0,0 +1,55 @@
...
+
+// HasPrefix returns true if the string starts with 0x.
+func HasPrefix(str string) bool {
+	return len(str) >= 2 && str[0] == '0' && (str[1] == 'x' || str[1] == 'X')
+}
+
+// TryParse parses the given hex string to bytes,
+// it can return error if the hex string is invalid.
+// Follows the semantic of ethereum's FromHex.
+func TryParse(s string) (b []byte, err error) {
```
jmank88, line 44: Why `Try*`? I think that just `Parse` or `Decode`/`DecodeString` would be clear on their own:
```suggestion
// Decode decodes the given hex string to bytes,
// it can return error if the hex string is invalid.
// Follows the semantic of ethereum's FromHex.
func Decode(s string) (b []byte, err error) {
```

pkg/utils/hex/hex.go
dimriou: Two places outside of tests:
- https://github.com/smartcontractkit/chainlink/blob/develop/core/services/keystore/keys/vrfkey/crypto.go#L22
- https://github.com/smartcontractkit/chainlink/blob/develop/core/services/vrf/v2/listener_v2_log_processor.go#L1185

dimriou: I'll update the function and fix the second case to check for an error. The first one will be modified to:

`FieldSize, _ := new(big.Int).SetString("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F", 16)`

jmank88: Oh if it is in init() code, then panic is OK. We can wrap it with a generic `func must[T any](v T, err error) T` or something :shrug:

pkg/utils/hex/hex.go
```diff
@@ -0,0 +1,55 @@
...
+// EnsurePrefix adds the prefix (0x) to a given hex string.
+func EnsurePrefix(str string) string {
+	if !strings.HasPrefix(str, "0x") {
+		str = "0x" + str
+	}
+	return str
+}
+
+// ToBig parses the given hex string and returns error if it is invalid.
+func ToBig(s string) (*big.Int, error) {
```
jmank88, line 20: WDYT about call this `ParseBig`?
```suggestion
// ParseBig parses the given hex string and returns error if it is invalid.
func ParseBig(s string) (*big.Int, error) {
```

dimriou: TBH `ParseBig` sounds like a function that parses big numbers to hex. Maybe `ParseToBig` instead?

jmank88: I really think that `hex.ParseBig` is the clear idiomatic name. The signature and the package already tell the rest of the story. Here are some std lib examples:
- https://pkg.go.dev/net/url#ParseQuery
- https://pkg.go.dev/net/url#Parse
- https://pkg.go.dev/github.com/google/uuid#Parse
- https://pkg.go.dev/time#Parse
- https://pkg.go.dev/time#ParseDuration
- https://pkg.go.dev/net/http#ParseTime

jmank88 (approved):

jmank88 (approved):

amit-momin (approved):

prashantkumar1982 (approved):

krehermann (approved):

PR #295, opened 2023-12-19
Add util methods from core

dimriou:
- Requires: https://github.com/smartcontractkit/chainlink-common/pull/291
- Required for: https://github.com/smartcontractkit/chainlink/pull/11622


jmank88:
If you use list items then you get fancy formatting with a status icon:

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/291

Required for:
- https://github.com/smartcontractkit/chainlink/pull/11622

pkg/utils/utils.go
```diff
@@ -4,8 +4,11 @@ import (
 	"context"
 	"math"
 	mrand "math/rand"
+	"sync"
 	"time"
 
+	"github.com/pkg/errors"
```
jmank88, line 10: We're trying to stick to std lib errors:
```suggestion
	"errors"
```

pkg/utils/utils.go
```diff
@@ -49,3 +52,76 @@ func IsZero[C comparable](val C) bool {
 	var zero C
 	return zero == val
 }
+
+// JustError takes a tuple and returns the last entry, the error.
+func JustError(_ interface{}, err error) error {
```
jmank88, line 57: Is this used in production?

pkg/utils/utils.go
```diff
@@ -49,3 +52,76 @@ func IsZero[C comparable](val C) bool {
...
+// be used with `defer` statements, like so:
+//
+//	func SomeFunction() (err error) {
+//	    defer WrapIfError(&err, "error in SomeFunction:")
+//
+//	    ...
+//	}
+func WrapIfError(err *error, msg string) {
+	if *err != nil {
+		*err = errors.Wrap(*err, msg)
```
jmank88, line 71: Oh maybe you don't need package errors:
```suggestion
		*err = fmt.Errorf("%s: %w", msg, *err)
```

pkg/utils/sleeper_task.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
+
+// SleeperTask represents a task that waits in the background to process some work.
+type SleeperTask interface {
+	Stop() error
+	WakeUp()
+	WakeUpIfStarted()
+}
```
jmank88, line 15: In the spirit of 'accept interfaces, return structs', let's ditch this interface:
```suggestion
```
And move the comment to the struct, which we can then export as `SleeperTask`:
```go
// SleeperTask represents a task that waits in the background to process some work.
type SleeperTask struct {
```
And then `NewSleeperTask` can return `*SleeperTask`:
```go
func NewSleeperTask(worker Worker) *SleeperTask {
```

pkg/utils/utils.go
dimriou: Mostly in tests, but there are a few exceptions like:
- https://github.com/smartcontractkit/chainlink/blob/develop/core/services/pg/lease_lock.go#L181
- https://github.com/smartcontractkit/chainlink/blob/develop/core/cmd/dkgencrypt_keys_commands.go#L27
- https://github.com/smartcontractkit/chainlink/blob/develop/core/cmd/admin_commands.go#L158

jmank88: It seems like overkill that we don't want to encourage but I guess that is out of scope :shrug:

pkg/utils/sleeper_task.go
dimriou: But that means in field or variable declarations, like [here](https://github.com/smartcontractkit/chainlink/blob/develop/core/chains/evm/monitor/balance.go#L45) we will have a struct instead of an interface. Not sure if that's the best approach.

jmank88: Yes. That is not problematic and callers are still welcome to use the same exact interface for their own purposes if they want.

jmank88: It is idiomatic Go, and should our default choice until there is a specific reason to do otherwise.

jmank88: (it also makes testing way easier, and should allow us to avoid a cast in this case, and a particularly dangerous one that crosses package boundaries and uses an anonymous type)
https://github.com/smartcontractkit/chainlink/blob/71d314fe1b6aa1c35fc7c7e6d65525fee3d18bdd/core/chains/evm/monitor/balance_helpers_test.go#L4

jmank88: https://go.dev/blog/module-compatibility
> Where possible, it is better to avoid this class of problem entirely. When designing constructors, for example, prefer to return concrete types. Working with concrete types allows you to add methods in the future without breaking users, unlike interfaces. That property allows your module to be extended more easily in the future.

prashantkumar1982 (dismissed):

pkg/utils/sleeper_task_test.go
jmank88, line 1: It would be nice to avoid gomega if possible, but I'm not aware of a drop-in replacement for eventually/consistently. It looks like we could just receive from a chan instead of using a counter in this case though :thinking:

jmank88: - https://github.com/smartcontractkit/chainlink-common/pull/300

go.mod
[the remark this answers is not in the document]
dimriou: That's ok, I'll clean up

jmank88 (approved):

PR #296, opened 2023-12-19
Extract core/store/models

DylanTinianov:
Add common models for `Duration` and `URL`


pkg/models/url.go
dimriou, line 1: There is already a URL package here: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/config/url.go

pkg/utils/utils.go
```diff
@@ -49,3 +50,11 @@ func IsZero[C comparable](val C) bool {
 	var zero C
 	return zero == val
 }
+
+// EnsureHexPrefix adds the prefix (0x) to a given hex string.
+func EnsureHexPrefix(str string) string {
```
jmank88, line 55: There is a `hex.EnsureHexPrexix` you can use.

pkg/models/duration.go
jmank88, line 1: Where is this duration used? We have tried to move away from it, and if we could isolate to to chainlink-evm that would be good.

If it is just for DB, then we have `package sqlutil`

jmank88: Oh right we have config.Duration too: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/utils/duration.go
This seems different though. I think probably sqlutil, or leave out of this repo (even if we have to have a copy in core and evm separately) :thinking:

dimriou (dismissed):
I saw you removed `EnsureHexPrefix` from the main repo, so just remove it from here as well. Otherwise looks good

jmank88 (approved):

krehermann (approved):

pkg/sqlutil/duration.go
jmank88, line 1: I think I steered you wrong here. Is this type actually different from `config.Duration`? If not, let's use that instead. If it is different, I think many of the core side uses should actually be using that (e.g. all the config related fields and iterfaces)

jmank88 (changes requested):

pkg/sqlutil/duration.go
jmank88: It looks super similar - maybe we just need to add a few methods or funcs?

DylanTinianov: I don't see anything in Chainlink using `config.Duration`. I think at this point it would be find to remove this and just use `sqlutil.Duration`. Are there other repos using `config.Duration`?

jmank88: `config.Duration` is the canonical type in the correct location. It is used across multiple repos. `sqlutil.Duration` should be removed.

pkg/config/duration.go
```diff
@@ -27,12 +27,51 @@ func MustNewDuration(d time.Duration) *Duration {
 	return &rv
 }
 
+func MakeDuration(d time.Duration) (Duration, error) {
+	if d < time.Duration(0) {
+		return Duration{}, fmt.Errorf("cannot make negative time duration: %s", d)
+	}
+	return Duration{d: d}, nil
+}
```
jmank88, line 35: This func looks the same as `NewDuration`. Can we call that instead?
```suggestion
```

pkg/config/duration.go
```diff
@@ -27,12 +27,51 @@ func MustNewDuration(d time.Duration) *Duration {
...
+	return MakeDuration(d)
+}
+
+func MustMakeDuration(d time.Duration) Duration {
+	rv, err := MakeDuration(d)
+	if err != nil {
+		panic(err)
+	}
+	return rv
+}
```
jmank88, line 52: This is basically the same as `MustNewDuration`, and callers could just call `*MustNewDuration(d)` to get the same result.
```suggestion
```
I don't love that `New` and `MustNew` are not aligned, but I think it would still be simpler to keep this API small by just using those directly in these cases.

DylanTinianov: Agreed, I'll make this update as well.

pkg/sqlutil/duration.go
DylanTinianov: Removed sql.Duration

jmank88 (dismissed):

pkg/config/duration_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+func TestDuration_Scan_Value(t *testing.T) {
+	t.Parallel()
+
+	d := MustNewDuration(100)
+	require.NotNil(t, d)
+
+	d.Duration()
```
amit-momin, line 39: nit:  I think this might be a stray line

DylanTinianov: Fixed

amit-momin (approved):

PR #297, opened 2023-12-20
Add automation types and services

ferglor:
The bulk of this PR is a cut/paste of the `v3/types` package in chainlink-automation

We're also moving the `MercuryCredentials` type from core, into the `RelayArgs`

Aside from that, I've added some new functions to the AutomationProvider that will expose services needed for automation in core

## Associated PRs

chainlink: https://github.com/smartcontractkit/chainlink/pull/11631
chainlink-automation: https://github.com/smartcontractkit/chainlink-automation/pull/299


pkg/types/automation/basetypes.go
```diff
@@ -0,0 +1,349 @@
...
+	"PerformData":"%s",
+	"FastGasWei":%s,
+	"LinkNative":%s,
+	"RetryInterval":%d
+}`
+
+func init() {
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, " ", "", -1)
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, "\t", "", -1)
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, "\n", "", -1)
```
jmank88, line 35: `ReplaceAll` is available:
```suggestion
	checkResultStringTemplate = strings.ReplaceAll(checkResultStringTemplate, " ", "")
	checkResultStringTemplate = strings.ReplaceAll(checkResultStringTemplate, "\t", "")
	checkResultStringTemplate = strings.ReplaceAll(checkResultStringTemplate, "\n", "")
```
Although this seems odd. Is the point to keep it in human readable form in source?

jmank88: Oh and there is a `Replacer`:
```suggestion
	checkResultStringTemplate = strings.NewReplacer(" ", "", "\t", "", "").Replace(checkResultStringTemplate)
```

pkg/types/automation/trigger.go
```diff
@@ -0,0 +1,87 @@
...
+
+// LogTriggerExtension is the extension used for log triggers,
+// It contains information of the log event that was triggered.
+// NOTE: This struct is sent on the p2p network as part of observations to get quorum
+// Any change here should be backwards compatible and should keep validation and
+// quorum requirements in mind. Please ensure to get a proper review along with an
+// upgrade plan before changing this
+type LogTriggerExtension struct {
+	// LogTxHash is the transaction hash of the log event
+	TxHash [32]byte
```
jmank88, line 59: I see a lot of `[32]byte` hash types. Is that something we will have to relax later on for other chains? like to a slice `[]byte`?

How are these values actually used? The approach taken with accounts is to just use a `string`, which for EVM happens to be the encoded hex format, but then never try to parse those values - only compare them directly. It would be nice to be able to make a similar kind of simplification here.

pkg/types/automation/interfaces.go
```diff
@@ -0,0 +1,139 @@
...
+	Close() error
+	HealthReport() map[string]error
+}
+
+type EventProvider interface {
+	Name() string
+	Start(_ context.Context) error
+	Close() error
+	Ready() error
+	HealthReport() map[string]error
```
jmank88, line 32: You can use `services.Service` for this common set.
```suggestion
	services.Service
```
You might want/need to upgrade anything with Start/Close to `services.Service` also, since that suggests they have background processes/state that could be monitored.

jmank88: Otherwise there is `services.StartClose`:
```go
type StartClose interface {
	Start(context.Context) error
	Close() error
}
```

pkg/types/automation/basetypes.go
ferglor: Yeah so the init implementation is just to collapse the more readable source string into a single line; my understanding here is that using a JSON encoder was deemed to be too expensive at the time, and using a string template was cheaper/faster. I didn't want to modify any of the code lifted out of automation in this PR; I wanted to keep everything as it was written in the automation repo to keep the impacts of these changes to a minimum

pkg/types/automation/trigger.go
ferglor: The TxHash is used to build a log identifier for a log event; I think we use the block hash as well and a few other fields maybe? I think we also hex encode this field for a string representation as well!

pkg/types/relayer.go
```diff
@@ -16,12 +16,20 @@ type PluginArgs struct {
...
-	New           bool // Whether this is a first time job add.
-	RelayConfig   []byte
-	ProviderType  string
+	ExternalJobID      uuid.UUID
+	JobID              int32
+	ContractID         string
+	New                bool // Whether this is a first time job add.
+	RelayConfig        []byte
+	ProviderType       string
+	MercuryCredentials *MercuryCredentials
```
krehermann, line 25: This looks odd to me. Everything else in this struct is at the system level - they are necessary field for any relayer. Whereas this seems very specific.

What is the use case for adding this here?

ferglor: MercuryCredentials can be provided when building the EVM Registry for 2.1 here: https://github.com/smartcontractkit/chainlink/blob/develop/core/services/ocr2/plugins/ocr2keeper/evmregistry/v21/registry.go#L84

In terms of passing additional config, I think we're limited to either relay args or plugin args; is there a more appropriate location where we could put this?

pkg/types/automation/interfaces.go
```diff
@@ -0,0 +1,136 @@
...
+	Unsubscribe(int) error
+	Start(context.Context) error
+	Close() error
+}
+
+type UpkeepStateUpdater interface {
+	SetUpkeepState(context.Context, CheckResult, UpkeepState) error
+}
+
+type RetryQueue interface {
```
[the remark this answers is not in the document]
ferglor: Would it be okay to add context and error as a separate PR immediately after this one? I wanted to get this set of changes merged exactly as they were in the automation code base, without any significant modifications, if anything to reduce the impact on testing?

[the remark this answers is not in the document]
jmank88: Yeah that should be fine.

pkg/types/relayer.go
krehermann: were are the mercury credentials stored? are they encrypted in the KeyStore, which is owned by the core node? rather than passing them as a struct here can it be pushed up the stack so that the Relayer has access to an interface that exposes something that can dish out mercury creds?

the evm relayer is already accessing subsets of the key store here https://github.com/smartcontractkit/chainlink/blob/ba841a51666daeb8252d034c5f64da2f637f5f7d/core/services/relay/evm/evm.go#L65C9-L65C9

would extending that work?

ferglor: So I want to say an earlier version of the original PR had modified the keystore interface; but i think that approach fell short at some point! I'll dig into the original PR and see if I can find it! I think I created something like a `CSAETHMercuryKeystore` or something similar 🤔

ferglor: Ah so this is the old PR I'm referring to: https://github.com/smartcontractkit/chainlink/pull/11647/files

Though, looking at those changes, I'm not sure exactly why I abandoned that changeset - I'll try this approach again to see if I can avoid updating the relay args! 👍

jmank88: This comes from the TOML config right? In that case I don't think we are violating anything here.

krehermann: i was assuming that the credentials are not in the TOML. if i'm understanding what you are saying, that sounds like security credential in a text file

jmank88: They are in the `secrets.toml`: https://github.com/smartcontractkit/chainlink/blob/develop/docs/SECRETS.md#mercurycredentialsname

ferglor: @krehermann so the more I look into it; I think the keystore changes I had made previously were addressing a separate feature rather than exposing the mercury credentials; the mercury credentials come from the `DelegateConfig`

krehermann: Ok, thanks for clarifying. Seems like we should merge this then add the ctx everywhere to the interfaces

krehermann (dismissed):

jmank88 (dismissed):

pkg/types/automation/interfaces.go
```diff
@@ -0,0 +1,79 @@
...
+	"io"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
+
+// UpkeepStateStore is the interface for managing upkeeps final state in a local store.
+type UpkeepStateStore interface {
+	UpkeepStateUpdater
+	UpkeepStateReader
+	Start(context.Context) error
```
amirylm, line 14: We can avoid `Start()` and `Close()` functions for all interfaces, as they are not really part of the interfaces (at least not the logical interface) and are only called from within the repo with the implementations

pkg/types/automation/trigger.go
infiloop2: Having these as concrete `[32]bytes` gives us a lot of safety as these values are sent within observations and outcomes in OCR, and having a tight control on the size reduces surface area for potential security issues and gives us better control over how many upkeeps we can include in an observation/outcome. 

We haven't run into chains where this has been an issue yet, but if we generalise this to byte slice then will need to do a proper review of any downstream implications.

PR #299, opened 2023-12-20
Allow bindings to be added to a chain reader

nolag:


pkg/loop/internal/chain_reader.go
```diff
@@ -132,6 +142,15 @@ func (c *chainReaderServer) GetLatestValue(ctx context.Context, request *pb.GetL
 	return &pb.GetLatestValueReply{RetVal: encodedRetVal}, nil
 }
 
+func (c *chainReaderServer) Bind(ctx context.Context, bindings *pb.BindRequest) (*emptypb.Empty, error) {
```
ilija42, line 145: why does bind have to return empty?

ilija42 (approved):

pkg/loop/internal/chain_reader.go
nolag: There's nothing to say to the caller.

jmank88 (approved):

PR #301, opened 2023-12-21
Add MercuryCredentials to RelayArgs

ferglor:
This PR reduces the responsibilities of the `AutomationProvider`, since the existing code in core will take care of some services needed for the delegate config, making `AutomationProvider` responsible only for the services passed to the `automationServices` slice

 This PR also introduces a `MercuryCredentials` struct that can be passed in the `RelayArgs`

### DIFF
--- a/pkg/types/relayer.go
+++ b/pkg/types/relayer.go
@@ -22,7 +22,7 @@ type RelayArgs struct {
 	New                bool // Whether this is a first time job add.
 	RelayConfig        []byte
 	ProviderType       string
-	MercuryCredentials MercuryCredentials
+	MercuryCredentials *MercuryCredentials
 }
 
 type MercuryCredentials struct {
--- a/pkg/types/relayer.go
+++ b/pkg/types/relayer.go
@@ -16,12 +16,20 @@ type PluginArgs struct {
 }
 
 type RelayArgs struct {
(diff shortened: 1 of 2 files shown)


pkg/config/toml.go
jmank88, line 1: Is this diff right?

pkg/types/relayer.go
```diff
@@ -16,12 +16,20 @@ type PluginArgs struct {
...
-	New           bool // Whether this is a first time job add.
-	RelayConfig   []byte
-	ProviderType  string
+	ExternalJobID      uuid.UUID
+	JobID              int32
+	ContractID         string
+	New                bool // Whether this is a first time job add.
+	RelayConfig        []byte
+	ProviderType       string
+	MercuryCredentials MercuryCredentials
```
jmank88, line 25: I had been thinking that this would be hidden inside of the opaque `RelayConfig []byte`, rather than be exposed for everyone like this. We might have to reconsider other approaches.
cc @krehermann @cedric-cordenier

jmank88 (changes requested):

jmank88:
Is there technical reason that mocks need to be in this repo? If not, I think we should leave them out. Anyone who needs them can generate them for themselves.

PR #302, opened 2023-12-21
Better error messages

nolag:


reductionista:
This is good. Ideally, we'd also have some information about whether this is during Encoding or Decoding, but maybe that's too difficult to know here.  Attaching a stack trace to the error could also help.

jmank88 (approved):

PR #306, opened 2024-01-05
Add an example for codec and updates some docs.

nolag:


pkg/codec/example_test.go
```diff
@@ -0,0 +1,177 @@
...
+const (
+	anyUnmodifiedTypeName     = "Unmodified"
+	anyModifiedStructTypeName = "SecondItem"
+)
+
+var _ types.RemoteCodec = &ExampleStructJsonCodec{}
+
+type ExampleStructJsonCodec struct{}
+
+func (j ExampleStructJsonCodec) Encode(_ context.Context, item any, _ string) ([]byte, error) {
```
jmank88, line 27: Could drop the method unused method receivers?
```suggestion
func (ExampleStructJsonCodec) Encode(_ context.Context, item any, _ string) ([]byte, error) {
```

pkg/codec/example_test.go
```diff
@@ -0,0 +1,177 @@
...
+// one that is modified before encoding for on-chain and modified after decoding for off-chain
+// the other is left unmodified during encoding and decoding.
+const (
+	anyUnmodifiedTypeName     = "Unmodified"
+	anyModifiedStructTypeName = "SecondItem"
+)
+
+var _ types.RemoteCodec = &ExampleStructJsonCodec{}
+
+type ExampleStructJsonCodec struct{}
```
jmank88, line 25:
```suggestion
type ExampleStructJSONCodec struct{}
```

pkg/codec/example_test.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Example demonstrates how to use the codec package.
+// It will make use of each [Modifier] provided in the package, along with their config.
+func Example() {
+	mods := createModsFromConfig()
+
+	c, err := codec.NewModifierCodec(&ExampleStructJsonCodec{}, mods)
+	if err != nil {
+		panic(err)
```
jmank88, line 91: I think panicing disrupts the rest of the test suite. How about something simpler?
```suggestion
		fmt.Println(err)
		return
```

pkg/codec/example_test.go
jmank88, line 1: How does this render in html? I know that whole file examples are supported, but I wonder about the unexported helper funcs - it might be cleaner to have those embedded in the body of `func Example` :shrug:

jmank88: > it might be cleaner to have those embedded in the body of `func Example`

I don't think they necessarily need to be completely inline, but it would read easier if all of the logging happened from `func Example`, instead of printing things along the way deep in the stack.

pkg/types/chain_reader.go
```diff
@@ -16,23 +16,28 @@ const (
...
+	// It must encode as an object via [json.Marshal] and [github.com/fxamacker/cbor/v2.Marshal].
+	// Typically, would be either a struct with field names mapping to arguments, or anonymous map such as `map[string]any{"baz": 42, "test": true}}`
 	//
-	// returnVal must [json.Unmarshal] as an object, and so should be a map, struct, or implement the [json.Unmarshaler] interface.
+	// returnVal must [json.Unmarshal] and and [github.com/fxamacker/cbor/v2.Marshal] as an object.
 	//
 	// Example use:
 	//  type ProductParams struct {
-	// 		Arg int `json:"arg"`
+	// 		Id int `json:"arg"`
```
jmank88, line 27:
```suggestion
	// 		ID int `json:"arg"`
```

pkg/codec/example_test.go
nolag: I'm ok inlining those two functions for readability, I had tried to put the expected output in each function, but that didn't work as expected.

pkg/types/chain_reader.go
```diff
@@ -16,23 +16,28 @@ const (
...
-	// 		Arg int `json:"arg"`
+	// 		ID int `json:"id"`
 	//  }
 	//  type ProductReturn struct {
 	// 		Foo string `json:"foo"`
 	// 		Bar *big.Int `json:"bar"`
 	//  }
 	//  func do(ctx context.Context, cr ChainReader) (resp ProductReturn, err error) {
-	// 		err = cr.GetLatestValue(ctx, bc, "method", ProductParams{Arg:1}, &resp)
+	// 		err = cr.GetLatestValue(ctx, "FooContract", "GetProduct", ProductParams{Id:1}, &resp)
```
jmank88, line 34:
```suggestion
	// 		err = cr.GetLatestValue(ctx, "FooContract", "GetProduct", ProductParams{ID:1}, &resp)
```

pkg/codec/example_test.go
```diff
@@ -0,0 +1,171 @@
...
+	ctx := context.Background()
+	b, err := c.Encode(ctx, input, anyUnmodifiedTypeName)
+	fmt.Println(string(b))
+
+	output := &OnChainStruct{}
+	if err = c.Decode(ctx, b, output, anyUnmodifiedTypeName); err != nil {
+		fmt.Println(err)
+		return
+	}
+	fmt.Println(reflect.DeepEqual(input, output))
```
jmank88, line 117: Printing a bool result from a `reflect.DeepEqual` isn't very human friendly. What if we just logged the result instead?
```suggestion
	fmt.Printf("%#v\n", output)
```
Then we can expect something more interesting than `true` in the `Output`.

pkg/codec/example_test.go
```diff
@@ -0,0 +1,171 @@
...
+	expected := *modifedInput
+
+	// Only the middle value was extracted, so decoding can only provide a single value back.
+	expected.Dd = []string{"great example"}
+	fmt.Println(reflect.DeepEqual(&expected, output2))
+	// Output:
+	// {"Aa":10,"Bb":"20","Cc":true,"Dd":"great example","Ee":631515600,"Ff":"dog"}
+	// true
+	// {"Aa":10,"Bb":"","Cc":true,"Dd":"great example","Ee":631515600,"Ff":"dog"}
+	// true
```
jmank88, line 151: Combined with the above suggestion, what if we prefixed each line with a description?
```suggestion
	// Output:
	// Encoded: {"Aa":10,"Bb":"20","Cc":true,"Dd":"great example","Ee":631515600,"Ff":"dog"}
	// Decoded: {Aa: 10, Bb 20, ...
	// Modified Encoded: {"Aa":10,"Bb":"","Cc":true,"Dd":"great example","Ee":631515600,"Ff":"dog"}
	// Modified Decoded: ...
```

pkg/types/chain_reader.go
reductionista: If we're changing the name of the golang field from Arg to ID, shouldn't we also change the json field from "arg" to "ID"?

pkg/codec/example_test.go
nolag: It has time.Time in it in one example.  Time is annoying because it formats with the time zone :(.

pkg/codec/example_test.go
nolag: Added prefix, see comment above for the why on true/false.

pkg/codec/element_extractor.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
 )
 
+// ElementExtractorLocation is used to determine which element to extract from a slice or array.
+// The default is ElementExtractorLocationMiddle, which will extract the middle element.
+// valid json values are "first", "middle", and "last".
```
reductionista, line 14: I'm confused by this whole setup.

Let's say you have a slice of 10 elements.  Which element gets extracted if you select "middle", and why is "middle" the default?  How would I extract, say, element 7?

ilija42: https://github.com/smartcontractkit/chainlink-common/pull/262#discussion_r1411196194

ilija42: https://github.com/smartcontractkit/chainlink-common/pull/262#discussion_r1411196194

ilija42 (approved):

pkg/codec/example_test.go
jmank88: Can't we control that?
- https://pkg.go.dev/time#Time.UTC
- https://pkg.go.dev/time#Time.In

pkg/codec/element_extractor.go
nolag: For now, we support first, middle, and last.  Middle was used for median and seemed to me to be the most likely one people would want.  I could not default, and simply throw.  I guess we could add index, but I'm not sure when it would be used or how easy it would be to use.  You may not always know the number of elements, my thought is it's mostly going to be based on the number of NOPs that reported.

Maybe I should change it to just be middle?  I didn't call it median, because it doesn't sort the data itself.

jmank88 (approved):

pkg/codec/element_extractor.go
reductionista: I see... maybe just add a mention in the docstring that it will use the lower of the two closest to the middle if it's an even number of elements?  Not sure how to say that succinctly.

reductionista: If "middle" is the only current use case, I think leaving it as default is fine

PR #307, opened 2024-01-05
Fix marshalling for config, also make naming more consistent 

nolag:


pkg/codec/config.go
```diff
@@ -72,9 +73,9 @@ func (m *ModifiersConfig) ToModifier(onChainHooks ...mapstructure.DecodeHookFunc
...
 
 const (
-	RenameModifier             ModifierType = "rename"
-	DropModifier               ModifierType = "drop"
-	HardCodeModifier           ModifierType = "hard code"
+	RenameModifierType         ModifierType = "rename"
+	DropModifierType           ModifierType = "drop"
+	HardCodeModifierType       ModifierType = "hard code"
 	ExtractElementModifierType ModifierType = "extract element"
 	EpochToTimeModifierType    ModifierType = "epoch to time"
```
jmank88, line 80: Alternatively, this shorter style is sometimes used when it's a prefix :shrug: 
```suggestion
	ModifierRename         ModifierType = "rename"
	ModifierDrop           ModifierType = "drop"
	ModifierHardCode       ModifierType = "hard code"
	ModifierExtractElement ModifierType = "extract element"
	ModifierEpochToTime    ModifierType = "epoch to time"
```

jmank88 (approved):

pkg/codec/config.go
```diff
@@ -117,30 +125,51 @@ func (d *DropModifierConfig) ToModifier(onChainHooks ...mapstructure.DecodeHookF
...
+}
+
+// ElementExtractorModifierConfig is used to extract an element from a slice or array
+type ElementExtractorModifierConfig struct {
 	// Key is the name of the field to extract from and the value is which element to extract.
 	Extractions map[string]*ElementExtractorLocation
 }
 
-func (e *ElementExtractorConfig) ToModifier(onChainHooks ...mapstructure.DecodeHookFunc) (Modifier, error) {
+func (e *ElementExtractorModifierConfig) ToModifier(_ ...mapstructure.DecodeHookFunc) (Modifier, error) {
```
reductionista, line 141: I'm curious, why are we ignoring the hook functions for drop & extract, but not for hard code?

nolag: dropping/rename change the name in an underlying struct.  They don't actually do anything with maps. 

extracting one takes an element from the value.  It doesn't need to know anything about how your types are made.  Your type is already there, or will later be converted at that point.

Hard-code will need to know how to sometimes convert types.  For example, an address is a string in config, but it doesn't if it's a hex encoding, base64 encoding, or something else.  Big int can overflow integers in json, so you may need to quote the value, it needs to have the big.Int hook, etc.

reductionista (approved):

PR #310, opened 2024-01-10
Nightly Fuzz CI

vyzaldysanchez:


.github/workflows/nightlyfuzz.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          path: "${{ steps.get-corpus-dir.outputs.corpus_dir }}"
+          # We need to ensure uniqueness of the key, as saving to a key more than once will fail (see Save corpus step).
+          # We never expect a cache hit with the key but we do expect a hit with the restore-keys prefix that is going
+          # to match the latest cache that has that prefix.
+          key: "nightlyfuzz-corpus-${{ github.run_id }}-${{ github.run_attempt }}"
+          restore-keys: "nightlyfuzz-corpus-"
+      - name: "Run native fuzzers"
+        # Fuzz for 5 hours, after 6 hours the action would time out
+        run: "cd fuzz && ./fuzz_all_native.py --seconds 18000"
```
jmank88, line 42: Might be too long :shrug:

jmank88:
LGTM!
Is there a plan to make this reusable for other repos, so we don't have to copy it around?

.github/workflows/nightlyfuzz.yml
vyzaldysanchez: What time would you suggest us to do?

vyzaldysanchez:
@jmank88 I'll be working with RelEng to make a reusable action.

.github/workflows/nightlyfuzz.yml
jmank88: Maybe 30m or 1h to start?

vyzaldysanchez: I think that's reasonable!

jmank88 (approved):

PR #312, opened 2024-01-12
add context

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2887

Supports:
- https://github.com/smartcontractkit/chainlink/pull/11773


pkg/loop/adapter.go
krehermann, line 1: what happened here? unused?

jmank88: Yeah it only served to bridge the gap with context.

pkg/monitoring/exporter_kafka_test.go
```diff
@@ -36,7 +36,7 @@ func TestKafkaExporter(t *testing.T) {
 		nodes := []NodeConfig{generateNodeConfig()}
 		exporter, err := factory.NewExporter(ExporterParams{chainConfig, feedConfig, nodes})
 		require.NoError(t, err)
-		envelope, err := generateEnvelope()
+		envelope, err := generateEnvelope(ctx)
```
topliceanu, line 39: I'm confused by the introduction of context here. Most of these methods have no concurrency in their implementation - ctx is not needed and won't help. 

Take `generateEnvelope`, it produces a structure filled with random values. https://github.com/smartcontractkit/chainlink-common/pull/312/files#diff-be6ec1c2691cf18a8a73b7ea25c78bab3dc41d5bccde1df3e9626310e849306bR454-R477

topliceanu (commented):
I only looked at the monitoring package. 
I'm confused by this change but I don't have context so I't only a "Comment" review.

pkg/monitoring/exporter_kafka_test.go
jmank88: What is the relevance of concurrency? We are motivated by carrying request scoped values for tracing.

This func calls `generateContractConfig`, which calls `median.StandardOnchainConfigCodec.Encode`, which now requires a context.

krehermann (approved):

PR #314, opened 2024-01-13
Add fuzz testing for codec and allow non-UTF-8 strings to be decoded

nolag:


.github/workflows/pkg.yml
```diff
@@ -20,6 +20,9 @@ jobs:
       - name: Unit Tests
         run: go test -race ./... -coverpkg=./... -coverprofile=pkg_coverage.out
       
+      - name: Fuzz Tests
+        run: go test github.com/smartcontractkit/chainlink-common/pkg/loop/internal -fuzz . -fuzztime 12s
```
jmank88, line 24: Maybe no harm to leave in as-is, but this quarter @vyzaldysanchez is working on a standard way to fuzz and starting with this repo.

pkg/types/interfacetests/codec_interface_fuzz_tests.go
```diff
@@ -0,0 +1,64 @@
...
+					},
+				},
+			}
+			codec := tester.GetCodec(t)
+			ctx := tests.Context(t)
+
+			encoded, err := codec.Encode(ctx, testStruct, TestItemType)
+			require.NoError(t, err)
+			decoded := &TestStruct{}
+			err = codec.Decode(ctx, encoded, decoded, TestItemType)
```
jmank88, line 54: I think would it would also be worthwhile to fuzz `func codec.Decode` directly, with raw encoded input, just to ensure it won't panic.

.github/workflows/pkg.yml
nolag: Ok, I can sync with him.  Maybe we get it into `Next` for now, and then I can see what the status is.  If it's close, I can wait or help, otherwise, he can get rid of this once he's ready :)

nolag: It's not going to be ready yet, he said I could check this in and he'll merge into what he's doing when it is.

pkg/types/interfacetests/codec_interface_fuzz_tests.go
```diff
@@ -0,0 +1,71 @@
...
+			// big.Int can represent zero in many ways, make it the same
+			if testStruct.BigField.Cmp(big.NewInt(0)) == 0 {
+				require.Equal(t, 0, decoded.BigField.Cmp(big.NewInt(0)))
+				decoded.BigField = testStruct.BigField
+			}
+			require.Equal(t, testStruct, decoded)
+
+			// Decode random bytes doesn't panic
+			// _ to get rid of compiler warnings, we just want to make sure of no panic!
+			_ = codec.Decode(ctx, decode, decoded, TestItemType)
```
jmank88, line 68: This works, but it couples the two unrelated tests together in to one corpus, even though they don't share any inputs. It would be more effective as a separate fuzz test.

jmank88 (approved):

pkg/types/interfacetests/codec_interface_fuzz_tests.go
nolag: Yeah, I agree.  I wanted to do `t.Run` for both, but it froze...   You also can't sub-test a fuzz :(.

I aim to have it as an interface test to ensure _**all**_ implementations do the same fuzzing.  If they are separate tests, then there's no way to make sure they all remember to call each of the fuzzes (especially if we ever add more).  This way, we can assure that much if they intended to.

jmank88: If we fight the tooling we are going to have a bad time. This approach has usability and scalability problems. It is important to be able to hone in a in a single test at a time, and we shouldn't have to throw away our historical corpus every time we add a new test that expands the parameter set.

PR #315, opened 2024-01-16
Add nop-friendly error logging

vyzaldysanchez:
This PR sends errors from the LOOP binary and GRPC calls to the core node so they can be presented in the UI to NOPs.

### DIFF
--- a/pkg/loop/internal/reporting.go
+++ b/pkg/loop/internal/reporting.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"time"
 
+	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 	"google.golang.org/protobuf/types/known/emptypb"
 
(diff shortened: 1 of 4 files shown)


pkg/loop/internal/reporting.go
```diff
@@ -5,23 +5,26 @@ import (
 	"math"
 	"time"
 
+	"github.com/pkg/errors"
```
jmank88, line 8:
```suggestion
	"errors"
```

pkg/loop/internal/reporting_plugin_service.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/mwitkow/grpc-proxy/proxy"
+	"github.com/pkg/errors"
```
jmank88, line 7:
```suggestion
	"errors"
```

pkg/loop/internal/reporting.go
```diff
@@ -56,9 +62,12 @@ func (r *reportingPluginFactoryClient) NewReportingPlugin(config libocr.Reportin
 	}
 	cc, err := r.brokerExt.dial(reply.ReportingPluginID)
 	if err != nil {
+		if saveErr := r.errorLog.SaveError(ctx, err.Error()); saveErr != nil {
```
vyzaldysanchez, line 65: I'd like to extract this into a utility function so we can save ourselves the duplication, but I wonder which would be the best place to put that function. It would be something like `TrySaveError(ctx, errorLog, err) error`

vyzaldysanchez: cc: @jmank88 @cedric-cordenier

jmank88: Can we slow down and frame the problem we are trying to solve here first?
All logs and errors are already returned to the host, so this seems redundant in principle.

vyzaldysanchez: Based on @cedric-cordenier's request, we would like these errors to popup in the UI by saving them when calling `errorLog`. I wonder if we can just modify the line in which we call `NewLOOPPService` on the node, and any error returned just save it in the error log instead. This would make them appear in the UI, right?

jmank88: An "error" log is a solution to a problem. What is the problem?

jmank88: If we _must_ act on every error for some reason, then we should use grpc interceptors to do it in just one place.

PR #316, opened 2024-01-17
[KS-14] Add values library

cedric-cordenier:
Add a values library to be used for the capability API for Keystone. This is heavily inspired by go's structpb, but adds special support for eg. decimals (which we use regularly) and bytes, to ensure the conversion isn't lossless of the type information.

This provides a basic set of types (ints, decimals, lists, maps, strings, bytes) which are serializable to protos by default. All of these types satisfy a common "Value" interface.


pkg/values/list.go
```diff
@@ -0,0 +1,49 @@
+package values
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/values/pb"
+)
+
+type List struct {
+	Value []Value
```
jmank88, line 8: Must all elements be the same type of `Value`?

cedric-cordenier: No, there's no requirement for that

pkg/values/bool.go
```diff
@@ -0,0 +1,21 @@
+package values
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/values/pb"
+)
+
+type Bool struct {
+	Value bool
```
krehermann, line 8: this had me confused for minute. here `Value` is a field name. other places `Value` is type.

1. does this field need to be public? it seem like the intention is to use the interface funcs, not the embedded field
2. can you change the name of the field to something else, maybe impl or data?

the readibility is even more difficult for me in the Map type ~ Value map[any]Value

krehermann (changes requested):

pkg/values/bool.go
cedric-cordenier: Yes, can rename :) I think leaving it public is valuable since if you type coerce it you can then get access to the underlying typed value.

PR #321, opened 2024-01-26
Add error value

cedric-cordenier:
For the callback channel, we want a way to signal that the capability has encountered an error.

We discussed two ways of addressing this:
* Extending values to support an error type
* Changing the capability API to return something like a Response struct with a value or an error in it.

Both solutions seem very similar since we'll have to handle both cases everywhere anyway, but the former has the advantage of fitting within the existing API, so we decided to play around with modelling errors as values.


cedric-cordenier:
Thinking about this more, I'm pretty sure this is a horrible idea.

PR #322, opened 2024-01-26
Bcf 2572 relayer fuzz test

EasterTheBunny:
This PR adds basic fuzz testing functions that apply fuzzing to `PluginRelayer.NewRelayer`
`Relayer.NewConfigProvider`. Since fuzzing does not currently support parallel runs, the fuzz
functions are designed in a granular way where a test constructor function that returns the
testable interface is provided to the fuzz testing functions.
    
Nested fuzz tests have not yet been tried, but this generally appears to be a bad idea
considering how fuzz testing works. ex:
    
```
func RunFuzzPluginRelayer(f *testing.F) {
    f.Fuzz(func(t *testing.T)) {
        f.Fuzz(func(t *testing.T)) {
    
        }
    }
}
```

### DIFF
--- a/pkg/loop/internal/test/relayer.go
+++ b/pkg/loop/internal/test/relayer.go
@@ -13,6 +13,8 @@ import (
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
@@ -381,8 +383,10 @@ func RunFuzzPluginRelayer(f *testing.F, relayerFunc func(*testing.T) internal.Pl
 			errStr:        fErr,
 		}
 
-		ctx := context.Background()
-		_, _ = relayerFunc(t).NewRelayer(ctx, fConfig, keystore)
+		ctx := tests.Context(t)
+		_, err := relayerFunc(t).NewRelayer(ctx, fConfig, keystore)
+
+		grpcUnavailableErr(t, err)
 	})
 }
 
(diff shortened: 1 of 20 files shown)


pkg/loop/internal/test/relayer.go
```diff
@@ -358,3 +361,145 @@ func RunRelayer(t *testing.T, relayer internal.Relayer) {
...
+
+		ctx := tests.Context(t)
+		_, err := relayerFunc(t).NewRelayer(ctx, fConfig, keystore)
+
+		grpcUnavailableErr(t, err)
+	})
+}
+
+func RunFuzzRelayer(f *testing.F, relayerFunc func(*testing.T) internal.Relayer) {
+	f.Add([]byte{}, int32(-1), "ABC\xa8\x8c\xb3G\xfc", false, []byte{}, "", "", []byte{})
```
jmank88, line 389: It would probably make sense to include a few realistic sets of args here, like one relayer/plugin config pair for each provider type.

EasterTheBunny: Like multiple `f.Add` calls with those pairs? The approach I went with was inputs that might be more likely to cause problems.

vyzaldysanchez: Agreed!

jmank88: The fuzz runtime will try to explore cases that expand the coverage of the code being fuzzed, so realistic test cases provides hints about how to get past validation and in to deeper parts of the stack early-on, without having to "discover" valid sets of inputs through randomization. We can still include problematic cases too.

EasterTheBunny: Ok. I just added an update to this. I think it is a little messy since it appears valid inputs are very common between all providers, but I'm not completely sure.

jmank88: They should each have particular differences in the RelayConfig or PluginConfig. There is also the chainreader/codec config to consider as well.

EasterTheBunny: Ok. All I have to go on right now is the values currently in test cases. Where can I see some valid inputs to some of these providers?

jmank88: We will probably have to look in the core tests for those. @nolag and @reductionista can help with chainreader/codec configs.

It is also possible that some products don't have any quirks. :shrug:

jmank88 (approved):

vyzaldysanchez (approved):

PR #323, opened 2024-02-05
KS-15: Add CapabilityRegistry GRPC implementation

cedric-cordenier:
- Move capabilities from core into chainlink-common.
- Add a nil value to the values package
- Implement the capabilities registry GRPC implementation, including
    support for calling the execution and metadata APIs of all of the
    capability types.


pkg/loop/internal/capabilities_registry.go
```diff
@@ -0,0 +1,655 @@
...
+type CapabilitiesRegistryClient struct {
+	*brokerExt
+	client pb.CapabilitiesRegistryClient
+}
+
+var _ capabilities.Registry = (*CapabilitiesRegistryClient)(nil)
+
+func NewCapabilitiesRegistryClient(broker Broker, brokerCfg BrokerConfig, conn *grpc.ClientConn) (*CapabilitiesRegistryClient, error) {
+	return &CapabilitiesRegistryClient{
+		client:    pb.NewCapabilitiesRegistryClient(conn),
```
cedric-cordenier, line 29: client -> pbclient.

pkg/loop/internal/pb/capabilities_registry.proto
```diff
@@ -0,0 +1,107 @@
...
+  EXECUTE_API_TYPE_CALLBACK = 2;
+}
+
+// Get has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.CapabilitiesRegistry.Get].
+message GetRequest {
+  string id = 1;
+}
+
+message GetReply {
+  uint32 capabilityID = 1;
```
cedric-cordenier, line 26: not a great name: it's actually the internal connection id.

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,263 @@
...
+	return CapabilityInfo{
+		ID:             id,
+		CapabilityType: capabilityType,
+		Description:    description,
+		Version:        version,
+	}, nil
+}
+
+// MustNewCapabilityInfo returns a new CapabilityInfo,
+// panicking if we could not instantiate a CapabilityInfo.
```
krehermann, line 190: nit: spelling `panick`

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,263 @@
...
+	return c
+}
+
+// TODO: this timeout was largely picked arbitrarily.
+// Consider what a realistic/desirable value should be.
+// See: https://smartcontract-it.atlassian.net/jira/software/c/projects/KS/boards/182
+var defaultExecuteTimeout = 10 * time.Second
+
+// ExecuteSync executes a capability synchronously.
+// We are not handling a case where a capability panics and crashes.
```
krehermann, line 211: what does this comment imply? if a panic occurs, what part of the system is taken offline?

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,263 @@
...
+// that then it should be executed asynchronously.
+func ExecuteSync(ctx context.Context, c CallbackExecutable, request CapabilityRequest) (*values.List, error) {
+	ctxWithT, cancel := context.WithTimeout(ctx, defaultExecuteTimeout)
+	defer cancel()
+
+	responseCh := make(chan CapabilityResponse)
+	setupCh := make(chan error)
+
+	go func(innerCtx context.Context, innerC CallbackExecutable, innerReq CapabilityRequest, innerCallback chan CapabilityResponse, errCh chan error) {
+		setupErr := innerC.Execute(innerCtx, innerCallback, innerReq)
```
krehermann, line 222: these 2 lines can be collapsed to one

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,263 @@
...
+
+			vs = append(vs, response.Value)
+
+		// Timeout when a capability panics, crashes, and does not close the channel.
+		case <-ctxWithT.Done():
+			return nil, fmt.Errorf("context timed out. If you did not set a timeout, be aware that the default ExecuteSync timeout is %f seconds", defaultExecuteTimeout.Seconds())
+		}
+	}
+
+	setupErr := <-setupCh
```
krehermann, line 249: why isn't this in the select loop? as i read it, if there is a setup error, you waste 10s waiting for something that can't happen

pkg/capabilities/triggers/on_demand_trigger.go
```diff
@@ -0,0 +1,75 @@
...
+var info = capabilities.MustNewCapabilityInfo(
+	"on-demand-trigger",
+	capabilities.CapabilityTypeTrigger,
+	"An example on-demand trigger.",
+	"v1.0.0",
+)
+
+type OnDemand struct {
+	capabilities.CapabilityInfo
+	chans map[string]chan capabilities.CapabilityResponse
```
krehermann, line 20: what is the semantic meaning of key? you can make a string alias to add meaning

pkg/loop/internal/capabilities_registry.go
```diff
@@ -0,0 +1,667 @@
...
+
+	switch resp.Type {
+	case pb.ExecuteAPIType_EXECUTE_API_TYPE_TRIGGER:
+		return newTriggerCapabilityClient(c.brokerExt, conn), nil
+	case pb.ExecuteAPIType_EXECUTE_API_TYPE_CALLBACK:
+		return newCallbackCapabilityClient(c.brokerExt, conn), nil
+	case pb.ExecuteAPIType_EXECUTE_API_TYPE_UNKNOWN:
+	}
+
+	return nil, fmt.Errorf("unknown api type")
```
krehermann, line 65: add value of unknown type to the error message

krehermann (commented):
minor comments on my first pass. need to read the grpc code.

pkg/capabilities/capabilities.go
cedric-cordenier: This comment just intends to clarify that we won't handle exceptional error conditions (eg. a capability crashing); i.e. we can assume that won't happen.

pkg/loop/internal/pb/capabilities_registry.proto
```diff
@@ -0,0 +1,107 @@
...
+  CAPABILITY_TYPE_TRIGGER = 1;
+  CAPABILITY_TYPE_ACTION = 2;
+  CAPABILITY_TYPE_CONSENSUS = 3;
+  CAPABILITY_TYPE_TARGET = 4;
+}
+
+message CapabilityInfoReply {
+  string id = 1; 
+  CapabilityType capability_type = 2;
+  string Description = 3;
```
krehermann, line 45: you have an odd admixture of upper case, lower case, snake case and camel case in the field defintions through this file. https://protobuf.dev/programming-guides/style/#message-field-names

pkg/loop/internal/pb/capabilities_registry.proto
```diff
@@ -0,0 +1,107 @@
...
+  EXECUTE_API_TYPE_TRIGGER = 1;
+  EXECUTE_API_TYPE_CALLBACK = 2;
+}
+
+// Get has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.CapabilitiesRegistry.Get].
+message GetRequest {
+  string id = 1;
+}
+
+message GetReply {
```
krehermann, line 25: perhaps add comment that this reply is a shim used to proxy connections (and therefore does not match the return type of Get go func referenced in the comment above)

pkg/capabilities/triggers/on_demand_trigger.go
```diff
@@ -0,0 +1,77 @@
+package triggers
```
bolekk, line 1: Not a huge fan of this file being in chainlink-common. I understand that this is for convenience but longer-term this should end up in a separate repo (or even core seems to be a better fit).

pkg/loop/internal/capabilities_registry.go
```diff
@@ -0,0 +1,667 @@
...
+
+// In the long-term, registration of capabilities should happen entirely
+// inside the core node, via well-defined registration paths. The canonical
+// example of this would be the standard capability registration path defined in
+// this design document:
+// https://docs.google.com/document/d/1i2m1BEytNUR-yBU5mLFVQLGv9FFgYWkVUllFGB062UQ/edit#heading=h.4e7ng0tekbkc
+
+// In the medium term, we need to define an OCR3 capability, and the best way
+// to do this is to pass in a capability registry when creating reporting plugin
+// loop.
```
bolekk, line 29: So are we creating a gRPC interface for Capability Registry only as a stop-gap?
Is the Registry going to exist inside the core node or will it be a LOOPP as well?
Why do we need to pass the Registry to a Capability so it can "register itself"? Can't we loading mechanism add Registry entries when spawning a new LOOPP?

pkg/loop/internal/pb/capabilities_registry.proto
cedric-cordenier: ugh, yes let me correct this

pkg/capabilities/triggers/on_demand_trigger.go
cedric-cordenier: I've posted a slack message to discuss in keystone-dev.

pkg/loop/internal/capabilities_registry.go
cedric-cordenier: I wouldn't describe it as a stop gap: the world we are describing here is the "ultimate" world where all capabilities are standard capabilities, and we don't have any non-standard ones like those based on relayers or OCR.

I think the registry is a first-class node concept so it makes sense to have it in the core node. We could consider loopifying it in the future, but since it's so tied to the core node I'm not clear on the benefits of that.

pkg/loop/internal/pb/capabilities_registry.proto
cedric-cordenier: Done :)

pkg/loop/internal/pb/capabilities_registry.proto
bolekk: Yes, it is very confusing, please change :)

cedric-cordenier:
Closing this for now; going to raise a new PR with just the capability components (since that will be required for standard capabilities and OCR3).

PR #304, opened 2024-01-04
BCF-2876:  gRPC adapter for Mercury LOOPP

krehermann:
BCF-2876
--

[BCF-2876](https://smartcontract-it.atlassian.net/browse/BCF-2876)

This change adapts the [MercuryProvider] (https://github.com/smartcontractkit/chainlink-common/blob/795676d23c7a5e9092f8ae36016f1e653e1dbca9/pkg/types/provider_mercury.go#L12)

to gRPC services and types. In follows the model laid out MedianProvider wrapper [provider] (https://github.com/smartcontractkit/chainlink-common/blob/795676d23c7a5e9092f8ae36016f1e653e1dbca9/pkg/types/provider_median.go#L11) -> [proto] (https://github.com/smartcontractkit/chainlink-common/blob/795676d23c7a5e9092f8ae36016f1e653e1dbca9/pkg/loop/internal/pb/median.proto#L1 )

the mapping between the `MercuryProvider` interface and gRPC adapter is:
- interface funcs -> gRPC service
- interface func args -> gRPC message ending in `Request`
- interface func return valuess -> gRPC message in `Response`
- intermediate  structs -> gRPC message of the same name
- golang pointers -> gRPC values (so that they can be serialized)  

[BCF-2876]: https://smartcontract-it.atlassian.net/browse/BCF-2876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #324, opened 2024-02-06
Node API OCR3 support

george-dorin:


pkg/loop/internal/reporting_ocr3.go
```diff
@@ -0,0 +1,393 @@
...
+		a, err := ocr3AttributedObservation(ao)
+		if err != nil {
+			return nil, err
+		}
+		o = append(o, a)
+	}
+	return
+}
+
+func reportsWithInfo(ri []*pb.OCR3ReportWithInfo) (rwi []ocr3types.ReportWithInfo[any]) {
```
cedric-cordenier, line 386: @george-dorin I thought you'd discussed passing in a protobuf rather than an any type in the reportWithInfo, is that right?

pkg/types/reporting_plugin_service.go
jmank88, line 1: Would it be simpler to have a separate package for ocr3? So that we have `ocr3.ReportingPlugin*` instead of avoiding name collisions with `OCR3ReportingPlugin`?

jmank88:
I think we need to be very careful with the naming here. I don't like the idea of OCR2 being implicit but OCR3 being explicit, and we should avoid the extra baggage of repeated prefixes/suffixes whenever possible. Is it time to start thinking about multiple proto packages?

pkg/loop/internal/pb/reporting_ocr3.proto
```diff
@@ -0,0 +1,149 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb";
```
jmank88, line 3: What if we moved these to a subpackage, and dropped the `OCR3` prefixes?
```suggestion
option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/ocr3";
```

pkg/loop/internal/reporting_ocr3.go
jmank88, line 1: Could this exist under `internal/ocr3/reporting.go` instead? Or does that create package import problems?

cedric-cordenier:
@jmank88 I agree -- I think we could move the OCR2 bits into internal/ocr2 and the OCR3 bits into an internal/OCR3. However, I want to avoid scope creep on this PR, since we need this for the first Keystone demo: would it be OK to move the OCR3 bits into their separate package and follow up on the OCR2 migration separately?

@george-dorin , I'd expect each OCRX package to contain:
* the GRPC server implementation for that flavour of OCR
* the loop constructor
* then the reporting plugin client/servers for that flavour of OCR as well.

pkg/loop/internal/pb/reporting_ocr3.proto
jmank88: Another option:
```suggestion
option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/ocr3/pb";
```
:shrug:

pkg/loop/internal/reporting_ocr3.go
krehermann: i think this implies refactoring the broker_ext to some degree. 

iirc, i tried to add sub packages to encapsulate all the mercury code, but couldn't do it easily because the broker_ext. it may be as simple mv'ing it a pkg and making it public, but i believe it's kind of tangled up.

pkg/loop/internal/pb/reporting_ocr3.proto
krehermann: i choose internal/pb/mercury here. let's keep it consistent https://github.com/smartcontractkit/chainlink-common/pull/304/files

pkg/loop/internal/reporting_ocr3.go
krehermann: to be clear, i would prefer to refactor to subpackages, but it will be a big scope creep in this PR. it should be a stand alone change to refactor the broker

pkg/loop/internal/pb/ocr3/ocr3_reporting.pb.go
cedric-cordenier, line 1: @george-dorin Could you remove the ocr3_X prefix from these proto files?

george-dorin: @cedric-cordenier not really, it leads to a proto file naming conflict

cedric-cordenier (approved):

PR #326, opened 2024-02-07
Add codec for big and little endian along with builders for an encoding and some more general purpose helpers

nolag:


pkg/codec/encodings/array.go
```diff
@@ -0,0 +1,60 @@
...
+	NumElements int
+	Field       TypeCodec
+}
+
+var _ TypeCodec = &array{}
+
+func (a *array) Encode(value any, into []byte) ([]byte, error) {
+	rValue := reflect.ValueOf(value)
+	if rValue.Kind() != reflect.Array && rValue.Kind() != reflect.Slice {
+		return nil, types.ErrNotASlice
```
jmank88, line 31: How about including the kind in the error?
```suggestion
	if kind := rValue.Kind(); kind != reflect.Array && kind != reflect.Slice {
		return nil, fmt.Errorf("%w: expected array or slice but got %s", types.ErrNotASlice, kind)
```

pkg/codec/encodings/binary/gen/main.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		res, err := format.Source(br.Bytes())
+		if err != nil {
+			panic(err)
+		}
+
+		res = []byte(
+			fmt.Sprintf(
+				"// DO NOT MODIFY: automatically generated from pkg/codec/raw/types/main.go using the template %s\n\n%s",
```
jmank88, line 36: Is this file name in sync? I think you could extract it from the runtime caller info to keep it up to date.

pkg/codec/modifier_codec.go
```diff
@@ -106,19 +106,21 @@ func (m *modifierCodec) Decode(ctx context.Context, raw []byte, into any, itemTy
...
-	return convert(reflect.ValueOf(offChain), rInto, m.hook)
+	return Convert(reflect.ValueOf(offChain), rInto, m.hook)
 }
 
 func (m *modifierCodec) GetMaxDecodingSize(ctx context.Context, n int, itemType string) (int, error) {
 	return m.codec.GetMaxDecodingSize(ctx, n, itemType)
 }
 
-func convert(from, to reflect.Value, hook mapstructure.DecodeHookFunc) error {
+func Convert(from, to reflect.Value, hook mapstructure.DecodeHookFunc) error {
```
jmank88, line 116: These could probably use some docs now that they are exported

pkg/types/interfacetests/codec_interface_fuzz_tests.go
```diff
@@ -32,6 +32,14 @@ func RunCodecInterfaceFuzzTests(f *testing.F, tester CodecInterfaceTester) {
 				fb[index] = fixedBytes[index]
 			}
 
+			// put a limit on the size of strings because some encodings will need to know ahead of time
+			if len(s) > 100 {
+				s = s[:100]
+			}
+			if len(differentField) > 100 {
+				differentField = differentField[:100]
+			}
```
jmank88, line 41: Would it make sense to just `t.Skip()` these cases instead? My thought was that rejecting the input instead of just mutating it would inform the fuzz runtime not to explore that kind of input.

nolag: I think I can just revert this.  I was thinking about chains, but this is just for the codec.  The codec test doesn't need to encode the exact same way as the chain test :).

jmank88 (dismissed):

jmank88 (approved):

PR #327, opened 2024-02-07
[KS-21] Implement mocked OCR3 Consensus Capability

DeividasK:


bolekk (changes requested):

pkg/capabilities/consensus/ocr3.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+var info = capabilities.MustNewCapabilityInfo(
+	"ocr3",
+	capabilities.CapabilityTypeConsensus,
+	"OCR3 consensus exposed as a capability.",
+	"v1.0.0",
+)
+
+type cache struct {
```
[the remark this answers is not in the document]
cedric-cordenier: @bolekk This is why there's a list of `workflowExecutionIDs` on line 35

pkg/capabilities/consensus/ocr3.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+
+	capabilityInputs := &CapabilityInputs{}
+	err = mapstructure.Decode(valuesMap, capabilityInputs)
+	if err != nil {
+		return err
+	}
+
+	// --- START OF MOCKED OCR3 LOGIC ---
```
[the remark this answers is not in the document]
cedric-cordenier: Our plan was to implement both sides of the interface in the same struct -- there's no point keeping them separate since they aren't reusable separately.

I don't have a strong feeling about this though, so happy to go with separating them.

pkg/capabilities/consensus/ocr3.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (o *OCR3) RegisterToWorkflow(ctx context.Context, request capabilities.RegisterToWorkflowRequest) error {
+	return nil
+}
+
+func (o *OCR3) UnregisterFromWorkflow(ctx context.Context, request capabilities.UnregisterFromWorkflowRequest) error {
+	return nil
+}
+
+type CapabilityInputs struct {
```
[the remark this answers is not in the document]
cedric-cordenier: This isn't intended to be a Test struct: this is supposed to be the strongly-typed version of whatever we send to the capability.

pkg/capabilities/consensus/ocr3.go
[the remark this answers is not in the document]
bolekk: Don't we need to have something to pass to ReportingPluginFactory so that the plugin can use it? I imagine that libocr's Oracle is a separate service from the one that handles Capability Interface calls...? Or were you just planning to use a struct for that (if so, having an interface makes it easier to mock stuff)...?

pkg/capabilities/consensus/ocr3_capability.go
```diff
@@ -0,0 +1,136 @@
...
+	evictedCh chan []*request
+}
+
+func newOCR3Capability(ctx context.Context, s *store) *ocr3Capability {
+	o := &ocr3Capability{
+		CapabilityInfo: info,
+		store:          s,
+		evictedCh:      make(chan []*request),
+	}
+	go o.evictionLoop(ctx)
```
bolekk, line 31: nit: Starting goroutines in constructors always worries me. Don't we prefer to follow the ServiceCtx interface and use Start/Stop methods?

pkg/capabilities/consensus/ocr3_store.go
```diff
@@ -0,0 +1,168 @@
...
+	if !ok {
+		return fmt.Errorf("request with id %s not found", requestID)
+	}
+
+	delete(s.requests, requestID)
+	return nil
+}
+
+func (s *store) evictExpiredBefore(ctx context.Context, now time.Time) ([]*request, error) {
+	idx, err := binarySearch(s.requests, s.requestIDs, 0, len(s.requestIDs)-1, now)
```
bolekk, line 100: There's no point in running binary search if you later iterate over the first segment of requestIDs array anyway. Just add a timestamp check in the loop below instead of index comparison.

pkg/capabilities/consensus/ocr3_store.go
```diff
@@ -0,0 +1,168 @@
...
+		delete(s.requests, r)
+	}
+
+	s.requestIDs = s.requestIDs[idx : len(s.requestIDs)-1]
+	return evicted, nil
+}
+
+// binarySearch searches sorted (in ascending order) array `array`
+// for the first index immediately preceding the searched for value.
+func binarySearch(mr map[string]*request, array []string, startIdx int, endIdx int, val time.Time) (int, error) {
```
bolekk, line 130: In a comment above I argue that you don't need this at all. Even if you needed it, it's available in standard library: https://pkg.go.dev/sort#Search

Pretty much whenever you find yourself re-implementing a textbook algorithm/DS, it should ring an alarm bell.

bolekk (changes requested):

pkg/capabilities/consensus/ocr3_store.go
cedric-cordenier: Wow... 🤦 I think I lost some perspective when writing this PR! Thanks for catching!

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,172 @@
...
+	return nil
+}
+
+func (o *capability) loop(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case r := <-o.newExpiryWorkerCh:
+			go o.expiryWorker(ctx, r)
```
bolekk, line 90: consider time.AfterFunc()

pkg/capabilities/consensus/ocr3_store.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	s.requestIDs = append(newRequestIDs, s.requestIDs[lastIdx+1:]...)
+	if len(got) == 0 {
+		return nil, errors.New("queue is empty")
+	}
+	return got, nil
+}
+
+func (s *store) evict(ctx context.Context, requestID string) error {
```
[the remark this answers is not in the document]
cedric-cordenier: Correct: this is done by the call to `response` in the ocr3 capability struct.

pkg/capabilities/consensus/ocr3_models.go
```diff
@@ -0,0 +1,28 @@
...
+	"time"
+
+	"github.com/shopspring/decimal"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+type request struct {
+	Observations      map[string]decimal.Decimal
```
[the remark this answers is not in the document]
cedric-cordenier: Yep, that's fine 👍

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: Hmmm... I tried using AfterFunc here but I don't think it works since we need to also listen to ctx.Done() in our case (to handle the capability being stopped).

bolekk: For that you can call Stop() on each object returned by AfteFunc
```
timer := time.AfterFunc
timer.Stop()
```

cedric-cordenier: Unless I'm missing something, I don't think that works: the docs for time.AfterFunc say that the call will block until the function has been called.

But in our case, we want to either exit without calling the function if its context is cancelled, or call the function.

I'm not able to account for that first case -- I think the only way to do that is a select on both the ctx.Done() and the timer's channel.

cedric-cordenier: (Also the channel for the returned timer is empty, so we can't do t := time.AfterFunc; and then <-t.C in a select)

bolekk: Hmm interesting. I rely on the Stop() call in another place so I need to check if this actually works as intended...
Feel free to go with you initial approach here.

pkg/capabilities/consensus/ocr3/types/generate.go
```diff
@@ -0,0 +1,2 @@
+//go:generate protoc --go_out=../../../../ --go_opt=paths=source_relative --go-grpc_out=../../../../ --go-grpc_opt=paths=source_relative --proto_path=../../../../ capabilities/consensus/ocr3/types/ocr3_types.proto values/pb/values.proto
```
jmank88, line 1: Could this be expressed more simply if it lived somewhere else?

pkg/capabilities/consensus/ocr3/ocr3.go
```diff
@@ -0,0 +1,22 @@
...
+	lggr logger.Logger
+}
+
+func NewOCR3(lggr logger.Logger) *OCR3 {
+	return &OCR3{lggr: lggr}
+}
+
+func (o *OCR3) NewReportingPluginFactory(ctx context.Context, config types.ReportingPluginServiceConfig, provider types.PluginProvider, pipelineRunner types.PipelineRunnerService, telemetry types.TelemetryClient, errorLog types.ErrorLog) (types.OCR3ReportingPluginFactory, error) {
+	factory, err := newFactoryService(nil)
+	// TODO capabilityRegistry.Add(factory.capability)
```
jmank88, line 20: This PR or later?

cedric-cordenier: Later -- this factory function doesn't currently accept a capabilityRegistry.

pkg/capabilities/consensus/ocr3/types/generate.go
cedric-cordenier: Yes, but I think that would require all protos living in the same package -- not necessarily a bad thing, but not the model we're following atm.

jmank88: I just mean the command. For example, would moving this generate line up the tree allow us to shed some of the `../`?

cedric-cordenier: Ah yes, it would. If we moved it to the top-level directory that encompassed all of the dependencies, then the .. would go away.

Do you think it's OK to have the commands detached from their physical location? That feels weird to me.

An alternative would be to have a make proto command that we can run rather than using `go generate` everywhere.

jmank88: > Do you think it's OK to have the commands detached from their physical location? That feels weird to me.

I agree, but I guess I'm not really understanding the issue. Why does only one of the existing generate commands have this complexity?
https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/loop/internal/pb/generate.go#L1-L10

pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -0,0 +1,73 @@
...
+
+	clock clockwork.Clock
+}
+
+type factoryService struct {
+	store *store
+	*capability
+	batchSize int
+
+	services.StateMachine
```
bolekk, line 25: Does it need to be a service? Do we expect it to run any internal goroutines?

pkg/capabilities/consensus/ocr3/models.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+type request struct {
+	Observations      values.Value `mapstructure:"-"`
+	AggregationMethod string
```
bolekk, line 13: I'm assuming that you're leaving this for now? If so maybe just add a comment that this will be extended or moved.

pkg/capabilities/consensus/ocr3/ocr3.go
```diff
@@ -0,0 +1,22 @@
+package consensus
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type OCR3 struct {
```
bolekk, line 10: IMO this could be called OCR3Capability or OCR3ConsensusCapability. It encapsulates all elements needed for the capability to work (interface + plugin + store).

bolekk (approved):
LGTM but please address @jmank88 's feedback

pkg/capabilities/consensus/ocr3/types/generate.go
cedric-cordenier: Ah it's because this generate command depends on the values protos which are defined in another directory. Without adding those protos to the list of files to generate, protoc can't resolve the dependency.

The solution to that is apparently to make sure that protoc is run from a common parent dir, hence the ...

pkg/capabilities/consensus/ocr3/factory.go
cedric-cordenier: Probably not, but it's required to be a service by the factory interface (and it should start the capability service as well)

pkg/capabilities/consensus/ocr3/models.go
cedric-cordenier: I've actually removed it :) I thought I'd forgotten to address some feedback.

pkg/capabilities/consensus/ocr3/ocr3.go
cedric-cordenier: Yep, renamed to OCR3Capability 👍

pkg/capabilities/consensus/ocr3/types/generate.go
cedric-cordenier: BTW we have encountered a similar problem with mercury, see: 

```
// NOTE: the relative paths in the proto_path are to ensure we find common utilities, like BigInt
//go:generate protoc --proto_path=.:../.. --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative datasource_v1.proto
```

cedric-cordenier: One way I can see around this is to add:

```
protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative --proto_path=. $(find . -type f -name "*.proto")
```

to `make generate`.

With some changes to our import paths (and possibly some further polish), this should automatically find all our protos with no further need to add each file one by file.

jmank88: If you think that dynamic approach is robust then that sounds great.


> The solution to that is apparently to make sure that protoc is run from a common parent dir, hence the dots.

Otherwise, doesn't this suggest that our assumption about having these commands live near the _output_ file is misguided? It seems like everything is smoother if the command is located based on the _input_ files, no?

cedric-cordenier: Cool, I'll follow up with this. In the meantime do you mind approving this PR so it's unblocked?

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,174 @@
...
+		store:             s,
+		newExpiryWorkerCh: make(chan *request),
+		clock:             clock,
+	}
+	return o
+}
+
+func (o *capability) Start(ctx context.Context) error {
+	return o.StartOnce("OCR3Capability", func() error {
+		innerCtx, c := context.WithCancel(ctx)
```
jmank88, line 44: The `Start` ctx is not meant to be held long term. Typically we would store a `StopChan` on `capabilty` and then derive contexts from that as needed.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,174 @@
...
+		o.cancel = c
+		go o.loop(innerCtx)
+		return nil
+	})
+}
+
+func (o *capability) Stop(ctx context.Context) error {
+	return o.StopOnce("OCR3Capability", func() error {
+		o.cancel()
```
jmank88, line 53: It is important to wait for any spawned goroutines before returning. I think we'll need a `WaitGroup` to track the `loop()` and `expiryWorker()` goroutines.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,183 @@
...
+		innerCtx, _ := utils.ContextFromChan(o.stopCh)
+		o.wg.Add(1)
+		go o.loop(innerCtx)
+		return nil
+	})
+}
+
+func (o *capability) Stop(ctx context.Context) error {
+	return o.StopOnce("OCR3Capability", func() error {
+		o.stopCh <- struct{}{}
```
jmank88, line 57: A `close()` will 'broadcast' to all listeners instead of just one:
```suggestion
		close(o.stopCh)
```

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,183 @@
...
+	capabilities.CapabilityTypeConsensus,
+	"OCR3 consensus exposed as a capability.",
+	"v1.0.0",
+)
+
+type capability struct {
+	services.StateMachine
+	capabilities.CapabilityInfo
+	store  *store
+	stopCh chan struct{}
```
jmank88, line 27: There is a `StopChan` type with built in methods:
```suggestion
	stopCh services.StopChan
```

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,183 @@
...
+	err = o.store.add(ctx, r)
+	if err != nil {
+		return err
+	}
+
+	o.newExpiryWorkerCh <- r
+	return nil
+}
+
+func (o *capability) loop(ctx context.Context) {
```
jmank88, line 89: I think it would be cleaner to create the context from here:
```suggestion
func (o *capability) loop() {
	ctx, cancel := o.stopChan.NewCtx()
	defer cancel()
```

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -0,0 +1,183 @@
...
+func (o *capability) Start(ctx context.Context) error {
+	return o.StartOnce("OCR3Capability", func() error {
+		innerCtx, _ := utils.ContextFromChan(o.stopCh)
+		o.wg.Add(1)
+		go o.loop(innerCtx)
+		return nil
+	})
+}
+
+func (o *capability) Stop(ctx context.Context) error {
```
jmank88, line 55: Why `Stop` instead of `Close`, to match our existing interfaces?
```suggestion
func (o *capability) Close() error {
```

bolekk (approved):

pkg/capabilities/consensus/ocr3/capability.go
DeividasK: @jmank88, is it Start/Close everywhere in the Core node? The asymmetry peeves me 😅

PR #330, opened 2024-02-08
ccip - define service provider interfaces and relevant types

dimkouv:
Defining CCIP related types and interfaces for eventually migrating to the LOOPP architecture.

Relevant CCIP PR: https://github.com/smartcontractkit/ccip/pull/489


pkg/types/cciptypes/models.go
```diff
@@ -0,0 +1,38 @@
+package cciptypes
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
```
makramkd, line 7: Not an expert on this but should we be importing go-ethereum here? Seems to defeat the purpose of chain-agnositicism.

pkg/types/cciptypes/models.go
```diff
@@ -0,0 +1,38 @@
...
+func (h Hash) String() string {
+	return hexutil.Encode(h[:])
+}
+
+type BlockMeta struct {
+	BlockTimestamp time.Time
+	BlockNumber    int64
+	TxHash         string
+	LogIndex       uint
+}
```
makramkd, line 38: These are EVM-specific types right?

pkg/types/cciptypes/models.go
krehermann: yes @makramkd is correct.

i thought we had defined agnostic types together @dimkouv ? did we just overlook this or is it more complicated?

pkg/types/cciptypes/models.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+type Address string
+
+func (a Address) Equals(addr2 Address) bool {
+	if common.IsHexAddress(string(a)) && common.IsHexAddress(string(addr2)) {
```
krehermann, line 14: ah, i see, given that Address is now a string, we can change this to return a == addr2

pkg/types/cciptypes/models.go
```diff
@@ -0,0 +1,38 @@
...
+		return common.HexToAddress(string(a)) == common.HexToAddress(string(addr2))
+	}
+	return a == addr2
+}
+
+func (a Address) ToEVM() (common.Address, error) {
```
krehermann, line 20: this needs to be a func in the evm relayer, not a method on the type in this repo

pkg/types/cciptypes/models.go
```diff
@@ -0,0 +1,38 @@
...
+	if !common.IsHexAddress(string(a)) {
+		return common.Address{}, fmt.Errorf("%s is not a hex evm address", a)
+	}
+	return common.HexToAddress(string(a)), nil
+}
+
+type Hash [32]byte
```
krehermann, line 27: @jmank88 do you have an advise here for generalizing the ccip types? specific do you know we already a common definition of a hash that is known to apply to various chains? my concern is that constraining to 32 bytes might be too limiting, but i'm not sure

jmank88: Yeah if the length assumption can be lifted, that would be helpful. Another angle would be to pass around a human readable string encoding, like with `types.Account`.

pkg/types/cciptypes/models.go
jmank88: Yeah maybe the ints should be `big.Int`? And time should be some integer with a specific precision?

makramkd: Also the `BlockMeta` is technically EVM-specific, has tx hash, log index, etc. Actually @dimkouv what is `BlockMeta` supposed to refer to? I don't think `LogIndex` nor `TxHash` are scoped to a block - should this be `TxMeta`?

pkg/types/cciptypes/models.go
jmank88: Code in this repo cannot make any assumptions about the format though. It has to work with any chain implementation, and it is their responsibility to format the strings canonically.

krehermann: you could add this functionality as a helper func in the evm relayer if you need it.

pkg/types/cciptypes/commitstore.go
```diff
@@ -0,0 +1,81 @@
+package cciptypes
```
krehermann, line 1: let's make this `ccip` rather than `cciptypes` to match the packages here

pkg/types/cciptypes/commitstore.go
```diff
@@ -0,0 +1,81 @@
...
+type CommitOffchainConfig struct {
+	GasPriceDeviationPPB   uint32
+	GasPriceHeartBeat      time.Duration
+	TokenPriceDeviationPPB uint32
+	TokenPriceHeartBeat    time.Duration
+	InflightCacheExpiry    time.Duration
+}
+
+type CommitStoreStaticConfig struct {
+	ChainSelector       uint64
```
krehermann, line 72: is this a chain id? other, non-evm, chains use string identifiers. it's probably best to change this now and adapt from string to int as needed in the evm relayer implementation. @jmank88 do you think?

if we do go this direction, then there are other chain selector fields in other types that should be changed, too.

for example, here is the configuration struct we use to represent a generic chain  https://github.com/smartcontractkit/chainlink-common/blob/5d1fecc0a6992a313c083ed17f53f574588b3af7/pkg/types/relayer.go#L36

pkg/types/cciptypes/gaspriceestimator.go
```diff
@@ -0,0 +1,49 @@
...
+// CommonGasPriceEstimator is abstraction over multi-component gas prices.
+type CommonGasPriceEstimator interface {
+	// GetGasPrice fetches the current gas price.
+	GetGasPrice(ctx context.Context) (*big.Int, error)
+	// DenoteInUSD converts the gas price to be in units of USD. Input prices should not be nil.
+	DenoteInUSD(p *big.Int, wrappedNativePrice *big.Int) (*big.Int, error)
+	// Median finds the median gas price in slice. If gas price has multiple components, median of each individual component should be taken. Input prices should not contain nil.
+	Median(gasPrices []*big.Int) (*big.Int, error)
+	// String converts the gas price to string.
+	String(p *big.Int) string
```
krehermann, line 38: can this String method be removed? it's not really doing anything since big.Int as a String func IIRC.

the reason to remove it is that when we loopify, this would require grpc to get the answer. it would be much better to just call the conversion locally

pkg/types/cciptypes/commitstore.go
jmank88: Nah it is intentionally not a semantic chain id. I think this is fine as-is.
https://github.com/smartcontractkit/chain-selectors

pkg/types/cciptypes/commitstore.go
krehermann: i'd prefer the existing pattern for a few reasons: it removes the from stutter `...types/cciptypes`, the call sites can alias to `cciptypes` upon import, it's maintains consistency. finally, the golang idiom here is `types/ccip` for the first two reasons.

krehermann (approved):

pkg/types/ccip/models.go
```diff
@@ -0,0 +1,25 @@
+package ccip
+
+import (
+	"encoding/hex"
+	"strings"
+)
+
+type Address string
```
jmank88, line 8: Can the libocr `Account` type be used instead? So that we don't create two different standards?

pkg/types/ccip/models.go
```diff
@@ -0,0 +1,25 @@
...
+	if !strings.HasPrefix(res, "0x") {
+		return "0x" + res
+	}
+	return res
+}
+
+type TxMeta struct {
+	BlockTimestampUnixMilli int64
+	BlockNumber             uint64
+	TxHash                  string
```
jmank88, line 23: Why do we use a `string` hash here, but a `[32]byte` hash in other places?

pkg/types/ccip/models.go
```diff
@@ -0,0 +1,25 @@
...
+func (h Hash) String() string {
+	res := hex.EncodeToString(h[:])
+	if !strings.HasPrefix(res, "0x") {
+		return "0x" + res
+	}
+	return res
+}
+
+type TxMeta struct {
+	BlockTimestampUnixMilli int64
```
krehermann, line 21: this can/should be uint64, right?

pkg/types/ccip/gaspriceestimator.go
```diff
@@ -0,0 +1,47 @@
...
+type GasPriceEstimator interface {
+	GasPriceEstimatorCommit
+	GasPriceEstimatorExec
+}
+
+type GasPriceEstimatorCommit interface {
+	CommonGasPriceEstimator
+
+	// Deviates checks if p1 gas price diffs from p2 by deviation options. Input prices should not be nil.
+	Deviates(p1 *big.Int, p2 *big.Int) (bool, error)
```
makramkd, line 18: Does this need to be an interface method? Feels like an odd calculation to do over an RPC call when it can be done prob 1000x faster locally.

pkg/types/ccip/gaspriceestimator.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// CommonGasPriceEstimator is abstraction over multi-component gas prices.
+type CommonGasPriceEstimator interface {
+	// GetGasPrice fetches the current gas price.
+	GetGasPrice(ctx context.Context) (*big.Int, error)
+	// DenoteInUSD converts the gas price to be in units of USD. Input prices should not be nil.
+	DenoteInUSD(p *big.Int, wrappedNativePrice *big.Int) (*big.Int, error)
+	// Median finds the median gas price in slice. If gas price has multiple components, median of each individual component should be taken. Input prices should not contain nil.
+	Median(gasPrices []*big.Int) (*big.Int, error)
```
makramkd, line 36: Ditto here, this is just a calculation, does this need to be an RPC call?

makramkd (commented):
General comment: can we use this as an opportunity to add informative doc comments on each interface and method? We're going to have a lot of people reading this as a reference as we onboard more folks and it would be nice to have some good docs here since they're not really available anywhere else.

pkg/types/ccip/models.go
krehermann: hmm, i guess there were considering times before 1970.

pkg/types/ccip/gaspriceestimator.go
krehermann: i had similar thoughts.

i started looking at the gas price estimator code in ccip to remove String and found some opportunities to refactor and make it cleaner independent of these methods.

personally, i would be ok with a TODO here referencing a ticket to clean this up. it's something i can pickup as part of implementing the grpc layer

makramkd: Fair enough!

pkg/types/ccip/models.go
jmank88: This is all part of one unified relayer API, so there is no sense having two different types for the same thing. An alias can work to bridge the existing code, but we should keep that as an implementation detail in core. There is no need for this common repo to inherit and spread that complexity from the service interface layer.

pkg/types/ccip/gaspriceestimator.go
jmank88: nit/ there is no need to waste time actually LOOPPifying a temporary interface. We can iterate on the interface, but we should do this before adding GRPC wrappers. Do you mean to circle back right away?

krehermann (approved):

pkg/types/ccip/gaspriceestimator.go
krehermann: yes, i'll do it as part of the grpc work

krehermann:
merging to unblock grpc work

PR #331, opened 2024-02-09
KS-34: Basic aggregator

bolekk:


pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
```diff
@@ -0,0 +1,137 @@
...
+	config       AggregatorConfig
+	mercuryCodec mercury.Codec
+	lggr         logger.Logger
+}
+
+var _ types.Aggregator = (*dataFeedsAggregator)(nil)
+
+// EncodableOutcome is a list of AggregatedPricePoints
+// Metadata is a map of feedID -> (timestamp, price) representing onchain state (see DataFeedsOutcomeMetadata proto)
+func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value) (*types.AggregationOutcome, error) {
```
DeividasK, line 36:
```suggestion
func (a *dataFeedsAggregator) Aggregate(currentOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value) (*types.AggregationOutcome, error) {
```

Given that we plan to merge previousOutcome with newOutcome, would it make sense to call the variable `currentOutcome`?

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		for feedID, report := range mercuryReportSet.Reports {
+			latest, ok := latestReportPerFeed[feedID]
+			if !ok || report.Info.Timestamp > latest.Info.Timestamp {
+				latestReportPerFeed[feedID] = report
+			}
+		}
+	}
+
+	onchainState := &DataFeedsOutcomeMetadata{}
```
DeividasK, line 61:
```suggestion
	currentState := &DataFeedsOutcomeMetadata{}
```

Similarly to above, given that this might not always be an on-chain state, would it be better to name this `currentState`?

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
```diff
@@ -0,0 +1,137 @@
...
+	"google.golang.org/protobuf/proto"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/types"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/mercury"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+type AggregatorConfig struct {
+	Tickers map[string]TickerConfig
```
DeividasK, line 18: Could we use feeds here, too, to match the mercury report output? Namely, we should consistently refer to Feed IDs, which will be uint64 (https://github.com/smartcontractkit/chainlink/pull/11901)

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
```diff
@@ -0,0 +1,137 @@
...
+	for nodeID, nodeObservations := range observations {
+		// we only expect a single observation per node - new Mercury data
+		if len(nodeObservations) == 0 {
+			a.lggr.Warnf("node %d contributed with empty observations", nodeID)
+			continue
+		}
+		if len(nodeObservations) > 1 {
+			a.lggr.Warnf("node %d contributed with more than one observation", nodeID)
+		}
+		mercuryReportSet, err := a.mercuryCodec.Unwrap(nodeObservations[0])
```
DeividasK, line 48: I think this should be done in the trigger, as otherwise, the codec would have to be used everywhere where the mercury report needs to be manipulated.

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
bolekk: Not really. Aggregate takes previous outcome and returns current outcome.

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
bolekk: Yes, I agree.

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
bolekk: Sure, makes sense. However I suspect we still need to pass the "raw" version to later put it onchain (let's get confirmation on that).

bolekk: On second thoughts, we actually need to do it here because observations can come from malicious nodes. We decided not to wait for quorum but trust the latest report, so we need to validate it.

pkg/capabilities/mercury/mercury.go
```diff
@@ -0,0 +1,31 @@
...
+	FullReport []byte     // full report, acceptable by the verifier contract
+}
+
+type ReportInfo struct {
+	Timestamp uint32
+	Price     float64
+}
+
+//go:generate mockery --quiet --name Codec --output ./mocks/ --case=underscore
+type Codec interface {
```
cedric-cordenier, line 23: IMO this is in the wrong place: it's considered idiomatic to declare interfaces where they are used, not in the "source" package. This is something we get wrong a lot at CLL and causes problems as we then introduce dependencies on the package hosting the concrete struct (admittedly that's not the case here).

pkg/capabilities/consensus/types/aggregator.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+import (
+	ocrcommon "github.com/smartcontractkit/libocr/commontypes"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+type Aggregator interface {
```
cedric-cordenier, line 9: Why do we need a separate types package for this? The aggregator should be an internal, private implementation detail and we shouldn't have to expose it elsewhere (either as a public interface, or in a types package).

cedric-cordenier: Ah, I guess your intent is to implement each aggregator in a separate package? Hmmm... wonder if that's overkill? IMO far simpler to just put everything in one package which gives us the flexibility to play around with package boundaries, and move things out a later date if that makes sense.

pkg/capabilities/consensus/datafeeds/feeds_aggregator.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/mercury"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+type AggregatorConfig struct {
+	Feeds map[mercury.FeedID]FeedConfig
+}
+
+type FeedConfig struct {
```
cedric-cordenier, line 22: @bolekk I don't think FeedConfig/AggregatorConfig needs to be public does it? They're only used here.

pkg/capabilities/consensus/types/aggregator.go
bolekk: Yeah, it's either everything in a single package or everything separated, otherwise we will run into circular dependencies (e.g. when we add a factory method). My slight preference is to keep stuff separated. Any particular worries about extra packages?

pkg/capabilities/mercury/mercury.go
bolekk: Where would you put it then? This is a Mercury Codec that can be used by multiple places in the code.

bolekk: Moving to datafeeds.

pkg/capabilities/consensus/types/aggregator.go
bolekk: Keeping separate for now.

bolekk:
@cedric-cordenier  I addressed latest comments and also changed FeedID type based on latest feedback from Austin. I'm using a hex-encoded string everywhere because (a) config will contain a human-readable hex and (b) protos don't allow bytes to be a map key. So this is simpler to avoid conversions back-and-forth everywhere. I added a validator.

bolekk:
Sonar is broken again but hopefully @chudilka1 fixes it separately.

cedric-cordenier (approved):

PR #334, opened 2024-02-09
[KS-15] GRPC support for the capability API

cedric-cordenier:
* Adds GRPC wrappers for the capability API only -- as this is required for OCR3 running in a LOOP and for standard capabilities going forward.
* Does not implement the capability registry.


pkg/loop/internal/capabilities.go
```diff
@@ -0,0 +1,587 @@
...
+
+func (c *baseCapabilityClient) Info(ctx context.Context) (capabilities.CapabilityInfo, error) {
+	resp, err := c.grpc.Info(ctx, &emptypb.Empty{})
+	if err != nil {
+		return capabilities.CapabilityInfo{}, err
+	}
+
+	var ct capabilities.CapabilityType
+	switch resp.CapabilityType {
+	case pb.CapabilityType_CAPABILITY_TYPE_TRIGGER:
```
jmank88, line 126: Is this the idiomatic way to name these? I guess we shouldn't favor Go over other languages in our protos?

cedric-cordenier: Yes, unfortunately! This is because proto enum member fields are globally named, so they recommend prepending the enum type in front of it. I could just add it as "TRIGGER", but that would get awkward with the first element, which by convention is called "UNKNOWN" and would cause issues for other enums in the package.

cedric-cordenier: I agree it's really awkward. I could reduce the stuttering less by prepending each member name with an abbreviation, eg. CT_TRIGGER/CT_ACTION/CT_TARGET etc?

jmank88: Could we put some aliases/consts in front of them, with cleaner names?
```go
const CapabilityTypeTrigger = pb.CapabilityType_CAPABILITY_TYPE_TRIGGER
```
Or maybe even from a file in the same package that is generated, so we can do `pb.CapabilityTypeTrigger` here?
```go
const CapabilityTypeTrigger = CapabilityType_CAPABILITY_TYPE_TRIGGER
```

pkg/loop/internal/capabilities.go
```diff
@@ -0,0 +1,587 @@
...
+			WorkflowExecutionId: req.Metadata.WorkflowExecutionID,
+		},
+		Inputs: inputsPb,
+		Config: configPb,
+	}, nil
+}
+
+func (c *callbackExecutableClient) Execute(ctx context.Context, callback chan<- capabilities.CapabilityResponse, req capabilities.CapabilityRequest) error {
+	cid, res, err := c.serveNew("Callback", func(s *grpc.Server) {
+		pb.RegisterCallbackServer(s, newCallbackServer(callback))
```
bolekk, line 436: Same question as on the previous PR - why do we need to create a new server for each call rather than re-using a single one per capability?

cedric-cordenier: OK, I did a bit of digging on this: under the hood what this does is just register a handler to a server, so it should be fairly inexpensive. However, we can't be guaranteed to receive the same underlying grpc.Server instance, so my original solution (to wrap the register call in a sync.Once) won't work.

With the understanding that this is fairly inexpensive and the expected way, are you happy to approve as is and we can revisit it if the overhead of these servers becomes a problem?

jmank88 (approved):

PR #341, opened 2024-02-14
[chore] Add clock to utils

cedric-cordenier:
Moving this from core into common.

Companion PR: https://github.com/smartcontractkit/chainlink/pull/12027

### DIFF
--- a/pkg/utils/clock.go
+++ b/pkg/utils/clock.go
@@ -1,0 +1,29 @@
+package utils
+
+import "time"
+
+type Clock interface {
+	Now() time.Time
+}
+
+type realClock struct{}
+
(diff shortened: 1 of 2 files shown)


DeividasK (approved):
Could do with a PR comment to explain why this is moved (I assume one of the capabilities will need it?)

jmank88:
Is there another package that would make more sense? We've been trying to dissolve `package utils` since it is an anti-pattern

cedric-cordenier:
@jmank88 I could move it to a faketime package in ./pkg? Or into utils/tests?

bolekk:
pkg/clock ?

PR #344, opened 2024-02-14
[KS-32] OCR3 Reporting Plugin

cedric-cordenier:


pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -0,0 +1,232 @@
...
+
+	ids := []*pbtypes.RequestId{}
+	for _, r := range batch {
+		ids = append(ids, &pbtypes.RequestId{
+			RequestId:   r.RequestID,
+			NamespaceId: r.NamespaceID,
+		})
+	}
+
+	return proto.Marshal(&pbtypes.Query{
```
bolekk, line 53: In each stage we need to be careful not to exceed max size in bytes from OCR config. For Query maybe it could be guaranteed by config validation?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	reqs, err := r.s.getN(ctx, rids)
+	if err != nil {
+		return nil, err
+	}
+
+	obs := &pbtypes.Observation{}
+	for _, r := range reqs {
+		obsPb, err := r.Observations.Proto()
```
bolekk, line 77: Side note: I feel that there's no reason for Proto() to ever return an error. If we have a valid Value object, it seems natural that it's always translatable into a valid proto. It looks like right now only Decimal can return an error because we depend on JSON marshaling. Can we fix it somehow and change this method to not return an error at all?

pkg/capabilities/consensus/ocr3/types/ocr3_types.proto
```diff
@@ -15,5 +15,24 @@ message AggregationOutcome {
...
-  repeated string requestIds = 1;
+  repeated RequestId requestIds = 1;
+}
+
+message RequestId {
+  string requestId = 1;
+  string namespaceId = 2;
+}
+
+message Request {
```
bolekk, line 26: This name "Request" is a little confusing. How about "SingleObservation"? Or "ObservationWithInfo"?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -0,0 +1,232 @@
...
+			if _, ok := m[rid]; !ok {
+				m[rid] = make(map[ocrcommon.OracleID][]values.Value)
+			}
+
+			val, err := values.FromProto(r.Observation)
+			if err != nil {
+				continue
+			}
+
+			m[rid][o.Observer] = append(m[rid][o.Observer], val)
```
bolekk, line 127: Could we rename "rid" to workflowID? Those request IDs are pretty confusing when reading.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -0,0 +1,232 @@
...
+		if !ok {
+			continue
+		}
+
+		namespaceOutcome, ok := o.Outcomes[rid.NamespaceId]
+		if !ok {
+			continue
+		}
+
+		outcome, err := r.aggregator.Aggregate(namespaceOutcome, requestObs)
```
bolekk, line 154: I wonder if we should check for minimum number of observation before calling or leave it to the aggregator... WDYT?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -0,0 +1,232 @@
...
+		o, ok := o.Outcomes[rid.NamespaceId]
+		if !ok {
+			continue
+		}
+
+		if !o.ShouldReport {
+			continue
+		}
+
+		mv, err := values.FromMapValueProto(o.EncodableOutcome.GetMapValue())
```
bolekk, line 185: I realized that this is incorrect. The fact that an ID was in query and has ShouldReport set doesn't mean it was even processed! It's possible that we didn't have enough observations or aggregate errored out. I think it's safer to just attach a list of IDs where the call to Aggregate returned ShouldReport=true in the latest round, instead of the query.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
cedric-cordenier: Yes, there's probably something we can do to fix this -- I've created a ticket [here](https://smartcontract-it.atlassian.net/jira/software/c/projects/KS/boards/182?selectedIssue=KS-50), and will follow up when I get back from vacation.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
cedric-cordenier: Yes, will do! I think I was over-optimising a bit here

pkg/capabilities/consensus/ocr3/reporting_plugin.go
cedric-cordenier: Hmmm... that feels like a responsibility of the reporting plugin, so I think it should go here.

pkg/capabilities/consensus/ocr3/types/encoder.go
```diff
@@ -0,0 +1,11 @@
+package types
```
bolekk, line 1: Why did this file disappear from the repo? :0

cedric-cordenier: That was me... I must have not moved it when I was rebasing my PR and gotten it lost that way.

george-dorin (approved):

PR #345, opened 2024-02-15
Make artifact available on failing fuzz tests

vyzaldysanchez:


.github/workflows/pkg.yml
```diff
@@ -25,6 +25,17 @@ jobs:
...
+      - name: Capture Fuzz Tests Failing Inputs
+        if: failure()
+        run: find . -type f|fgrep '/testdata/fuzz/'|while read f; do echo $f; cat $f; done > failing_fuzz_inputs.txt 2>&1
+
+      - name: Upload Fuzz Tests Failing Inputs
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: failing-fuzz-inputs
+          path: ./failing_fuzz_inputs.txt
```
jmank88, line 38: Does this accept a wild card pattern? If so, we could do both things at once and retain the file structure.

vyzaldysanchez: That I don't know

jmank88 (dismissed):

jmank88 (approved):

PR #350, opened 2024-02-16
Add standalone plugin provider

cedric-cordenier:


george-dorin (dismissed):

pkg/loop/standalone_provider.go
```diff
@@ -21,6 +21,7 @@ func RegisterStandAloneProvider(s *grpc.Server, p types.PluginProvider, pType ty
 		internal.RegisterStandAloneMedianProvider(s, mp)
 		return nil
 	default:
```
george-dorin, line 23: This will break a test

bolekk (dismissed):

jmank88 (dismissed):

bolekk (approved):

PR #355, opened 2024-02-19
OCR3 Capability: Don't error on empty queue

bolekk:
Empty request queue is not an error. In fact, it will be a very common state when overall load is low.


bolekk:
FYI @george-dorin failing test seems to be related to one of the previous changes...?

george-dorin:
> FYI @george-dorin failing test seems to be related to one of the previous changes...?

Data race is inside the OCR2 tests which are untouched by the previous changes

archseer (approved):

george-dorin (approved):

PR #356, opened 2024-02-20
Make report format a non-alias and add string methods

samsondav:


pkg/types/llo/types.go
jmank88, line 1: Alternatively, you could implement this with a string array to simplify the implementation and maintenance. I think performance should be comparable: https://go.dev/play/p/tcmgf326DI3

jmank88: Oh I went too quick. This will match empty string on 0, but that should actually error. Can just Index over [1:] instead though: https://go.dev/play/p/fdJKjdb_qLi

samsondav: Will that work if we e.g. remove one in the future?

jmank88 (approved):

PR #358, opened 2024-02-21
Minor improvements to consensus capability reporting plugin

bolekk:
1. Better logging
2. Non-zero limits for phase output sizes
3. Change capability type name
4. Fix unnecessary exit on empty previous outcome
5. Fix incorrect clearing of ReportsToGenerate
6. Add missing map initialization inside Outcome()
7. Fix aggregator/encoder config parsing

### DIFF
--- a/pkg/capabilities/consensus/ocr3/capability.go
+++ b/pkg/capabilities/consensus/ocr3/capability.go
@@ -18,7 +18,7 @@ import (
 )
 
 var info = capabilities.MustNewCapabilityInfo(
-	"ocr3",
+	"offchain_reporting",
 	capabilities.CapabilityTypeConsensus,
 	"OCR3 consensus exposed as a capability.",
 	"v1.0.0",
@@ -76,10 +76,10 @@ func (o *capability) Close() error {
 }
 
 type workflowConfig struct {
-	AggregationMethod       string `mapstructure:"aggregation_method"`
-	AggregationMethodConfig map[string]any
-	Encoder                 string
-	EncoderConfig           map[string]any
+	AggregationMethod string         `mapstructure:"aggregation_method"`
+	AggregationConfig map[string]any `mapstructure:"aggregation_config"`
+	Encoder           string         `mapstructure:"encoder"`
+	EncoderConfig     map[string]any `mapstructure:"encoder_config"`
 }
 
 func (o *capability) RegisterToWorkflow(ctx context.Context, request capabilities.RegisterToWorkflowRequest) error {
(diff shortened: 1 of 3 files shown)


archseer (dismissed):

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -18,10 +18,10 @@ import (
 )
 
 var info = capabilities.MustNewCapabilityInfo(
-	"ocr3",
+	"offchain_reporting",
 	capabilities.CapabilityTypeConsensus,
 	"OCR3 consensus exposed as a capability.",
-	"v1.0.0",
+	"v3.0.0",
```
DeividasK, line 24: This is exposed to end users, so the versioning is different from our internal versioning. I suggest we stick with v1

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -76,10 +76,10 @@ func (o *capability) Close() error {
...
 
 type workflowConfig struct {
-	AggregationMethod       string `mapstructure:"aggregation_method"`
-	AggregationMethodConfig map[string]any
-	Encoder                 string
-	EncoderConfig           map[string]any
+	AggregationMethod string         `mapstructure:"aggregation_method"`
+	AggregationConfig map[string]any `mapstructure:"aggregation_config"`
+	Encoder           string         `mapstructure:"encoder"`
+	EncoderConfig     map[string]any `mapstructure:"encoder_config"`
```
DeividasK, line 82: To just `Aggregation` and `Encoder` (`Config` seems redundant), which takes in a map with config.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -97,9 +97,18 @@ func (o *capability) RegisterToWorkflow(ctx context.Context, request capabilitie
 		return err
 	}
 
+	if c.AggregationConfig == nil {
```
DeividasK, line 100: We should default to `data_feeds_2_0` and `EVM` encoder. This will keep the demo Workflow specs smaller and when we need something different (Solana) we'll be able to showcase this much better.

pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -32,6 +32,8 @@ type factoryService struct {
 const (
 	defaultRequestExpiry time.Duration = 1 * time.Hour
 	defaultBatchSize                   = 1000
+	maxPhaseOutputBytes                = 100000
+	maxReportCount                     = 20
```
DeividasK, line 36: Where are these numbers coming from? If they are arbitrary, could this be called out in a comment?

DeividasK (dismissed):

pkg/capabilities/consensus/ocr3/capability.go
bolekk: OCR versioning is also somewhat public (we release papers or give public talks about OCR3). If we refer to OCR I'd prefer to use v3. An alternative is to call is generically "consensus" and then we can version independently.

pkg/capabilities/consensus/ocr3/capability.go
bolekk: I like it but do you mind if I change in a separate PR? That change will span multiple files.

pkg/capabilities/consensus/ocr3/factory.go
bolekk: hmm do you want a comment that says "arbitrary"? How about I rename to "defaultMax.."? The existing two ones above are also arbitrary.

DeividasK: > hmm do you want a comment that says "arbitrary"?

Yes, if that's the case. When someone looks at this later, we don't want them to spend hours trying to understand why this was set to these specific numbers 😅

DeividasK (approved):

PR #359, opened 2024-02-21
Add Must* value functions

DeividasK:
A preferred alternative would be to update the original functions to not return an error.

### DIFF
--- a/pkg/values/bool.go
+++ b/pkg/values/bool.go
@@ -12,6 +12,11 @@ func NewBool(b bool) (*Bool, error) {
 	return &Bool{Underlying: b}, nil
 }
 
+func MustNewBool(b bool) *Bool {
+	value, _ := NewBool(b)
+	return value
+}
+
 func (b *Bool) Proto() (*pb.Value, error) {
 	return pb.NewBoolValue(b.Underlying)
 }
--- a/pkg/values/value.go
+++ b/pkg/values/value.go
@@ -45,6 +45,32 @@ func Wrap(v any) (Value, error) {
 	return nil, fmt.Errorf("could not wrap into value: %+v", v)
 }
 
+func MustWrap(v any) Value {
+	switch tv := v.(type) {
+	case map[string]any:
+		return MustNewMap(tv)
+	case string:
+		return MustNewString(tv)
+	case bool:
+		return MustNewBool(tv)
+	case []byte:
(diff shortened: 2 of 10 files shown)


pkg/values/value.go
```diff
@@ -45,6 +45,32 @@ func Wrap(v any) (Value, error) {
 	return nil, fmt.Errorf("could not wrap into value: %+v", v)
 }
 
+func MustWrap(v any) Value {
```
jmank88, line 48: Where would this be used?

pkg/values/bool.go
```diff
@@ -12,6 +12,11 @@ func NewBool(b bool) (*Bool, error) {
 	return &Bool{Underlying: b}, nil
 }
 
+func MustNewBool(b bool) *Bool {
+	value, _ := NewBool(b)
```
jmank88, line 16: Shouldn't they all panic?

pkg/values/value_test.go
```diff
@@ -23,6 +24,13 @@ func Test_Value(t *testing.T) {
 				mv, err := NewMap(m)
 				return m, mv, err
 			},
+			mustNewValue: func() (any, Value) {
+				m := map[string]any{
+					"hello": "world",
+				}
+				mv := MustNewMap(m)
```
jmank88, line 31: Did you consider declaring one generic func?
```suggestion
				mv := must(NewMap(m))
```
```go
func must[T any](t T, err error) T {
	if err != nil { panic(err) }
	return t
}
```
And if the usages are only from this package, we should prefer not to export it.

pkg/values/value.go
DeividasK: Initially, I thought that this would only be used in tests, but is it even possible for err to be not nil in map and list cases? If I understand correctly, we should always be able to match one of the types, so it should never error.

jmank88: Maybe but the callers only have the signature that includes an error to work with and shouldn't be assuming any implementation details (that could change in the future). If the panic case never executes, then there is no difference in practice, and if it does, we certainly want to know about it.

bolekk:
I suggested a similar change to @cedric-cordenier  - the only type that can currently return an error from New*() is Decimal so if we could somehow prevent that, we can just simplify New*() instead of adding new methods.

PR #361, opened 2024-02-21
Intercept stderr to log panics

george-dorin:
- Add a custom runner
- Log panics

### DIFF
--- a/pkg/loop/internal/runner/runner.go
+++ b/pkg/loop/internal/runner/runner.go
@@ -76,7 +76,7 @@ func (i *stderrWrapper) Read(p []byte) (n int, err error) {
 }
 
 func (i *stderrWrapper) Close() error {
-	i.Close()
+	i.r.Close()
 	return nil
 }
 
--- a/pkg/loop/internal/runner/runner.go
+++ b/pkg/loop/internal/runner/runner.go
@@ -1,0 +1,227 @@
+package runner
+
+import (
+	"context"
+	"debug/elf"
+	"debug/macho"
+	"debug/pe"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"os/user"
+	"runtime"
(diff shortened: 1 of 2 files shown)


pkg/loop/internal/runner/runner.go
```diff
@@ -0,0 +1,227 @@
...
+		return nil, err
+	}
+
+	c.logger = logger
+	c.cmd = cmd
+	c.cmd.Path = c.cmdPath
+	c.cmd.Args = c.cmdArgs
+	c.cmd.Env = append(c.cmd.Env, c.cmdEnv...)
+	c.stdout = stdout
+	c.stderr = NewIOWrapper(logger, stderr)
```
krehermann, line 109: i don't think we need a custom runner. instead, i think we can specific the stderr writer in the client config.

look at https://github.com/smartcontractkit/chainlink-common/blob/d1a00d5221f3328e7b74ccb17f891ba3bd227aa3/pkg/loop/plugin_relayer.go#L60

for example. and https://github.com/hashicorp/go-plugin/blob/8d2aaa458971cba97c3bfec1b0380322e024b514/client.go#L197

that client config lets us declare a stderr handle. you could add a 'panicWriter` config wrapper 

~ NewPanicWriterConfig(in *clientConfig) *clientConfig {

return &ClientConfig {
...
Stderr: NewPanicIOWrapper(in.logger, in.stderr)
}

and use this `panicWriter` everywhere to achieve the same thing that your custom runner is doing

krehermann: another thought is that you might be able to intercept the panic here, in our hclogger wrapper https://github.com/smartcontractkit/chainlink-common/blob/d1a00d5221f3328e7b74ccb17f891ba3bd227aa3/pkg/loop/logger.go#L69

if you know that `panics` are written a debug level, then in principle you might be able to put a parser in that switch and re-route then to Error level.

PR #362, opened 2024-02-21
[KS-28] create mercury trigger

patrick-dowell:
Initial draft of the mercury trigger for the keystone demo


pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,172 @@
...
+	"fmt"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+// TODO: Register the capiability in the NewDelegateFunction in chalink/core/services/workflows.delegate.go
```
HenryNguyen5, line 16:
```suggestion
// TODO: Register the capability in the NewDelegateFunction in chainlink/core/services/workflows.delegate.go
```

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,172 @@
...
+	defer o.mu.Unlock()
+
+	currentTime := time.Now()
+	unixTimestampMillis := currentTime.UnixNano() / int64(time.Millisecond)
+	feedId := report.feedId
+
+	// Create a new CapabilityResponse with the MercuryTriggerEvent
+	triggerEvent := MercuryTriggerEvent{
+		triggerType: "mercury",
+		id:          sha256Hash(strconv.FormatUint(feedId, 10) + strconv.FormatUint(uint64(report.observationTimestamp), 10)),
```
HenryNguyen5, line 71: I think it'd be good to extract this into a `generateTriggerEventId` method. I'd imagine we'd want this to eventually be part of any trigger service, since its a requirement for trigger events to have a unique ID. This will also make testing this function easier

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,172 @@
...
+		payload:     report,
+	}
+
+	val, err := values.Wrap(triggerEvent)
+	if err != nil {
+		return err
+	}
+
+	// Create a new CapabilityResponse with the MercuryTriggerEvent
+	event := capabilities.CapabilityResponse{
```
HenryNguyen5, line 82:
```suggestion
	capabilityResponse := capabilities.CapabilityResponse{
```

Nit: IMO this is just a capability response, the event itself in contained within it

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,172 @@
...
+type MercuryReport struct {
+	feedId               uint64
+	fullreport           []byte
+	benchmarkPrice       int64
+	observationTimestamp uint32
+}
+
+type MercuryTriggerEvent struct {
+	triggerType string // "mercury"
+	id          string // "sha256 of payload.feedId + payload.timestamp"
```
bolekk, line 43: I left this comment on the design doc too - I believe that a trigger event needs to contain multiple reports. Otherwise we won't be able to batch feed updates in onchain transactions - workflow executions are independent.

In consequence, I don't think that trigger ID can be based on a single feedID.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Create a new CapabilityResponse with the MercuryTriggerEvent
+	triggerEvent := MercuryTriggerEvent{
+		triggerType: "mercury",
+		id:          sha256Hash(strconv.FormatUint(feedId, 10) + strconv.FormatUint(uint64(report.observationTimestamp), 10)),
+		timestamp:   strconv.FormatInt(unixTimestampMillis, 10),
+		payload:     report,
+	}
+
+	val, err := values.Wrap(triggerEvent)
```
bolekk, line 76: Can we Wrap() structs / nested structs?

patrick-dowell: We'd probably have to add those structs to values.wrap here:

https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/values/value.go#L23

Unless go has some sort of indirection that would allow us to do it without having to implement each struct individually.

bolekk:
Looks good overall!  What can we do to make it work in the short term? Can we work around struct wrapping by translating into a nested map?

bolekk:
Ping - any updates here?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,199 @@
...
+
+	return nil
+}
+
+// Get array of feedIds from CapabilityRequest req
+func (o *MercuryTriggerService) GetFeedIds(req capabilities.CapabilityRequest) []uint64 {
+	feedIds := make([]uint64, 0)
+	// Unwrap the inputs which should return pair (map, nil) and then get the feedIds from the map
+	if inputs, err := req.Inputs.Unwrap(); err == nil {
+		if feeds, ok := inputs.(map[string]interface{})["feedIds"].([]uint64); ok {
```
bolekk, line 167: According to Austin, in current Mercury version we should use [32]byte as feed IDs (or just a byte array and we can validate len separately).

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,199 @@
...
+
+	currentTime := time.Now()
+	unixTimestampMillis := currentTime.UnixNano() / int64(time.Millisecond)
+	triggerIdsForReports := make(map[string][]int)
+	reportIndex := 0
+
+	for _, report := range reports {
+		// for each feed id, we need to find the triggerId associated with it.
+		reportFeedId := report.feedId
+		for triggerId, feedIds := range o.feedIdsForTriggerId {
```
bolekk, line 73: Did you consider a reverse mapping to speed up this lookup? We expect ProcessReport() calls to be much more frequent than RegisterTrigger() ones.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	return feedIds
+}
+
+// Get the triggerId from the CapabilityRequest req map
+func (o *MercuryTriggerService) GetTriggerId(req capabilities.CapabilityRequest) string {
+	var triggerId string
+	// Unwrap the inputs which should return pair (map, nil) and then get the triggerId from the map
+	if inputs, err := req.Inputs.Unwrap(); err == nil {
+		if id, ok := inputs.(map[string]interface{})["triggerId"].(string); ok {
```
bolekk, line 180: nit: instead of trusting that the triggerID will be globally unique, we should probably combine it with workflowID ourselves. Not important now, we can change it in the future.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,199 @@
...
+	return triggerId
+}
+
+func sha256Hash(s string) string {
+	hash := sha256.New()
+	hash.Write([]byte(s))
+	return hex.EncodeToString(hash.Sum(nil))
+}
+
+// TODO: The generated id should probably be the sha256 of all the report feed ids and timestamps associated with this trigger event
```
bolekk, line 193: Yeah, I think that should work.

pkg/values/value.go
```diff
@@ -139,3 +152,38 @@ func FromDecimalValueProto(decStr string) (*Decimal, error) {
 	}
 	return NewDecimal(dec)
 }
+
+func createMapFromStruct(v any) (Value, error) {
```
cedric-cordenier, line 156: @patrick-dowell Looks good 👍 

At first glance I was going to propose using mapstructure for this, since that's what we use elsewhere but it seems mapstructure doesn't have an Encode method 🤷 

I do think we should handle float64 natively if there's a use case for it -- since you've had to workaround it, I'm assuming there is. Could you add a float62 type instead, rather than adding the convert method?

pkg/capabilities/mercury/mercury.go
```diff
@@ -51,6 +52,21 @@ type ReportInfo struct {
 	Price     float64
 }
 
+// TODO: fix this by adding support for uint64 in value.go
+type FeedReport struct {
+	FeedID               int64  `json:"feedId"`
+	Fullreport           []byte `json:"fullreport"`
+	BenchmarkPrice       int64  `json:"benchmarkPrice"`
+	ObservationTimestamp int64  `json:"observationTimestamp"`
```
HenryNguyen5, line 60: We probably want the expiresAt and validFrom fields, the observation timestamp by itself isnt very useful to us

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,219 @@
...
+func sha256Hash(s string) string {
+	hash := sha256.New()
+	hash.Write([]byte(s))
+	return hex.EncodeToString(hash.Sum(nil))
+}
+
+func generateTriggerEventID(reports []mercury.FeedReport) string {
+	// Let's hash all the feedIds and timestamps together
+	s := ""
+	for _, report := range reports {
```
HenryNguyen5, line 182: We should apply a sort before concatting them all so we get deterministic event IDs. I'd imagine we'd want to sort by feedId then timestamp afterwards since we could have identical timestamps for two different reports under two different feed IDs

pkg/capabilities/triggers/mercury_trigger.go
HenryNguyen5: @bolekk triggerID is assumed to be under the workflowID namespace, i was under the assumption that the workflow engine will construct the fully namespaced workflow execution id

pkg/capabilities/triggers/mercury_trigger.go
HenryNguyen5: @bolekk why [32]byte rather than uint64? I know @samsondav is using the latter for LLO

pkg/capabilities/triggers/mercury_trigger.go
bolekk: oh ok then we're good!

pkg/capabilities/mercury/mercury.go
HenryNguyen5: NVM, this might be wrong. Confirming with austin first. Ignore for now.

pkg/capabilities/mercury/mercury.go
```diff
@@ -51,6 +52,21 @@ type ReportInfo struct {
 	Price     float64
 }
 
+// TODO: fix this by adding support for uint64 in value.go
+type FeedReport struct {
```
bolekk, line 56: Feel free to remove the old structures above and update the aggregator (we can also do it in a separate PR to not grow this one too much).

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,219 @@
...
+	capabilities.CapabilityInfo
+	chans                 map[workflowID]chan<- capabilities.CapabilityResponse
+	feedIDsForTriggerID   map[string][]int64 // TODO: switch this to uint64 when value.go supports it
+	mu                    sync.Mutex
+	triggerIDToWorkflowID map[string]string
+}
+
+var _ capabilities.TriggerCapability = (*MercuryTriggerService)(nil)
+
+func NewMercuryTriggerService() *MercuryTriggerService {
```
bolekk, line 37: Add logger and use it for some debug-level logs across all methods?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return nil
+}
+
+// Get array of feedIds from CapabilityRequest req
+func (o *MercuryTriggerService) GetFeedIDs(req capabilities.CapabilityRequest) []int64 {
+	feedIDs := make([]int64, 0)
+	// Unwrap the inputs which should return pair (map, nil) and then get the feedIds from the map
+	if inputs, err := req.Inputs.Unwrap(); err == nil {
+		if feeds, ok := inputs.(map[string]interface{})["feedIds"].([]int64); ok {
```
bolekk, line 153: Here you'll need to cast to []any instead of []int64.

pkg/values/value.go
patrick-dowell: This worked! Thanks!

pkg/capabilities/mercury/mercury.go
patrick-dowell: Will do in a separate PR

pkg/capabilities/triggers/mercury_trigger.go
patrick-dowell: Done!

pkg/capabilities/triggers/mercury_trigger.go
patrick-dowell: I will handle this in a separate PR.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/mercury"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+// TODO: Register the capability in the NewDelegateFunction in chainlink/core/services/workflows.delegate.go
```
bolekk, line 18: nit: remove this comment

pkg/capabilities/triggers/mercury_trigger.go
bolekk: can be done in a separate PR

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+	return &MercuryTriggerService{
+		CapabilityInfo:        mercuryInfo,
+		chans:                 map[string]chan<- capabilities.CapabilityResponse{},
+		feedIDsForTriggerID:   make(map[string][]int64),
+		triggerIDsForFeedID:   make(map[int64][]string),
+		triggerIDToWorkflowID: make(map[string]string),
+	}
+}
+
+// maybe use mercury-pipline reports.go report struct instead of MercuryReport struct, and then we can use the pipeline report generators
```
bolekk, line 49: This comment is confusing, maybe remove it and put it in some follow-up JIRA ticket with more details?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	currentTime := time.Now()
+	unixTimestampMillis := currentTime.UnixNano() / int64(time.Millisecond)
+	triggerIDsForReports := make(map[string][]int)
+	reportIndex := 0
+
+	for _, report := range reports {
+		for _, triggerID := range o.triggerIDsForFeedID[report.FeedID] {
```
bolekk, line 60: This "_" can be your reportIndex

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// maybe use mercury-pipline reports.go report struct instead of MercuryReport struct, and then we can use the pipeline report generators
+func (o *MercuryTriggerService) ProcessReport(reports []mercury.FeedReport) error {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	currentTime := time.Now()
+	unixTimestampMillis := currentTime.UnixNano() / int64(time.Millisecond)
+	triggerIDsForReports := make(map[string][]int)
```
bolekk, line 56: Maybe rename to triggerIDsToReports to emphasize the direction of the mapping? Now it's inconsistent with other ones like "feedIDsForTriggerID"

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// Get the triggerId from the CapabilityRequest req map
+func (o *MercuryTriggerService) GetTriggerID(req capabilities.CapabilityRequest) (string, error) {
+	// Unwrap the inputs which should return pair (map, nil) and then get the triggerId from the map
+	inputs, err := req.Inputs.Unwrap()
+	if err != nil {
+		return "", err
+	}
+	if id, ok := inputs.(map[string]interface{})["triggerId"].(string); ok {
```
bolekk, line 183: nit: more "triggerId" string to a const

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+
+	return nil
+}
+
+// Get array of feedIds from CapabilityRequest req
+func (o *MercuryTriggerService) GetFeedIDs(req capabilities.CapabilityRequest) []int64 {
+	feedIDs := make([]int64, 0)
+	// Unwrap the inputs which should return pair (map, nil) and then get the feedIds from the map
+	if inputs, err := req.Inputs.Unwrap(); err == nil {
+		if feeds, ok := inputs.(map[string]interface{})["feedIds"].([]any); ok {
```
bolekk, line 164: nit: move "feedIds" string to a const

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+			return err
+		}
+
+		// Create a new CapabilityResponse with the MercuryTriggerEvent
+		capabilityResponse := capabilities.CapabilityResponse{
+			Value: val,
+		}
+
+		wfID := o.triggerIDToWorkflowID[triggerID]
+		ch, ok := o.chans[wfID+triggerID]
```
bolekk, line 96: Extract the logic of building an internal key (wfID+triggerID) to a helper functions

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+		if ok {
+			ch <- capabilityResponse
+		} else {
+			return fmt.Errorf("no registration for %s", triggerID)
```
bolekk, line 100: I'm not convinced that returning an error from this function is very useful.
- What is the caller going to do with it? The caller just pushes the data that it received over the network.
- We might succeed with some responses and error on other ones so it's a bit weird that the order in the loop could determine what went through.

Maybe just log a loud error here?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return err
+	}
+	// If triggerId is already registered, return an error
+	if _, ok := o.chans[wid+triggerID]; ok {
+		return fmt.Errorf("triggerId %s already registered", triggerID)
+	}
+	feedIDs := o.GetFeedIDs(req) // TODO: what if feedIds is empty? should we throw an error or allow it?
+
+	o.chans[wid+triggerID] = callback
```
bolekk, line 122: Would it simplify things if we just identified a trigger by (wid, triggerID) throughout the whole file?
Or maybe just assume and validate that triggerIDs are globally unique like @HenryNguyen5 suggested earlier (IIRC)?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -0,0 +1,256 @@
...
+	return nil
+}
+
+func sha256Hash(s string) string {
+	hash := sha256.New()
+	hash.Write([]byte(s))
+	return hex.EncodeToString(hash.Sum(nil))
+}
+
+func removeFromSlice(list []string, element string) []string {
```
bolekk, line 244: Maybe change the slice to a set (map[string]struct{})? Then you wouldn't need to have this helper.

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -0,0 +1,201 @@
...
+	assert.Len(t, callback, 1)
+	select {
+	case msg := <-callback:
+		unwrapped, _ := mercury.Codec{}.UnwrapMercuryTriggerEvent(msg.Value)
+		assert.Equal(t, "mercury", unwrapped.TriggerType)
+		assert.Equal(t, GenerateTriggerEventID(fr), unwrapped.ID)
+		assert.Len(t, unwrapped.Payload, 1)
+		assert.Equal(t, fr[0], unwrapped.Payload[0])
+	default:
+		// The channel is empty; proceed without blocking
```
bolekk, line 57: Should we have this at all? We always expect something to be in that channel in this test.

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -0,0 +1,201 @@
...
+	case msg := <-callback2:
+		unwrapped, _ := mercury.Codec{}.UnwrapMercuryTriggerEvent(msg.Value)
+		assert.Equal(t, "mercury", unwrapped.TriggerType)
+		payload := make([]mercury.FeedReport, 0)
+		payload = append(payload, fr2[0])
+		assert.Equal(t, GenerateTriggerEventID(payload), unwrapped.ID)
+		assert.Len(t, unwrapped.Payload, 1)
+		assert.Equal(t, fr2[0], unwrapped.Payload[0])
+	default:
+		// The channel is empty; proceed without blocking
```
bolekk, line 193: Same as above

pkg/capabilities/mercury/mercury.go
```diff
@@ -92,6 +107,34 @@ func (m Codec) Wrap(reportSet ReportSet) (values.Value, error) {
...
+	}
+	event := val.(map[string]any)
+	mercuryTriggerEvent.TriggerType = event["TriggerType"].(string)
+	mercuryTriggerEvent.ID = event["ID"].(string)
+	mercuryTriggerEvent.Timestamp = event["Timestamp"].(string)
+	mercuryTriggerEvent.Payload = make([]FeedReport, 0)
+	for _, report := range event["Payload"].([]any) {
+		reportMap := report.(map[string]any)
+		mercuryReport := FeedReport{
+			FeedID:               reportMap["FeedID"].(int64),
```
bolekk, line 128: I think you can use the mapstructure library here too.

pkg/capabilities/triggers/mercury_trigger.go
patrick-dowell: If we assumed that triggers were globally unique, then you do lose a little security in that right now, workflow A can't remove workflow Bs triggers since they are effectively namespaced. I'm okay with identifying the trigger as (wid, triggerID) throughout the file though if that simplifies the code throughout (which I think it would).

bolekk (approved):

PR #371, opened 2024-02-26
[KS-59] Add Start to capability service; add some unit tests

cedric-cordenier:


jmank88 (approved):

pkg/capabilities/testing/mock_registry.go
```diff
@@ -0,0 +1,99 @@
+package testing
+
+import (
+	"context"
+	"errors"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+)
+
+type MockRegistry struct {
```
bolekk, line 10: Can we use mockery for this? Generate from CapabilitiesRegistry interface.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -17,8 +17,12 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
+const (
+	offchainReporting = "offchain_reporting"
```
bolekk, line 21: Maybe call it "ocrCapabilityType"?

pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -76,15 +47,11 @@ func (o *factoryService) NewReportingPlugin(config ocr3types.ReportingPluginConf
 }
 
 func (o *factoryService) Start(ctx context.Context) error {
-	return o.StartOnce("plugin factory service", func() error {
-		return o.capability.Start(ctx)
-	})
+	return nil
```
bolekk, line 50: Does the factory need to be a service any more? I actually find it a bit odd.

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: Hmmm... it doesn't really feel like a type though. WDYT of `ocrCapabilityId`?

bolekk: You're right, that's not type but ID, thanks.

pkg/capabilities/testing/mock_registry.go
cedric-cordenier: Hmmm... I've gone ahead and used mockery for this but I'm actually not sure this is the right way to proceed. There are some tests which depend on a concrete registry implementation rather than on a mock, so that would leave us with some tests using the generated mock and others using this mockRegistry implementation.

I'm not wild on the fragmentation so was thinking of using the mock_registry everywhere. WDYT?

bolekk: How hard is it to convert all tests to use the generated mock? I'm fine with leaving it instead of generating one if it's actually useful, your call.

jmank88 (dismissed):

pkg/loop/plugin.go
```diff
@@ -25,6 +26,20 @@ func (p *Plugin) SubService(s services.Service) {
 	p.mu.Unlock()
 }
 
+func (p *Plugin) Start(ctx context.Context) error {
```
bolekk, line 29: Where is this called from?

bolekk (approved):

pkg/capabilities/testing/mock_registry.go
cedric-cordenier: Turns out not that hard 🤷 I've gone ahead and changed.

jmank88 (dismissed):

cedric-cordenier:
@bolekk The tests should be fixed now 👍

bolekk (approved):

PR #373, opened 2024-02-26
BCF-3006: refactor loop test suite

krehermann:
BCF-3006: test suite refactoring to mirror the natural domain separation in our interfaces

Prior to this change our loop test harness packed all the logic into one package. It was becoming increasing difficult to add new tests and to reason about tests because they shared values that where logically distinct.

This work uncovered a number of copy and paste typos as well as missing tests which were mostly like caused by intermixed test implementations.

There is a README in the change that explains the general pattern.

Note the none of the calling tests change other than to reference the new, decoupled test implementations


pkg/loop/internal/test/README.md
krehermann, line 1: To review this change, I suggest starting here, then minimizing the `test` dir and reading the changes file in `internal` ie `internal/median_service_test.go` ...
and then looking at the details of the new test pkgs and impls

pkg/loop/internal/test/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+package x_test
+
+var FooImpl = staticFoo{
+    expectedStr = "a"
+    expectedInt = 1
+}
+
+type FooEvaluator interface {
```
krehermann, line 25: if i were more ambitious, i could likely turn this into a generic. that could be nice to enforce a bit more rigidity for new test implementations.

however, that is out of scope for now though as the real goal of this PR is to unblock CCIP. we can iterate on the implement details of the test suite after delivering OKR

pkg/loop/internal/test/README.md
```diff
@@ -0,0 +1,65 @@
...
+    foo // some interface in the used in the composition of a loop
+    // run methods of other, returns first error or error if
+    // result of method invocation does not match expected value of embedded foo
+    Evaluate(ctx context.Context, other foo) error
```
cedric-cordenier, line 29: I'd suggest `IsEqual` or `Equals` here

cedric-cordenier:
@krehermann Some comments but otherwise the direction seems good to me 👍

pkg/loop/internal/test/README.md
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type staticFoo struct {
+    expectedStr string
+    expectedInt int
+    ...
+}
+
+var _ FooEvaluator = staticFoo
```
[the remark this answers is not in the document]
krehermann: yes. i'll add it.

pkg/loop/internal/mercury/common/test/chain_reader.go
```diff
@@ -2,14 +2,57 @@ package mercury_common_test
...
+			},
+			{
+				Hash:   []byte{2},
+				Number: 2,
+			},
+		},
+	},
+}
+
+type MercuryChainReaderEvaluator interface {
```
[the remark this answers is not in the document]
krehermann: that makes sense, i'll rework the packaging

nit: i can't define the interface where it's use and do the `var _ iface = x` trick, right, because that would be a circular dependence. i'd need a pkg just for the interfaces

pkg/loop/standalone_provider_test.go
```diff
@@ -7,13 +7,13 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/loop"
-	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test"
+	pluginprovider_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test/ocr2/plugin_provider"
```
jmank88, line 10: Why the alias?

pkg/loop/reportingplugins/loopp_service_test.go
```diff
@@ -10,6 +10,9 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/loop"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test"
+	reportingplugin_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test/ocr2/reporting_plugin"
+	resources_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test/resources"
+	coreapi_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test/resources/core_api"
```
jmank88, line 15: I'm not a fan of underscores in package names, but let's align the aliases either way:
```suggestion
	core_api_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test/resources/core_api"
```
```suggestion
	coreapi_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/test/resources/coreapi"
```

pkg/loop/internal/test/README.md
```diff
@@ -0,0 +1,95 @@
+This package contains test implementations for all the components that compose OCR2 and OCR3 LOOPPs.
+
+The directory layout is intended to mirror the natural interfaces composition of libocr:
+- ocr2: contains static test implementations for reuseable reporting plugin, factory, and plugin provider
+- ocr3: implements the ocr3 generic-based reporting plugin and reused common components from ocr2
+- resources: represents core-node provided resources used to in calls to core node api(s) eg pipeline_runner, keystore, etc
```
jmank88, line 6: Did you consider other names? `core` seems straightforward to me :shrug: 
```suggestion
- core: represents core-node provided resources used to in calls to core node api(s) eg pipeline_runner, keystore, etc
```
Or would it make sense to split them up further? e.g. if we promote the keystore to a plugin later, outside of core-node provided things, will the organization still make sense?

pkg/loop/internal/test/README.md
```diff
@@ -0,0 +1,95 @@
...
+- mercury: product specific test implementation required to fulfill MercuryProvider and constructors
+
+Every test implementation follows the pattern wrapping an interface and provider one or more funcs to compare to another
+instance of the interface. Every package attempts to share exposed the bare minimum surface area to avoid entanglement with logically seperated tests and domains.
+
+In practice this is accomplished by exporting a static implementation and an interface that the implementation satisfies. The interface is used by other packages to declaration dependencies and the implementation is used in the test implementation of those other packaged
+
+Example
+
+```
```
jmank88, line 17:
```suggestion
```go
```

pkg/loop/internal/test/README.md
jmank88: It looks like the code block formatting is off

pkg/loop/standalone_provider_test.go
krehermann: i find it easier to read. if we did eventually move to individual packages with sub packages for `/test` wouldn't it read like this?

pkg/loop/internal/test/README.md
krehermann: i originally had all the `resources` in there own packages. that felt too granular at this stage because it led to tiny packages. i consider `core` but decided to be a little more agnostic precisely because of questions about things like the keystore.

pkg/loop/standalone_provider_test.go
jmank88: > if we did eventually move to individual packages with sub packages for /test wouldn't it read like this?

Yes but it would be aligned.


Another common pattern requiring no aliases is:
```suggestion
	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/ocr2/pluginprovider/pluginprovidertest"
```

krehermann:
> I'm a little concerned about the scalability still. Having multiple packages under a `test/` dir helps but still has tradeoffs, like potentially requiring a redundant sub-tree under `test/`, duplicating all the normal packages. A more scalable approach, and something that we have some precedent for already (with mercury) is to have `<thing>/test` packages instead. This would help align package import aliases too.

I hear you. We don't have good top level packages to embedded the `/test` though. I think this change will make that kind of organization easier if we invest the time to refactor those other packages. the goal here is make the CCIP dev go faster and i think this is a reasonable and useful step to both cleanup our tests and helping the next stage of development

pkg/loop/reportingplugins/loopp_service_test.go
jmank88: https://go.dev/blog/package-names

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

pkg/loop/standalone_provider_test.go
krehermann: i've moved the code to pkgXXX/test where ever possible, instead of test/XXX eg median/test, ocr3/test, ...

i filed this as follow up work https://smartcontract-it.atlassian.net/browse/BCF-3039 and put a reference in the README to clean it up further.

the fact that is was possible for me to move the code around so quickly is strong evidence that this change is both good progress and sufficient for the current purposes

pkg/loop/internal/mercury/common/test/chain_reader.go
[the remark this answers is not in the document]
jmank88: You can do the interface check from package x_test files too

cedric-cordenier (approved):

PR #374, opened 2024-02-27
[fix] Fix panic when instantiating factory service

cedric-cordenier:


pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -1,28 +1,34 @@
...
+	services.StateMachine
 }
 
 const (
 	defaultMaxPhaseOutputBytes = 100000
 	defaultMaxReportCount      = 20
 )
 
-func newFactory(s *store, batchSize int, lggr logger.Logger) (*factory, error) {
+func newFactory(s *store, c *capability, batchSize int, lggr logger.Logger) (*factory, error) {
```
cedric-cordenier, line 26: @bolekk Some context here: previously we were embedding capability (which is a service), but I evidently forgot to actually plug it in. This made factory implement the service interface even though I had removed that in my previous PR.

Unfortunately, factory does need to be a service -- that's a mandate of the ReportingPluginFactory type, so I've stubbed out the relevant functions and removed the embedding of the capability. The capability is instantiated elsewhere and I don't want to conflate calling start on the factory (which rightfully should noop) with calling start on the capability.

cedric-cordenier: It's part of the LOOPP reporting plugin definition (chainlink-common/pkg/types/reporting_plugin_service.go) and is used by the LOOPP framework.

PR #377, opened 2024-02-27
[KS-60] Return a response regardless of whether a report should be em…

cedric-cordenier:
…itted


pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -187,18 +187,16 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
...
 
-		if outcome.ShouldReport {
-			report := &pbtypes.Report{
-				Outcome: outcome,
-				Id:      weid,
-			}
-			o.ReportsToGenerate = append(o.ReportsToGenerate, report)
+		report := &pbtypes.Report{
+			Outcome: outcome,
+			Id:      weid,
```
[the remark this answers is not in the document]
cedric-cordenier: It's to handle the case where there are two outcomes for a workflow in a given round. In that case the outcomes map would only store the latest Outcome (which would of course depend on the first Outcome).

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -200,31 +198,39 @@ func (r *reportingPlugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]oc
...
+			}
+
+			enc, err := r.r.getEncoder(id.WorkflowId)
+			if err != nil {
+				r.lggr.Errorw("could not retrieve encoder for workflow", "error", err, "workflowID", id.WorkflowId)
+				continue
+			}
+
+			mv := values.FromMapValueProto(newOutcome.EncodableOutcome)
+			report, err = enc.Encode(context.Background(), *mv)
```
bolekk, line 224: This feels a little risky. We require every aggregator to always produce an encodable report even if should_report was set to true. It will work with DataFeeds (empty array) but is it generally safe?
It's especially weird given that ShouldAccept later ignores the encoded report altogether :) 

I'm not sure what an alternative could be, though... a signed empty byte array (with attached IDs)...?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -239,23 +245,35 @@ func (r *reportingPlugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]oc
...
+	if err != nil {
+		r.lggr.Error("could not unmarshal info")
+		return false, err
+	}
+
+	resp := map[string]any{}
+	if info.ShouldReport {
+		resp["report"] = []byte(rwi.Report)
+	} else {
+		resp["report"] = nil
```
bolekk, line 259: Maybe we should keep the report for observability purposes? If not then why even generate it (see my earlier comment)...

bolekk (approved):

pkg/capabilities/consensus/ocr3/reporting_plugin.go
bolekk: ugh, of course, sorry for a silly comment

PR #379, opened 2024-02-28
[fix] Synchronize closing the callback channel to avoid panics

cedric-cordenier:
Move each timer to its own goroutine, and add a `transmitCh` to synchronize all transmissions and associated channel closures. This avoids races with respect to the various channels used in the capability.

In particular, the use of `store.evict()` in the transmit handler guarantees that we'll only ever issue one transmission response and prevents races between the reporting plugin transmission and a request's expiration.

Also do some cleanup of the store: previously we set the request expiry there, but since none of the other logic around expiry happens there, I've moved this to the capability.


jmank88 (approved):

jmank88 (approved):

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -199,14 +205,30 @@ func (o *capability) loop() {
...
+			o.handleTransmitMsg(ctx, resp)
 		}
 	}
 }
 
-func (o *capability) expiryWorker(ctx context.Context, r *request) {
+func (o *capability) handleTransmitMsg(ctx context.Context, resp *response) {
+	req, wasPresent := o.store.evict(ctx, resp.WorkflowExecutionID)
+	if !wasPresent {
+		o.lggr.Debugf("failed to clean up request with id %s", resp.WorkflowExecutionID)
```
bolekk, line 220: We will see this log for every single request, right? IIUC we never cancel expiry timers early.

pkg/capabilities/consensus/ocr3/models.go
```diff
@@ -13,14 +13,14 @@ type request struct {
 	ExpiresAt    time.Time
 
 	CallbackCh chan<- capabilities.CapabilityResponse
+	CleanupCh  chan struct{}
```
bolekk, line 16: I can't see this used anywhere...?

bolekk (commented):
Looks good!

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -199,14 +205,31 @@ func (o *capability) loop() {
...
+		case resp := <-o.transmitCh:
+			o.handleTransmitMsg(ctx, resp)
 		}
 	}
 }
 
-func (o *capability) expiryWorker(ctx context.Context, r *request) {
+func (o *capability) handleTransmitMsg(ctx context.Context, resp *response) {
+	req, wasPresent := o.store.evict(ctx, resp.WorkflowExecutionID)
+	if !wasPresent {
```
bolekk, line 219: The debug log that was here can still be useful, I just thought that the message should be different if this is something expected. Or maybe no log for now is better.

bolekk (approved):

PR #384, opened 2024-03-04
Convert panics from log level debug to log level error

george-dorin:


pkg/loop/internal/test/logger_loop.go
```diff
@@ -51,3 +52,31 @@ func PluginLoggerTestHandshakeConfig() plugin.HandshakeConfig {
...
+	plugin.NetRPCUnsupportedPlugin
+
+	logger.Logger
+}
+
+func (g *GRPCPluginLoggerTestPanic) GRPCServer(*plugin.GRPCBroker, *grpc.Server) (err error) {
+	//Simulate panic after GRPC is started
+	go func() {
+		time.Sleep(time.Second)
+		panic("test panic")
```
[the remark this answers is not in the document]
jmank88: Actually, that's the only way I would expect us to be able to recover the panic at all. How does this work currently?

pkg/loop/logger.go
```diff
@@ -52,6 +52,14 @@ func removeArg(args []interface{}, key string) ([]interface{}, string) {
 	return args, ""
 }
 
+func logDebug(msg string, l logger.Logger, args ...interface{}) {
```
krehermann, line 55: let's add a comment about what we are trying to intercept panics here. IIRC you said something like if hashicorp can't parse a msg, then it falls back to logging that message at Debug. we need to capture that context in a comment here

george-dorin: Addressed in [`bc1e8fa` (#384)](https://github.com/smartcontractkit/chainlink-common/pull/384/commits/bc1e8fa5dc95420d53d217baca37818f5b266eef)

pkg/loop/internal/test/logger_loop.go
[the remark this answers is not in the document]
jmank88: Oh I see, you aren't actually "recovering" the panic. Can we tweak the title? I think others will assume catch == recover, since panic/recover is analogous to throw/catch.

pkg/loop/logger.go
```diff
@@ -52,6 +52,16 @@ func removeArg(args []interface{}, key string) ([]interface{}, string) {
 	return args, ""
 }
 
+// logDebug will parse msg and figure out if it's a panic, this is done here because the hashicorp plugin will push any
+// unrecognizable message from stderr as a debug statement
+func logDebug(msg string, l logger.Logger, args ...interface{}) {
+	if len(msg) > 7 && msg[:6] == "panic:" {
```
jmank88, line 58:
```suggestion
	if strings.HasPrefix(msg, "panic:") {
```

jmank88:
Can we do something similar for `system.Exit`?

pkg/loop/logger.go
george-dorin: Changed in [`c7fa511` (#384)](https://github.com/smartcontractkit/chainlink-common/pull/384/commits/c7fa511229b14237e6951cb76d428de6db76baf4)

jmank88 (approved):

krehermann (approved):

vyzaldysanchez (approved):

pkg/loop/internal/test/logger_loop.go
```diff
@@ -20,15 +20,32 @@ const LoggerTestName = "server-side-logger-name"
...
-	err = errors.New("test error")
-	g.Logger.Errorw("Error!", "err", err)
-	err = errors.Join(err, g.Logger.Sync())
-	time.Sleep(time.Second)
+	//Simulate panic/error/log after GRPC is started, if a panic is thrown before the GRPC server is initialized
+	//it will not be caught as stderr will be closed before HashiCorp plugin will have a change to read from it
+	go func() {
+		time.Sleep(time.Second)
+		switch g.ErrorType {
+		case 0:
```
jmank88, line 33: Maybe worth an enum so these can be named?

pkg/loop/logger.go
```diff
@@ -52,6 +55,81 @@ func removeArg(args []interface{}, key string) ([]interface{}, string) {
...
+}
+
+// logDebug will parse msg and figure out if it's a panic, fatal or critical log message, this is done here because the hashicorp plugin will push any
+// unrecognizable message from stderr as a debug statement
+func logDebug(msg string, l logger.Logger, args ...interface{}) {
+	if strings.HasPrefix(msg, "panic:") {
+		l.Errorw(msg, args...)
+	} else if log, err := parseJSON(msg); err == nil {
+		if log.Level == "dpanic" || log.Level == "fatal" || log.Level == "critical" {
+			l.Errorw(log.Message, flattenExtraArgs(log)...)
```
jmank88, line 124: Should we add a prefix here? Otherwise they just look like regular error level, right?

jmank88: Oh wait, we can just forward to the correct method, right?
```suggestion
			l.Fatalw(log.Message, flattenExtraArgs(log)...)
```

pkg/loop/internal/test/logger_loop.go
george-dorin: Added in [`11a37f3` (#384)](https://github.com/smartcontractkit/chainlink-common/pull/384/commits/11a37f37be63708efaad9535f23a70d4d848e584)

pkg/loop/logger.go
krehermann: which process would be exit'd? the loop or the core node? seems correct to exit the loop and definitely wrong to exit the core node

vyzaldysanchez: Wouldn't be helpful to at least state when it is a panic, a fatal or a critical log level in the log when we forward the message?

I feel it would be good to understand the severity and not encapsulate/group all 3 error levels in a single error, if that makes sense.

george-dorin: @krehermann it's going to exit the core node
@vyzaldysanchez we can try prefixing it, it will come out a bit ugly like `[ERROR] [CRIT] some message` but if that's fine we can go with it

vyzaldysanchez:
There's a flake on this [run](https://github.com/smartcontractkit/chainlink-common/actions/runs/8174802961/job/22350443029?pr=384). Can we track it somewhere?

pkg/loop/logger.go
jmank88: Oh right, no fatal. But what about critical?

And I think we could escalate the panic logs to ciritical level too.

george-dorin: Added prefixes and escalated the logs

pkg/loop/logger.go
```diff
@@ -52,6 +56,83 @@ func removeArg(args []interface{}, key string) ([]interface{}, string) {
...
+// logDebug will parse msg and figure out if it's a panic, fatal or critical log message, this is done here because the hashicorp plugin will push any
+// unrecognizable message from stderr as a debug statement
+func logDebug(msg string, l logger.SugaredLogger, args ...interface{}) {
+	if strings.HasPrefix(msg, "panic:") {
+		l.Criticalw(fmt.Sprintf("[PANIC] %s", msg), args...)
+	} else if log, err := parseJSON(msg); err == nil {
+		switch log.Level {
+		case "dpanic", "fatal":
+			l.Criticalw(fmt.Sprintf("[%s] %s", strings.ToUpper(log.Level), log.Message), flattenExtraArgs(log)...)
+		case "critical":
```
jmank88, line 126: `critical` is just our own renaming of zap's `dpanic`, so they can be mapped together. Could simplify the format a bit then too:
```suggestion
		case "fatal":
			l.Criticalw(fmt.Sprintf("[FATAL] %s", log.Message), flattenExtraArgs(log)...)
		case "critical", "dpanic":
```

george-dorin: Nice, added in [`fd68ba6` (#384)](https://github.com/smartcontractkit/chainlink-common/pull/384/commits/fd68ba61d00e8a18f97e9ee2441ecec7993594bf)

jmank88 (approved):

PR #388, opened 2024-03-05
[KS-66/KS-67/KS-68] More values library improvements

cedric-cordenier:
* Add support for big.Int
* Support UnwrapTo on all values; make this part of the Unwrappable interface
* More complex type support, eg. []int, []string, []decimal.Decimal etc.


pkg/values/value.go
```diff
@@ -38,9 +40,99 @@ func Wrap(v any) (Value, error) {
...
+		}
+		return NewMap(nm)
+	case map[string]big.Int:
+		nm := map[string]any{}
+		for k, v := range tv {
+			nm[k] = v
+		}
+		return NewMap(nm)
+
+	// Better complex type support for lists
```
DeividasK, line 92: Could this be made more efficient using reflection and recursion? I wanted to explore this but didn't get to it.

cedric-cordenier: Good point :) I reworked those sections, please take a look

pkg/values/map.go
```diff
@@ -73,10 +74,34 @@ func (m *Map) UnwrapTo(toStruct any) error {
...
+			if err != nil {
+				return nil, err
+			}
+
+			d[k] = unw
+		}
+
+		return d, nil
+	}
+	return data, nil
```
bolekk, line 97: Shouldn't this be an error? When does returning back unmodified data make sense?

bolekk (approved):

pkg/values/map.go
cedric-cordenier: An error means that this decode hook should have handled the transformation but couldn't; returning the data unmodified means we fallback to the next decode hook and/or whatever other decode logic mapstructure uses by default.

PR #393, opened 2024-03-06
Property Extract Modifier

EasterTheBunny:
This commit adds a modifier that extracts a single property from a struct to an off-chain value.

### DIFF
--- a/pkg/codec/property_extractor.go
+++ b/pkg/codec/property_extractor.go
@@ -80,35 +80,11 @@ func (e *propertyExtractor) RetypeToOffChain(onChainType reflect.Type, itemType 
 }
 
 func (e *propertyExtractor) TransformToOnChain(offChainValue any, _ string) (any, error) {
-	rItem := reflect.ValueOf(offChainValue)
-
-	toType, ok := e.offToOnChainType[rItem.Type()]
-	if !ok {
-		return reflect.Value{}, fmt.Errorf("%w: cannot retype %v", types.ErrInvalidType, rItem.Type())
-	}
-
-	output, err := expandWithMapsHelper(rItem, toType, e.fieldName)
-	if err != nil {
-		return reflect.Value{}, err
-	}
-
-	return output.Interface(), nil
+	return extractOrExpandWithMaps(offChainValue, e.offToOnChainType, e.fieldName, expandWithMapsHelper)
 }
 
 func (e *propertyExtractor) TransformToOffChain(onChainValue any, _ string) (any, error) {
-	rItem := reflect.ValueOf(onChainValue)
-
-	toType, ok := e.onToOffChainType[rItem.Type()]
-	if !ok {
-		return reflect.Value{}, fmt.Errorf("%w: cannot retype %v", types.ErrInvalidType, rItem.Type())
-	}
-
-	output, err := extractWithMapsHelper(rItem, toType, e.fieldName)
-	if err != nil {
-		return reflect.Value{}, err
-	}
-
-	return output.Interface(), nil
+	return extractOrExpandWithMaps(onChainValue, e.onToOffChainType, e.fieldName, extractWithMapsHelper)
 }
 
 func (e *propertyExtractor) getPropTypeFromStruct(onChainType reflect.Type) (reflect.Type, error) {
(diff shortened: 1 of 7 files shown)


pkg/codec/property_extractor.go
```diff
@@ -0,0 +1,98 @@
...
+	fieldName string
+}
+
+func (e *propertyExtractor) TransformToOnChain(offChainValue any, _ string) (any, error) {
+	parts := strings.Split(e.fieldName, ".")
+	fieldName := parts[len(parts)-1]
+
+	fields := []reflect.StructField{
+		{Name: fieldName, Type: reflect.TypeOf(offChainValue)},
+	}
```
nolag, line 43: Consider that the way modifiers work today, they could extract 

A.B.C and expect to get C on-chain.  The off-chain would need to nest C correctly on the way back to be an A with a B with a C not just a B with a C

pkg/codec/property_extractor.go
```diff
@@ -0,0 +1,98 @@
...
+
+	_, ok := e.onToOffChainType[rItem.Type()]
+	if !ok {
+		return reflect.Value{}, fmt.Errorf("%w: cannot retype %v", types.ErrInvalidType, rItem.Type())
+	}
+
+	switch rItem.Kind() {
+	case reflect.Struct:
+	case reflect.Pointer:
+		rItem = rItem.Elem()
```
nolag, line 64: this will die if it's a struct?  It's easier to use recursion sometimes.

pkg/codec/property_extractor.go
```diff
@@ -0,0 +1,98 @@
...
+		if rItem.Kind() != reflect.Struct {
```
nolag, line 65: Today, you can do all the Modifiers on lists and get back a list of modified items.  It would be nice to allow that in all modifiers.

Ie: If you had []struct{A : string, B: int} and you extract A you should get back a []string with values from A

pkg/codec/property_extractor.go
EasterTheBunny: It doesn't die if its a struct. Its just a non-error case where nothing special needs to be done to `rItem`. I'll admit the code isn't intuitive on this one.

pkg/codec/property_extractor.go
EasterTheBunny: Oh that's an interesting case. I can add that one in.

pkg/codec/property_extractor.go
EasterTheBunny: This was changed to support other features anyway.

pkg/codec/property_extractor.go
EasterTheBunny: done

pkg/codec/property_extractor_test.go
```diff
@@ -0,0 +1,250 @@
...
+				A: true,
+				B: 42,
+			},
+		}
+
+		offChainValue, err := nestedExtractor.TransformToOffChain(onChainValue, "")
+		require.NoError(t, err)
+		require.Equal(t, int64(42), offChainValue)
+
+		lossyOnChain, err := nestedExtractor.TransformToOnChain(int64(3), "")
```
nolag, line 113: nit: it's nice to use the same value that you got from the result to show it works with the other transform.

nolag (commented):
Just double-check that you can't reuse more from base, or refactor to share more.

I do wonder if you want to also have extract properties that would wrap them in a struct.  That way if you have
{
   A: {B: "foo", C: "bar"},
   C: {D: "X", "E": "zzz"}
}
You could say I want A.C and C.E. to get {C: "bar", "E": "zzz"}  If you don't need it for solana, ignore me.  I'm not sure if you will.

nolag (commented):
Can you add this to example_test too please :)

pkg/codec/property_extractor_test.go
EasterTheBunny: Like using the `offChainValue` to transform to `lossyOnChain`?

pkg/codec/utils.go
```diff
@@ -152,24 +152,55 @@ func SliceToArrayVerifySizeHook(from reflect.Type, to reflect.Type, data any) (a
...
+		if !ok {
+			return data, nil
+		}
+
+		unixTime, ok := rawUnixTime.(int64)
+		if !ok {
+			return data, nil
+		}
+
+		return convertToEpoch(to, time.Unix(unixTime, 0)), nil
```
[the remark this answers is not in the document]
EasterTheBunny: That is correct. I'll add tests for those.

pkg/codec/property_extractor_test.go
nolag: yeah

nolag (approved):

PR #394, opened 2024-03-06
ccip method updates and address json overrides

dimkouv:
- Add new methods for CCIP provider interfaces.
- Override JSON marshal/unmarshal functionality for CCIP backwards compatibility. This change is required for compatibility with nodes running old CCIP version since `common.Address` type is encoded to `lower-case` and the string representation that we were using in some places was `EIP55`. Apart from that `MarshalText` and `UnmarshalText` also need to be implemented for encoding addresses in map keys. For reference CCIP uses https://github.com/goccy/go-json 

In order not to use `go-ethereum` just for EIP55, I added a custom implementation. I followed the approach described in [the book mastering ethereum](https://www.oreilly.com/library/view/mastering-ethereum/9781491971932/ch04.html) (search for: Hex Encoding with Checksum in Capitalization (EIP-55)) since I couldn't find some nice implementation anywhere.


pkg/types/ccip/models.go
```diff
@@ -2,10 +2,49 @@ package ccip
...
+	"fmt"
+	"strings"
+	"unicode"
+
+	"golang.org/x/crypto/sha3"
 )
 
 type Address string
 
+func (a *Address) UnmarshalJSON(bytes []byte) error {
```
krehermann, line 14: add a code comment about why we need this custom un/marshaling. iiuc, it boils down to backward compatibility

pkg/types/ccip/models.go
```diff
@@ -2,10 +2,55 @@ package ccip
...
+
+	"golang.org/x/crypto/sha3"
 )
 
+// Address is the generic address type used across CCIP plugins.
+//
+// NOTE: JSON codec had to be overridden for CCIP backwards compatibility.
+// Before this generic address type, CCIP was using common.Address from go-ethereum library which marshals
+// to lower-case and prints as EIP55. We have to maintain this behavior to keep nodes that run different
+// versions to come in consensus.
```
makramkd, line 17:
```suggestion
// versions to come to consensus.
```

pkg/types/ccip/models.go
```diff
@@ -18,3 +63,40 @@ type TxMeta struct {
 	TxHash                  string
 	LogIndex                uint64
 }
+
+// EIP55 converts the provided address to EIP55.
+//
+// If the ith digit is a letter (ie. it’s one of abcdef) print it in uppercase if the 4*ith bit of the hash of the
+// lowercase hexadecimal address is 1 otherwise print it in lowercase.
+func EIP55(addr string) (string, error) {
```
makramkd, line 71: The "standard" implementation of this in Go is go-ethereum's `checksumHex`: https://github.com/ethereum/go-ethereum/blob/e31709db6570e302557a9bccd681034ea0dcc246/common/types.go#L261

pkg/types/ccip/models.go
```diff
@@ -18,3 +63,40 @@ type TxMeta struct {
...
+	addr = strings.ToLower(strings.TrimPrefix(addr, "0x"))
```
makramkd, line 72: Since `0x` hex strings as addresses are pretty much exclusively Ethereum addresses, I wonder if we can gate this function on that as well, i.e `if !strings.StartsWith(addr, "0x") { return error }`

pkg/types/ccip/models.go
makramkd, line 71: This function does multiple things right now - its basically our check for whether an address is an Ethereum address or not, and secondarily, it tries to convert it to EIP55 format.

Not necessarily bad but I'm wondering if we need more robust checks here to avoid e.g EIP-55'ing a Solana or Starknet address.

dimkouv: I will also have to copy some more stuff if I copy this.
```
BytesToAddress(FromHex(s))
```

I think the difference is only in performance but current implementation is cleaner.

dimkouv: cleaned this up

pkg/types/ccip/models.go
dimkouv: fixed

makramkd (approved):

jmank88 (changes requested):
This approach will not work. This module/repo is chain-agnostic. The chain-specifics, like the format used for Account/Address types, is not known here - it is a chain-relayer implementation detail.

PR #396, opened 2024-03-08
pkg/sqlutil: add QueryHook for wrapping Queryers with timeouts, prometheus metrics and slow query logging

jmank88:
This PR introduces `sqlutil.QueryHook`s based on the core `pq.Q` type's behavior: https://github.com/smartcontractkit/chainlink/blob/5f212bbc8485fc1fb20076ef5df7842f7bbfc6c1/core/services/pg/q.go#L117

https://smartcontract-it.atlassian.net/browse/BCF-2978

image
image

### DIFF
--- a/pkg/sqlutil/hook.go
+++ b/pkg/sqlutil/hook.go
@@ -1,0 +1,126 @@
+package sqlutil
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/jmoiron/sqlx"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+var _ DB = &WrappedDB{}
+
+// WrappedDB is a [DB] which invokes a [QueryHook] on each call.
+type WrappedDB struct {
+	db   DB
+	lggr logger.Logger
+	hook QueryHook
+}
+
+// QueryHook is a func that is executed for each query, providing an opportunity to measure, log, inspect/modify errors, etc.
+// The do func *must* be called.
+// Logs emitted through the provided logger.Logger will have the caller's line info.
+//
+// See [MonitorHook] and [TimeoutHook] for examples.
+type QueryHook func(ctx context.Context, lggr logger.Logger, do func(context.Context) error, query string, args ...any) error
+
+// NewWrappedDB returns a new [WrappedDB] that calls each [QueryHook] in the provided order.
+func NewWrappedDB(db DB, l logger.Logger, hs ...QueryHook) *WrappedDB {
+	iq := WrappedDB{db: db,
+		lggr: logger.Helper(logger.Named(l, "WrappedDB"), 2), // skip our own wrapper and one interceptor
+		hook: noopHook,
+	}
+	switch len(hs) {
+	case 0:
+		return &iq
+	case 1:
+		iq.hook = hs[0]
+		return &iq
+	}
+
+	// Nest the QueryHook calls so that they are wrapped from first to last.
+	// Example:
+	// 	[A, B, C] => A(B(C(do())))
+	for i := len(hs) - 1; i >= 0; i-- {
+		next := hs[i]
+		prev := iq.hook
+		iq.hook = func(ctx context.Context, lggr logger.Logger, do func(context.Context) error, query string, args ...any) error {
+			// opt: cache the construction of these loggers
+			lggr = logger.Helper(lggr, 1) // skip one more for this wrapper
+			return next(ctx, lggr, func(ctx context.Context) error {
+				lggr = logger.Helper(lggr, 2) // skip two more for do() and this extra wrapper
+				return prev(ctx, lggr, do, query, args...)
+			}, query, args...)
+		}
+	}
+	return &iq
+}
+
+func noopHook(ctx context.Context, lggr logger.Logger, do func(context.Context) error, query string, args ...any) error {
+	return do(ctx)
+}
+
+func (w *WrappedDB) DriverName() string {
+	return w.db.DriverName()
+}
+
+func (w *WrappedDB) Rebind(s string) string {
+	return w.db.Rebind(s)
+}
+
+func (w *WrappedDB) BindNamed(s string, i interface{}) (string, []any, error) {
+	return w.db.BindNamed(s, i)
+}
+
+func (w *WrappedDB) QueryContext(ctx context.Context, query string, args ...any) (rows *sql.Rows, err error) {
+	err = w.hook(ctx, w.lggr, func(ctx context.Context) (err error) {
+		rows, err = w.db.QueryContext(ctx, query, args...) //nolint
+		return
+	}, query, args...)
+	return
+}
+
+func (w *WrappedDB) QueryxContext(ctx context.Context, query string, args ...any) (rows *sqlx.Rows, err error) {
+	err = w.hook(ctx, w.lggr, func(ctx context.Context) (err error) {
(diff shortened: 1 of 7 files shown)


pkg/sqlutil/log.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+var promSQLQueryTime = promauto.NewHistogram(prometheus.HistogramOpts{
+	Name:    "sql_query_timeout_percent",
+	Help:    "SQL query time as a percentage of timeout.",
+	Buckets: []float64{10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120},
```
patrickhuie19, line 24: For my understanding: why stop/end at 120?

jmank88: The response to context cancellation is generally very fast, so we don't see values beyond this point very often :shrug: 
Or at least we didn't when I calibrated this scale for the original one that I copied here. Could double check in the grafana stats.

pkg/sqlutil/log.go
```diff
@@ -0,0 +1,227 @@
+package sqlutil
```
patrickhuie19, line 1: This PR made me think of `core/services/pg/sqlx.go` - did you imagine we would consolidate the `Queryer` interface there into the one in this package?

jmank88: Yeah, this is meant to replace it.

samsondav (approved):

patrickhuie19 (dismissed):

pkg/sqlutil/interceptor.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+	"database/sql"
+
+	"github.com/jmoiron/sqlx"
+)
+
+var _ Queryer = &InterceptedQueryer{}
+
+type Interceptor func(ctx context.Context, do func(context.Context) error, query string, args ...any) error
```
patrickhuie19, line 12: Does the `database/sql` package or `sqlx` package provide an Interceptor type + APIs we can use? Seems like we aren't the only ones who would want this functionality as well.

jmank88: I had the same thought but couldn't find one for sqlx

DylanTinianov (dismissed):

pkg/sqlutil/hook.go
```diff
@@ -0,0 +1,123 @@
...
+// HookingQueryer is a Queryer which invokes a QueryHook on each call.
+type HookingQueryer struct {
+	q    Queryer
+	lggr logger.Logger
+	hook QueryHook
+}
+
+// An QueryHook is a func that is executed for each query, providing an opportunity to measure, log, inspect/modify errors, etc.
+// The do func *must* be called.
+// Logs emitted through the provided logger.Logger will have the caller's line info.
```
jmank88, line 23: I ended up adding a logger to get  the caller info right.

jmank88:
Updated naming, added tests for caller line info.

pkg/sqlutil/monitor.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+const slowMsg = "SLOW SQL QUERY"
+
+var promSQLQueryTime = promauto.NewHistogram(prometheus.HistogramOpts{
+	Name:    "sql_query_timeout_percent",
```
jmank88, line 23: I don't think we will be able to swap this in for everything at once in core, so we'll have to find a way to share this prom metric during the transition. Maybe temporarily exporting is the simplest thing to do :thinking:

pkg/sqlutil/hook.go
```diff
@@ -0,0 +1,122 @@
...
+// NewHookingQueryer returns a new [HookingQueryer] that calls each [QueryHook] in the provided order.
+func NewHookingQueryer(q Queryer, l logger.Logger, hs ...QueryHook) *HookingQueryer {
+	iq := HookingQueryer{q: q,
+		lggr: logger.Helper(l, 2), // skip our own wrapper and one interceptor
+		hook: noopHook,
+	}
+	switch len(hs) {
+	case 0:
+		return &iq
+	case 1:
```
patrickhuie19, line 37: /nit `case default` ?

samsondav (dismissed):

pkg/sqlutil/hook.go
```diff
@@ -0,0 +1,122 @@
...
+		iq.hook = hs[0]
+		return &iq
+	}
+
+	for i := len(hs) - 1; i >= 0; i-- {
```
krehermann, line 42: please add a comment to guide future readers. i can't figure out what this is doing after 3 minutes of reading it.

jmank88: Will do. It is chaining them by exploiting the `do func` that they are already expected to call. The tricky part from getting the logger skip levels right.

pkg/sqlutil/hook_test.go
```diff
@@ -0,0 +1,124 @@
...
+	getDur = 10 * time.Millisecond
+	selDur = 200 * time.Millisecond
+)
+
+func TestNewInterceptedQueryer(t *testing.T) {
+	lggr, ol := logger.TestObserved(t, zapcore.InfoLevel)
+	var q Queryer = &queryer{}
+	var sentinelErr = errors.New("intercepted error")
+	const fakeError = "fake warning"
+	q = NewHookingQueryer(q, lggr, TimeoutHook(selDur/2), noopHook, MonitorHook(func() bool { return true }), noopHook, func(ctx context.Context, lggr logger.Logger, do func(context.Context) error, query string, args ...any) error {
```
krehermann, line 31: does this test validate the order of the hook executions? if so, can you tell me where that is happening?

jmank88: It does insofar as it wouldn't work the same if they were reversed, but I'll try to find a way to make it more clear.

DylanTinianov (dismissed):
Looks good

samsondav (approved):

pkg/sqlutil/monitor.go
```diff
@@ -0,0 +1,147 @@
...
+// PromSQLQueryTime is exported temporarily while transitioning the core ORMs.
+var PromSQLQueryTime = promauto.NewHistogram(prometheus.HistogramOpts{
+	Name:    "sql_query_timeout_percent",
+	Help:    "SQL query time as a percentage of timeout.",
+	Buckets: []float64{10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120},
+})
+
+// MonitorHook returns a [QueryHook] that measures the timing of each query and logs about slow queries at increasing levels of severity.
+// When logAll returns true, every query and error will be logged.
+func MonitorHook(logAll func() bool) QueryHook {
```
patrickhuie19, line 31: Could `MonitorHook` accept a straight `bool` arg?

patrickhuie19 (approved):

pkg/sqlutil/monitor.go
jmank88: No, it is one of the few settings that can be changed at runtime, so we call this every time.

PR #397, opened 2024-03-08
[chore] Some more improvements; follow up on review comments

cedric-cordenier:


pkg/values/list.go
```diff
@@ -34,39 +34,46 @@ func (l *List) proto() *pb.Value {
...
+			if err != nil {
+				return err
+			}
+			newList = reflect.Append(newList, reflect.Indirect(reflect.ValueOf(ptrEl)))
+		}
+		reflect.Indirect(val).Set(newList)
+		return nil
+	default:
+		dl := []any{}
+		err := l.UnwrapTo(&dl)
```
bolekk, line 67: I don't understand this - if "to" is not a slice, why are we still trying to unwrap rather than fail?

bolekk (approved):
Left one question in a comment.

pkg/values/list.go
cedric-cordenier: This handles the case where you pass in an uninstantiated `any` pointer.

Without that this test case fails: https://github.com/smartcontractkit/chainlink-common/pull/397/files#diff-4c5e50901d7b6e0f0668549fa0a61acb044d57a42a82e2020fbaf601934afbecR67

PR #398, opened 2024-03-11
Chain reader keys querying Alternative DSL definition (OUTDATED)

ilija42:


pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -9,9 +9,112 @@ import "google/protobuf/empty.proto";
...
+}
+
+message LimitAndSort{
+  uint64 limit = 1;
+  repeated SortBy sort_by = 2;
+}
+
+message QueryKeysFilter {
+  oneof filter {
+    AndFilter and_filter = 1;
```
krehermann, line 100: i'm i reading this correctly: it's an infinite recursion?

QueryKeysFilter -> AndFilter -> QueryKeysFilter -> ...

can you remove the AndFilter and have something like ~QueryKeysFilter {
// repeated filters are AND'd
repeated filters {
	oneof  {
	AddressFilter
	...
	}
	
	not sure what the syntax is supposed to be

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -9,9 +9,112 @@ import "google/protobuf/empty.proto";
...
+message Confirmations {
+  int32 confirmation = 1;
+}
+
+message ConfirmationsFilter{
+  Confirmations confirmations = 1;
+}
+
+message ComparisonOperator {
+  int32 comparisonOperator = 1;
```
krehermann, line 51: what is this? should it be an enum of things like EQ, NOTEQ LT, GT, LTE, GTE?

ilija42: Yeah it should, just haven't done that yet

pkg/loop/internal/pb/chain_reader.proto
ilija42: It just means that we can have nested ands. This doesn't enter recursion anywhere or break anything



`

pkg/loop/internal/pb/chain_reader.proto
ilija42: added now

pkg/loop/internal/pb/chain_reader.proto
krehermann: it's going to be hard to use and understand because it's mixing primitive operators and derived concepts into the same thing.

ilija42: I've looked at some orms, and this is similar to how it usually works, except that I'm not allowing arbitrary data to be used with conditions(=,>=,!=...) but predefined conditions(block,timestamp ...)

krehermann: ok, i think i get the idea. i would have been tempted to do ~

```
message QueryFilter {
repeated Expr expr
}


message Expr {
oneof evualator {
Primitive primitive
Composite composite}
}

// first class primitives for comparing chain data
message Primitive {
	oneof compatior {
	AddressFilter address_filter = 3; // hard coded to check for equality
    ConfirmationsFilter confirmations_filter = 4;
    BlockFilter block_filter = 5; // allows comparision with eq, neq,...
    TimestampFilter timestamp_filter = 6;
    TxHashFilter tx_hash_filter = 7;
}
}

//composite enables us to chain  exprs
message Composite {
 oneof comparitor {
 And and
 Or or
}
}

// and checks the both exprs are true
message And {
  Expr arg1
  Expr arg2
}

// or checks at either expr is true
message Or {
 Expr arg1
 Expr arg2
}
```

i believe this expresses the same thing as what you have. it's easier for me to think it terms of these seperated building blocks. i think it would read in more or less the same way you've written the example go code

ilija42: Its really hard to make this intuitive and also parse-able at the end destination.

I provided an alternative here https://github.com/smartcontractkit/chainlink-common/commit/73ef07d6a97e0f91a73109f12afb1ffcf85d6dd5

ilija42: It all ends up looking very similar from the perspective of the end user using functions to build out the query.

pkg/types/chain_reader.go
```diff
@@ -51,21 +68,212 @@ type BoundContract struct {
...
-	BlockHash         string
-	BlockNumber       int64
-	BlockTimestamp    time.Time
-	CreatedAt         time.Time
-	Keys              []string
-	Data              []byte
+type Sequence struct {
+	// TODO SequenceCursor, this should be a unique sequence identifier that chain reader impl. understands.
+	// This way we can retrieve past/future sequences (EVM log events) very granularly but still hiding the chain detail.
+	SequenceCursor string
```
EasterTheBunny, line 74: Will the `SequenceCursor` be required to be unique? Is it made of a block number or a combination of data?

pkg/types/chain_reader.go
```diff
@@ -51,21 +68,212 @@ type BoundContract struct {
...
+
+func (f *TxHashFilter) Accept(visitor Visitor) {
+	visitor.VisitTxHashFilter(*f)
 }
 
-type EventFilter struct {
-	AddressList []string   // contract address
-	KeysList    [][]string // 2D list of indexed search keys, outer dim = AND, inner dim = OR.  Params[0] is the name of the event (or "event type"), rest are any narrowing parameters that may help further specify the event
-	Retention   time.Duration
+type Visitor interface {
```
EasterTheBunny, line 272: Is the expectation that a `Visitor` is some arbitrary filter implementation? Does this introduce chain-specific code on the product side?

pkg/types/chain_reader.go
```diff
@@ -41,8 +41,25 @@ type ChainReader interface {
...
+
+	//TODO Rebind binding address
+	//ReBind(ctx context.Context, name, address string)
+	// TODO some filters have to be dynamic, so this has to override chain reader bind that comes from config?
+	// Maube this should just be Bind?
+	// RegisterFilter()
+	// UnRegisterFilter()
+
+	QueryKey(ctx context.Context, keys string, queryFilters []QueryFilter, limitAndSort LimitAndSort, sequenceDataType any) ([]Sequence, error)
+	QueryKeys(ctx context.Context, keys []string, queryFilters []QueryFilter, limitAndSort LimitAndSort, sequenceDataTypes []any) ([][]Sequence, error)
```
EasterTheBunny, line 55: I am initially confused on the return type here being a multi-dimensional slice. Of the existing log poller functions, none of them have this return type. They return either a single log value or a slice of values.

pkg/types/chain_reader.go
ilija42: Visitor should have a chain specific implementation in the chain specific loop, but this won't be exposed to the product

pkg/types/chain_reader.go
ilija42: It will be unique since its a combination of different things that make a sequence(log event) unique on that chain

pkg/types/chain_reader.go
ilija42: Since the caller can't differentiate between sequences they have to do it by relying on dimensions, because they can know to which dimension of key and sequenceDataTypes it belongs to

PR #399, opened 2024-03-11
Add null int64

dimriou:


pkg/utils/null/int64.go
jmank88, line 1: This looks like it belongs in `package sqlutil` or `package config`. Is it the first null value?

Another option is the new generic std lib type: https://pkg.go.dev/database/sql#Null

jmank88: I would think that the DB marshalling would be fine, but we could double check for JSON issues with a test.

dimriou:
Closed since we can utilize `database/sql` package

PR #402, opened 2024-03-14
Plugin config validation service

george-dorin:
Add a validation service that we call to validate plugin config
Chainlink PR: https://github.com/smartcontractkit/chainlink/pull/12430


pkg/loop/internal/test/types/interfaces.go
```diff
@@ -75,6 +75,12 @@ type ErrorLogEvaluator interface {
 	Evaluator[types.ErrorLog]
 }
 
+// ValidationEvaluator is a helper interface for testing ErrorLogs
```
cedric-cordenier, line 78: @george-dorin I think your comment is mistaken here :) ErrorLogs -> ValidationService?

pkg/loop/reportingplugins/grpc.go
```diff
@@ -73,7 +81,14 @@ func (g *GRPCService[T]) GRPCServer(broker *plugin.GRPCBroker, server *grpc.Serv
 		tc := internal.NewTelemetryClient(ts)
 		return g.PluginServer.NewReportingPluginFactory(ctx, cfg, provider, pr, tc, el)
 	}
-	return internal.RegisterReportingPluginServiceServer(server, broker, g.BrokerConfig, serverAdapter(adapter))
+
+	validateConfigFn := func(ctx context.Context) (types.ValidationService, error) {
+		return g.PluginServer.NewValidationService(ctx)
+	}
+	return internal.RegisterReportingPluginServiceServer(server, broker, g.BrokerConfig, serverAdapter{
```
cedric-cordenier, line 88: @george-dorin I don't think we should extend the serverAdapter for this; can we instead just embed the PluginServer?

All the adapter should do is translate between the providerConn based NRP function and the provider struct based one. All other methods should be passed through as it.

pkg/loop/internal/test/types/interfaces.go
george-dorin: Fixed in [`75ca4bb` (#402)](https://github.com/smartcontractkit/chainlink-common/pull/402/commits/75ca4bb87e3b7d6a714ebda797351f090c972082)

pkg/loop/reportingplugins/grpc.go
george-dorin: @cedric-cordenier not sure on what you mean when saying embedding the PluginServer, the `GRPCService` is shared between NRP and `NewValidationService`

pkg/capabilities/consensus/ocr3/validation_service.go
```diff
@@ -0,0 +1,38 @@
...
+
+var _ commontypes.ValidationService = (*validationService)(nil)
+
+type validationService struct {
+	lggr logger.Logger
+	services.StateMachine
+}
+
+func (v *validationService) ValidateConfig(ctx context.Context, config map[string]interface{}) error {
+	return nil
```
krehermann, line 19: why isn't this implemented?

krehermann (commented):
i'm missing a lot of context to review this PR. is there a ticket that describes the goal that i can read?

krehermann (commented):
I found this https://smartcontract-it.atlassian.net/browse/BCF-2777

AFAIK, it was an open question about how to do the validation WRT startup.

I don't understand what is implemented here and the corresponding change in core.

I would have expected static validation (for things like the path of the binary and whatever can be known upfront) and then good error handling, perhaps will explicit error types, on startup of the plugin.

as far as a i can tell we are starting the plugin twice - once for validation and once for really. if that is correct, then i don't like it.

it if passes validation great, all things are equal.

if it fails validation, we need good error handling in two cases instead of one. and, passing validation doesn't actually guarantee that the loopp with start a second time. nor are you guaranteed that if it fails validation it won't start later. these are facts are because of the port registration -- more generally, starting ( and as validation  as written), depend on ephemeral external resources. 

to summarize: i confused about this change and currently i don't believe it validates plugin as intended. instead i propose all dynamic dependencies are encapsulated with good error handling in the actual one-time startup of the loopp.

krehermann (changes requested):
forgot to click the request changes button. @cedric-cordenier @jmank88 @george-dorin can you read my last comment and tell me if you think that i'm way off base?

george-dorin:
@krehermann The reason for starting the plugin twice and doing it like this instead of a one-time start is that job validation and job starts are done in different places (before job creation, when starting job services), the goal of this validation is to:
a) do static checks like does the binary exists, and if the known configuration required to start the plugin is present and valid(OCR version, provider, etc), these are done before the binary is started
b) plugin-specific config, for example, a plugin could require multiple pipeline specs or any arbitrary, plugin-specific configuration. The core node does not know about these requirements so the solution is to start the binary and let it validate that the config is valid and all requirements are met.

Doing this inside the validation method has the benefit of not creating a job, if we did it when the LOOPP is started for real then we would have to delete the job if it's impropered configuration, this is what we have right now and it's confusing, the job is created successfully but you start seeing errors in the logs and you have just created a job that will never run because it does not have the required configuration fields

george-dorin:
@krehermann To summarise, this is more about checking if the plugin config inside the job spec is valid not necessarily about ensuring that the LOOPP will start, as you pointed out in your comment, this cannot be guaranteed as we depend on external resources

krehermann (commented):
Hmm. I see your point. And yet starting and stopping, unregistering and registering in the name of validataion, only to do it all again with the possibility of it failing is wrong.

Maybe we need notion of pool or something? you could optimistically start the loop (with some idle time out), use it to validate your thing, and ensure that this specific started instance is reused when the given job needs it. or maybe we need to expand the registry a bit to help us manage this kind of dependency?

george-dorin:
@krehermann could you expand on why starting and stopping a LOOPP and the register-unregister is bad, I don't get why it would be a problem, we free the Prometheus port when we unregister it. I can understand that it's not efficient but I don't get why it's a problem

krehermann:
Because you can pass validation and then fail to start. Thats not reasonable behavior.

krehermann (dismissed):
after offline discussion, we decided that my objections point to a lower level problem with our loopp management.

@george-dorin please add a comment in the validation code referencing this ticket https://smartcontract-it.atlassian.net/browse/BCF-3126

cedric-cordenier (approved):

PR #405, opened 2024-03-15
Implement Price Getter server/client and tests

krehermann:
This change adds the grpc layer for PriceRegistry and plumbs it into the ExecutionProvider.

The most significant change from other gRPC implementations is the `Close()` handling for the PriceRegistry. The PriceRegistry is the only existing CCIP interface that has an explicit `Close()`. This enables us to hook into the lifecycle of the spawned gRPC server and ensure that the server is terminated when the client calls Close(). This a specific instance and solution to the problem documented in BCF-3067. It provides a general pattern to use to resolve that issue.


pkg/loop/internal/ccip_execution.go
```diff
@@ -286,3 +299,25 @@ func (e *execProviderServer) NewOnRampReader(ctx context.Context, req *ccippb.Ne
...
+	// wrap the reader in a grpc server and serve it
+	priceRegistryHandler := ccipinternal.NewPriceRegistryGRPCServer(reader)
+	// the id is handle to the broker, we will need it on the other side to dial the resource
+	priceReaderID, spawnedServer, err := e.ServeNew("PriceRegistryReader", func(s *grpc.Server) {
+		ccippb.RegisterPriceRegistryReaderServer(s, priceRegistryHandler)
+	})
+	if err != nil {
+		return nil, err
+	}
+	// There is a chicken-and-egg problem here. Our broker is responsible for spawning the grpc server.
```
krehermann, line 317: @jmank88 please specific examine this logic and associated test in `price_registry_test.go`

pkg/loop/internal/ccip_execution.go
```diff
@@ -286,3 +299,25 @@ func (e *execProviderServer) NewOnRampReader(ctx context.Context, req *ccippb.Ne
...
+	// that server needs to be shutdown when the priceRegistry is closed. We don't have a handle to the
+	// grpc server until we after we have constructed the priceRegistry, so we can't configure the shutdown
+	// handler up front.
+	priceRegistryHandler.WithCloseHandler(func() error { return spawnedServer.Close() })
```
jmank88, line 321:
```suggestion
	priceRegistryHandler.WithCloseHandler(spawnedServer.Close)
```

pkg/loop/internal/ccip/test/price_registry_test.go
```diff
@@ -0,0 +1,129 @@
...
+	// that the server has stopped, which is what we care about.
+	require.NoError(t, client.Close(ctx), "failed to close client %T, %v", err, status.Code(err))
+	wg.Wait()
+}
+
+// roundTripPriceRegistryTests tests the round trip of the client<->server.
+// it should exercise all the methods of the client.
+// do not add client.Close to this test, test that from the driver test
+func roundTripPriceRegistryTests(ctx context.Context, t *testing.T, client *ccip.PriceRegistryGRPCClient) {
+	t.Parallel()
```
jmank88, line 85: This is a rather odd place for `t.Parallel()` to be called. Why not at the start of `TestPriceRegistryGRPC`?

pkg/loop/internal/ccip/price_registry.go
```diff
@@ -0,0 +1,354 @@
...
+
+//
+// PriceRegistryGRPCServer implementation
+//
+
+// Close implements ccippb.PriceRegistryReaderServer.
+func (p *PriceRegistryGRPCServer) Close(context.Context, *emptypb.Empty) (*emptypb.Empty, error) {
+	// TODO close this server and any resources it holds
+	// if i close the server, how do i make as response?
+	// close(p.done)
```
jmank88, line 142: Is this still relevant?

pkg/loop/internal/ccip/test/exec_provider.go
```diff
@@ -60,106 +65,169 @@ type staticExecProvider struct {
...
+	/*
+		otherOffRamp, err := other.NewOffRampReader(ctx, "ignored")
+		if err != nil {
+			return fmt.Errorf("failed to create other off ramp reader: %w", err)
+		}
+		err = s.offRampReader.Evaluate(ctx, otherOffRamp)
+		if err != nil {
+			return evaluationError{err: err, component: offRampComponent}
+		}
+	*/
```
jmank88, line 115: Restore?

pkg/loop/internal/ccip/test/exec_provider.go
```diff
@@ -60,106 +65,169 @@ type staticExecProvider struct {
...
 }
 
 // AssertEqual implements ExecProviderTester.
-func (s *staticExecProvider) AssertEqual(ctx context.Context, t *testing.T, other types.CCIPExecProvider) {
-	// TODO BCF-2979 other components of exec provider
-	assert.NoError(t, s.Evaluate(ctx, other))
-}
+func (s staticExecProvider) AssertEqual(ctx context.Context, t *testing.T, other types.CCIPExecProvider) {
+	t.Parallel()
+	t.Run("StaticExecProvider", func(t *testing.T) {
```
jmank88, line 194: Is the subtest meant to be parallel?
```suggestion
	t.Run("StaticExecProvider", func(t *testing.T) {
		t.Parallel()
```

pkg/loop/internal/ccip/test/exec_provider.go
krehermann: no, this will be a quick followup. it's not ready yet, i had an oversight earlier. i could delete it altogether, but expect to have it implemented in the next 1-2 days

pkg/loop/internal/ccip/price_registry.go
krehermann: no, good catch. thanks

pkg/loop/internal/ccip/test/price_registry_test.go
krehermann: oops. only the top level test should be parallel. fixed

pkg/loop/internal/ccip/test/exec_provider.go
krehermann: no, just a mistake. fixed, thanks

jmank88: This still looks the same to me?

krehermann: now it fixed. parallel wasn't supposed to be there. thought i had deleted it

jmank88 (approved):

PR #406, opened 2024-03-18
[BCF-3048] - Chain reader keys querying definition

ilija42:
Related PRs:
1. [Core Chain Reader Implementation](https://github.com/smartcontractkit/chainlink/pull/12486)
2. [DSL EVM Log Poller implementation(Merges into the PR above)](https://github.com/smartcontractkit/chainlink/pull/12793)


pkg/types/chain_reader.go
```diff
@@ -51,21 +75,260 @@ type BoundContract struct {
...
+
+func (f *TxHashFilter) Accept(visitor Visitor) {
+	visitor.VisitTxHashFilter(*f)
+}
+
+// Where eg. usage:
+// queryFilter, err := Where(
+//
+//		NewTxHashPrimitive("0xHash"),
+//		NewBooleanExpression("OR",
```
jmank88, line 303: Some helpers for `OR` & `AND` would be nice:
```suggestion
//		NewOr(
```
And external callers will have to include the package, so it might be worth declaring these in a different place, e.g. `package expr`?
```suggestion
//		expr.NewOr(
```
I think dropping `New` would be fine in that context too:
```suggestion
//		expr.Or(
```

Or maybe `package query`? e.g. `query.Expression`, `query.Or(`, `query.Filter`

pkg/types/chain_reader.go
```diff
@@ -51,21 +75,260 @@ type BoundContract struct {
...
+type Expression struct {
+	Primitive         Primitive
+	BooleanExpression BooleanExpression
+}
+
+func (expr Expression) IsPrimitive() bool {
+	return expr.Primitive != nil
+}
+
+type BooleanOperator int
```
jmank88, line 183: nit/ I think `Bool` is more go like?
```suggestion
type BoolOperator int
```

pkg/types/chain_reader.go
```diff
@@ -51,21 +75,260 @@ type BoundContract struct {
...
+type Head struct {
+	Number    uint64
+	Hash      []byte
+	Timestamp uint64
+}
+
+type Sequence struct {
+	// TODO SequenceCursor, this should be a unique sequence identifier that chain reader impl. understands.
+	// This way we can retrieve past/future sequences (EVM log events) very granularly, but still hide the chain detail.
+	SequenceCursor string
```
jmank88, line 89: Having `Sequence.Sequence*` is usually discouraged. WDYT about once of these?
```suggestion
	// TODO Cursor, this should be a unique sequence identifier that chain reader impl. understands.
	// This way we can retrieve past/future sequences (EVM log events) very granularly, but still hide the chain detail.
	Cursor string
```
```suggestion
	// TODO Cursor, this should be a unique sequence identifier that chain reader impl. understands.
	// This way we can retrieve past/future sequences (EVM log events) very granularly, but still hide the chain detail.
	Cursor SequenceCursor
```

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -9,9 +9,202 @@ import "google/protobuf/empty.proto";
 
 service ChainReader {
   rpc GetLatestValue (GetLatestValueRequest) returns (GetLatestValueReply) {}
+  rpc QueryKey (QueryKeyRequest) returns (QueryKeyReply) {}
+  rpc QueryKeys (QueryKeysRequest) returns (QueryKeysReply) {}
```
krehermann, line 13: do we need both the singular and pural to be exposed?

at least in the grpc layer, can't we expose the pural and, if needed elsewhere, adapt a signular use case to the pural?

that is, we could have golang API the for the singular and pural if needed, but only the plural here and shim the golang-grpc as needed.

pkg/types/chain_reader.go
```diff
@@ -41,8 +41,40 @@ type ChainReader interface {
...
-	// contract is not known by the ChainReader, or if the address is invalid
+	// contract is not known by the ChainReader, or if the Address is invalid
 	Bind(ctx context.Context, bindings []BoundContract) error
+	// TODO UnBind?, doesn't seem like its needed?
+
+	//TODO ReBindByKey, UnBindByKey, this is needed in some places where addresses change but the contract stays the same
+	// ReBindByKey(ctx context.Context, key, address string)
+	// UnBindByKey() or UnBindByKey(ctx context.Context, key, address string)
+
+	QueryKey(ctx context.Context, keys string, queryFilter QueryFilter, limitAndSort LimitAndSort, sequenceDataType any) ([]Sequence, error)
```
krehermann, line 52: keys -> key ?

pkg/types/codec.go
```diff
@@ -40,6 +40,7 @@ type TypeProvider interface {
 
 type ContractTypeProvider interface {
 	CreateContractType(contractName, itemType string, forEncoding bool) (any, error)
+	CreateContractTypeByKey(key string, forEncoding bool) (any, error)
```
nolag, line 43: What does a "key" look like?  I feel like `CreateContractType` should just call `CreateContractTypeByKey` using some character or something to combine the contractName and itemType

ilija42: Key can be anything. For EVM its one of these:
1. Event signature:  "0x..."
2. Event signature and one of the topics "0x-value" which for `QueryKeyByValue()` could indicate that we want all USDC transfer logs and we want to filter them based on their topic value
3. Event signature and EVM word bytes eg. "0x-word=1" or something similar which for `QueryKeyByValue()`  means that we are querying for USDC transfer logs and filtering them based on their word value. 

CreateContractType isn't always tied to an event, since it is also used for contract methods. We could map these to a key also, but I feel like it adds extra complexity

pkg/loop/internal/pb/chain_reader.proto
ilija42: Most of the use cases are singular, so we would have to have an adapter everywhere.

pkg/types/chain_reader.go
```diff
@@ -51,21 +83,260 @@ type BoundContract struct {
...
+)
+
+type SortBy interface {
+	GetDirection() SortDirection
+}
+
+type LimitAndSort struct {
+	SortBy []SortBy
+	Limit  uint64
+}
```
reductionista, line 127: I have a feeling we may want to expand this later to allow for more result modification options.  Maybe we should name it something more general like `OutputMods`, `ResultMods` or `ResultTransforms`, `SequenceTransforms`, `OutputOpts`, `ResultFormatOpts`, etc?

It could be anything that rearranges the results or filters them in some way.  For example, maybe you want to drop both the first 5 results and the last 5 results, like `Head` & `Tail` operations instead of just `Limit`.  Or what if you want only results that when sorted are at an odd or even index? Can't think of specific use cases for any of these off the top of my head, just seems like at some point we're bound to want to add something beyond `Sort` & `Limit`

pkg/types/chain_reader.go
```diff
@@ -51,21 +83,260 @@ type BoundContract struct {
...
+	return QueryFilter{expressions}, nil
 }
 
-type EventFilter struct {
-	AddressList []string   // contract address
-	KeysList    [][]string // 2D list of indexed search keys, outer dim = AND, inner dim = OR.  Params[0] is the name of the event (or "event type"), rest are any narrowing parameters that may help further specify the event
-	Retention   time.Duration
+type Visitor interface {
+	VisitAddressFilter(filter AddressFilter)
+	VisitBlockFilter(filter BlockFilter)
```
reductionista, line 338: Maybe I'm confused about how this works, but it seems to me that we need a `KeyFilter` here somewhere, analogous to `KeysList`.

Also, what happens to `Retention`?  We need both of those for `RegisterEventFilter()` or whatever it's going to be named now (`RegisterSequenceFilter`?)

pkg/types/chain_reader.go
reductionista: I don't have a problem with including plural and singular, but I would prefer we name these `QuerySequenceByKey`, `QuerySequenceByKeys`, etc.

(Or of course, `QueryEventsByKey`, etc. but I assume we've already banned the word "event" due to confusion with standard blockchain terminology.)

I guess it could also just be `QueryChainByKey(s)`, or simply `QueryByKey)s)`

reductionista: My reasoning for that:  usually a key is used to narrow down a query, but the thing you're actually querying for is some subset of the set of all things that might get returned. And thing thing you're querying is usually a container holding those things, like a database or a blockchain.

pkg/types/chain_reader.go
```diff
@@ -41,8 +41,40 @@ type ChainReader interface {
...
+	// TODO make EVM words map to a key and then do this through the QueryByEquality methods.
+	// GetCommitReportMatchingSeqNum()
+	// GetSendRequestsBetweenSeqNums()
+	// GetCommitReportGreaterThanSeqNum()
+}
+
+type ValueEquality struct {
+	Value    string
+	Operator ComparisonOperator
+}
```
reductionista, line 68: If `ComparisonOperator` can be anything (equality, inequality, negation, etc.) then why is it called `ValueEquality` and `QueryKeyValueEquality` rather than `ValueComparison` and `QueryKeyValueComparison`?

reductionista: Actually, I think the main suggestion I'd have for this whole interface is that it could be simplified a lot by just adding a primitive like this:

```
type KeyValuePrimitive struct {
	Key string
	Value string
	Operator ComparisonOperator
}
```
Where the effect of it is to say "KeyValueMap[Key] = Value" (for equality) or eg "KeyValueMap[Key] > Value" for Gt, etc.

If we add that, then I think there is no need to have 6 different versions of `QueryKey` there just needs to be one, or maybe two. We wouldn't need singular and plural for each though.

For example, if you want to use 2 different keys to filter the results, instead of using `QueryByKeysEquality` you'd just pass a filter expression to `QueryByKeys` (or whatever we decide to call it) that's a Bool AND type, with each of its children a `KeyValuePrimitive` enforcing that these two keys equal particular values.  Similarly, instead of using `QueryByKeyValuesIn` you could either AND together a `KeyValuePrimitive` for each value the key can match, or better yet just define another primitive like this:

```
type KeyValueInPrimitive struct {
	Key string
	Values []string
}
```
which would mean KeyValueMap[Key] = one of the values in the set of Values

It's late, and off the top of my head I'm not sure quite what to do with the other 2 methods (`QueryKey(s)`) but it seems like they could probably also be dealt with in the same way.

ilija42: 1. Fair point on naming
2. It can't be a primitive and be part of the Query Filter interface. I am keeping keys separate from QueryFilter on purpose for a couple of reasons: 
- Its impossible to query for multiple keys with the same QueryFIlter. From what I've seen the usual pattern for querying for multiple keys (multiple eventSigs) is to query for them in the same way (same time range, block range etc ...). If you put keys in the QueryFilter you cannot turn that into sensible db query because everything is jumbled up together. This is similar to how ORMs work, you provide what you are filtering for separately eg. `Where(a.Equal(x), b.Gte(y)...).Fetch(SomeData)`. So providing the key separately is necessary and is less complex.
- Another point is that it provides a better UX because it makes it is obvious that the key must be provided whereas if it was part of QueryFilter it can't be enforced at compile time.
3. Having separate Querying methods makes it way less complex on the implementation. Each of the methods serves a different purpose. QueryKey/s is just requesting sequences by keys and the other 4 are for requesting sequences by keys and providing a way to filter over data stored in the sequence. I can see merging `QueryByKeyValuesIn` and `QueryByKeysValuesEquality`, but couldn't think of a clean way to express that with proto

pkg/types/chain_reader.go
ilija42: I don't know what you mean by `KeysList`.  Keys are kept as a separate concept from QueryFilter. This is why there are multiple different QueryKey methods. I see key as a way to point to a result set which then gets filtered by QueryFilter. 
Lets give an evm example.
- QueryKey() would be used to say give me all sequences that belong to a "usdcTransferKey", which then is 
  narrowed down by `QueryFilter` block range, time range ...
- `QueryByKeyValuesIn` would be used to say give me all sequences where "usdcTransfer-value" is in [1000, 2000...], which returns all usdc transfer sequences where value is in xy, which then is narrowed down by `. 
  QueryFilter` block range, time range ...
- QueryByKeyValueEquality would be used to say give me all sequences where "usdcTransfer-value" is 
   [{500, >=, {1500, <=}] or just[{500, >=], which returns all usdc transfer sequences where value is in some range, 
    which then is narrowed down by `QueryFilter` block range, time range ...

ilija42: I can see `QueryByKeyValuesIn()` and `QueryByKeyValueEquality()` somehow being merged as they are doing a similar thing, but I haven't yet found a way to express it through proto in a clean way.

ilija42: I will add Retention, just haven't started on that yet. Retention would be tied to a key during binding/registering

pkg/types/chain_reader.go
ilija42: I can't think of any useful use case other than `LimitAndSort`, so thats why I called it like that.

nolag: I happen to agree with @ilija42 here.  It's difficult in SQL to drop the first five results (could do it with joins and some hackery).  I can't think of when you would want to drop head and tail during a query.  Dropping head usually is just continuing from a cursor, dropping tail is just a lower limit.

pkg/loop/internal/pb/chain_reader.proto
nolag: Plural can check incoming requests, if it's one it can call a different method internally.

pkg/types/codec.go
nolag: I don't follow the logic.  They need to define a contract that the event is on (contractName) and the event emitted (itemType) for us to know what they asking to encode or decode into.  How would having a different field that you filter on impact what the data is decoded into?

ilija42: contractName and itemType are derived from the key itself since the key has a mapping to an event binding [link](https://github.com/smartcontractkit/chainlink/blob/c466c600ae0d5c2b9379f513e9460e69355ababc/core/services/relay/evm/chain_reader.go#L289). This allows product to call QueryKey... by just providing the key and it still knowing how to decode the data.

ilija42: I am just saying that I don't think `CreateContractType` should call `CreateContractTypeByKey` since that would mean defining keys for methods which don't need keys in the first place

pkg/loop/internal/pb/chain_reader.proto
ilija42: What i wanted to say is that most of the use cases are singular which means that almost every call would use an adapter which I think would look ugly.
On the other hand if you make an adapter over singular for plural it could have performance implications since you aren't making a potentially optimised plural call, but multiple singular calls through an adapter.

pkg/types/codec.go
ilija42: Scratch that, I refreshed my memory on the evm impl. Yeah, what you said in the first comment makes sense

pkg/types/chain_reader.go
reductionista: > It's difficult in SQL to drop the first five results (could do it with joins and some hackery).

I thought we were trying to design this in a way that doesn't assume the cache will be a sql db... it could be in-memory storage, key-value storage, etc. right? For now, since we are using postgres on the backend it's a practical concern... but we could still implement it in the relay if there was a need, by dropping some results after the db query comes back, before it gets sent back over gRPC.  (I guess the product plugin itself could also filter the results after receiving them though, which wouldn't be too much different.)

> Dropping head usually is just continuing from a cursor, dropping tail is just a lower limit.

Ah, continuing from a cursor actually seems like it might be an important use though, isn't it?  If there are a large number of events, maybe the product wants to fetch them in batches of 1000 at a time or something. How would you say "give me the next 1000 events that match this filter"? If we implement it with cursors, then we'd have to introduce some sort of statefulness where you make one query and it sends you back a token (like the "fingerprints" in the Tron API) which can be used to request the next one.  If we implement it in terms of Head & Tail (allowing you to say something like "give me events 5000 - 6000 when sorted by this key" with the `SortAndLimit` param), that would allow us to keep the API stateless... which seems better, right?

reductionista: In postgresql, it's not terribly hard to have an upper and lower limit like that... you can use `ROW_NUMBER() OVER(ORDER BY key) as row_id` to generate a column in any query that numbers the row results, then use that as a subquery of a query with a  WHERE clause which filters by `row_id >= min AND row_id <= max`

reductionista: Admittedly, even if we add support for that, the name `SortAndLimit` isn't too bad... you're still sorting and limiting the results. I'm oikay with keeping it.

pkg/types/chain_reader.go
reductionista: > I merged QueryByKeyValuesIn() and QueryByKeyValueEquality() into QueryByKeyValuesComparison()

Nice, thanks!

> I don't know what you mean by KeysList

I'm referring to the field of `EventFilter` we defined in the CLIP:
https://github.com/smartcontractkit/CLIPs/tree/2023-06-31-chain-reader/clips/2023-08-25-chain-reader
```
type EventFilter struct {
    AddressList []string    // contract address
    KeysList    [][]string  // 2D list of indexed search keys, outer dim = AND, inner dim = OR.  Params[0] is the name of the event (or "event type"), rest are any narrowing parameters that may help further specify the event
    Retention   time.Duration
}

type EventQuery struct{
    FromBlock      BlockID // either a hash or a big int
    ToBlock        BlockID // either a hash or a big int
    Filter         EventFilter
    Pending        bool
}

type Event struct {
    ChainId             string
	EventIndexInBlock   string
	Address             string
    TxHash              string
    BlockHash           string
    BlockNumber         int64
    BlockTimestamp      time.Time
    CreatedAt           time.Time
    Keys                []string
    Data                []byte
}
```
In the original design, `EventQuery` was what gets passed to `QueryEvents()` and a list of `Event`'s was what gets returned.  `EventQuery` filtered by address, keys, and block numbers... where Key[0] would usually be the "event type" or "event sig" and the rest of the keys would be topics or indices, or whatever else is needed to narrow things down.

It seems like you've taken that single method and split it in two, so that `Key[0]` can be treated differently from the rest of the keys. The disadvantage of that is having two separate methods, one for filtering just on `Key[0]` and one for filtering on any subset of keys including `Key[0]`.  I'm not sure I see what the advantage is of splitting it in two, is it motivated by trying to make the interface more user friendly, or is there another reason behind it?

ilija42: It is motivated by the ux. If you merge it into one function then all of the complexity lives in the QueryFilter, which means that there are a lot of invalid QueryFilters that the caller can come up with. I originally tried to make it into one function but the impl. is a lot more complex and it is a worse ux for the end user.

pkg/types/chain_reader.go
ilija42: It doesn't assume what the underlying db impl. is, but if we know that it would mostly be SQL then it is important that something is hard to do with SQL.

I was imagining the cursor to be a mandatory unique identifier attached  to every sequence. This would be just a concat of all things that make it unique (blocknumber-hash-topicId-whatever). This cursor could be added as a `CursorFilter` to `QueryFilter` or to the `LimitAndSort`.  Functionally it would do the same thing where we can ask to continue from the cursor or limit to next/previous 1000 sequences.

pkg/types/chain_reader.go
ilija42: If you merge `QueryKey()` and `QueryByKeyValuesComparison()` into a single method then the key value filtering would have to live in `QueryFilter`. This means that we would have to pass a slice of `QueryFilter` to `QueryKeys` since each `QueryFilter` would become bound to a key.

ilija42: I am trying to figure out the best UX, but I think that pushing this complexity into the filter could be more intuitive than this

ilija42: After revisiting this I think it may make more sense to merge these into one function. Even if the UX is worse it makes sense to allow more granular filtering which makes the issue of having to deal with a slice of `query.FIlter` a necessity.

reductionista: I might be missing something here, but I don't see why we should need to pass a slice of filters. We should only be passing a single filter which is the root node of a tree of filters constructed with the DSL.

Each `KeyFilter` primitive would either be for a single key, or for a collection of key/value pairs.  I could see pros and cons for either of those approaches.  But either way, the way to aggregate different filters is not through slices but through trees of logical operations. That's already how it works for all the other primitives right? Why should it be different for the `KeyFilter` primitive?

reductionista: I know things have evolved somewhat from here, but this section of that Rationale doc I wrote has some pseudo-code an examples of the general idea I have in mind:
https://smartcontract-it.atlassian.net/wiki/spaces/BCF/pages/583729325/Rationale+Behind+ChainReader+Design

I'm not sure how different that is from what you have now. If it's completely different then don't worry about it, but if there is a way to adjust what you have to go more in this direction I think it might make for a cleaner interface.

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +77,142 @@ message BoundContract {
...
+}
+
+// Expression encapsulates a single unit of filtering logic, which can be a common blockchain primitive or a composite of boolean expressions. 
+// This allows for both simple and more complex nested expressions.
+message Expression {
+  oneof evaluator {
+    Primitive primitive = 1;
+    BooleanExpression boolean_expression = 2;
+  }
+}
```
reductionista, line 118: This is really nice by the way!

I had thought we'd have to resort to just a generic json blob to be able to serialize a DSL tree like this, but this is much better. Didn't realize protobuf was this powerful, fits our needs here perfectly, well done!

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +77,142 @@ message BoundContract {
...
+// Primitive defines the basic building blocks for filter conditions based around fundamental blockchain concepts.
+message Primitive {
+  oneof comparator {
+    Address address = 3;
+    Confirmations confirmations = 4;
+    Block block = 5;
+    Timestamp timestamp = 6;
+    TxHash tx_hash = 7;
+  }
+}
```
reductionista, line 174: I was trying to understand a different comment, where we were discussing `KeyLIst` and `KeyFilter` and it led me here.

I was picturing when we DSL-ified the original `QueryEvents` proposal in the CLIP with KeysList was that there would be another primitive in this list for key filtering, along with `Address`, `Confirmations`, `Block`, `Timestamp`, and `TxHash`.  I was mostly thinking it would just be `Key` but it could also be a`KeyList` if we wanted to handle automatically AND'ing together a list of keys in the same primitive.

Before I mentioned that you split off Key[0] as having special significance from the other keys, but another thing I'm noticing now is that all of the keys are treated different from the other fields you can filter on. Not sure if that's better or worse, but wanted to remind you in case you hadn't thought about it this way, that they don't have to be treated so differently. I'm not sure what the disadvantages may be, but it would make more sense to me to just be able to use keys in the DSL in the same way you use any other of these fields.

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +62,133 @@ message BoundContract {
   string name = 2;
   bool pending = 3;
```
reductionista, line 63: We still need to update this to be finalized or confidence, right?

ilija42: Yes, this will be in a separate PR to make this easier to review

ilija42: I started throwing everything into this PR and it became a mess

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -172,10 +229,249 @@ func (c *Server) Bind(ctx context.Context, bindings *pb.BindRequest) (*emptypb.E
...
+		return query.LimitAndSort{}, status.Errorf(codes.InvalidArgument, "Limit cursor and cursor direction must both be defined or undefined")
+	}
+
+	return query.NewLimitAndSort(query.CountLimit(limit.Count), sortByArr...), nil
+}
+
+func convertSequencesFromProto(pbSequences []*pb.Sequence, sequenceDataType any) ([]types.Sequence, error) {
+	var sequences []types.Sequence
+	for _, pbSequence := range pbSequences {
+		cpy := reflect.New(reflect.TypeOf(sequenceDataType).Elem()).Interface()
```
ilija42, line 461: We can also check if sequenceDataType is a non pointer and then just do reflect.New(sequenceDataType). I don't think that changes performance that much though. I ran some benchmark and this doesn't impact performance too much which was surprising.

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -9,23 +9,51 @@ import "google/protobuf/empty.proto";
...
+// Head is gRPC adapter for the Head struct [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Head].
+message Head {
+  string identifier = 1;
+  bytes hash = 2;
+  uint64 timestamp = 3;
+}
+
+// Head is gRPC adapter for the Head struct [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Sequence].
+message Sequence {
+  string sequence_cursor = 1;
```
krehermann, line 54: out of curiosity why is the cursor a string? is it like a nextPage token?

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -9,23 +9,51 @@ import "google/protobuf/empty.proto";
...
-  repeated BoundContract bindings = 1;
+// QueryKeyReply has return arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.QueryKey].
+message QueryKeyReply {
+  repeated Sequence sequences = 1;
+}
+
+// Head is gRPC adapter for the Head struct [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Head].
+message Head {
+  string identifier = 1;
+  bytes hash = 2;
```
krehermann, line 48: are there any assumptions in the backend about the length of this? if so, add a comment here about it.

for example in ccip, all the hashes are 32 bytes but since grpc doesn't constrain the length, we just document the expectation next to the definition.

pkg/types/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+
+type CursorDirection int32
+
+const (
+	Previous CursorDirection = iota
+	Following
+)
+
+type Limit struct {
+	Cursor          *string
```
jmank88, line 234: Is empty string meaningful? If not, could avoid the complications of pointers.

pkg/types/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+	CursorDirection *CursorDirection
```
jmank88, line 235: Is it important for this to be a pointer?

pkg/types/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+
+type SortBy interface {
+	GetDirection() SortDirection
+}
+
+type CursorDirection int32
+
+const (
+	Previous CursorDirection = iota
+	Following
```
jmank88, line 230: There kinds of values are typically prefixed or suffixed with the type name in some way to distinguish and group them.
```suggestion
	CursorPrevious CursorDirection = iota
	CursorFollowing
```

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -179,3 +234,242 @@ func getContractEncodedType(contractName, itemType string, possibleTypeProvider
...
+					Timestamp: primitive.Timestamp,
+					Operator:  pb.ComparisonOperator(primitive.Operator),
+				}}
+		case *query.TxHashPrimitive:
+			pbExpression.GetPrimitive().Primitive = &pb.Primitive_TxHash{
+				TxHash: &pb.TxHash{
+					TxHash: primitive.TxHash,
+				}}
+		default:
+			return nil, status.Errorf(codes.InvalidArgument, "Unknown expression type")
```
krehermann, line 290: add the type to the error message here for debugging ease

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -179,3 +234,242 @@ func getContractEncodedType(contractName, itemType string, possibleTypeProvider
...
+		case *pb.Primitive_Confirmations:
+			return query.Confirmation(query.ConfirmationLevel(primitive.Confirmations.Confirmations)), nil
+		case *pb.Primitive_Block:
+			return query.Block(primitive.Block.BlockNumber, query.ComparisonOperator(primitive.Block.Operator)), nil
+		case *pb.Primitive_TxHash:
+			return query.TxHash(primitive.TxHash.TxHash), nil
+		case *pb.Primitive_Timestamp:
+			return query.Timestamp(primitive.Timestamp.Timestamp, query.ComparisonOperator(primitive.Timestamp.Operator)), nil
+		default:
+			return query.Expression{}, status.Errorf(codes.InvalidArgument, "Unknown expression type")
```
krehermann, line 422: here too

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -179,3 +234,242 @@ func getContractEncodedType(contractName, itemType string, possibleTypeProvider
...
+		}
+	default:
+		return query.Expression{}, status.Errorf(codes.InvalidArgument, "Unknown expression type")
```
krehermann, line 425: and here

pkg/types/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+
+// ComparerPrimitive is used to filter over values that belong to key data.
+type ComparerPrimitive struct {
+	Name           string
+	ValueComparers []ValueComparer
+}
+
+// Comparer is used for filtering through specific key values.
+// e.g. of filtering for key that belongs to a token transfer by values: Comparer("transferValue", [{"150",LTE}, {"300",GTE}])
+func Comparer(name string, valueComparers ...ValueComparer) Expression {
```
jmank88, line 100: Should it technically be s/Comparer/Comparator?
```suggestion
func Comparator(name string, valueComparators ...ValueComparator) Expression {
```
Although an altogether different word might be nice :thinking:

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -179,3 +234,242 @@ func getContractEncodedType(contractName, itemType string, possibleTypeProvider
...
+	for _, sortBy := range limitAndSort.SortBy {
+		switch sort := sortBy.SortBy.(type) {
+		case *pb.SortBy_SortByTimestamp:
+			sortByArr = append(sortByArr, query.NewSortByTimestamp(query.SortDirection(sort.SortByTimestamp.GetSortDirection())))
+		case *pb.SortBy_SortByBlock:
+			sortByArr = append(sortByArr, query.NewSortByBlock(query.SortDirection(sort.SortByBlock.GetSortDirection())))
+		case *pb.SortBy_SortBySequence:
+			sortByArr = append(sortByArr, query.NewSortBySequence(query.SortDirection(sort.SortBySequence.GetSortDirection())))
+		default:
+			return query.LimitAndSort{}, status.Errorf(codes.InvalidArgument, "Unknown order by type")
```
krehermann, line 440: same

pkg/types/query/query.go
```diff
@@ -0,0 +1,294 @@
+package query
+
+import "fmt"
+
+// Visitor should have a per chain per db type implementation that converts primitives to db queries.
+type Visitor interface {
+	ComparerPrimitive(primitive ComparerPrimitive)
```
krehermann, line 7: nit/ if these are always going to be Primitives, you can drop the suffix and make a primitives pkg in this `query` dir

pkg/loop/internal/pb/chain_reader.proto
ilija42: EVM cursor would be "blockHash-txHash-logIndex" it will be used both as nextPage token, but also so that you can say give me 100 events that happened before this one

pkg/types/query/query.go
ilija42: Proto generates these as pointers so it was easier to cast and to have nil as uninitialized.
Also iota for `CursorDirection` default is 0, didn't know that I can start it from 1 and keep 0 as uninitialized.

Changed them to non pointers

pkg/types/query/query.go
ilija42: sgtm

pkg/types/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+//		 ==> `txHash = txHash AND (
+//									 block > startBlock AND block < endBlock
+//									 AND (
+//										 (timestamp > someTs1 And timestamp < otherTs1)
+//										 OR
+//										 (timestamp > someTs2 And timestamp < otherTs2)
+//									 )
+//							 )`
+//		if err != nil{return nil, err}
+//		QueryKey(key, queryFilter)...
```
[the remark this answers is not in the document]
ilija42: Formatted it better now. Examples in tests will exist, but will be on the relayer side. I thought it would be nice to have an example on hover since it may be confusing at first glance

pkg/types/query/query.go
ilija42: Changed to `Comparator`, don't have any fitting naming ideas that are better tbh

pkg/loop/internal/pb/chain_reader.proto
ilija42: All of details like these should get validated in the chain specific implementation since that is always the destination or origin. I don't think that there is a universal fixed length that can work here

pkg/types/query/query.go
```diff
@@ -0,0 +1,208 @@
...
+func CursorLimit(cursor string, cursorDirection CursorDirection, count uint64) Limit {
+	return Limit{
+		Cursor:          cursor,
+		CursorDirection: cursorDirection,
+		Count:           count,
+	}
+}
+
+func CountLimit(count uint64) Limit {
+	return Limit{Count: count}
```
jmank88, line 162: Without the pointer fields, do these constructors still add value? Or would directly initializing the structs at the calls sites be more clear?
```diff
-limit := query.CursorLimit("foo", query.CursorPrevious, 10)
+limit := query.Limit{Cursor: "foor", CursorDirection: query.CursorPrevious, Count: 10)
```
```diff
-limit := query.CountLimit(10)
+limit := query.Limit{Count: 10)
```

ilija42: Almost everything else is initialised via constructros, I am leaning towards leaving it like this for consitency

krehermann (dismissed):

pkg/types/chain_reader.go
```diff
@@ -40,9 +40,13 @@ type ChainReader interface {
 	// use across chains and contract versions.
 	// Similarly, when using a struct for returnVal, fields in the return value that are not on-chain will not be set.
 	GetLatestValue(ctx context.Context, contractName, method string, params, returnVal any) error
+
+	// Bind TODO BCF-3139 can't handle log poller filters,explore handling this through chain reader config.
 	// Bind will override current bindings for the same contract, if one has been set and will return an error if the
 	// contract is not known by the ChainReader, or if the Address is invalid
 	Bind(ctx context.Context, bindings []BoundContract) error
+
+	// QueryKey TODO BCF-3174 find a way to abstract EVM Log Poller IndexedLogsWithSigsExcluding, should be possible by adding a new filter.
```
krehermann, line 49: nit/ add a real description before the TODO

krehermann (approved):
exciting to see this progress. good work.

pkg/types/query/query.go
jmank88: Usually constructors are the exception. They also will become inflexible once we cut v1.0.0 of this repo FWIW :shrug:

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -9,23 +9,51 @@ import "google/protobuf/empty.proto";
...
+}
+
+// BindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind].
+message BindRequest {
+  repeated BoundContract bindings = 1;
+}
+
 // GetLatestValueReply has return arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.GetLatestValue].
 message GetLatestValueReply {
   VersionedBytes retVal = 1;
```
EasterTheBunny, line 37: Everything else appears to be using snake case. Maybe update this one too?

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +62,133 @@ message BoundContract {
...
+  repeated ValueComparator value_comparators = 2;
+}
+
+message Block{
+  uint64 block_number = 1;
+  ComparisonOperator operator = 2;
+}
+
+enum ConfirmationLevel {
+  Finalized   = 0;
```
EasterTheBunny, line 124: We changed the defaults for these. Should they or can they be changed here too to match? Not sure what can be done in protobufs with negative numbers.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
EasterTheBunny: Because we don't actually write to the `sequenceDataType` and only use it for typing future elements, it doesn't really matter if it comes in a pointer or not. I would say leave it alone but if we do normalize, normalize to a non-pointer.

EasterTheBunny (commented):
A few minor considerations, but otherwise it looks great.

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +62,133 @@ message BoundContract {
...
+
+message SortBySequence {
+  SortDirection sort_direction = 1;
+}
+
+message SortBy {
+  oneof sort_by {
+    SortByTimestamp sort_by_timestamp = 1;
+    SortByBlock sort_by_block = 2;
+    SortBySequence sort_by_sequence = 3;
```
nolag, line 186: Why not just make them all the same type, SortDirection and give them different names?

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +62,133 @@ message BoundContract {
...
+  string key = 1;
+  // Expressions are lightweight orm like DSL defined for filtering over common blockchain primitives.
+  repeated Expression expression = 2;
+}
+
+// Expression encapsulates a single unit of filtering logic, which can be a common blockchain primitive or a composite of boolean expressions. 
+// This allows for both simple and more complex nested expressions.
+message Expression {
+  oneof evaluator {
+    Primitive primitive = 1;
```
mateusz-sekara, line 85: Does it mean there will be a separate struct for filtering by topics/data words etc?

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,3 +62,133 @@ message BoundContract {
...
+enum SortDirection {
+  Asc = 0;
+  Desc = 1;
+}
+
+message SortByTimestamp {
+  SortDirection sort_direction = 1;
+}
+
+message SortByBlock {
```
mateusz-sekara, line 174: All of current LogPoller queries are sorted by block_number and log_index. Why don't we have that included? `log_index` is too specific to evm chain family?

pkg/types/query/query.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+}
+
+func Or(expressions ...Expression) Expression {
+	return Expression{
+		BoolExpression: BoolExpression{Expressions: expressions, BoolOperator: OR},
+	}
+}
+
+// Where is a helper function for building KeyFilter, eg. usage:
```
mateusz-sekara, line 96: 👏

mateusz-sekara (approved):
Couple minor comments, besides that it looks good. I would love to see some real Postgres queries generated by this tooling ;)

pkg/loop/internal/pb/chain_reader.proto
EasterTheBunny: That is correct. An example of filtering by data words:

```
type eventByWordFilter struct {
	EventSig       common.Hash
	WordIndex      uint8
	ValueComparers []primitives.ValueComparator
}

func (f *eventByWordFilter) Accept(visitor primitives.Visitor) {
	switch v := visitor.(type) {
	case *pgDSLParser:
		v.VisitEventByWordFilter(f)
	}
}
```

More detail is in the EVM implementation in the `chainlink` repo.

pkg/loop/internal/pb/chain_reader.proto
EasterTheBunny: I simplified the proto definition for sort by as a message with two properties `SortType` and `Direction` such that each are enums and the `SortBy` no longer has the `oneof` property type.

krehermann (approved):

mateusz-sekara (approved):

PR #407, opened 2024-03-18
Implement Gas Estimator server/client and tests

krehermann:
Adds the grpc layer for the gas estimators.

Sub divide follow up tasks to potentially simple the interface and wire it thru the OffRampReader


pkg/loop/internal/ccip/test/exec_gas_estimator_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// test the client
+	roundTripGasPriceEstimatorExecTests(ctx, t, client)
+}
+
+// roundTripGasPriceEstimatorExecTests tests the round trip of the client<->server.
+// it should exercise all the methods of the client.
+// do not add client.Close to this test, test that from the driver test
+func roundTripGasPriceEstimatorExecTests(ctx context.Context, t *testing.T, client *ccip.ExecGasEstimatorGRPCClient) {
+	t.Parallel()
```
jmank88, line 73: Stray `t.Parallel()`

pkg/loop/internal/ccip/test/commit_gas_estimator_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// test the client
+	roundTripGasPriceEstimatorCommitTests(ctx, t, client)
+}
+
+// roundTripGasPriceEstimatorCommitTests tests the round trip of the client<->server.
+// it should exercise all the methods of the client.
+// do not add client.Close to this test, test that from the driver test
+func roundTripGasPriceEstimatorCommitTests(ctx context.Context, t *testing.T, client *ccip.CommitGasEstimatorGRPCClient) {
+	t.Parallel()
```
jmank88, line 72: And here

PR #408, opened 2024-03-18
BCF-3073: add gas estimator service to off ramp

krehermann:
This change completes the OffRamp grpc implementation.

It solves and interesting problem: The OffRamp serves a GasEstimator, which in turn is a service.

The most straightforward way to handle this is to statically allocate the GasEstimator server when an OffRamp is created. Then, in order to hydrate the OffRamp interface we use a shared BrokerExt between the OffRamp client and OffRamp server. This enables us to interpret the requests and responses that are based on a GasEstimator service id into the backing implementations as needed.

In order to take this all, I need to write a very simple test Broker.


pkg/loop/internal/ccip/test/offramp_test.go
```diff
@@ -2,23 +2,219 @@ package test
...
+		assert.Equal(t, OffRampReader.getExecutionStateResponse, state)
+	})
+
+	t.Run("GetExecutionStateChangesBetweenSeqNums", func(t *testing.T) {
+		state, err := client.GetExecutionStateChangesBetweenSeqNums(ctx, OffRampReader.getExecutionStateChangesBetweenSeqNumsRequest.seqNumMin, OffRampReader.getExecutionStateChangesBetweenSeqNumsRequest.seqNumMax, OffRampReader.getExecutionStateChangesBetweenSeqNumsRequest.confirmations)
+		require.NoError(t, err)
+		if !reflect.DeepEqual(OffRampReader.getExecutionStateChangesBetweenSeqNumsResponse.executionStateChangedWithTxMeta, state) {
+			t.Errorf("expected %v, got %v", OffRampReader.getExecutionStateChangesBetweenSeqNumsResponse, state)
+		}
+		//assert.Equal(t, OffRampReader.getExecutionStateChangesBetweenSeqNumsResponse, state)
```
jmank88, line 182: Remove? Restore?

pkg/loop/internal/ccip/price_registry.go
```diff
@@ -40,7 +41,7 @@ type PriceRegistryGRPCServer struct {
 }
 
 func NewPriceRegistryGRPCServer(impl cciptypes.PriceRegistryReader) *PriceRegistryGRPCServer {
-	return &PriceRegistryGRPCServer{impl: impl}
+	return &PriceRegistryGRPCServer{impl: impl, onClose: emptyCloser}
```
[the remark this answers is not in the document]
krehermann: changed the pattern

pkg/loop/internal/ccip/offramp.go
```diff
@@ -211,8 +277,20 @@ func (o *OffRampReaderServer) ChangeConfig(ctx context.Context, req *ccippb.Chan
...
+// Close implements ccippb.OffRampReaderServer.
+func (o *OffRampReaderGRPCServer) Close(ctx context.Context, req *emptypb.Empty) (*emptypb.Empty, error) {
+	closeAll := func() error {
+		var err error
+		for _, f := range o.onClose {
+			err = errors.Join(err, f())
+		}
+		return err
+	}
+	return &emptypb.Empty{}, closeAll()
```
jmank88, line 289: Could you use a `services.MultiCloser`?

krehermann: yes, done

jmank88 (approved):

PR #409, opened 2024-03-19
ccip updates

dimkouv:
Summary:

- 3 New `OnRamp` methods.
- 1 New `CommitStore` method.
- 1 New field in `CommitOffchainConfig`.

### DIFF
--- a/pkg/types/ccip/onramp.go
+++ b/pkg/types/ccip/onramp.go
@@ -59,4 +59,14 @@ type OnRampReader interface {
 	Address() (Address, error)
 
 	GetDynamicConfig() (OnRampDynamicConfig, error)
+
+	// IsSourceChainHealthy returns true if the source chain is healthy.
+	IsSourceChainHealthy(ctx context.Context) (bool, error)
+
+	// IsSourceCursed returns true if the source chain is cursed. OnRamp communicates with the underlying RMN
+	// to verify if source chain was cursed or not.
+	IsSourceCursed(ctx context.Context) (bool, error)
+
+	// SourcePriceRegistryAddress returns the address of the current price registry configured on the onRamp.
+	SourcePriceRegistryAddress(ctx context.Context) (Address, error)
 }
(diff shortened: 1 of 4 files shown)


pkg/types/ccip/onramp.go
```diff
@@ -59,4 +59,14 @@ type OnRampReader interface {
 	Address() (Address, error)
 
 	GetDynamicConfig() (OnRampDynamicConfig, error)
+
+	// IsSourceChainHealthy returns true if the source chain is healthy.
+	IsSourceChainHealthy(ctx context.Context) (bool, error)
+
+	// IsSourceCursed returns true if the source chain is cursed. OnRamp communicates with the underlying RMN
+	// to verify if source chain was cursed or not.
+	IsSourceCursed(ctx context.Context) (bool, error)
```
jmank88, line 68: Why is RMN communication chain-specific?

pkg/types/ccip/onramp.go
```diff
@@ -59,4 +59,14 @@ type OnRampReader interface {
 	Address() (Address, error)
 
 	GetDynamicConfig() (OnRampDynamicConfig, error)
+
+	// IsSourceChainHealthy returns true if the source chain is healthy.
+	IsSourceChainHealthy(ctx context.Context) (bool, error)
```
jmank88, line 64: Is it important to say `SourceChain` here? Or is that already implied since we are interacting with an `OnRampReader`? e.g. `onRamp.IsHealthy(ctx)`, `sourceChainOnRamp.IsHealthy(ctx)`
```suggestion
	// IsHealthy returns true if the source chain is healthy.
	IsHealthy(ctx context.Context) (bool, error)
```

Same question for `IsDestChainHealthy` vs. `IsHealthy`. e.g. `commitStore.IsHealth(ctx)`, `destChainCommitStore.IsHealth(ctx)`

pkg/types/ccip/onramp.go
dimkouv: There's one RMN per chain

pkg/types/ccip/onramp.go
dimkouv: IMO it's not really important but I'm fine either way, the people that worked on this changes chose this names, I'm just migrating their changes to this repo.

krehermann (approved):

PR #413, opened 2024-03-20
BCF-3108: implement ccip token pool grpc layer

krehermann:
implement the grpc plumbing for the TokenPoolBatchedReader interface.

fix minor typos and consolidate some func calls and naming conventions while in the area


jmank88 (approved):

pkg/loop/internal/pb/ccip/exec_factory.proto
```diff
@@ -99,3 +99,16 @@ message NewTokenDataResponse {
     int32 token_data_reader_service_id = 1;
 }
 
+// NewTokenPoolBatchedReaderResponse is a gRPC adapter for the output arguments of
+// [github.com/smartcontractkit/chainlink-common/pkg/types/CCIPExecProvider.NewTokenPoolBatchedReader]
+// It contains the id of the created token pool batched reader service that was created
+message NewTokenPoolBatchedReaderResponse {
+    int32 token_pool_batched_reader_service_id = 1;
```
patrickhuie19, line 106: Is this value ever in danger of hitting the max for a signed int32?

krehermann: no, but it would be worthwhile to go through all proto defns ensure we are using the same type. i wouldn't be surprised if some use uint32 and iirc, our broker uses uint32.

low priority though, @jmank88 has a ticket to make our id's string based

PR #415, opened 2024-03-20
BCF-3067: ensure grpc server shutdown

krehermann:
This addresses all the outstanding instances of BCF-3067. 

The point is to ensure that grpc servers that are spawned by a provider (via interfaces like `Provider.NewXYZ`) are shutdown when clients call `Close`

Fix some ctx.TODOs, tidy up naming conventions and the interface declarations while in the area.


pkg/loop/internal/ccip_execution.go
```diff
@@ -360,18 +358,16 @@ func (e *execProviderServer) NewOnRampReader(ctx context.Context, req *ccippb.Ne
...
 	})
 	if err != nil {
 		return nil, err
 	}
-	// TODO BCF-3067 LEAKS!!!
-	// this dependency needs to be closed when the onramp reader is closed, which
-	// should happen when the calling reporting plugin is closed/goes out of scope
-	e.deps.Add(onRampResource)
+	// ensure the grpc server is closed when the onRamp is closed. See comment in NewPriceRegistryReader for more details
+	srv.AddDep(onRampResource)
```
krehermann, line 370: this is the meat of the change. same pattern is implemented for the all the other services already

jmank88 (approved):

PR #419, opened 2024-03-22
logical packages for internal/loop components

krehermann:
_Motivation_

the loop/internal pkg was a catch all. it forced all the Provider grpc implementations into the same package because they all shared private grpc servers such as the contract transmitter.

This change refactors into package structure

```
loop/internal/core                                         // core node api grpc wrappers; to be expand as needed
loop/internal/core/services
loop/internal/core/services/pipeline
loop/internal/core/services/capability
loop/internal/core/services/reportingplugin
loop/internal/core/services/reportingplugin/ocr3
loop/internal/core/services/reportingplugin/ocr2
loop/internal/core/services/telemetry
loop/internal/core/services/errorlog
loop/internal/relayer                                       // loopp relayer grpc wrappers
loop/internal/relayer/pluginprovider
loop/internal/relayer/pluginprovider/chainreader
loop/internal/relayer/pluginprovider/ocr2                   // common services like config tracker, offchain digester...
loop/internal/relayer/pluginprovider/ext                    // product specific provider extension; eventually phased out
loop/internal/relayer/pluginprovider/ext/mercury
loop/internal/relayer/pluginprovider/ext/mercury/v1
loop/internal/relayer/pluginprovider/ext/mercury/v2
loop/internal/relayer/pluginprovider/ext/mercury/v3
loop/internal/relayer/pluginprovider/ext/median
loop/internal/relayer/pluginprovider/ext/ccip
loop/internal/goplugin                                      // hashicorp goplugin wrappers; base layer for all loopps
loop/internal/reportingplugin                               // product specific grpc wrappers
loop/internal/reportingplugin/mercury
loop/internal/reportingplugin/median
loop/internal/reportingplugin/ccip
```

the two extention points for a new product are
`loop/internal/relayer/pluginprovider/ext`, which will house the custom `Provider`
and 
`loop/internal/reportingplugin` will with house the reporting plugin and product-specific implements of services needed to construct a factory. An example today is median/Datasource. In the future it might include a product-specific keyring implementation


pkg/loop/internal/chainreader/codec.go
krehermann, line 1: potentially codec should be its own pkg

pkg/loop/internal/core/api/error_log.go
krehermann, line 1: `api` is intended to house all the resource definitions provided by the core node api for loop devs

pkg/loop/internal/core/plugin.go
krehermann, line 1: i'm not sure if `core` is the best package for the `plugin, plugin_service, service` but i couldn't think of a better one @jmank88 wdyt?

pkg/loop/internal/relayer/providerext/ccip/ccip_execution.go
krehermann, line 1: one of the beautiful things about this change is that now all the custom provider details can be co-located

pkg/loop/internal/core/plugin.go
krehermann: maybe `goplugin`? and remove the incurring stutters. this thing are the base layer for all the hashi-corp client/server. i do not want to make `plugin` pkg because that word is way too overloaded. maybe `base`? other ideas?

pkg/loop/plugin_relayer.go
```diff
@@ -6,15 +6,17 @@ import (
 	"github.com/hashicorp/go-plugin"
 	"google.golang.org/grpc"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/net"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer"
+	internal "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer"
+	looptypes "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/types"
```
jmank88, line 12: Double import of `package relayer` as `internal`, which could be used for `internal/types`:
```suggestion
	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer"
	internal "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/types"
```

pkg/loop/plugin_mercury_test.go
```diff
@@ -9,8 +9,7 @@ import (
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop"
-	mercury_common_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/mercury/common/test"
-	mercury_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/mercury/common/test"
+	mercury_test "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer/providerext/mercury/test"
```
jmank88, line 12: Can we try to avoid underscores?
```suggestion
	mercurytest "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer/providerext/mercury/test"
```
https://go.dev/doc/effective_go#names

pkg/loop/internal/relayer/providerext/mercury/mercury_reporting.go
```diff
@@ -10,22 +10,23 @@ import (
...
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/net"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb"
 	mercurypb "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/mercury"
 )
 
-type mercuryPluginFactoryClient struct {
+type PluginFactoryClient struct {
 	*net.BrokerExt
-	*ServiceClient
+	*core.ServiceClient
```
jmank88, line 21: Should we unexport these fields since the struct is now exported?

pkg/loop/internal/relayer/providerext/mercury/mercury.go
```diff
@@ -8,41 +8,44 @@ import (
...
 	mercury_v2 "github.com/smartcontractkit/chainlink-common/pkg/types/mercury/v2"
 	mercury_v3 "github.com/smartcontractkit/chainlink-common/pkg/types/mercury/v3"
 )
 
-type MercuryAdapterClient struct {
-	*PluginClient
-	*ServiceClient
+type AdapterClient struct {
+	*core.PluginClient
+	*core.ServiceClient
```
jmank88, line 33: Unexport?

pkg/loop/internal/relayer/providerext/ccip/ccip_execution.go
```diff
@@ -139,30 +140,30 @@ func (r *ExecutionLOOPServer) NewExecutionFactory(ctx context.Context, request *
...
-	_ types.CCIPExecProvider = (*execProviderClient)(nil)
-	_ GRPCClientConn         = (*execProviderClient)(nil)
+	_ types.CCIPExecProvider = (*ExecProviderClient)(nil)
+	_ core.GRPCClientConn    = (*ExecProviderClient)(nil)
 )
 
-type execProviderClient struct {
-	*pluginProviderClient
+type ExecProviderClient struct {
+	*ocr2.PluginProviderClient
```
jmank88, line 148: Unexport? `BrokerExt` too?

pkg/loop/internal/ocr2/config.go
```diff
@@ -7,45 +7,47 @@ import (
...
-	_ types.ConfigProvider = (*configProviderClient)(nil)
-	_ GRPCClientConn       = (*configProviderClient)(nil)
+	_ types.ConfigProvider = (*ConfigProviderClient)(nil)
+	_ core.GRPCClientConn  = (*ConfigProviderClient)(nil)
 )
 
-type configProviderClient struct {
-	*ServiceClient
+type ConfigProviderClient struct {
+	*core.ServiceClient
```
jmank88, line 23: Unexport?

jmank88: Oh wait are all of these embedded to inherit methods? :thinking: Maybe no problem then, but still feels a bit odd.

pkg/loop/internal/relayer/providerext/mercury/mercury_reporting.go
krehermann: yes, there is cleanup work to unexport both fields, structs and funcs.

i would prefer to do this as a second pass and use the PR to align on the pkg names and organization

i filed a ticket here, which includes my reasoning for doing this as a second pass (which i would start immediately after this is mergd) https://smartcontract-it.atlassian.net/browse/BCF-3114

pkg/loop/plugin_mercury_test.go
krehermann: i have this ticket to make another cleanup of the tests https://smartcontract-it.atlassian.net/browse/BCF-3068

pkg/loop/internal/relayer/providerext/mercury/mercury_reporting.go
jmank88: I didn't notice these were all still internal too, so I'm not actually concerned and follow up sounds good :+1:

pkg/loop/internal/core/api/pipeline_runner.go
cedric-cordenier, line 1: @krehermann I wouldn't create a separate core/api package for these because the dividing line seems a bit arbitrary. Couldn't we say that OCR2/OCR3 and the relayer APIs are also part of the core/api? And if so, why aren't they in this package?

IMO just separating these three components into their own package is a better alternative even if the package is really small as a result.

krehermann: i moved the code around, inspired by this conversation. i choose something slightly different, make the top level dirs `core, relayer, reportingplugin` to mirror the logical separation of the runtime

jmank88 (approved):

cedric-cordenier (approved):

george-dorin (approved):

PR #423, opened 2024-03-23
golangci-lint: add depguard

jmank88:


.golangci.yml
```diff
@@ -80,6 +80,35 @@ linters-settings:
       # - name: confusing-results
       - name: bool-literal-in-expr
       - name: atomic
+  depguard:
+    rules:
+      main:
+        list-mode: lax
+        deny:
+          - pkg: cosmossdk.io/errors
```
krehermann, line 88: can we add pkgerrors?  i know it's everywhere in core, but we can keep it out of here

krehermann (approved):

PR #426, opened 2024-03-25
[fix] FeedID int -> string

cedric-cordenier:


pkg/capabilities/mercury/mercury.go
```diff
@@ -54,7 +54,7 @@ type ReportInfo struct {
 
 // TODO: fix this by adding support for uint64 in value.go
 type FeedReport struct {
-	FeedID               int64  `json:"feedId"`
+	FeedID               string `json:"feedId"`
```
patrick-dowell, line 57: What's the motivation behind this change by the way? I ask mainly because I have a task open to convert this to bytes[] which maybe we won't need to do if we use strings here instead?

https://smartcontract-it.atlassian.net/jira/software/c/projects/KS/boards/182?selectedIssue=KS-99

bolekk: String is used to represent a hex-encoded byte array.  Internally it might be better to use []byte or [32]byte, I'm not sure...

bolekk: If we decide to use a string here then we would need to add some validations (starting with "0x", correct number of bytes, all lowercase)

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -21,8 +21,9 @@ type aggregatorConfig struct {
 }
 
 type feedConfig struct {
-	Deviation decimal.Decimal
-	Heartbeat int
+	Deviation       decimal.Decimal `mapstructure:"-"`
+	Heartbeat       int
+	DeviationString string `mapstructure:"deviation"`
```
bolekk, line 26: What's the problem with Decimal? Values support it...

cedric-cordenier: Yaml doesn't :/

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -156,15 +157,20 @@ func ParseConfig(config values.Map) (aggregatorConfig, error) {
 		if err != nil {
```
cedric-cordenier, line 157: @bolekk line 155 just above this validates the feedID string.

pkg/capabilities/mercury/mercury.go
cedric-cordenier: Now using [32]byte wherever possible :)

pkg/capabilities/mercury/mercury.go
```diff
@@ -55,8 +69,8 @@ type ReportInfo struct {
 
 // TODO: fix this by adding support for uint64 in value.go
```
patrick-dowell, line 70: You can remove this by the way since FeedID was the field that was supposed to be uint

pkg/capabilities/mercury/mercury.go
```diff
@@ -94,15 +108,41 @@ func (m Codec) Unwrap(raw values.Value) (ReportSet, error) {
...
 				"price":     decimal.NewFromFloat(100.00),
 			},
 		},
 	)
 }
 
+// feedReport is a values-compatible version of <FeedReport>, where
+// <FeedID> is a string rather than a custom type.
+type feedReport struct {
+	FeedID               string `json:"feedId"`
```
patrick-dowell, line 122: Given that we need FeedID to be a string for the values framework, would it make sense to just have the field be a string in the FeedReport struct and convert from byte[] to string when creating the FeedReport? I have to check where we are using it but I assume we could convert it back to byte[] for things that need it to be a byte[] too.

patrick-dowell (commented):
It feels like this PR would be a lot simpler if we converted FeedID to a string instead of a byte[], mainly because we treat it like a string throughout this entire PR :)

But if that doesn't make sense, I'm more or less ok with approving as is. The change just looks a bit messy.

bolekk (commented):
Yeah it looks like using a string right away will be simpler. I'm fine either way.

pkg/capabilities/mercury/mercury.go
```diff
@@ -24,6 +30,11 @@ func (id FeedID) String() string {
 	return string(id)
 }
 
+func (id FeedID) Bytes() [32]byte {
+	b, _ := hex.DecodeString(string(id)[2:])
+	return [32]byte(b)
```
bolekk, line 35: This will panic it b is shorter than 32 bytes - is that ok or should we pad with zeros (only to prevent a panic)?

pkg/capabilities/mercury/mercury.go
```diff
@@ -24,6 +30,11 @@ func (id FeedID) String() string {
 	return string(id)
 }
 
+func (id FeedID) Bytes() [32]byte {
```
bolekk, line 33: Do you want to use FeedIDBytesLen in place of 32 everywhere?

bolekk (approved):
looks good with minor comments

PR #428, opened 2024-03-25
Add new method to monitor without Kafka exporter

augustbleeds:
From RTSP's request, for Starknet we are going to do monitoring without exporting to Kafka because the Kafka topic is unused. As long as we emit metrics to prometheus we have a good enough monitoring solution.

### DIFF
--- a/pkg/monitoring/config/parse.go
+++ b/pkg/monitoring/config/parse.go
@@ -21,6 +21,20 @@ func Parse() (Config, error) {
 	return cfg, err
 }
 
+func ParseWithoutKafka() (Config, error) {
+	cfg := Config{}
+
+	if err := parseEnvVars(&cfg); err != nil {
+		return cfg, err
+	}
+
+	applyDefaults(&cfg)
+
+	err := validateConfigWithoutKafka(cfg)
+
+	return cfg, err
+}
+
 func parseEnvVars(cfg *Config) error {
 	if value, isPresent := os.LookupEnv("KAFKA_BROKERS"); isPresent {
 		cfg.Kafka.Brokers = value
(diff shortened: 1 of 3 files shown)


pkg/monitoring/config/parse.go
```diff
@@ -21,6 +21,20 @@ func Parse() (Config, error) {
 	return cfg, err
 }
 
+func ParseWithoutKafka() (Config, error) {
```
krehermann, line 24: i'd refactor this so they share the same initialization helper and only apply different validation ~
~
```
func defaultCfg() Config, error {
cfg := Config{}

	if err := parseEnvVars(&cfg); err != nil {
		return cfg, err
	}

	applyDefaults(&cfg)
}

func Parse() Config, error {
 cfg, err := defaultCfg()
 if err != nil {
 	return Config{}, fmt.Errorf("failed to create default configuration: %w", err)
 	}
 	return cfg, validateConfig(cfg)
 	}
 	
 func ParseWithoutKafka error ...
 return cfg, validateConfigWithoutKafka
 ```

File: pkg/monitoring/config/parse.go [hunk 2]
```diff
@@ -105,6 +119,31 @@ func applyDefaults(cfg *Config) {
...
+			return fmt.Errorf("%s='%s' is not a valid URL: %w", envVarName, currentValue, err)
+		}
+	}
+
+	return nil
+}
+
 func validateConfig(cfg Config) error {
 	// Required config
 	for envVarName, currentValue := range map[string]string{
```
krehermann, line 149: similarly here, can this be rewritten in terms of the common (without kafka) validation and the extra (kafka-specific) validation?

~
```

func validateConfigWithKafka() {
 err := baseValidation(cfg)
 ...
 // kafka stuff
 }	
 
 ```
 
 where i've renamed validateConfigWithoutKafka to baseValidation

File: pkg/monitoring/monitor_test.go [hunk 3]
```diff
@@ -17,6 +17,40 @@ import (
...
+	_, err := NewMonitorPrometheusOnly(
+		ctx,
+		newNullLogger(),
+		chainConfig,
+		sourceFactory1,
+		sourceFactory2,
+		func(buf io.ReadCloser) ([]FeedConfig, error) { return []FeedConfig{}, nil },
+		func(buf io.ReadCloser) ([]NodeConfig, error) { return []NodeConfig{}, nil },
+	)
+	require.NoError(t, err)
```
topliceanu, line 51: I still think this test needs to make sure data goes through the system ok.

topliceanu (commented):
I'm confused as to why does this need to be in chainlink-common? 
Does this change need to be deployed for Solana and Starknet as well as Cosmos? 

If the answer is that we need this only for Cosmos, all you're doing in this PR is ignoring the Kafka configuration and removing the kafka exporter, that can be done in the Cosmos-specific monitor.

If the answer is "we need this in all monitors", then I agree with @krehermann, validation needs to be split up: ValidateBaseCfg, ValidatePrometheusCfg, ValidateKafkaCfg. Also, how about instead of `NewMonitorPrometheusOnly`, we stick with `NewMonitor` and supply the list of exporter names, something like:

```go
func NewMonitor(
	rootCtx context.Context,
	log Logger,
	chainConfig ChainConfig,
	envelopeSourceFactory SourceFactory,
	txResultsSourceFactory SourceFactory,
	feedsParser FeedsParser,
	nodesParser NodesParser,
        exporters []ExporterType, // <----------
) (*Monitor, error) {
```

Also, if we ditch kafka completely, an option is removing kafka from the monitor altoghether?!

augustbleeds:
@topliceanu good point, i didn't think of just creating the new monitor in the starknet repo. I'll close this PR and open a new one there. the only downside is I have to redefine the config parsing logic.

PR #430, opened 2024-03-25
BCF-3068 reorganize internal tests to match dir struct

krehermann:
This moves code out of `internal/test` into the newly minted individual packages.

It is the last of major outstanding internal refactoring work.


pkg/loop/internal/test/ocr2/plugin_provider/codec.go
krehermann, line 1: this was a dupe; there is already code in chainreader/test/codec to cover it

jmank88 (approved):

PR #431, opened 2024-03-25
BCF-3071: implement price getter grpc layer

krehermann:
Implements the price getter interface in service of the Commit Plugin


pkg/loop/internal/relayer/pluginprovider/ext/ccip/test/pricegetter_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func TestPriceGetterGRPC(t *testing.T) {
+	t.Parallel()
+	ctx := tests.Context(t)
+	// create a price registry server
+	port := freeport.GetOne(t)
+	addr := fmt.Sprintf("localhost:%d", port)
+	lis, err := net.Listen("tcp", addr)
+	require.NoError(t, err, "failed to listen on port %d", port)
+	t.Cleanup(func() { lis.Close() })
```
jmank88, line 42: Is there an error to check here? If not, then can shorten:
```suggestion
	t.Cleanup(lis.Close)
```

jmank88 (approved):

PR #436, opened 2024-03-27
add keyvalue store to the list of services received by generic plugins

ettec:
A few things I need to check with Cedric, hence in draft


pkg/loop/internal/core/services/keyvaluestore/keyvalue_store.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var _ types.KeyValueStore = (*Client)(nil)
+
+type Client struct {
+	grpc pb.KeyValueStoreClient
+}
+
+func (k Client) Store(ctx context.Context, key string, val any) error {
+	jsonVal, err := json.Marshal(val)
```
cedric-cordenier, line 22: @ettec Hmmm... this doesn't seem quite right as I'm sure you've figured out :) Could we instead make the underlying service just store arbitrary bytes --potentially with some validation that the bytes are valid JSON? That way the GRPC layer would be relatively simple and could just pass the bytes along.

pkg/types/keyvalue_store.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import "context"
+
+type KeyValueStore interface {
```
cedric-cordenier, line 5: @ettec As I said above, I would change the root interface to just store bytes, and maybe add some validation that the bytes are valid JSON.

pkg/loop/internal/pb/keyvalue_store.proto
```diff
@@ -0,0 +1,25 @@
...
+message StoreValueRequest {
+  string key = 1;
+  bytes value = 2;
+}
+
+message GetValueRequest {
+  string key = 1;
+}
+
+message GetValueResponse {
```
cedric-cordenier, line 18: @ettec Usually these names should mirror the method and add Request/Response, so:
- StoreRequest
- GetRequest
etc.

pkg/loop/internal/core/services/keyvaluestore/keyvalue_store.go
ettec: right, I was a little unsure about that, and whether the type passed in should maintain the interface as outlined in the orm_kv.go. or just have the bytes methods.  Question -  would it not be reasonable for the client of the store to want to store some arbitrary bytes, and perhaps we add a convenience method for the the case where they want to serialise the object via json?

pkg/loop/internal/pb/keyvalue_store.proto
ettec: sorted that,  renamed the methods to be a bit more verbose to avoid name clash with other services

PR #438, opened 2024-03-27
[KS-74] Make TriggerEvent struct generic

bolekk:
It's not specific to Mercury - all triggers will use it

### DIFF
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -75,6 +75,13 @@ type CapabilityRequest struct {
 	Inputs   *values.Map
 }
 
+type TriggerEvent struct {
+	TriggerType    string
+	ID             string
+	Timestamp      string
+	BatchedPayload map[string]any
+}
+
 type RegisterToWorkflowRequest struct {
 	Metadata RegistrationMetadata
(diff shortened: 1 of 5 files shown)


pkg/capabilities/capabilities.go
```diff
@@ -75,6 +75,13 @@ type CapabilityRequest struct {
 	Inputs   *values.Map
 }
 
+type TriggerEvent struct {
+	TriggerType    string
+	ID             string
+	Timestamp      string
+	BatchedPayload map[string]any
```
HenryNguyen5, line 82: Do we consider these always to be batched payload? Would it be more useful to consider this just a `payload` of type `map[string]any`? This would make it more generic, so that triggers can either accept batched payloads or single ones.

HenryNguyen5 (approved):

cedric-cordenier (approved):

pkg/capabilities/capabilities.go
bolekk: Yeah we can call it payload and have a default constant "batch" name (or rather "shard" name?) for single-item payloads. I can do that separately since Cedric merged this one already.

PR #439, opened 2024-03-28
add keyvalue store to the list of services received by generic plugins

ettec:
Please do not merge, review only - requires coordinated merge with chainlink repo


cedric-cordenier (dismissed):

pkg/loop/internal/core/services/keyvaluestore/keyvalue_store_test.go
```diff
@@ -0,0 +1,61 @@
+package keyvaluestore
+
+import (
+	"context"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+	"testing"
```
jmank88, line 9: Is the linter complaining about the unsorted imports? Typically we sort/group by std lib, third parties, smartcontractkit, then local repo/module. Configuring your IDE to use goimports with `-local github.com/smartcontractkit` does most of the work for you.

ettec: It looks to be passing the lint (though perhaps I should still update goimports?)  the failing build step looks like it is because I need to push the corresponding changes on the chainlink repo (?) - I understand I just need to ensure that both PRs are ready to go to minimise impact and then merge both as close together as possible, there is some issue on the chainlink PR build -  I'll look to unblock that shortly.

jmank88: > It looks to be passing the lint (though perhaps I should still update goimports?) 

Still worth updating. Maybe we have test files omitted from the linter, but we should update that. Many rules are always important.

pkg/loop/internal/core/services/keyvaluestore/keyvalue_store.go
```diff
@@ -0,0 +1,67 @@
+package keyvaluestore
```
cedric-cordenier, line 1: @ettec Nit: your package naming is stuttering a bit here. Alternatives could be keyvalue/store.go or similar.

pkg/loop/internal/pb/keyvalue_store.proto
```diff
@@ -0,0 +1,25 @@
...
+message GetValueForKeyRequest {
+  string key = 1;
+}
+
+message GetValueForKeyResponse {
+  bytes value = 1;
+}
+
+service KeyValueStore {
+  rpc StoreKeyValue(StoreKeyValueRequest) returns (google.protobuf.Empty);
```
cedric-cordenier, line 23: @ettec Not super keep on the lack of symmetry here between the GRPC methods and the actual interface. Doubly so since you're using Store to mean two different things.

Could we either shorten or try to match the interface as closely as possible?

eg. Store(StoreRequest)
Get(GetRequest)

Bonus points would be to rename Store everywhere with Put, but that's maybe veering towards a stylistic preference.

ettec: It was this originally but conflicted with other services 'GetRequest' message - hence the more verbose name

cedric-cordenier (approved):

pkg/loop/internal/pb/keyvalue_store.proto
cedric-cordenier: @ettec Test

PR #440, opened 2024-03-28
[Capabilities] Proto conversion helpers

bolekk:
Expose public helpers to convert between types and simplify some gRPC methods.

### DIFF
--- a/pkg/capabilities/pb/capabilities_helpers.go
+++ b/pkg/capabilities/pb/capabilities_helpers.go
@@ -1,5 +1,14 @@
 package pb
 
+import (
+	"errors"
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
 const (
 	CapabilityTypeUnknown   = CapabilityType_CAPABILITY_TYPE_UNKNOWN
 	CapabilityTypeTrigger   = CapabilityType_CAPABILITY_TYPE_TRIGGER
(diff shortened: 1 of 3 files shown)


pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -1,9 +1,101 @@
...
+	}
+	config := &values.Map{Underlying: map[string]values.Value{}}
+	if req.Config != nil {
+		config = req.Config
+	}
+	return &CapabilityRequest{
+		Metadata: &RequestMetadata{
+			WorkflowId:          req.Metadata.WorkflowID,
+			WorkflowExecutionId: req.Metadata.WorkflowExecutionID,
+		},
```
HenryNguyen5, line 37: What's the reasoning behind creating another struct here vs passing in the old one?

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -1,9 +1,101 @@
...
+
 const (
 	CapabilityTypeUnknown   = CapabilityType_CAPABILITY_TYPE_UNKNOWN
 	CapabilityTypeTrigger   = CapabilityType_CAPABILITY_TYPE_TRIGGER
 	CapabilityTypeAction    = CapabilityType_CAPABILITY_TYPE_ACTION
 	CapabilityTypeConsensus = CapabilityType_CAPABILITY_TYPE_CONSENSUS
 	CapabilityTypeTarget    = CapabilityType_CAPABILITY_TYPE_TARGET
 )
+
+func MarshalCapabilityRequest(req capabilities.CapabilityRequest) ([]byte, error) {
```
HenryNguyen5, line 20: Why do we choose to make these functions rather than making them struct methods?

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -1,9 +1,101 @@
...
+func CapabilityRequestFromProto(pr *CapabilityRequest) capabilities.CapabilityRequest {
+	md := pr.Metadata
+	config := values.FromProto(pr.Config)
+	inputs := values.FromProto(pr.Inputs)
+
+	return capabilities.CapabilityRequest{
+		Metadata: capabilities.RequestMetadata{
+			WorkflowID:          md.WorkflowId,
+			WorkflowExecutionID: md.WorkflowExecutionId,
+		},
```
HenryNguyen5, line 60: Could we direct assign here rather than make a new struct?

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -1,9 +1,101 @@
...
+
+func UnmarshalCapabilityRequest(raw []byte) (capabilities.CapabilityRequest, error) {
+	var cr CapabilityRequest
+	if err := proto.Unmarshal(raw, &cr); err != nil {
+		return capabilities.CapabilityRequest{}, err
+	}
+	return CapabilityRequestFromProto(&cr), nil
+}
+
+func CapabilityRequestFromProto(pr *CapabilityRequest) capabilities.CapabilityRequest {
```
HenryNguyen5, line 51: Nit: would be good if this was moved above `UnmarshalCapabilityRequest` to pair like functions together

pkg/capabilities/pb/capabilities_helpers.go
bolekk: Marshal could be a method but Unmarshal would be a little more awkward because you would need to create an empty object first and then it could still fail.

Functions are how json and proto do it.

pkg/capabilities/pb/capabilities_helpers.go
bolekk: There is no old one. One is proto-generated, one is not.

pkg/capabilities/pb/capabilities_helpers.go
bolekk: ok!

pkg/capabilities/pb/capabilities_helpers.go
bolekk: we can't - these are different structs even though they are named the same

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -1,9 +1,101 @@
...
 const (
 	CapabilityTypeUnknown   = CapabilityType_CAPABILITY_TYPE_UNKNOWN
 	CapabilityTypeTrigger   = CapabilityType_CAPABILITY_TYPE_TRIGGER
 	CapabilityTypeAction    = CapabilityType_CAPABILITY_TYPE_ACTION
 	CapabilityTypeConsensus = CapabilityType_CAPABILITY_TYPE_CONSENSUS
 	CapabilityTypeTarget    = CapabilityType_CAPABILITY_TYPE_TARGET
 )
+
+func MarshalCapabilityRequest(req capabilities.CapabilityRequest) ([]byte, error) {
+	return proto.Marshal(CapabilityRequestToProto(req))
```
DeividasK, line 21: Nit: I would inline `CapabilityRequestToProto(req)` here or move `proto.Marshal` into `CapabilityRequestToProto` (pref for former). Seems like the function split is unnecessary and makes it harder to read. Same for the other three functions below, i.e., I'd expect 4 functions total - Marshal/Unmarshal * Request/Response

DeividasK (approved):

pkg/capabilities/pb/capabilities_helpers.go
bolekk: ToProto/FromProto are used in capabilities.go after this refactor and they don't need un/marshaling to bytes. Marshal/Unmarshal will be useful in my next core PR.

PR #442, opened 2024-04-02
[KS-96] Remove old report structs in mercury capabilities

patrick-dowell:
This is some cleanup work coming out of https://smartcontract-it.atlassian.net/browse/KS-28. We just need to cleanup the old structs now that we’ve replaced them with new ones.

### DIFF
--- a/pkg/capabilities/mercury/mercury.go
+++ b/pkg/capabilities/mercury/mercury.go
@@ -4,10 +4,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"strings"
-	"time"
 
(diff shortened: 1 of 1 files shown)


cedric-cordenier:
@patrick-dowell I think Unwrap/Wrap are used by the OCR3 capability

patrick-dowell:
@cedric-cordenier this is definitely a WIP. I probably should have put this up in review. I wanted to see what got affected by removing the structs and the wrap/unwrap which depends on them. I'm looking through the errors and thinking through how to do this.

PR #444, opened 2024-04-02
Replace Pending boolean with Confidence

reductionista:
This flag was being used to determine whether to use latest block or finalized block for event reads (originally named Finalized in a draft of the CLIP, but renamed to Pending during review due to some miscommunications around its purpose).

Having it be a Confidence level from 0 to 10 allows for more granularity but still avoids products needing to pass a chain-specific # of confirmations.  Instead, the Confidence level should be translated into # of confirmations based on the chain, in the RelayConfig.

Choosing int32 instead of uint32 leaves the door open for special values like Confidence = -1 which could be used to represent the Pending block in the tx pool. (Presently this is not supported, and there are no known use cases, but better to be more flexible in case the need arises.)

Along with this, we should also start using it for method reads to make things more parallel between events & state, the distinction between which the product plugins are not supposed to know or care about.


jmank88:
Can you please open PRs for the other repos that will need changes so we can review the whole set at once?

reductionista:
> Can you please open PRs for the other repos that will need changes so we can review the whole set at once?

Ah, I forgot this was still open. I think @ilija42 is deciding on the exact approach, and might already be including changes to this effect (or something different?) in his present work.  @ilija42 should I just close this one? If it's still relevant, I can add a linked PR to chainlink repo.

ilija42 (changes requested):
We can include this now, but there is no need since there are still a lot of issues with this approach that need to be addressed. I don't think it makes sense to set confidence level for the whole contract, it needs to be per event and per method

PR #445, opened 2024-04-03
Capability validation API

HenryNguyen5:
Slimmed this down to just the validation struct, will add the integration into capabilities into a diff PR


pkg/capabilities/capabilities.go
```diff
@@ -105,6 +105,28 @@ type CallbackExecutable interface {
...
+	//  - `RegisterTrigger` and `UnregisterTrigger`
+	//
+	// They take in a request object that has a `Config` field that may conflate
+	// with the `Config` field within `Execute`.
+	GetRequestConfigJsonSchema() (*CapabilityResponse)
+	// ValidateRequestConfig(config *values.Map) error
+	// ValidateRequestInput(inputs *values.Map) error
+	// NOTE: Should we need this? Or should we just document this within
+	// the capability itself?
+	// ExampleOutput() *values.Map
```
DeividasK, line 127: We need this to validate input/output compatibility. We should be able to do something like
```
SrcCapability.Validate(DestCapability.ExampleOutput())
```

The above can catch some basic mismatches between something like
```
consensus:
  - type: offchain_reporting@1
    ref: evm_median
    inputs:
      observations:
        - triggers.report_data.outputs
    config:
      aggregation_method: data_feeds_2_0
      aggregation_config:
        123: # ETHUSD
          deviation: "0.005"
          heartbeat: 24h
        456: # LINKUSD
          deviation: "0.001"
          heartbeat: 24h
        789: # USDBTC
          deviation: "0.002"
          heartbeat: 6h
      encoder: EVM
      encoder_config:
          abi: "mercury_reports bytes[]"

targets:
  - type: write_polygon_mainnet@1
    inputs:
      report:
        - consensus.evm_median.outputs # SHOULD BE consensus.evm_median.outputs.report
    config:
      address: "0xaabbcc"
      method: "updateFeedValues(report bytes, role uint8)"
      params: [$(inputs.report), 1] # THIS IS AN ERROR, BECAUSE REPORT IS report.report
```

pkg/capabilities/capabilities.go
```diff
@@ -105,6 +105,28 @@ type CallbackExecutable interface {
 	Execute(ctx context.Context, callback chan<- CapabilityResponse, request CapabilityRequest) error
 }
 
+type Validatable interface {
```
DeividasK, line 108: Could this be on `BaseCapability`?

pkg/capabilities/capabilities.go
```diff
@@ -105,6 +105,28 @@ type CallbackExecutable interface {
...
+	// the config field against the schema.
+	//
+	// NOTE: What to do with method calls like:
+	//
+	//  - `RegisterToWorkflow` and `UnregisterFromWorkflow` 
+	//  - `RegisterTrigger` and `UnregisterTrigger`
+	//
+	// They take in a request object that has a `Config` field that may conflate
+	// with the `Config` field within `Execute`.
+	GetRequestConfigJsonSchema() (*CapabilityResponse)
```
DeividasK, line 122: I expected
```suggestion
	ValidateConfig() (any, error)
```

Not sure if any can be made into a generic that returns the capability-specific interface.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -92,10 +93,10 @@ func (o *capability) HealthReport() map[string]error {
 }
 
 type workflowConfig struct {
-	AggregationMethod string      `mapstructure:"aggregation_method"`
-	AggregationConfig *values.Map `mapstructure:"aggregation_config"`
-	Encoder           string      `mapstructure:"encoder"`
-	EncoderConfig     *values.Map `mapstructure:"encoder_config"`
+	AggregationMethod string      `mapstructure:"aggregation_method" json:"aggregation_method,omitempty"`
```
DeividasK, line 96: Why `omitempty`?

pkg/capabilities/pb/capabilities.proto
```diff
@@ -49,6 +49,7 @@ message UnregisterTriggerRequest {
 service TriggerExecutable {
   rpc RegisterTrigger (RegisterTriggerRequest) returns (google.protobuf.Empty) {}
   rpc UnregisterTrigger (UnregisterTriggerRequest) returns (google.protobuf.Empty) {}
+  rpc GetRequestConfigJsonSchema(google.protobuf.Empty) returns (CapabilityResponse) {}
```
DeividasK, line 52: Hmmm, I was thinking that this would be something that's internal to the capability, i.e., when `RegisterTrigger` is called, it (internally) calls `ValidateConfig` and returns an error if config is invalid, otherwise it stores the CapabilitySpecific config internally.

pkg/capabilities/capabilities.go
HenryNguyen5: Yes it should be on `BaseCapability`! But I was making incremental changes first. I didn't want to make a change across all capability types at the same time.

pkg/capabilities/consensus/ocr3/capability.go
HenryNguyen5: No particular reason, I can remove it

pkg/capabilities/pb/capabilities.proto
HenryNguyen5: I was originally thinking what you are proposing here too, but what I realized is that by exposing this via endpoint we can provide a much better DX.

With this endpoint, we can do something like creating a full JSON schema that can be used to lint a YAML workflow file.

HenryNguyen5: I also want to internally call `ValidateConfig` during entrypoints like `RegisterTrigger`, then we get both:
1. The ability to provide workflow authors with a JSON schema meant for linting
2. Capabilities always validate their config and inputs upon execution

pkg/capabilities/capabilities.go
HenryNguyen5: I was thinking of exposing another endpoint called `ValidateConfig` that returns nothing or a list of errors associated with the config.

pkg/capabilities/capabilities.go
HenryNguyen5: Do we expect this to be a common use case? For some reason I thought that was an exception rather than the norm. I was thinking that often when we're chaining capabilities we might be passing in some other inputs too.

pkg/capabilities/consensus/ocr3/capability_test.go
```diff
@@ -20,6 +23,15 @@ import (
 const workflowTestID = "consensus-workflow-test-id-1"
 const workflowExecutionTestID = "consensus-workflow-execution-test-id-1"
 
+var transformJSON = cmp.FilterValues(func(x, y []byte) bool {
```
HenryNguyen5, line 26: I ended up copy / pasting this multiple times... where should i put this if i want to have test helpers?

pkg/capabilities/capabilities.go
```diff
@@ -105,10 +105,33 @@ type CallbackExecutable interface {
...
+	// the config field against the schema.
+	//
+	// NOTE: What to do with method calls like:
+	//
+	//  - `RegisterToWorkflow` and `UnregisterFromWorkflow`
+	//  - `RegisterTrigger` and `UnregisterTrigger`
+	//
+	// They take in a request object that has a `Config` field that may conflate
+	// with the `Config` field within `Execute`.
+	GetRequestConfigJSONSchema() *CapabilityResponse
```
cedric-cordenier, line 122: @HenryNguyen5 No objection to using JSONSchema for validation, I think that's great, but I'm a bit concerned about the API here.

Since this is truly generic (i.e. all capabilities) will provide us with a JSONSchema to validate against, I think we can be very concrete with our API and not use something like a CapabilityResponse. Some concrete JSONSchema type would be fine 👍

pkg/values/map.go
```diff
@@ -12,6 +13,12 @@ type Map struct {
 	Underlying map[string]Value
 }
 
+func (m Map) JSONSchema() *jsonschema.Schema {
```
cedric-cordenier, line 16: @HenryNguyen5 Why is this needed?

pkg/capabilities/capabilities.go
cedric-cordenier: On your comment about multiple method calls, I think having a single schema per call is fine 👍  or alternatively having a single method which returns a map of method names to schemas.

pkg/capabilities/consensus/ocr3/capability_test.go
cedric-cordenier: If these are test helpers my preference is to leave them copied since I don't think it harms much.

However there is a test utils package in common IIRC, so you could put it there too.

pkg/capabilities/capabilities.go
cedric-cordenier: Ah, the name feels pretty clunky too!

Getters in golang usually omit the GetX prefix, so would rename this to InputsSchema(), and maybe add a ConfigSchema() for config fields.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -26,6 +26,7 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
 // This Trigger Service allows for the registration and deregistration of triggers. You can also send reports to the service.
 type MercuryTriggerService struct {
 	capabilities.CapabilityInfo
+	*capabilities.Validator
```
cedric-cordenier, line 29: Nit: if you make Validator a value type and add generics to it, then you can one line this addition into a fairly neat:

```
capabilities.Validator[mercury.FeedReport]
```

with no instantiation required.

HenryNguyen5: whoa, where can i read about this distinction / use case?

pkg/capabilities/consensus/ocr3/capability_test.go
HenryNguyen5: yeah they're test helpers. Okay, I'll keep copy pasting them for now then.

pkg/capabilities/capabilities.go
HenryNguyen5: Added.

go.mod
```diff
@@ -39,6 +39,13 @@ require (
...
 )
 
+require (
+	github.com/bahlo/generic-list-go v0.2.0 // indirect
+	github.com/buger/jsonparser v1.1.1 // indirect
+	github.com/mailru/easyjson v0.7.7 // indirect
+	github.com/wk8/go-ordered-map/v2 v2.1.8 // indirect
+)
+
 require (
```
jmank88, line 49: Please combine and tidy these `indirect` blocks:
```suggestion
```

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -92,10 +97,10 @@ func (o *capability) HealthReport() map[string]error {
 }
 
 type workflowConfig struct {
-	AggregationMethod string      `mapstructure:"aggregation_method"`
```
jmank88, line 95: Are we losing portability by using irregular struct tags like `jsonschema`?

Does the experimental v2 API meet our needs?
- https://pkg.go.dev/github.com/go-json-experiment/json
- https://github.com/golang/go/discussions/63397

HenryNguyen5: The `jsonschema` struct tag doesn't affect json (de)serialization, it instructs the jsonschema package on how to generate a JSON schema. Unfortunately that new API doesn't help us here

jmank88: Are the schemas a standard format, or specific to `jsonschema`?

HenryNguyen5: They are a standard https://json-schema.org/

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,8 +23,44 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
 
 const defaultTickerResolutionMs = 1000
 
+type MercuryTriggerOutput struct {
```
bolekk, line 26: This is the same as TriggerEvent, which is a generic struct to support all triggers, not only Mercury. Why do we need a separate one here?

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -29,9 +28,17 @@ var info = capabilities.MustNewCapabilityInfo(
 	"v1.0.0",
 )
 
+type CapabilityConfig = workflowConfig
```
cedric-cordenier, line 31: No need for the type alias here I think: just rename `workflowConfig` to `CapabilityConfig`

I think you can simplify the naming btw: CapabilityConfig -> Config; CapabilityInputs -> Inputs, etc

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -50,10 +57,13 @@ type capability struct {
 }
 
 var _ capabilityIface = (*capability)(nil)
+var _ capabilities.ConsensusCapability = (*capability)(nil)
```
cedric-cordenier, line 60: @HenryNguyen5 How come you added this? I think this should do the same job as capabilityIface (my intent was to loosely couple this to the consensus interface so we only pulled in what we needed)

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -50,10 +57,13 @@ type capability struct {
...
+var OCR3CapabilityValidator = capabilities.NewValidator[CapabilityConfig, CapabilityInputs, CapabilityOutputs](capabilities.ValidatorArgs{Info: info})
```
cedric-cordenier, line 61: Hmmmm... does this need to be public, and do we need to instantiate at the package level (i.e. can we just move it into newCapability)?

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -92,22 +102,14 @@ func (o *capability) HealthReport() map[string]error {
...
-	AggregationMethod string      `mapstructure:"aggregation_method"`
-	AggregationConfig *values.Map `mapstructure:"aggregation_config"`
-	Encoder           string      `mapstructure:"encoder"`
-	EncoderConfig     *values.Map `mapstructure:"encoder_config"`
-}
-
-func newWorkflowConfig() *workflowConfig {
-	return &workflowConfig{
-		EncoderConfig:     values.EmptyMap(),
-		AggregationConfig: values.EmptyMap(),
```
cedric-cordenier, line 104: @HenryNguyen5 Did anything break when you removed this? IIRC I added this to ensure the map was properly initialized (with its internal map[string]any)

pkg/capabilities/consensus/ocr3/capability_test.go
```diff
@@ -136,7 +175,12 @@ func TestOCR3Capability_Registration(t *testing.T) {
 	cp := newCapability(s, fc, 1*time.Second, mockEncoderFactory, lggr)
 	require.NoError(t, cp.Start(ctx))
 
-	config, err := values.NewMap(map[string]any{"aggregation_method": "data_feeds_2_0"})
+	config, err := values.NewMap(map[string]any{
+		"aggregation_method": "data_feeds_2_0",
+		"aggregation_config": map[string]any{},
```
cedric-cordenier, line 180: @HenryNguyen5 This is what caused me to add a values.EmptyMap -- is this guaranteed to be present because of the schema validation? (I assume so)

pkg/capabilities/generate_schemas/main.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/triggers"
+)
+
+func main() {
```
cedric-cordenier, line 11: Since this is a `cmd` can we put it in `cmd/capabilities/generate_schemas/main.go` or similar?

pkg/capabilities/pb/capabilities.proto
```diff
@@ -26,6 +26,7 @@ message CapabilityInfoReply {
   string version = 4;
 }
 
+
```
cedric-cordenier, line 29: Extra line :)

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -13,43 +16,56 @@ import (
...
-	feedTwo   = "0x2222222222222222222200000000000000000000000000000000000000000000"
-	feedThree = "0x3333333333333333333300000000000000000000000000000000000000000000"
-	feedFour  = "0x4444444444444444444400000000000000000000000000000000000000000000"
-	feedFive  = "0x5555555555555555555500000000000000000000000000000000000000000000"
-)
-
-func TestMercuryTrigger(t *testing.T) {
-	ts := NewMercuryTriggerService(100, logger.Nop())
-	ctx := tests.Context(t)
+func registeredTriggerHelper(
```
cedric-cordenier, line 19: @HenryNguyen5 Can we call this something more descriptive that says what it does, eg. registerTrigger?

pkg/capabilities/validate.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return Validator[Config, Input, Output]{
+		ValidatorArgs: args,
+		schemas:       make(map[string]string),
+	}
+}
+
+func (v *Validator[Config, Input, Output]) Schema() (string, error) {
+	type Combined struct {
```
cedric-cordenier, line 77: This (and CombinedWithoutInput below) can be private since they're declared inside a function.

pkg/capabilities/validate.go
```diff
@@ -0,0 +1,210 @@
...
+
+	schemas map[string]string
+}
+
+var uriPrefix = "https://github.com/smartcontractkit/chainlink/capabilities/"
+
+var _ Validatable = (*Validator[any, any, any])(nil)
+
+type ValidatorArgs struct {
+	Info CapabilityInfo
```
cedric-cordenier, line 30: @HenryNguyen5 It would be nice to not depend on CapabilityInfo if we can here. It looks like we only need it for the ID, is that correct? If so, maybe we can just pass in the ID?

pkg/capabilities/validate.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, err
+	}
+
+	err = jsonSchema.Validate(raw)
+	if err != nil {
+		return nil, err
+	}
+
+	var decodedValue *DecodedValue
+	decoder, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
```
cedric-cordenier, line 160: If we're decoding `value` into a struct here, we should try to use UnwrapTo if we can -- same thing with the custom decode hook: if we're not handling an edge case correctly, let's put it in the values library rather than here.

pkg/capabilities/validate.go
```diff
@@ -0,0 +1,210 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	err = decoder.Decode(raw)
+	return decodedValue, err
+}
+
+func schemaWith(reflector *jsonschema.Reflector, schemaType any, schemaCache map[string]string, key string, info CapabilityInfo) (string, error) {
+	if reflector == nil {
```
cedric-cordenier, line 172: @HenryNguyen5 Can you avoid this by making reflector a `jsonschema.Reflector` to guarantee it's not nil.

pkg/capabilities/consensus/ocr3/capability.go
HenryNguyen5: Surprisingly not? Unless somehow it'd not caught by the test suite we currently have. I made sure all tests were passing as I was refactoring

pkg/capabilities/consensus/ocr3/capability_test.go
HenryNguyen5: Yeah, its required to be specfied now, a YAML representation would be:
```yaml
aggregation_method: data_feeds_2_0
aggregation_config: {}
```

pkg/capabilities/generate_schemas/main.go
HenryNguyen5: Going to remove this file entirely, or rework it so that it regenerates all the test fixtures

pkg/capabilities/consensus/ocr3/capability.go
HenryNguyen5: Another reason for public exposure + package level is to have an easy way to generate all of the test fixtures by importing all of the validators in a single file, rather than have a generate script in each package. WDYT?

pkg/capabilities/triggers/mercury_trigger_test.go
HenryNguyen5: Yeah, this deserves some comments

pkg/capabilities/validate.go
HenryNguyen5: We use `ID` and `Description`. I figured it was good to pass in the entire struct since we can store a lot of useful metadata in the schema.

pkg/capabilities/validate.go
HenryNguyen5: Sure, could you explain why to me? I'm not knowledgable on when I should choose using `UnwrapTo` rather than just using mapstructure directly

pkg/capabilities/validate.go
HenryNguyen5: Would I just deref all of the args to it then?

pkg/capabilities/capabilities.go
```diff
@@ -106,6 +106,20 @@ type CallbackExecutable interface {
...
+	// This schema is used within registration calls, which is when the capability instance is created.
+	ConfigSchema() (string, error)
+	// InputSchema returns the JSON schema for the inputs of the capability.
+	// This schema is used within execution calls, which is when the capability instance is being invoked
+	InputSchema() (string, error)
+	// OutputSchema returns the JSON schema for the output of the capability.
+	// This schema is used to describe the return value of an execution call.
+	OutputSchema() (string, error)
+	// Schema is the combination of all of the above schemas. It is used to describe the capability as a whole.
+	Schema() (string, error)
```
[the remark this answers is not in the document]
HenryNguyen5: I can make it just `Schema`, then all of the other methods will be used just for internal validation inside capabilities.

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: No, not necessarily -- but in that case I think you don't need line 59, so I'd prefer we pick one and remove the other

pkg/capabilities/validate.go
cedric-cordenier: Yep, at the point at which this is called I'd expect you to be guaranteed of having a reflector

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: Yeah if you want to test things in isolation I think it's fine to leave it at the package level -- still not sure it needs to be public though? In the end we embed the validator into the capability, and I think it's nice to ensure that we only access it via there.

HenryNguyen5: The value I see exposing it publically would be to generate all test fixtures without having to create an instance of the capability first.

Maybe this isn't a big deal though, I was just thinking it was extra steps if some capabilities are only creatable via a `NewCapability` type function that didn't allow you to pass in empty args.

pkg/capabilities/consensus/ocr3/capability.go
HenryNguyen5: I'm confused since capabilityIface contains completely separate methods (used to implement a reporting plugin) whereas ConsensusCapability is to make sure that this struct really is a capability.

I removed the line I added for now. I'd love to understand why we don't enforce the intersection of the two interfaces against this struct.

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: Hmmm... fair point -- you can ignore this then 👍

cedric-cordenier (dismissed):

pkg/values/list.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 
+
```
cedric-cordenier, line 7: Could you remove this additional line, both here and in map.go @HenryNguyen5 ? Otherwise LGTM

bolekk (approved):

PR #450, opened 2024-04-04
pkg/sqlutil/pg: create package; expand env config

jmank88:
Supports:
- https://github.com/smartcontractkit/chainlink/pull/12701
- https://github.com/smartcontractkit/chainlink-cosmos/pull/415
- https://github.com/smartcontractkit/chainlink/pull/14180


pkg/loop/config.go
```diff
@@ -6,10 +6,18 @@ import (
...
 	"strings"
+	"time"
 )
 
 const (
-	envDatabaseURL            = "CL_DATABASE_URL"
+	envDatabaseURL                           = "CL_DATABASE_URL"
+	envDatabaseDefaultIdleInTxSessionTimeout = "CL_DATABASE_DEFAULT_IDLE_IN_TX_SESSION_TIMEOUT"
+	envDatabaseDefaultLockTimeout            = "CL_DATABASE_DEFAULT_LOCK_TIMEOUT"
+	envDatabaseDefaultQueryTimeout           = "CL_DATABASE_DEFAULT_QUERY_TIMEOUT"
```
jmank88, line 16: Does `DEFAULT` add value here?
```suggestion
	envDatabaseIdleInTxSessionTimeout = "CL_DATABASE_IDLE_IN_TX_SESSION_TIMEOUT"
	envDatabaseLockTimeout            = "CL_DATABASE_LOCK_TIMEOUT"
	envDatabaseQueryTimeout           = "CL_DATABASE_QUERY_TIMEOUT"
```

krehermann: have these been released? will we break someone if they are changed? if not, then yes rename them

krehermann: yeah, default does not add anything

pkg/loop/config.go
```diff
@@ -150,3 +196,27 @@ func getDatabaseURL() (*url.URL, error) {
 	}
 	return u, nil
 }
+
+func getDatabaseIdleInTxSessionTimeout() (time.Duration, error) {
+	return time.ParseDuration(os.Getenv(envDatabaseDefaultIdleInTxSessionTimeout))
```
krehermann, line 201: do these need default values? if i'm reading the code correctly, i think this will error if the env vars aren't set? are we assuming that the core node sets them? if so, i think we need a comment somewhere in this repo, probably where they are defined, and a test integration test to ensure all the value are coming through the env correctly

jmank88: They are only required if the DB URL is included. Yeah the dummy plugin on the core side is meant to address it.

pkg/sqlutil/pg/connection.go
jmank88, line 1: - [ ] revisit this file since core has changed since this was started

pkg/sqlutil/pg/connection.go
```diff
@@ -0,0 +1,118 @@
...
+		})}
+
+	// Set default connection options
+	lockTimeout := config.DefaultLockTimeout.Milliseconds()
+	idleInTxSessionTimeout := config.DefaultIdleInTxSessionTimeout.Milliseconds()
+	stmt := fmt.Sprintf(`SET TIME ZONE 'UTC'; SET lock_timeout = %d; SET idle_in_transaction_session_timeout = %d; SET default_transaction_isolation = %q`,
+		lockTimeout, idleInTxSessionTimeout, defaultIsolation.String())
+
+	var sqldb *sql.DB
+	if dialect == DialectTransactionWrappedPostgres {
```
jmank88, line 64: Alternatively we could have this:
- trigger based on a switch in `ConnectionConfig`, so that we don't have to make any assumptions about specific dialects from here.
- split out as a separate methods, perhaps `Open` and `OpenDB` to align with `sql` and `sqlx`

Both cases burden the caller slightly, but trivially and with little to no risk. FWIW the config version would be less risky, since the default would be to use a connector, and there is no decision to be made unless you are looking for it.

pkg/loop/server.go
```diff
@@ -90,6 +106,27 @@ func (s *Server) start() error {
 		return fmt.Errorf("error starting health checker: %w", err)
 	}
 
+	if envCfg.DatabaseURL != nil {
```
patrickhuie19, line 109: is the same db username being used in core?

jmank88: This code doesn't have an opinion on that and should be flexible enough to work with any changes on the core side.

jmank88: Yes, but what do you mean by locked out? Currently, only the core host holds a lease lock, but that could change in the future.

pkg/sqlutil/pg/connection.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type ConnectionConfig struct {
+	DefaultIdleInTxSessionTimeout time.Duration
+	DefaultLockTimeout            time.Duration
+	MaxOpenConns                  int
+	MaxIdleConns                  int
+}
+
+func (config ConnectionConfig) NewDB(ctx context.Context, uri string, driver Driver) (*sqlx.DB, error) {
```
patrickhuie19, line 57: I'm trying to understand NewDB as a method on a `ConnectionConfig`, as opposed to it being global for the loop and accepting a connection config. What guided you towards defining it this way?

pkg/sqlutil/pg/stats.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	return r
+}
+
+func (r *StatsReporter) Start() {
+	startOnce := func() {
+		r.wg.Add(1)
+		r.lggr.Debug("Starting DB stat reporter")
+		ctx, cancelFunc := context.WithCancel(context.Background())
```
patrickhuie19, line 97: should this be a context.Background? is there an overarching resource lifecycle that would make sense to drill through here?

pkg/sqlutil/pg/url.go
```diff
@@ -0,0 +1,13 @@
+package pg
+
+import "net/url"
+
+func SetApplicationName(u *url.URL, name string) {
+	// trim to postres limit
```
patrickhuie19, line 6: /nit postgres

pkg/sqlutil/pg/stats.go
jmank88: It would still be a `context.Background()` if we wired up a `StopChan`. But since there is just one routine, that complexity is not necessary.

jmank88: Cancel and Wait() still happen normally

pkg/sqlutil/pg/connection.go
jmank88: Can you elaborate on the alternative you are referring to?
This is extracted from the logic in core, and repurposed to be used by LOOPPs, but still used by core as well.

patrickhuie19:
```
func NewDB(ctx context.Context, config ConnectionConfig, uri string, driver Driver) (*sqlx.DB, error) {
```

patrickhuie19 (dismissed):

pkg/sqlutil/pg/connection.go
jmank88: Typically using a method is less verbose since you don't have to repeat the package and the func doesn't need to name the type.
```go
v := packge.NewThing(package.ThingConfig{})
```
```go
v := package.ThingConfig{}.New()
```
However, in this case the payoff is minor given the short package name `pg` and the difference in config and returned type names.
```go
db, err := pg.ConnectionConfig{...}.NewDB(ctx, uri, driver)
```
```go
db, err := pg.NewDB(ctx, pg.ConnectionConfig{...}, uri, driver)
```
Although sometimes struct at the end is friendlier for inline calls:
```go
db, err := pg.NewDB(ctx, uri, driver, pg.ConnectionConfig{
	Foo: "bar",
	Baz: 42,
})
```
Although, now I am questioning my config type naming - why not call this `DBConfig`? So then we have:
```go
db, err := pg.DBConfig{...}.New(ctx, uri, driver)
```
Or:
```go
db, err := pg.NewDB(ctx, uri, driver, pg.DBConfig{...})
```

jmank88: Updated to `DBConfig.New()`

pkg/config/build/build.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	if Version == "" {
+		Version = Unset
+	}
+	if Checksum == "" {
+		Checksum = Unset
+	}
+	ChecksumPrefix = Checksum
+	if len(ChecksumPrefix) > 7 {
+		ChecksumPrefix = ChecksumPrefix[:7]
```
pavel-raykov, line 41: ChecksumPrefix = ChecksumPrefix[:min(7, len(ChecksumPrefix)]?

pkg/loop/cmd/example-relay.go
```diff
@@ -0,0 +1,123 @@
+package main
```
pavel-raykov, line 1: Could you probably add a single sentence describing this example?

pkg/loop/server.go
```diff
@@ -138,6 +174,12 @@ func (s *Server) Register(c services.HealthReporter) error { return s.checker.Re
 
 // Stop closes resources and flushes logs.
 func (s *Server) Stop() {
+	if s.dbStatsReporter != nil {
+		s.dbStatsReporter.Stop()
+	}
+	if s.db != nil {
+		s.Logger.ErrorIfFn(s.db.Close, "Failed to close database connection")
+	}
 	s.Logger.ErrorIfFn(s.checker.Close, "Failed to close health checker")
```
pavel-raykov, line 183: sorry, is it clear why, for example, s.checker is not nil at this point, whereas s.db can still be nil and need to be checked?

pkg/sqlutil/pg/url.go
```diff
@@ -0,0 +1,13 @@
+package pg
+
+import "net/url"
+
+func SetApplicationName(u *url.URL, name string) {
+	// trim to postgres limit
+	if len(name) > 63 {
```
pavel-raykov, line 7: name = name[:min(63, len(name)]?

pkg/loop/server.go
jmank88: The checker is always present but the db is optional and so can be nil if not configured.

jmank88: - [ ] I'll make a note on the fields

pkg/loop/cmd/example-relay.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (p *pluginRelayer) Name() string { return p.lggr.Name() }
+
+func (p *pluginRelayer) NewRelayer(ctx context.Context, config string, keystore core.Keystore, cr core.CapabilitiesRegistry) (loop.Relayer, error) {
+	return &relayer{lggr: logger.Named(p.lggr, "Relayer"), ds: p.ds}, nil
+}
+
+func (p *pluginRelayer) Close() error { return nil }
+
+type relayer struct {
```
patrickhuie19, line 66: how is this consumed?

pkg/loop/server.go
```diff
@@ -44,10 +51,13 @@ func MustNewStartedServer(loggerName string) *Server {
...
 type Server struct {
-	GRPCOpts   GRPCOpts
-	Logger     logger.SugaredLogger
-	promServer *PromServer
-	checker    *services.HealthChecker
+	GRPCOpts        GRPCOpts
+	Logger          logger.SugaredLogger
+	db              *sqlx.DB
+	dbStatsReporter *pg.StatsReporter
+	DataSource      sqlutil.DataSource
```
patrickhuie19, line 58: what do we intend `db` to be used for vs `DataSource`? a comment may be helpful

pkg/sqlutil/pg/connection.go
patrickhuie19: Why move some of this postgres setup over to common? Core no longer needs to control all the connection strings that are passed to LOOPs?

pkg/loop/server.go
reductionista: I've also been wondering if we need to do anything special to handle the lease locks for LOOP plugins. As I understand it, the lease lock is only advisory... so for now the LOOP can just connect to the db and ignore it.  However, it seems like that might leave the door open for two different LOOP plugins of the same type clobbering each other if somehow they both get pointed at the same db.

Maybe this can't happen, because they would have had to have been passed the db url from two different chainlink nodes configured for the same db... and whichever one doesn't get the lock would abort before ever starting a LOOP plugin?

Alternatively, is there any way the same chainlink node could end up starting two LOOP plugins of the same type? What if it starts one, then gets restarted itself but for some reason the first LOOP plugin executable is still running, and then it starts another one? I'm guessing that would cause more problems than just the db.

reductionista: > What if it starts one, then gets restarted itself but for some reason the first LOOP plugin executable is still running

Seems like this would have to be possible if the chainlink executable received a SIGKILL instead of a SIGTERM, but then probably the second LOOP plugin wouldn't start due to port conflicts?

pkg/sqlutil/pg/connection.go
jmank88: What sort of control did you have in mind?

pkg/loop/server.go
jmank88: We hang to `db` in order to call `Close`. We could store `dbCloseFn func() error = db.Close` instead, but it is not exported so seems like nbd :shrug:

pkg/loop/cmd/example-relay.go
jmank88: It is just an example, but returned from `(*pluginRelayer).NewRelayer`

pkg/loop/server.go
jmank88: I agree we will have to consider these things and should do defensive locking, but I'm not sure we can address it yet in the short term until we establish stricter isolation :thinking:

pavel-raykov (dismissed):

pavel-raykov (dismissed):

pkg/sqlutil/pg/connection.go
```diff
@@ -0,0 +1,121 @@
...
+type Driver string
+
+const (
+	// DriverPostgres represents the postgres dialect.
+	DriverPostgres Driver = "pgx"
+	// DriverTxWrappedPostgres is useful for tests.
+	// When the connection is opened, it starts a transaction and all
+	// operations performed on the DB will be within that transaction.
+	DriverTxWrappedPostgres Driver = "txdb"
+)
```
reductionista, line 35: I notice some overlap here with this PR:
https://github.com/smartcontractkit/chainlink-common/pull/910

I assume we only want the dialects defined in one place--so we'll need to decide whether it's here in `pg/connection.go` or in `pg/dialects.go`:
https://github.com/smartcontractkit/chainlink-common/pull/910/files#diff-57a42a4e715a6082c3fd24971c8516609094df5e15be2ae9fe49f302a2cc3efa

I'm fine with either, just pointing out that this is going to cause conflicts.

jmank88: Why don't we combine packages?

pkg/sqlutil/pg/connection_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	"github.com/google/uuid"
+	_ "github.com/jackc/pgx/v4/stdlib"
+	"github.com/jmoiron/sqlx"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_disallowReplica(t *testing.T) {
+	db, err := sqlx.Open(string(DriverTxWrappedPostgres), uuid.New().String())
```
reductionista, line 15: Is this test working already?  Or is the plan for it to work once the PR that moves txdb into common is merged?

(As I understand it, this shouldn't work unless `sql.Register()` is called first, to register the txdb driver and implement it.)

jmank88: It is moved from core so it is a working test, but it is behind a build tag so it is not enabled in CI currently (or when invoking plain `go test`). I don't recall having a particular plan, but we can activate it with your PR

pkg/sqlutil/pg/connection.go
jmank88: I had tried to ignore txdb in this PR, but the switch for the different API was required :shrug:

reductionista: Oh, yeah I hadn't noticed this but your is `pkg/sqlutil/pg` and mine is `pkg/pg`... I don't have a strong preference so if you feel it should be the first one I can move mine.

jmank88: I think it would be good to nest the packages but I don't have strong preferences on type names

pkg/sqlutil/pg/connection.go
reductionista: I think I understand what @patrickhuie19 was asking here and can answer it:

For the short term, the evm, cosmos, & solana schemas are being kept in the main chainlink db, but the long term plan is that each relay will have its own separate db. @krehermann worked out a system (at least partially?) where goose migrations can be managed by each chain separately (even different evm chains), the implementation of which I think is the main blocker for each relay having its own db. Once each relay has its own db it will also use its own config to determine the connection string to it, rather than receiving it from the core executable.  This PR seems like it gets us a step closer to doing that, but there are other pieces that have to fall into place before we can do that.

pkg/sqlutil/pg/pg.go
```diff
@@ -1,35 +1,132 @@
...
+
+	if val == "replica" {
+		return fmt.Errorf("invalid `session_replication_role`: %s. Refusing to connect to replica database. Writing to a replica will corrupt the database", val)
+	}
+
+	return nil
+}
+
+// Deprecated: Use Driver* constants
+type DialectName string
```
jmank88, line 123: This type was forcing type conversions and wasn't restricting what could be used, so it is cleaner to use plain string instead. Leaving it here temporarily for compatibility though.

jmank88: (also it seems to be called 'driver name' everywhere? not sure where the term 'dialect' orginated from)

pkg/loop/internal/example-relay/main.go
```diff
@@ -0,0 +1,124 @@
...
+type relayer struct {
+	lggr logger.Logger
+	ds   sqlutil.DataSource
+}
+
+func (r *relayer) Name() string { return r.lggr.Name() }
+
+func (r *relayer) Start(ctx context.Context) error {
+	var names []string
+	err := r.ds.SelectContext(ctx, names, "SELECT table_name FROM information_schema.tables WHERE table_schema='public'")
```
krehermann, line 76: add comment that this is a dummy query to ensure db connectivity and nothing more

krehermann (dismissed):

pavel-raykov (approved):

PR #451, opened 2024-04-05
Add CCIP price getter IsTokenConfigured method

justinkaseman:


pkg/loop/internal/pb/ccip/pricegetter.proto
```diff
@@ -11,10 +11,23 @@ import "google/protobuf/empty.proto";
 // PriceGetter is a service that returns the price of a token in USD. It is a gRPC service adapter for the interface
 // [github.com/smartcontractkit/chainlink-common/chainlink-common/pkg/types/ccip/PriceGetter]
 service PriceGetter {
+	rpc IsTokenConfigured(TokenConfiguredRequest) returns (TokenConfiguredResponse) {}
```
krehermann, line 14: weird intention.

krehermann (dismissed):

pkg/loop/internal/pb/ccip/pricegetter.proto
justinkaseman: Weird, Github diff being formatted like that

Actual code is as expected
image

justinkaseman: Oh, it's a tab vs. spaces from my code editor formatter. Fixed this

krehermann (approved):

PR #452, opened 2024-04-05
KS-120 remove callback servers and use streams/move channel creation into channel owners/ensure liveness of triggers in event of slow consumer

ettec:
There are related parts to this change:

1. Removing the callback server and using streams
2. Moving ownership of channel lifecycle into the sender objects (to avoid panic on sending to closed channels)
3. Ensuring liveness of the trigger objects by handling slow consumers.


pkg/capabilities/capabilities.go
```diff
@@ -212,6 +212,7 @@ func MustNewCapabilityInfo(
 // TODO: this timeout was largely picked arbitrarily.
 // Consider what a realistic/desirable value should be.
 // See: https://smartcontract-it.atlassian.net/jira/software/c/projects/KS/boards/182
+// TODO: CHECK THIS IS AS EXPECTED ->After this time has passed the call will return - so this is a maximum not a default, caller can choose to cancel the context before this point, but not after, so should be maximumExecuteTimeout, but is this the desired behaviour.
```
cedric-cordenier, line 215: @ettec This looks correct to me -- we called it a `defaultExecuteTimeout` with a view to making overridable at a later stage. I agree maximum makes more sense, and happy for you to modify the code accordingly 👍

pkg/capabilities/capabilities.go
```diff
@@ -222,12 +223,10 @@ func ExecuteSync(ctx context.Context, c CallbackExecutable, request CapabilityRe
 	ctxWithT, cancel := context.WithTimeout(ctx, defaultExecuteTimeout)
 	defer cancel()
 
-	responseCh := make(chan CapabilityResponse)
-	setupCh := make(chan error)
-
-	go func(innerCtx context.Context, innerC CallbackExecutable, innerReq CapabilityRequest, innerCallback chan CapabilityResponse, errCh chan error) {
-		setupCh <- innerC.Execute(innerCtx, innerCallback, innerReq)
-	}(ctxWithT, c, request, responseCh, setupCh)
+	responseCh, err := c.Execute(ctxWithT, request)
```
cedric-cordenier, line 226: This is much cleaner!

pkg/capabilities/capabilities.go
```diff
@@ -102,7 +102,7 @@ type CallbackExecutable interface {
 	// Request specific configuration is passed in via the request parameter.
 	// A successful response must always return a value. An error is assumed otherwise.
 	// The intent is to make the API explicit.
-	Execute(ctx context.Context, callback chan<- CapabilityResponse, request CapabilityRequest) error
```
cedric-cordenier, line 105: @ettec Note -- this will require some changes to capabilities in chainlink and also to the OCR3 capability in this repo. I'm surprised the typechecker hasn't caught the latter case.

pkg/capabilities/capabilities.go
```diff
@@ -245,15 +244,10 @@ outerLoop:
 			}
 
 			vs = append(vs, response.Value)
-
 		// Timeout when a capability panics, crashes, and does not close the channel.
 		case <-ctxWithT.Done():
+			// Not necessarily, the context could have been cancelled by the caller
```
cedric-cordenier, line 249: Correct -- could you modify the comment to clarify that this also applies if the context is canceled?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -91,29 +100,35 @@ func (o *MercuryTriggerService) ProcessReport(reports []mercury.FeedReport) erro
 		if !ok {
 			return fmt.Errorf("no registration for %s", triggerID)
 		}
-		ch <- capabilityResponse
+
+		select {
+		case ch <- capabilityResponse:
+		default:
+			o.log.Warn("dropping response due to slow consumer", "triggerID", triggerID)
```
cedric-cordenier, line 107: Rather than dropping in this case, why not just return an error? That means the caller can choose how to handle it.

pkg/capabilities/triggers/on_demand_trigger.go
```diff
@@ -18,50 +19,69 @@ var info = capabilities.MustNewCapabilityInfo(
...
 
 type OnDemand struct {
+	log logger.Logger
 	capabilities.CapabilityInfo
 	chans map[workflowID]chan<- capabilities.CapabilityResponse
 	mu    sync.Mutex
+
+	// TO REVIEW: Need to ensure that slow consumer does not block this service as most method lock the services mutex
+	// including the send methods. We use a buffer to achieve this and drop (with logging) client messages if the buffer is full.
+	sendChannelBufferSize int
```
cedric-cordenier, line 29: @ettec You can drop this optimization for the trigger IMO -- this is really only meant as a example; we wouldn't ever use this in production.

cedric-cordenier: Discussed offline: we'll leave this in but remove the comment

pkg/capabilities/capabilities.go
```diff
@@ -113,7 +113,7 @@ type BaseCapability interface {
 }
 
 type TriggerExecutable interface {
-	RegisterTrigger(ctx context.Context, callback chan<- CapabilityResponse, request CapabilityRequest) error
+	RegisterTrigger(ctx context.Context, request CapabilityRequest) (<-chan CapabilityResponse, error)
```
bolekk, line 116: We discussed changing CapabilityResponse to TriggerEvent here. Maybe it make sense to include in this PR right away?

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -47,11 +47,16 @@ type capability struct {
...
+	callbackChannelBufferSize int
 }
 
 var _ capabilityIface = (*capability)(nil)
 
-func newCapability(s *store, clock clockwork.Clock, requestTimeout time.Duration, encoderFactory EncoderFactory, lggr logger.Logger) *capability {
+var _ capabilities.ActionCapability = (*capability)(nil)
+
+func newCapability(s *store, clock clockwork.Clock, requestTimeout time.Duration, encoderFactory EncoderFactory, lggr logger.Logger,
+	callbackChannelBufferSize int) *capability {
```
bolekk, line 59: I wonder if we should create a dedicated OCR3CapabilityConfig struct at this point.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -39,7 +42,7 @@ type MercuryTriggerService struct {
 var _ services.Service = &MercuryTriggerService{}
 
 type subscriberConfig struct {
-	FeedIds        []string
+	FeedIDs        []string
```
cedric-cordenier, line 45: @ettec I think you're missing a JSON tag here -- this is populated from line 102, and the incoming data will use `FeedIds` as the key.

pkg/loop/internal/core/services/capability/capabilities.go
```diff
@@ -406,55 +371,27 @@ func (c *callbackExecutableClient) RegisterToWorkflow(ctx context.Context, req c
...
-		case resp, isOpen := <-callbackChannel:
-			if !isOpen {
-				_, err := client.CloseCallback(ctx, &emptypb.Empty{})
-				if err != nil {
-					logger.Error("could not close upstream callback", err)
-				}
+	go func() {
+		defer func() {
+			close(responseCh)
+		}()
```
cedric-cordenier, line 380: @ettec Does `defer close(responseCh)` not work?

pkg/capabilities/triggers/mercury_trigger.go
ettec: I've reverted this rename - was showing up as an issue when I locally linted but seems fine on CI

cedric-cordenier (approved):

PR #454, opened 2024-04-08
feature/capability config validation

HenryNguyen5:
- **Add workflow registration config validation for ocr3**
- **Remove unused config unmarshalling**
- **Add config validation to mercury trigger**


pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -111,8 +111,14 @@ func newWorkflowConfig() *workflowConfig {
 }
 
 func (o *capability) RegisterToWorkflow(ctx context.Context, request capabilities.RegisterToWorkflowRequest) error {
+	err := o.ValidateConfig(request.Config)
```
DeividasK, line 114:
```suggestion
	config, err := o.UnwrapAndValidateConfig(request.Config)
```

DeividasK: I suspect that in most cases you'll want to unwrap before validating

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -281,6 +281,39 @@ func TestMultipleMercuryTriggers(t *testing.T) {
...
+	err = ts.ValidateConfig(config)
+	require.Error(t, err)
+
+	config = newConfigSingleFeed(t, "0x123zzz")
+	err = ts.ValidateConfig(config)
+	require.Error(t, err)
+
+	config = newConfigSingleFeed(t, "0x0001013ebd4ed3f5889FB5a8a52b42675c60c1a8c42bc79eaa72dcd922ac4292")
+	err = ts.ValidateConfig(config)
+	require.Error(t, err)
```
DeividasK, line 310: It's not clear here what your expected error is. Could you use `require.ErrorIs` instead?

cc @cedric-cordenier @bolekk what's the expected best practice around this?

pkg/capabilities/consensus/ocr3/capability.go
HenryNguyen5: nice, having unwrap + validate in a single call would be great for DX

pkg/capabilities/triggers/mercury_trigger_test.go
HenryNguyen5: i copy / pasted the test cases from `FeedId.validate()` :see_no_evil:

HenryNguyen5:
Closing out in favour of https://github.com/smartcontractkit/chainlink-common/pull/445

PR #455, opened 2024-04-09
[KS-79] Mercury Codec improvements

bolekk:
1. Revert BatchedPayload back to a single value.
2. Remove unnecessary ReportSet and related structs.
3. Implement Wrap/Unwrap in Codec.
4. Change previous outcome types to integers.
5. Add logger to Mercury Trigger.


pkg/capabilities/consensus/ocr3/datafeeds/data_feeds_types.proto
```diff
@@ -9,6 +9,8 @@ message DataFeedsOutcomeMetadata {
 }
 
 message DataFeedsMercuryReportInfo {
-  uint32 timestamp = 1;
-  double price = 2;
+  // DEPRECATED = 1;
+  // DEPRECATED = 2;
+  int64 observation_timestamp = 3;
+  int64 benchmark_price = 4;
```
cedric-cordenier, line 15: nit: @bolekk since we're not actually persisting these fields anywhere (and we're not migrating server + client separately) you can safely overwrite the field.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -47,25 +48,31 @@ var _ types.Aggregator = (*dataFeedsAggregator)(nil)
 // Metadata is a map of feedID -> (timestamp, price) representing onchain state (see DataFeedsOutcomeMetadata proto)
 func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value) (*types.AggregationOutcome, error) {
 	// find latest valid Mercury report for each feed ID
-	latestReportPerFeed := make(map[mercury.FeedID]mercury.Report)
+	latestReportPerFeed := make(map[mercury.FeedID]mercury.FeedReport)
 	for nodeID, nodeObservations := range observations {
 		// we only expect a single observation per node - new Mercury data
-		if len(nodeObservations) == 0 {
+		if len(nodeObservations) == 0 || nodeObservations[0] == nil {
```
cedric-cordenier, line 54: @bolekk Hmmm... curious as to why this or was needed? Do we ever get nil values here?

cedric-cordenier (dismissed):

pkg/capabilities/capabilities.go
```diff
@@ -76,10 +76,10 @@ type CapabilityRequest struct {
...
 
 type TriggerEvent struct {
-	TriggerType    string
-	ID             string
-	Timestamp      string
-	BatchedPayload map[string]any
+	TriggerType string
+	ID          string
+	Timestamp   string
+	Payload     values.Value
```
DeividasK, line 82: Nit/Unrelated: These could use comments. This whole file basically defines the Capability API but is sparsely commented - not ideal.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
DeividasK: The answer to the above should be a code comment!

DeividasK: In addition, what are the circumstances when a node would contribute more than one observation? Do we look at those warnings? What's the expected course of action when this happens?

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -47,25 +48,31 @@ var _ types.Aggregator = (*dataFeedsAggregator)(nil)
...
 			a.lggr.Warnf("node %d contributed with empty observations", nodeID)
 			continue
 		}
 		if len(nodeObservations) > 1 {
```
DeividasK, line 58: What are the circumstances when a node would contribute more than one observation? Do we look at those warnings? What's the expected course of action when this happens? Trying to understand why this is a `warn` and not an `error`/`critical`?

DeividasK (changes requested):
Let's chat about how we can make this more generic keeping Mercury and OCR3 capability domains separated.

DeividasK (commented):
On second thought, my comment is quite general, which I don't think is worth blocking the PR on. Still important to resolve though.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: I somehow got it during tests. Does engine guarantee non-nil? A malicious node could send it. Doesn't hurt to have an extra check.

pkg/capabilities/capabilities.go
bolekk: This struct is documented in the design docs. I can add a comment saying that it's a trigger-specific payload but does that add any value?

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: A malicious node could send us any crap and we need to continue processing hoping to have a quorum. Here more than 1 obs can come from an invalid workflow too, which could be caught during validation (depends on how smart it is). Error logs might be too noisy. Soon I'm planning to also sprinkle the code with a bunch of Prom counters so we could easier detect anomalies coming from single nodes.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -47,25 +48,31 @@ var _ types.Aggregator = (*dataFeedsAggregator)(nil)
...
 			a.lggr.Warnf("node %d contributed with more than one observation", nodeID)
 		}
-		mercuryReportSet, err := a.mercuryCodec.Unwrap(nodeObservations[0])
+		event := capabilities.TriggerEvent{}
```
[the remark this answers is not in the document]
bolekk: You're right, I need to move this unwrapping a layer above, into the engine.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -47,25 +48,31 @@ var _ types.Aggregator = (*dataFeedsAggregator)(nil)
...
+		err := nodeObservations[0].UnwrapTo(&event)
+		if err != nil || event.Payload == nil {
+			a.lggr.Warnf("node %d contributed with invalid TriggerEvent", nodeID)
+			continue
+		}
+		mercuryReports, err := a.mercuryCodec.Unwrap(event.Payload)
```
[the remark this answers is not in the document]
bolekk: The aggregator is Mercury-specific, not the capability.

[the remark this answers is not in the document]
bolekk: This evolved into a longer discussion :) Let's make a decision and address it separately.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
DeividasK: `nil` would mean that a node didn't get an observation, right?

pkg/capabilities/consensus/ocr3/datafeeds/data_feeds_types.proto
```diff
@@ -9,6 +9,8 @@ message DataFeedsOutcomeMetadata {
 }
 
 message DataFeedsMercuryReportInfo {
-  uint32 timestamp = 1;
-  double price = 2;
+  // DEPRECATED = 1;
+  // DEPRECATED = 2;
```
cedric-cordenier, line 13: nit: proto supports the reserved keyword for this: https://protobuf.dev/programming-guides/proto3/#fieldreserved

cedric-cordenier (approved):

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: It's different. If a node didn't send on observation during OCR protocol, it wouldn't appear in this map `observations map[ocrcommon.OracleID][]values.Value`.

I think a node could still send us a list of nils and the Outcome() implementation would pass it to the aggregator. I'm not sure if it makes sense to filter it out earlier (maybe a nil could make sense for other aggregators?)

pkg/capabilities/consensus/ocr3/datafeeds/data_feeds_types.proto
bolekk: oh nice, let me change :)

cedric-cordenier (approved):

PR #457, opened 2024-04-09
Fetching nonces for multiple senders

mateusz-sekara:
Backporting reader changes from CCIP https://github.com/smartcontractkit/ccip/pull/646

### DIFF
--- a/pkg/types/ccip/offramp.go
+++ b/pkg/types/ccip/offramp.go
@@ -28,6 +28,7 @@ type OffRampReader interface {
 	GetExecutionStateChangesBetweenSeqNums(ctx context.Context, seqNumMin, seqNumMax uint64, confirmations int) ([]ExecutionStateChangedWithTxMeta, error)
 	GetRouter(ctx context.Context) (Address, error)
 	GetSenderNonce(ctx context.Context, sender Address) (uint64, error)
+	GetSendersNonce(ctx context.Context, senders []Address) (map[Address]uint64, error)
 	GetSourceToDestTokensMapping(ctx context.Context) (map[Address]Address, error)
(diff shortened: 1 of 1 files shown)


pkg/loop/internal/pb/ccip/offramp.proto
```diff
@@ -19,6 +19,7 @@ service OffRampReader {
     rpc OnchainConfig(google.protobuf.Empty) returns (OnchainConfigResponse) {}
     rpc GasPriceEstimator(google.protobuf.Empty) returns (GasPriceEstimatorResponse) {}
     rpc GetSenderNonce(GetSenderNonceRequest) returns (GetSenderNonceResponse) {}
+    rpc GetSendersNonce(GetSendersNonceRequest) returns (GetSendersNonceResponse) {}
```
krehermann, line 22: two nits: 

1/ can you reorder these rpcs to be in alphabetical order while you in this code? i did that in other places, but missed it here.

2/ i suggest calling this ListNonces. it's inline with standard api naming (ex Googlehttps://cloud.google.com/apis/design/naming_convention), and clearly distinguishes the singular case from the plural case. methods that differ my one letter in the middle are very easy to misread and therefore require extra attention.

mateusz-sekara: 1/ will do

2/ actually, we can drop `GetSenderNonce` (singular), because its usage is replaced by the implementation that accepts list. So it's no longer used in CCIP codebase. If we decide to fetch single nonce in the future, it can be done with `GetSendersNonce` (plural) version by passing a single element list. wdyt?

krehermann (approved):

dimkouv (approved):

PR #461, opened 2024-04-12
Fix non-deterministic encoding of Outcomes in OCR3 capability

bolekk:
We can't use maps in protos as they are encoded non-deterministically


pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -175,14 +182,14 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
 			continue
 		}
 
-		agg, err := r.r.getAggregator(weid.WorkflowId)
-		if err != nil {
+		agg, err2 := r.r.getAggregator(weid.WorkflowId)
+		if err2 != nil {
```
HenryNguyen5, line 186: Should this be `err`?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -193,11 +200,25 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
...
 		o.CurrentReports = append(o.CurrentReports, report)
 
-		o.Outcomes[weid.WorkflowId] = outcome
+		outcomeMap[weid.WorkflowId] = outcome
 	}
 
-	r.lggr.Debugw("Outcome complete", "len", len(o.Outcomes), "nCurrentReports", len(o.CurrentReports))
-	return proto.Marshal(o)
+	o.Outcomes = toProto(outcomeMap)
+	raw, err := proto.Marshal(o)
```
HenryNguyen5, line 207: Could we just use the `Deterministic` option for `MarshalOptions` instead?

```
	// Deterministic controls whether the same message will always be
	// serialized to the same bytes within the same binary.
....
	// If deterministic serialization is requested, map entries will be
	// sorted by keys in lexographical order. This is an implementation
	// detail and subject to change.
```
The only possible issue I see here is the tidbit about:
```
// Note that the deterministic serialization is NOT canonical across
	// languages.
```
But I don't forsee us making such guarantees in the first place. WDYT?

HenryNguyen5 (changes requested):

pkg/capabilities/consensus/ocr3/reporting_plugin.go
bolekk: Nice! thanks a lot :)

PR #462, opened 2024-04-12
[WIP] Typed capability API

cedric-cordenier:


pkg/capabilities/v2/capability.go
```diff
@@ -0,0 +1,164 @@
...
+
+	schema, err := jsonvalidate.CompileString("<uriPrefix>", string(schemab))
+	if err != nil {
+		return err
+	}
+	return schema.Validate(mapping)
+}
+
+func (c *capability[I, O, C]) Execute(ctx context.Context, callback chan<- capabilities.CapabilityResponse, request capabilities.CapabilityRequest) error {
+	tcb := make(chan CapabilityResponse[O])
```
HenryNguyen5, line 101: What does `tcb` mean?

pkg/capabilities/v2/capability.go
```diff
@@ -0,0 +1,164 @@
...
+			}
+
+			callback <- capabilities.CapabilityResponse{
+				Value: v,
+			}
+		}
+	}
+}
+
+func NewCapability[I, O, C any](cap Capability[I, O, C]) capabilities.CallbackExecutable {
```
HenryNguyen5, line 160: This looks great, and is the direction I wanted to take the Validation API efforts to :)
What we're building represents a RPC framework of sorts.

 I'd imagine there's a lot of value to be had by allowing the user to inject middlewares that are called during request/response lifecycle

pkg/capabilities/v2/capability.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+type CapabilityResponse[O any] struct {
+	Value O
+	Err   error
+}
+
+type CapabilityRequest[I, C any] struct {
```
bolekk, line 19: How do those work in a remote setting? Does the receiver always know what type to expect or is there some type info encoded too?

HenryNguyen5: I'd imagine it's the same, capabilities should have a input/config/output schema they abide to.

pkg/capabilities/v2/capability.go
cedric-cordenier: Yes agreed :) We could use this for observability and logging for example :)

pkg/capabilities/v2/capability.go
cedric-cordenier: @bolekk This is purely a client-side feature (i.e. inside the capability), anything accessing the capability (either via the Remote API or via the workflow engine would continue to do so via the lower-level, values-based API.

PR #466, opened 2024-04-12
[KS-78] Mercury Trigger rate-limiting and correct ID generation

bolekk:


pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,98 +22,61 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
...
+	workflowID string
+	config     SubscriberConfig
 }
 
 var _ capabilities.TriggerCapability = (*MercuryTriggerService)(nil)
 
-func NewMercuryTriggerService(lggr logger.Logger) *MercuryTriggerService {
+// Mercury Trigger will send events to each subscriber every MaxFrequencyMs (configurable per subscriber).
+// Event generation happens whenever local unix time is a multiple of tickerResolutionMs. Therefore,
+// all subscribers' MaxFrequencyMs valies need to be a multiple of tickerResolutionMs.
```
HenryNguyen5, line 57:
```suggestion
// all subscribers' MaxFrequencyMs values need to be a multiple of tickerResolutionMs.
```

HenryNguyen5 (commented):
I think this style of time sync only works under the condition where the mercury trigger is remote. If each node within the DON is uses the trigger locally, they will no longer be in sync, since the ticker will be based on when the workflow is uploaded.

I guess a more simple way of thinking of it is, if we use this remotely, then it works since the ticker acts as a external time sync. If its used locally, we no longer have a single source of truth. This is why in the DD we use CRON instead, since CRON is independent of subscription time

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -68,27 +60,28 @@ func TestMercuryTrigger(t *testing.T) {
 			ObservationTimestamp: 3,
 		},
 	}
-	err = ts.ProcessReport(fr)
+	err = ts.ProcessReport(mfr)
 	assert.NoError(t, err)
-	assert.Len(t, callback, 1)
+	//assert.Len(t, callback, 1)
```
HenryNguyen5, line 65:
```suggestion
```

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -225,19 +171,76 @@ func (o *MercuryTriggerService) GetTriggerID(req capabilities.CapabilityRequest,
...
+		case <-o.stopCh:
+			return
+		case <-time.After(time.Duration(nextWait) * time.Millisecond):
+			startTs := time.Now().UnixMilli()
+			// find closest timestamp that is a multiple of o.tickerResolutionMs
+			aligned := (startTs + o.tickerResolutionMs/2) / o.tickerResolutionMs * o.tickerResolutionMs
+			o.process(aligned)
+			currTs := time.Now().UnixMilli()
+			desiredNext := aligned + o.tickerResolutionMs
+			nextWait = desiredNext - currTs
```
HenryNguyen5, line 190: I think having this as a private function and explicitly table tested would be good. It's hard as a reviewer to see exactly what's happening here. Its a good candidate for unit testing

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -225,19 +171,76 @@ func (o *MercuryTriggerService) GetTriggerID(req capabilities.CapabilityRequest,
...
+
+			// Create a new CapabilityResponse with the MercuryTriggerEvent
+			capabilityResponse := capabilities.CapabilityResponse{
+				Value: eventVal,
+			}
+
+			o.lggr.Debugw("ProcessReport pushing event", "nReports", len(reportList), "eventID", triggerEvent.ID)
+			select {
+			case sub.ch <- capabilityResponse:
+			default:
```
HenryNguyen5, line 239: Don't we want subscribers to always get the latest feed? I'd imagine that we want to keep a small buffer and aggresively drop the head elements

HenryNguyen5 (changes requested):

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -225,19 +171,76 @@ func (o *MercuryTriggerService) GetTriggerID(req capabilities.CapabilityRequest,
...
+			val, err := mercury.Codec{}.Wrap(reportList)
+			if err != nil {
+				o.lggr.Errorw("Error encoding reportList", "err", err)
+				continue
+			}
+
+			timestampStr := strconv.FormatInt(timestamp, 10)
+			triggerEvent := capabilities.TriggerEvent{
+				TriggerType: "mercury",
+				ID:          timestampStr, // use timestamp as EventID
```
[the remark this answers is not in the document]
bolekk: It's better to keep any IDs human-readable whenever possible. I buy the argument of constant size so I'm going to align it to 32 bytes.

pkg/capabilities/triggers/mercury_trigger.go
bolekk: That shouldn't be a concern of the trigger. It's up to the consumer to handle it, if they want to. In fact, in the remote setting, we already have a ring buffer of messages.

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -47,19 +48,10 @@ func TestMercuryTrigger(t *testing.T) {
 		Config: configWrapped,
 		Inputs: inputsWrapped,
 	}
-	callback := make(chan capabilities.CapabilityResponse, 10)
+	callback := make(chan capabilities.CapabilityResponse, 1000)
```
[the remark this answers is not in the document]
bolekk: No, just a large value to make sure we don't start dropping expected messages if the test decided to run some goroutines slower than the others.

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -246,26 +198,26 @@ func TestMultipleMercuryTriggers(t *testing.T) {
...
-	assert.Equal(t, GenerateTriggerEventID(payload), triggerEvent.ID)
-	assert.Len(t, reports, 1)
-	assert.Equal(t, mfr2[0], reports[0])
+	for rMsg := range callback2 {
+		triggerEvent, reports = upwrapTriggerEvent(t, rMsg.Value)
+		require.NoError(t, err)
+		require.Len(t, reports, 2)
+		require.Equal(t, "mercury", triggerEvent.TriggerType)
+		if reports[1].BenchmarkPrice == 50 {
+			// expect to eventually get updated feed value
```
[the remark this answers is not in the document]
bolekk: Added a simple max-retry limit. Every test has some timeout in CI too so maybe it's OK to just rely on that...?
A fancier way to do it is to use gomega library like core does but I'm not sure if it will work well with waiting on a channel.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,98 +22,61 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
...
 
-type FeedReport struct {
-	FeedID               [mercury.FeedIDBytesLen]byte `json:"feedId"`
-	FullReport           []byte                       `json:"fullReport"`
-	BenchmarkPrice       int64                        `json:"benchmarkPrice"`
-	ObservationTimestamp int64                        `json:"observationTimestamp"`
-}
-
-func (o *MercuryTriggerService) ProcessReport(reports []FeedReport) error {
+func (o *MercuryTriggerService) ProcessReport(reports []mercury.FeedReport) error {
```
cedric-cordenier, line 73: @bolekk Are we using string-based FeedIDs then? (The reason these are different is because ProcessReport uses a byte based FeedID, but internally we use a string).

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -136,25 +97,37 @@ func (o *MercuryTriggerService) RegisterTrigger(ctx context.Context, callback ch
...
 	}
 
+	o.subscribers[triggerID] =
+		&subscriber{
+			ch:         callback,
+			workflowID: wid,
+			config:     cfg,
+		}
+
+	o.chans[triggerID] = callback
```
cedric-cordenier, line 130: @bolekk Do we need to store the channel twice in maps that are keyed to the same thing? Can we just use o.subscribers everywhere?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,98 +22,61 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
...
+	latestReports      map[mercury.FeedID]mercury.FeedReport
+	mu                 sync.Mutex
+	stopCh             services.StopChan
+	wg                 sync.WaitGroup
+	lggr               logger.Logger
+}
+
+var _ services.Service = &MercuryTriggerService{}
+
+type SubscriberConfig struct {
```
cedric-cordenier, line 42: Does this need to be public?

cedric-cordenier: Ignore: I assume there's a follow up PR that will use this 👍

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -225,19 +171,85 @@ func (o *MercuryTriggerService) GetTriggerID(req capabilities.CapabilityRequest,
...
+				continue
+			}
+
+			// Create a new CapabilityResponse with the MercuryTriggerEvent
+			capabilityResponse := capabilities.CapabilityResponse{
+				Value: eventVal,
+			}
+
+			o.lggr.Debugw("ProcessReport pushing event", "nReports", len(reportList), "eventID", triggerEvent.ID)
+			select {
```
cedric-cordenier, line 246: Hmmmm... Note that the workflow engine currently uses an unbuffered channel, and that this channel is read by a synchronising thread that also does other things. I'm not sure we should make the assumption that the channel will be buffered. It feels like this begs the possibility of message loss, which will lead to consensus issues right?

pkg/capabilities/triggers/mercury_trigger.go
bolekk: I think whoever calls ProcessReport can just convert to a string. I didn't notice before that we have two almost identical FeedReport structs, I think it's OK to simplify.

pkg/capabilities/triggers/mercury_trigger.go
bolekk: Yes, good point, I'll change.

pkg/capabilities/triggers/mercury_trigger.go
bolekk: no, changing

pkg/capabilities/triggers/mercury_trigger.go
bolekk: OK then would you rather add buffering on the trigger side? Why is that better than buffering on the engine side?

bolekk: Agreed to re-think and address separately (either in the trigger or in engine - tbd).

cedric-cordenier (approved):

pkg/capabilities/triggers/mercury_trigger_test.go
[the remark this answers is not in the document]
HenryNguyen5: maxRetry is fine, i dont like relying on CI timeouts because they burn $$ and waste developer time

PR #467, opened 2024-04-15
feat: support network level monitoring

aalu1418:


pkg/monitoring/monitor.go
```diff
@@ -168,21 +167,37 @@ func (m Monitor) Run() {
...
 		instrumentedSourceFactories,
 		m.ExporterFactories,
 		100, // bufferCapacity for source pollers
 	)
-
 	subs.Go(func() {
 		m.Manager.Run(rootCtx, func(localCtx context.Context, data RDDData) {
 			m.ChainMetrics.SetNewFeedConfigsDetected(float64(len(data.Feeds)))
 			monitor.Run(localCtx, data)
 		})
```
topliceanu, line 182: Run() doesn't work as you expect - ie. it does NOT "broadcast" rdd data to all the callbacks registered with Run. It's designed to only be run once! I suggest an easy alternative:

```suggestion
		m.Manager.Run(rootCtx, func(localCtx context.Context, data RDDData) {
			m.ChainMetrics.SetNewFeedConfigsDetected(float64(len(data.Feeds)))
			monitor.Run(localCtx, data)
			networkMonitor.Run(localCtx, data)
		})
```

pkg/monitoring/multi_feed_monitor.go
```diff
@@ -49,67 +49,102 @@ func (m *multiFeedMonitor) Run(ctx context.Context, data RDDData) {
...
-			pollers,
-			exporters,
+			},
 		)
+	}
+}
+
+// createMonitor is a reusable method for creating pollers, exporters, and monitors
+// sources are created outside because they may require different input parameters
+func createMonitor(
```
topliceanu, line 93: WDYT of renaming this to `runMonitor()`. Also since this is a shared function, how about moving it to a new file? maybe `run_monitor.go`.

topliceanu (changes requested):
It's looking good. One critical change is that `Manager.Run()` should only be called once! I left more details in the review.

topliceanu (approved):

pkg/monitoring/source.go
```diff
@@ -24,3 +24,8 @@ type SourceFactory interface {
 	// GetType should return a namespace for all the source instances produced by this factory.
 	GetType() string
 }
+
+type NetworkSourceFactory interface {
+	NewSource(chainConfig ChainConfig, nodeConfig []NodeConfig) (Source, error)
```
topliceanu, line 29: Why is this taking in a list of nodes, it should be just one, surely!

pkg/monitoring/source.go
```diff
@@ -24,3 +24,8 @@ type SourceFactory interface {
 	// GetType should return a namespace for all the source instances produced by this factory.
 	GetType() string
 }
```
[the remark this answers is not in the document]
aalu1418: wouldn't it be cleaner to have a separate interface?

otherwise, every existing source would need to implement a new method even if it's not relevant?

PR #468, opened 2024-04-16
create a grpc test scaffold

krehermann:
There was a lot of boiler plate copy pasta to setup grpc client-server tests.

This adds a light weight scaffolding that takes a constructor for the server and client and injects them into the appropriate places for the boilerplate.

It should be possible to refactor other loop internals to use this scaffolding, ie create grpc tests for ocr2/3, median, etc


pkg/loop/internal/relayer/pluginprovider/ext/ccip/test/exec_provider_test.go
```diff
@@ -31,3 +38,68 @@ func TestStaticExecProvider(t *testing.T) {
...
+func TestExecProviderGRPC(t *testing.T) {
+	t.Parallel()
+	ctx := tests.Context(t)
+
+	grpcScaffold := looptest.NewGRPCScaffold(t, setupExecProviderServer, ccip.NewExecProviderClient)
+	t.Cleanup(grpcScaffold.Close)
+	roundTripExecProviderTests(ctx, t, grpcScaffold.Client())
+}
+
+func roundTripExecProviderTests(ctx context.Context, t *testing.T, client types.CCIPExecProvider) {
```
jmank88, line 51: Contexts should be derived from the most narrowly scoped `*testing.T`. Otherwise they aren't cancelled until some ancestor test ends, at which point any goroutines leftover running from a sub-test have leaked.
```suggestion
func roundTripExecProviderTests(t *testing.T, client types.CCIPExecProvider) {
```

pkg/loop/internal/relayer/pluginprovider/ext/ccip/test/commit_store_test.go
```diff
@@ -41,63 +34,25 @@ func TestStaticCommitStore(t *testing.T) {
...
+		assert.True(t, d.IsClosed())
+	})
 }
 
 // roundTripCommitStoreTests tests the round trip of the client<->server.
 // it should exercise all the methods of the client.
 // do not add client.Close to this test, test that from the driver test
-func roundTripCommitStoreTests(ctx context.Context, t *testing.T, client cciptypes.CommitStoreReader) {
+func roundTripCommitStoreTests(t *testing.T, client cciptypes.CommitStoreReader) {
+	ctx := tests.Context(t)
```
jmank88, line 54: It would be safer to make a context within each sub-test

pkg/loop/internal/relayer/pluginprovider/ext/ccip/test/exec_gas_estimator_test.go
```diff
@@ -34,43 +30,18 @@ func TestStaticExecGasEstimator(t *testing.T) {
...
+	t.Cleanup(scaffold.Close)
+	roundTripGasPriceEstimatorExecTests(t, scaffold.Client())
 }
 
 // roundTripGasPriceEstimatorExecTests tests the round trip of the client<->server.
 // it should exercise all the methods of the client.
 // do not add client.Close to this test, test that from the driver test
-func roundTripGasPriceEstimatorExecTests(ctx context.Context, t *testing.T, client *ccip.ExecGasEstimatorGRPCClient) {
+func roundTripGasPriceEstimatorExecTests(t *testing.T, client *ccip.ExecGasEstimatorGRPCClient) {
+	ctx := tests.Context(t)
```
jmank88, line 43: Same here

pkg/loop/internal/relayer/pluginprovider/ext/ccip/test/exec_provider_test.go
```diff
@@ -31,3 +37,69 @@ func TestStaticExecProvider(t *testing.T) {
...
+func TestExecProviderGRPC(t *testing.T) {
+	t.Parallel()
+
+	grpcScaffold := looptest.NewGRPCScaffold(t, setupExecProviderServer, ccip.NewExecProviderClient)
+	t.Cleanup(grpcScaffold.Close)
+	roundTripExecProviderTests(t, grpcScaffold.Client())
+}
+
+func roundTripExecProviderTests(t *testing.T, client types.CCIPExecProvider) {
+	ctx := tests.Context(t)
```
jmank88, line 50: :heavy_plus_sign:

pkg/loop/internal/relayer/pluginprovider/ext/ccip/test/commit_store_test.go
krehermann: why? the are all the same `t *testing.T` 

if we make a ctx for each subtest, wouldn't every single one of them behave in the exact same way?

jmank88: Every sub-test derives a new `t *testing.T`:
```go
t.Run("", func(t *testing.T) {
```

jmank88 (approved):

PR #470, opened 2024-04-16
Various Keystone bugfixes

bolekk:
1. MaxFrequencyMs can't be set to 0 in Mercury Trigger config.
2. Previous outcome initialization that supports changes in aggregator config.
3. Observation phase can't fail when store.getN() can't find some of the IDs - it should still process the one that were found.

### DIFF
--- a/pkg/capabilities/consensus/ocr3/store.go
+++ b/pkg/capabilities/consensus/ocr3/store.go
@@ -36,21 +36,20 @@ func (s *store) add(ctx context.Context, req *request) error {
 	return nil
 }
 
-func (s *store) getN(ctx context.Context, requestIDs []string) ([]*request, error) {
+// best-effort, doesn't return requests that are not in store
+func (s *store) getN(ctx context.Context, requestIDs []string) []*request {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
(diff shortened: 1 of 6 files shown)


pkg/capabilities/consensus/ocr3/store.go
```diff
@@ -43,11 +43,9 @@ func (s *store) getN(ctx context.Context, requestIDs []string) ([]*request, erro
 	o := []*request{}
 	for _, r := range requestIDs {
 		gr, ok := s.requests[r]
-		if !ok {
-			return nil, fmt.Errorf("request with id %s not found", r)
+		if ok {
```
cedric-cordenier, line 46: @bolekk This can't fail any more -- can we remove the `err` from the return? (And maybe add a comment that it's a "best-effort" get that isn't guaranteed to return everything requested.

cedric-cordenier: (I'll approve in the meantime, this is a nit, so if you want to defer to later, please do)

cedric-cordenier (dismissed):

cedric-cordenier (approved):

HenryNguyen5 (approved):

PR #471, opened 2024-04-17
Create a new type codec to be lenient on extra bits.

nolag:
Eg: Solana returns all bits allocated regardless of what's used. Vectors (like slices in go) may require extra bits so they can grow.


EasterTheBunny (approved):
LGTM

PR #472, opened 2024-04-17
Introduce test helper for mercury trigger setup

HenryNguyen5:


cedric-cordenier (approved):

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -290,6 +291,13 @@ func ExampleOutput() (values.Value, error) {
 }
 
 func (o *MercuryTriggerService) Start(ctx context.Context) error {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if o.started {
```
bolekk, line 296: I don't think we need that here, or if we do there's a standard way to implement it (see other services).

HenryNguyen5: I had it as part of the DD, I'd want to eventually make this part of the capability framework such that all triggers automatically noop https://docs.google.com/document/d/1J5iRxoB9Nb4jzkTgsWuTMN_voat4RzPrUaBblgOOOR0/edit#heading=h.uaws7yj4nf70

For now though, are we OK with a copy/paste across triggers?

krehermann: all the services in core are leveraging `chainlink-common.StateMachine` to `StartOnce` you can search the code base to see the pattern and we should use it for consistently unless there is a good reason not to do so

state machine code defined here https://github.com/smartcontractkit/chainlink-common/blob/ec6d48a178ae86e680cd6d53862b4221a8272f7e/pkg/services/state.go#L77

HenryNguyen5: Reverted, thanks for the explanation!

PR #474, opened 2024-04-17
Create poller

DylanTinianov:


pkg/utils/poller/poller.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Poller is a component that polls a function at a given interval
+// and delivers the result to subscribers
+type Poller[T any] struct {
+	services.StateMachine
+	pollingInterval time.Duration
+	pollingFunc     func() (T, error)
+	logger          logger.Logger
+
+	subscribers []*mailbox.Mailbox[T]
```
dhaidashenko, line 20: We should not use mailbox as it drops old messages, if we exceed queue capacity. Each instance of the Poller should be used by single subscriber. If subscriber does not read msg from the channel, it's fine to block the polling goroutine

pkg/utils/poller/poller.go
```diff
@@ -0,0 +1,91 @@
...
+	stopCh      chan struct{}
+	wg          sync.WaitGroup
+}
+
+// NewPoller creates a new Poller instance
+func NewPoller[T any](pollingInterval time.Duration, pollingFunc func() (T, error), logger logger.Logger) *Poller[T] {
```
dhaidashenko, line 26: We'll use this component to implement methods with [following signatures](https://github.com/smartcontractkit/chainlink/blob/89abd726b6c3f29a84e0fc5d230a1324f622755b/core/chains/evm/client/rpc_client.go#L344)
`Subscribe(ctx context.Context, channel chan<- *evmtypes.Head, args ...interface{}) (commontypes.Subscription, error)`

We should try to simplify this process as much as possible. For instance we should right away return type that implements [Subscription interface.](https://github.com/smartcontractkit/chainlink/blob/578cc6b73acf323740a73abfd8e96b2ba444462b/common/types/subscription.go#L8)

dhaidashenko (changes requested):

DylanTinianov:
Moving to `chainlink/common`

PR #475, opened 2024-04-18
[chore] Move core service types to types/core to mirror package structure; Add RelayerSet

cedric-cordenier:
* Move the core API services (keystore, capabilities_registry, kv store, telemetry, etc) into core.
* Also move the API "integration path" methods relating to integration reporting plugins into core.
* Add a RelayerSet API. This duplicates some structs (PluginArgs, RelayArgs), but omits job-related fields.

NOTE: Working on preparing the corresponding PRs to core/chainlink-solana, and will hold off on merging this until those are ready. I'd still appreciate some early feedback at this stage though!

### DIFF
--- a/pkg/types/core/median.go
+++ b/pkg/types/core/median.go
@@ -1,0 +1,14 @@
+package core
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/libocr/offchainreporting2/reportingplugin/median"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type PluginMedian interface {
+	// NewMedianFactory returns a new ReportingPluginFactory. If provider implements GRPCClientConn, it can be forwarded efficiently via proxy.
+	NewMedianFactory(ctx context.Context, provider types.MedianProvider, dataSource, juelsPerFeeCoin median.DataSource, errorLog ErrorLog) (types.ReportingPluginFactory, error)
+}
--- a/pkg/types/core/relayerset.go
+++ b/pkg/types/core/relayerset.go
@@ -1,0 +1,33 @@
+package core
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type RelayerSet interface {
+	Get(ctx context.Context, relayID types.RelayID) (Relayer, error)
+
+	// List lists the relayers corresponding to `...types.RelayID`
+	// returning all relayers if len(...types.RelayID) == 0.
+	List(ctx context.Context, relayIDs ...types.RelayID) ([]Relayer, error)
+}
+
+type PluginArgs struct {
+	TransmitterID string
+	PluginConfig  []byte
+}
+
+type RelayArgs struct {
+	ContractID         string
+	RelayConfig        []byte
+	ProviderType       string
+	MercuryCredentials *types.MercuryCredentials
+}
+
+type Relayer interface {
+	services.Service
+	NewPluginProvider(rargs RelayArgs, pargs PluginArgs) (types.PluginProvider, error)
+}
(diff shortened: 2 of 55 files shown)


pkg/types/core/relayerset.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type RelayerSet interface {
+	Get(ctx context.Context, relayID types.RelayID) (Relayer, error)
+	GetAll(ctx context.Context) ([]Relayer, error)
```
krehermann, line 12: prefer List to GetAll. 

additionally if you want to complete, you can pass a vardiac filter `List(ctx, ids ...types.RelayID)` and return everything when len(ids) == 0

krehermann (commented):
lgtm up to nit about api naming

pkg/types/relayer.go
```diff
@@ -7,6 +7,14 @@ import (
 	"github.com/google/uuid"
 )
 
+type Network = string
+type ChainID = string
```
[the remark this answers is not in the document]
cedric-cordenier: @jmank88 This actually lives inside chainlink at the moment, but we'd need to move it to common so we can build out the RelayerSet API. 

I agree that having these as separate types seems overkill: looking at the current implementation we don't attach any methods to them, so happy to remove 👍

pkg/types/relayer.go
```diff
@@ -2,11 +2,71 @@ package types
...
+
+func (i *RelayID) String() string {
+	return i.Name()
+}
+func NewRelayID(n string, c string) RelayID {
+	return RelayID{Network: n, ChainID: c}
+}
+
+var idRegex = regexp.MustCompile(
+	fmt.Sprintf("^((%s)|(%s)|(%s)|(%s))\\.", NetworkEVM, NetworkCosmos, NetworkSolana, NetworkStarkNet),
```
krehermann, line 44: as said below, i think the regex is more complex than needed. however, if you need to keep to it, then it should be constructed from SupportNetwork rather than individual consts

cedric-cordenier: @krehermann Do you mind if I defer this? This is just the existing implementation copied from core to common

pkg/types/relayer.go
krehermann, line 1: does this file make sense here? if we are mirroring the core types up from internal, shouldn't we do the same for the other building blocks?

imho it would make the core side easier to read.

currently it is ~`types.NetworkEVM` and `core.Keystore` etc. 

`relayer.EVM` and `core.Keystone` would be more consistent. or `relayer.EVMNetwork`

pkg/types/core/median.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/libocr/offchainreporting2/reportingplugin/median"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type PluginMedian interface {
```
krehermann, line 11: why is this a core type? it seems out of place relative to the other things

pkg/types/relayer.go
cedric-cordenier: I don't disagree, but I was trying to just do what I needed to get the RelayerSet interface into common as I promised Dimitris I'd do that by EOW. There's already been a bit of scope creep and I want to avoid further creep here, but happy to follow up with moving the remaining types into types/relayer.

pkg/types/core/median.go
cedric-cordenier: It's analogous to the reporting plugin factories methods, except median-specific (dating from when median was a custom loopp type.

pkg/types/relayer.go
krehermann: ok. can you add some TODOs with ticket references for the deferred maintenance? i'm fine with postponing, but i want to make sure we do the cleanup. if there are TODOs and unassigned tickets in the backlog we can clean them during planning.

krehermann (approved):

PR #473, opened 2024-04-17
Add validator usage to ondemand trigger

HenryNguyen5:

PR #478, opened 2024-04-19
Add validation to mercury trigger service

HenryNguyen5:


pkg/capabilities/mercury/mercury.go
```diff
@@ -30,6 +30,11 @@ func (id FeedID) Bytes() [FeedIDBytesLen]byte {
 	return [FeedIDBytesLen]byte(b)
 }
 
+// NOTE: Torn about this since we already have JSON Schema validation
+// on mercury trigger side.
+//
+// feeds_aggregator uses this, but on the previous round outcome
+// which should have already clean data?
```
HenryNguyen5, line 37: @bolekk could you help me confirm this?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,8 +23,22 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
 
 const defaultTickerResolutionMs = 1000
 
+type MercuryTriggerConfig struct {
```
HenryNguyen5, line 26: Make private

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,8 +23,22 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
...
+	// strings should be hex-encoded 32-byte values, prefixed with "0x", all lowercase, minimum 1 item
+	FeedIDs []string `json:"feedIds" jsonschema:"pattern=^0x[0-9a-f]{64}$,minItems=1"`
+	// must be greater than 0
+	MaxFrequencyMs int `json:"maxFrequencyMs" jsonschema:"minimum=1"`
+}
+
+type MercuryTriggerInputs struct {
```
HenryNguyen5, line 33: Make private

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -85,46 +92,31 @@ func (o *MercuryTriggerService) RegisterTrigger(ctx context.Context, callback ch
...
 	}
 
-	if len(feedIDs) == 0 {
-		return errors.New("no feedIDs to register")
-	}
-	if cfg.MaxFrequencyMs <= 0 {
-		return errors.New("MaxFrequencyMs must be greater than 0")
-	}
-	if int64(cfg.MaxFrequencyMs)%o.tickerResolutionMs != 0 {
+	if int64(config.MaxFrequencyMs)%o.tickerResolutionMs != 0 {
```
HenryNguyen5, line 111: Sucks that we cant encode this in native JSON Schema. We can technically add extensions like https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema/v5#example-package-UserDefinedFormat but IMO i dont think its worth doing for now.

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -202,7 +185,7 @@ func (o *MercuryTriggerService) process(timestamp int64) {
 	for _, sub := range o.subscribers {
 		if timestamp%int64(sub.config.MaxFrequencyMs) == 0 {
 			reportList := make([]mercury.FeedReport, 0)
-			for _, feedID := range sub.config.FeedIds {
+			for _, feedID := range sub.config.FeedIDs {
```
HenryNguyen5, line 188: Made this change since revive linter was complaining

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -250,11 +233,6 @@ func wrapReports(reportList []mercury.FeedReport, eventID string, timestamp int6
 	}, nil
 }
 
-func ValidateInput(mercuryTriggerEvent values.Value) error {
-	// TODO: Fill this in
-	return nil
-}
-
 func ExampleOutput() (values.Value, error) {
```
HenryNguyen5, line 236: Should remove this since we have an output schema now

pkg/capabilities/triggers/mercury_trigger_test.go
```diff
@@ -285,3 +288,61 @@ func upwrapTriggerEvent(t *testing.T, wrappedEvent values.Value) (capabilities.T
...
+	require.Error(t, err)
+	require.Empty(t, conf)
+
+	rawConf = newConfigSingleFeed(t, "0x0001013ebd4ed3f5889FB5a8a52b42675c60c1a8c42bc79eaa72dcd922ac4292")
+	conf, err = ts.ValidateConfig(rawConf)
+	require.Error(t, err)
+	require.Empty(t, conf)
+
+	rawConf = newConfigSingleFeed(t, "0x0001013ebd4ed3f5889fb5a8a52b42675c60c1a8c42bc79eaa72dcd922ac4292")
+	conf, err = ts.ValidateConfig(rawConf)
```
HenryNguyen5, line 326: Add a test for a `maxFrequencyMs` less than 1

pkg/utils/testutils.go
```diff
@@ -2,12 +2,29 @@ package utils
...
 func Context(t *testing.T) context.Context {
 	return tests.Context(t)
 }
+
+// A custom transformer for comparing JSON strings.
+//
+// Use this with the cmp.Diff function to compare JSON strings.
+//
+// E.g. cmp.Diff(expected, actual, transformJSON)
+var TransformJSON = cmp.FilterValues(func(x, y []byte) bool {
```
HenryNguyen5, line 23: PR #473 should be refactored after this is merged to use this rather than duplicate the same func

cedric-cordenier: Nit: if this only exists to be used with cmp.Diff, then maybe we could create an assertEqualsWithDiff that combines the two?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -24,8 +23,22 @@ var mercuryInfo = capabilities.MustNewCapabilityInfo(
...
 const defaultTickerResolutionMs = 1000
 
+type mercuryTriggerConfig struct {
+	// strings should be hex-encoded 32-byte values, prefixed with "0x", all lowercase, minimum 1 item
+	FeedIDs []string `json:"feedIds" jsonschema:"pattern=^0x[0-9a-f]{64}$,minItems=1"`
+	// must be greater than 0
+	MaxFrequencyMs int `json:"maxFrequencyMs" jsonschema:"minimum=1"`
+}
+
+type mercuryTriggerInputs struct {
```
cedric-cordenier, line 33: Nit: you can avoid some stuttering here since these are private:
mercuryTriggerConfig/Inputs -> config/inputs

cedric-cordenier (commented):
Looks good but I'll let Bolek approve :)

pkg/utils/testutils.go
HenryNguyen5: Good idea, I can make a CmpJSON wrapper!

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -26,9 +25,22 @@ const defaultTickerResolutionMs = 1000
...
+	TriggerID string `json:"triggerId"`
+}
+
+var mercuryTriggerValidator = capabilities.NewValidator[mercuryTriggerConfig, mercuryTriggerInputs, capabilities.TriggerEvent](capabilities.ValidatorArgs{Info: mercuryInfo})
```
bolekk, line 39: Why create it here and not inside NewMercuryTriggerService?

pkg/capabilities/mercury/mercury.go
bolekk: I'm OK replacing with your validator.

pkg/capabilities/triggers/mercury_trigger.go
bolekk: +1

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -153,19 +145,9 @@ func (o *MercuryTriggerService) UnregisterTrigger(ctx context.Context, req capab
...
-	if err != nil {
-		return "", err
-	}
-	if id, ok := inputs.(map[string]interface{})["triggerId"].(string); ok {
-		// TriggerIDs should be namespaced to the workflowID
-		return workflowID + "|" + id, nil
-	}
-
-	return "", fmt.Errorf("triggerId not found in inputs")
+func (o *MercuryTriggerService) GetTriggerID(triggerID string, wid string) string {
```
bolekk, line 148: I think this doesn't have to be public.

pkg/capabilities/triggers/mercury_trigger.go
HenryNguyen5: I was originally writing tests that used just the validator. I don't think I have any here. I can either inline it or keep the pattern consistent. I thought it may be useful when we have complex capabilities where we dont want to bother settting up the capability just to test validation

pkg/capabilities/mercury/mercury.go
HenryNguyen5: Leaving it in for now. Validators should be left to capabilities only. @bolekk I'd only be comfortable with removing this if the data coming into `Aggregate` is guaranteed to be validated. Then we dont need this check at all.

cedric-cordenier (approved):

PR #488, opened 2024-04-22
math: unsigned average

aalu1418:


pkg/utils/mathutil/mathutil.go
```diff
@@ -21,3 +25,25 @@ func Min[V constraints.Ordered](first V, vals ...V) V {
 	}
 	return min
 }
+
+func Avg[V constraints.Unsigned](arr []V) (V, error) {
```
jmank88, line 29: Why limit to `Unsigned`? Should work the same for signed, no?
```suggestion
func Avg[V constraints.Integer](arr []V) (V, error) {
```
And what about variadic, to align with the others? (although min/max are built in now...)
```suggestion
func Avg[V constraints.Unsigned](arr ...V) (V, error) {
```

aalu1418: the overflow checking logic is only for unsigned right now - let me see how hard it is to make it work for signed

jmank88 (approved):

PR #493, opened 2024-04-25
monitoring: allow managing multiple blocking functions

aalu1418:
`monitor.Run` and `networkMonitor.Run` are blocking functions (which prevented `networkMonitor.Run` from being called)

PR allows `manager` to manage multiple functions in parallel

### DIFF
--- a/pkg/monitoring/manager_test.go
+++ b/pkg/monitoring/manager_test.go
@@ -144,22 +144,25 @@ func TestManager(t *testing.T) {
 			<-ctx.Done()
 			closeWG.Done()
 		}
+		managedNonBlocking := func(_ context.Context, _ RDDData) {
+			createWG.Done()
+			closeWG.Done()
+		}
 		subs.Go(func() {
-			manager.Run(ctx, managed, managed)
+			manager.Run(ctx, managed, managed, managedNonBlocking)
 		})
 
 		// send RDD update to create multiple managed funcs
-		createWG.Add(2) // expect to see two created
+		createWG.Add(3) // expect to see 3 created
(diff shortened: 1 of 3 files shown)


pkg/monitoring/manager_test.go
```diff
@@ -123,4 +125,48 @@ func TestManager(t *testing.T) {
...
+
+		// send RDD update to create multiple managed funcs
+		createWG.Add(3) // expect to see 3 created
+		closeWG.Add(3)  // expect to see 3 closed on restart
+		poller.ch <- RDDData{Feeds: []FeedConfig{generateFeedConfig()}}
+		createWG.Wait() // wait for created
+
+		// ensure stop + restarting works
+		createWG.Add(3)                                                 // expect 3 new created
+		closeWG.Add(3)                                                  // expect 3 closed on shutdown
```
augustbleeds, line 163: right before this point the closeWG should be 1 right? since one of them hasn't returned yet. so how does adding another 3 work here?

augustbleeds (approved):

pkg/monitoring/manager_test.go
aalu1418: technically `closeWG = 2` because 1 func is nonblocking and two are

but then when an new config is received, the remaining 2 are closed down and 3 more are immediately spun up

I added 3 more before the new config trigger, otherwise there could be a race where the nonblocking call would call `Done()` and trigger a panic because the wg group hasn't accounted for the new ones. so instead i opted to preallocate the count before triggering a new config

augustbleeds: oh ok gotcha! i see that the context is controlled by the manager so that makes sense. thanks

PR #494, opened 2024-04-25
Change Confirmations to Confidence

EasterTheBunny:
`Confidence` is a different approach to mapping a chain agnostic concept of confidence that an event or transaction can or should be acted on. `Confidence` is intended to map to defined levels of confirmations on EVM chains and other states on other chains.

Only two levels of `Confidence` exist by default to define the boundaries: `Highest` and `Lowest`. This creates an inclusive range for all chain configurable `Confidence` levels and can be treated as % confidence that an event or transaction can or should be acted on.


ilija42 (changes requested):
This PR diff got broken when Querying PR was merged

pkg/types/query/primitives/primitives.go
```diff
@@ -50,22 +50,21 @@ func (f *Block) Accept(visitor Visitor) {
...
 }
 
-type ConfirmationLevel int32
+type ConfidenceLevel float64
 
-// TODO come up with a proper chain agnostic solution BCF-3156
 const (
-	Finalized   = ConfirmationLevel(-1)
-	Unconfirmed = ConfirmationLevel(-2)
+	Highest = ConfidenceLevel(1)
```
mateusz-sekara, line 56: I noticed that the most often used values for confirmations are (in block confirmations):
* 0 block confirmations
* finalized (based on finality depth or tag)
* 1 block confirmation

I believe that 0 block confirmations and finalized are very easy to model with that approach - it would be Lowest and Highet, but how about 1 block confirmation? AFAIK in CCIP we don't use that, but I saw some core components relying on 1 block confirmation, e.g. config_poller, contract_transmitter

pkg/types/query/primitives/primitives.go
```diff
@@ -50,22 +50,21 @@ func (f *Block) Accept(visitor Visitor) {
 	visitor.Block(*f)
 }
 
-type ConfirmationLevel int32
+type ConfidenceLevel float64
```
mateusz-sekara, line 53: Just thinking out loud, but the set of real numbers is infinite (of course float64 has limitations, but that's not the case). This means that the user can specify any float number from [0..1]. Does it really make sense to make it that granular? What would be the point of passing the confidence level `0.12345` here? Probably anything after 1st or 2nd digit after decimal point would be ignored anyway

Wouldn't it be simpler from the user and implementation perspective to specify that as a finite set? For instance integers `[0..10]` or `[0..100]`?

EasterTheBunny: I think that's a great point and @ilija42 brought it up too. The reason I went with a `float64` and infinite granularity is because we don't lose anything significant computation-wise and we gain configurability. I think we need to avoid confining this too early before we know how it will be used. I think worst case, the granularity freedom is ignored and 1st or 2nd digit numbers are the top choice, like you say.

jmank88: The std lib `package slog` approach to this sort of problem was to use int but leave space between: https://pkg.go.dev/log/slog#Level
```go
const (
	LevelDebug Level = -4
	LevelInfo  Level = 0
	LevelWarn  Level = 4
	LevelError Level = 8
)
```

pkg/types/query/primitives/primitives.go
EasterTheBunny: The case of a 1 block confirmation would be a configuration in the chain reader somewhere on the spectrum of `[0..1]` in confidence level. The goal of switching to confidence level is to decouple the concept of finality and confirmations from the abstract confidence of being able to act on any particular event/log on any chain without reorgs or the like.

Therefore, the mappings of example confidence levels like `Medium`, or `KindaHigh` might map to very different concepts such as actual confirmations or a strict finality tag. Each chain can do something different AND each contract/event can be mapped different, allowing products a bit more freedom.

pkg/types/query/primitives/primitives.go
EasterTheBunny: @jmank88 our situation is a little different where the API defines the lower and upper bounds leaving the space between to be configured per chain/product. The issue is how we represent those numbers: `int` or `float`. The value of `int` would be fewer 'in-between' options to confine implementation complexity. The value of `float` would be not having to be concerned in the future about not having enough 'in-between' options (too confined).

jmank88: Only upper and lower bounds? I thought some of the EVM levels would map to common chain-agnostic levels.

EasterTheBunny: The common package only defines the upper and lower bounds. The EVM implementation maps those bounds to common EVM concepts like `Unconfirmed` and `Finalized`. At the moment, those are the only common EVM concepts (each mapping to a different SQL query). Anything else is treated as an exact # of confirmations (though technically Unconfirmed is zero confirmations).

We can create more chain-agnostic levels going forward, but at the moment this is all that is needed.

EasterTheBunny: The API is intended to not allow exact # of confirmations through the API. This completely abstracts confirmations into the concept of `Confidence` and pushes mappings into the chain implementation and configuration.

jmank88: That makes sense, but superficially this system seems less flexible than the EVM system based on positive integers mapping to # of confirmations and negative integers having a special meaning. Isn't it possible that we'll need to support types of confidence that don't fall in a perfect sliding scale like this?

EasterTheBunny: As for sliding scale, I don't think the number itself matters, but I think confidence is always a sliding scale. If we define an absolute highest and an absolute lowest then inserted values are relative to the endpoints representing confidence as a gradient.

EVM finality concepts include: `latest`, `earliest`, `pending`, `safe`, `justified`, `finalized`, and exact block confirmation count

The above string values represent a sliding scale of confidence for values being pulled from the chain, with the exception of exact block confirmations, which can be arbitrary points. However, even using arbitrary points like block confirmations still fit somewhere on the scale of finality, though they could in practice be higher than the highest string representation `finalized`.

If an application defined 3 points of granularity to `Confidence` such as `Highest`, `Medium`, and `Lowest`, the EVM implementation would simply need to map those three points of granularity to various meaningful points in EVM. Maybe the following mappings would be used: `Highest == finalized`, `Medium == safe`, and `Lowest == confirmations(0)` and all others are ignored.

With the same API granularity, a Solana mapping might be `Highest == finalized`, `Medium == finalized`, `Lowest == confirmed` allowing the API usage to be the same and the mapping of confidence to be interpreted exclusively by the relay implementation.

In this way, application code remains the same and expected behavior is a representation of intent instead of direct chain code and the relay interprets the intent.

@jmank88 pointed out that this opens us up to confusion when developing against the API because of the potential for opaque interpretations of `ConfidenceLevel` intent. For example, if the product code continues to use the 3 points of granularity and the EVM implementation switches `Medium` to map to `latest`, the product would then begin to behave differently.

jmank88: I think this makes sense, but the more interesting case to consider is when a new chain integration shows up which has a `Medium` level which is not logically like `safe`. If `Medium` maps to `safe` on EVM but something with completely different guarantees on another network, then things start to break down. Product code needs to be able to choose a level based on documented guarantees that the chain implementations follow. For this reason, I don't think that high/medium/low are useful names. I think we need to use common terms, like `finalized`, with clearly defined guarantees. I don't know enough about `safe` to say what a good general name could be, or how that should be documented, but I think that is the question to answer. Although, I notice that we don't actually use it anywhere in the core node - do we have any good example uses?

ilija42: I think that the most likely thing to happen is that someone would define a new reserved word that maps to some concept like CCIP did with `DestOptimisticConfirmations `. Sounds hard to define some concepts and put them on a viable abstract scale using integers/floats. Maybe just having Constants that map to config is enough?

EasterTheBunny: If we use reserved words that map directly to specific finality concepts on a chain, I think we'll begin to leak chain specific details through the interface. The granularity of finality can be different per chain family and actual finality depth can be different per chain in the same family.

While `DestOptimisticConfirmations` does exist at this point, I don't think it is something we can support anyway with the reader interface because exact confirmations leak chain details. We need another way to do the same thing for CCIP.

EasterTheBunny: As a general concept, finality is a scale on all chains (so far) and can be represented as such. I do agree that the mappings from the abstract scale to the finite scale can be error prone. How can we address those potential issues without leaking chain specific details?

jmank88: > that map directly

They don't need to map directly, just to at-least the level specified.

ilija42: But they don’t map directly thats the whole point. All of these would just directly map to whatever is set in the config. EVM finality depth is different for all evm chains

ilija42: A lot of use cases aren’t on a scale but are a direct value from config, so it makes sense to have those as reserved words. We can still also provide a scale

EasterTheBunny: Thought experiment:

Assume an abstract 5-point finality scale for the interface with defined points `lowest`, `low`, `medium`, `high`, and `highest`. Each point is an arbitrary integer or string value (not on a linear scale) therefore, no numeric relationship between each point.

Assume chain  A implementation has 4 points of finality on an abstract scale (non-numeric) `unconfirmed`, `1`, `safe`, and `finalized`. The 5-point interface scale would need to map to the 4-point chain scale, which would be the responsibility of the chain implementation.

Assume initial mappings as:
- `lowest` --> `unconfirmed`
- `low` --> `1`
- `medium` --> `safe`
- `high` --> `finalized`
- `highest` --> `finalized`

From here, we add a new chain that has 7 points of finality `a`, `b`, `c`, `d`, `e`, `f`, and `g`. A decision would need to be made on adding another point of granularity in the interface scale or not. Assume we add `kindalow` to the interface to help accommodate the new chain. We still don't have a 1:1 between the interface scale and the chain scale and the previous implementation would need to map the new interface point before being able to use it in a product.

However, with a numeric linear scale on the interface with defined boundaries, a product can specify any point in between and combined with chain configurations on mapped finality points, introducing arbitrary points in the interface becomes a product decision without leaking chain specific details.

In both cases, either defined points or numeric scale on the interface, there is a challenge in mapping to finite finality points on a chain.

jmank88: Solana and Ethereum both have a two-thirds majority commitment level. Is that one common enough to name?

| Commitment | On-Chain | 2/3 Majority | 99.99%
| ----|--------|--------------|---------| 
| Solana | Processed | Confirmed | Finalized
| EVM | Confirmed | Safe | Finalized
| Generic | ? | ? | Finalized

If a product needs the 2/3 level, and a particular chain doesn't have it, then it would have to wait for Finality.

jmank88: > a product can specify any point in between and combined with chain configurations on mapped finality points, introducing arbitrary points in the interface becomes a product decision without leaking chain specific details.

We must be able to write product code without thinking about chain details though. It is OK to configure the chain to meet the products' needs, but that product code has to already be referencing something well defined in the API, not just a number or a word like 'medium' that doesn't come with specific guarantees about the commitment level.

EasterTheBunny: Currently we're not working with chains that don't operate on two-thirds commitment levels. For a very basic set, that could work. I see that it provides explicit guarantees, but provides no configuration between those strict points. Maybe that's sufficient?

EasterTheBunny: @mateusz-sekara Do the defined commitment level points offer enough granularity for CCIPv2 to function?

mateusz-sekara: @dimkouv

EasterTheBunny: We resolved this conversation by agreeing to use enums in the interface, allowing the relay to configure the chain meaning of the enums (Unconfirmed, Finalized), and keeping the concept of a range such that new enums could be introduced in the future when more granularity is needed.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -307,6 +306,17 @@ func convertExpressionToProto(expression query.Expression) (*pb.Expression, erro
 	return pbExpression, nil
 }
 
+func confidenceToProto(level primitives.ConfidenceLevel) pb.Confidence {
+	switch level {
```
ilija42, line 310: Did you check how `DestOptimisticConfirmations` is used? Does it need its own enum?

pkg/types/query/primitives/primitives.go
reductionista: I still have some reservations about making it a continuous parameter. I like the idea of letting people configure different custom levels like `Medium` or `KindaHigh`, but the case of 1-block confirmations is a good example of where a continuous confidence level doesn't make a whole lot of sense.

You could define a confidence level called `OneBlock` either way I guess. But if it were up to me, I'd avoid giving people the misleading impression that this is an actual "confidence" value in the statistical sense (a p-value between 0 and 1, representing Bayesian likelihood).  We could just have it be an integer between 0 and 10... very unlikely anyone would need to define more than 3 or 4 different levels... so even allowing 10 different values gives us plenty more flexibility than we'll ever need.

reductionista: Oh nevermind... I read this comment before looking over the code. Looks like you have already changed it to an enum, great!

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
EasterTheBunny: Yes, it is it's own configuration component, but I get conflicting information on whether this will be included in CCIPv2. Since both endpoints are configurable, I suggest we leave it with the two enums and add a new one in the case it's needed for v2.

ilija42 (approved):

pkg/types/query/primitives/primitives.go
EasterTheBunny: Yup, this was changed to enums. I'll close this conversation.

ilija42 (approved):

PR #495, opened 2024-04-26
BCF-3168 add a relayset server to generic plugins

ettec:


pkg/loop/internal/pb/relayerset/relayerset.proto
```diff
@@ -0,0 +1,88 @@
...
+  rpc NewPluginProvider(NewPluginProviderRequest) returns (NewPluginProviderResponse) {}
+
+  rpc StartRelayer(RelayerId) returns (google.protobuf.Empty) {}
+  rpc CloseRelayer(RelayerId) returns (google.protobuf.Empty) {}
+  rpc RelayerReady(RelayerId) returns (google.protobuf.Empty) {}
+  rpc RelayerHealthReport(RelayerId) returns (RelayerHealthReportResponse) {}
+  rpc RelayerName(RelayerId) returns (RelayerNameResponse) {}
+
+}
+
```
cedric-cordenier, line 81: Extra white space :)

pkg/loop/internal/pb/relayerset/relayerset.proto
```diff
@@ -0,0 +1,88 @@
...
+message RelayerNameResponse {
+  string name = 1;
+}
+
+service RelayerSet {
+  rpc Get(GetRelayerRequest) returns (GetRelayerResponse) {}
+  rpc List(ListAllRelayersRequest) returns (ListAllRelayersResponse) {}
+  rpc NewPluginProvider(NewPluginProviderRequest) returns (NewPluginProviderResponse) {}
+
+  rpc StartRelayer(RelayerId) returns (google.protobuf.Empty) {}
```
cedric-cordenier, line 74: @ettec Why do we need the word Relayer over and over again in the method names?

pkg/loop/internal/relayer/relayer.go
```diff
@@ -336,21 +347,24 @@ func (r *relayerServer) NewConfigProvider(ctx context.Context, request *pb.NewCo
 }
 
 func (r *relayerServer) NewPluginProvider(ctx context.Context, request *pb.NewPluginProviderRequest) (*pb.NewPluginProviderReply, error) {
-	exJobID, err := uuid.FromBytes(request.RelayArgs.ExternalJobID)
+	reqRelayArgs := request.RelayArgs
+	reqPluginArgs := request.PluginArgs
```
cedric-cordenier, line 351: nit: as these are local variables, consider rargs, pargs since that's more idiomatic :)

pkg/loop/inprocess_provider_server.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayerset/inprocessprovider"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+)
+
+type ProviderServer interface {
+	Start(context.Context) error
+	Close() error
+	GetConn() (*grpc.ClientConn, error)
```
cedric-cordenier, line 20: @ettec *grpc.ClientConn -> grpc.ClientConnInterface?

pkg/loop/internal/pb/relayerset/relayerset.proto
ettec: ah, yeah, was because originally this sat in the same package as the other protos, hence would clash - will shorten these

pkg/loop/internal/relayer/relayer.go
ettec: I'm a recovering java coder and struggle with brevity of naming ;-) will shorten

pkg/loop/internal/pb/relayerset/relayerset.proto
ettec: IIRC This was because without the addition relayer there was a name clash with the embedded ServiceClient in the client impl,  I suppose because the relayerset itself is a service, so start stop on the service is different from start/stop on an underlying relayer:  

`type Client struct {
	*net.BrokerExt
	*goplugin.ServiceClient

	log logger.Logger

	relayerSetClient relayerset.RelayerSetClient
}`

cedric-cordenier (approved):

pkg/loop/internal/core/services/relayerset/test/relayerset.go
jmank88, line 1: Unused?

ettec: it is used, though not in the containing package but in reporting plugin tests, this is a similar pattern used to other types in the services package, I assume this is to avoid having to declare test types more than once

cedric-cordenier (approved):

pkg/loop/internal/core/services/relayerset/test/relayerset.go
jmank88: Sorry, how is it used? The methods both panic, so does that mean that we are missing coverage?

PR #507, opened 2024-05-09
[chore] Add OCR3ContractTransmitter type

cedric-cordenier:


ettec (approved):

pkg/codec/encodings/binary/int_gen_test.go
```diff
@@ -666,6 +666,7 @@ func TestUint64(t *testing.T) {
 }
 
 func TestBuilderInt(t *testing.T) {
```
cedric-cordenier, line 668: `make generate` added this and was required for CI to pass.

PR #509, opened 2024-05-13
BCI-1426: Gas Price Subunits Changes

augustbleeds:
Since last time this change was worked on, the code base has changed significantly so I created a new branch. 

> PluginMedian Interface needs to be modified to include the gasPriceSubunits data source which tells us how much the gas price is on the chain during a report transmission (for chains which do not support onchain methods of retrieving gas price). It's part of the transmission structure similar to juelsPerFee so it's actually a value which is transmitted to the block-chain as a part of the median plugin.

Merge Order
1. This PR
2. https://github.com/smartcontractkit/chainlink-feeds/pull/28
3. https://github.com/smartcontractkit/chainlink/pull/13200

Please see original PR description here for more information https://github.com/smartcontractkit/chainlink-common/pull/166/files


pkg/loop/internal/relayer/pluginprovider/ext/median/test/datasource.go
```diff
@@ -52,7 +59,14 @@ func (s staticDataSource) Evaluate(ctx context.Context, ds median.DataSource) er
 		return fmt.Errorf("failed to observe dataSource: %w", err)
 	}
 	if gotVal.Cmp(s.Value) != 0 {
-		return fmt.Errorf("expected Value %s but got %s", value, gotVal)
+		return fmt.Errorf("expected Value %s but got %s", s.Value, gotVal)
```
augustbleeds, line 62: this was a prior typo where it was using the global variable value instead of the static data source's value

pkg/loop/internal/relayer/pluginprovider/ext/median/test/median.go
```diff
@@ -93,6 +104,12 @@ func (s staticMedianFactoryServer) NewMedianFactory(ctx context.Context, provide
 		return nil, fmt.Errorf("NewMedianFactory: juelsPerFeeCoinDataSource does not equal a static test juels per fee coin data source implementation: %w", err)
 	}
 
+	err = s.gasPriceSubunitsDataSource.Evaluate(ctx, gasPriceSubunitsDataSource)
+	// allow for testing 0 value with the same staticMedianFactoryServer (only defined once)
+	if err != nil && !strings.HasSuffix(err.Error(), "got 0") {
```
augustbleeds, line 109: i'd have to rework a lot of logic if I'd want to create a staticMedianFactoryServer with 0 static data source value. Since this is just a test assertion i allow for 0 to come back as a valid value during the evaluation phase

.tool-versions
```diff
@@ -1,2 +1,3 @@
 golang 1.21.4
 golangci-lint 1.55.2
+mockery 2.38.0
```
jmank88, line 3: Is this functional?
We have a make command that installs directly: https://github.com/smartcontractkit/chainlink-common/blob/main/Makefile#L17-L18

augustbleeds: yeah it works , i use asdf locally so I wanted to install this. it's the same version as the makefile

pkg/loop/internal/relayer/pluginprovider/ext/median/test/median.go
krehermann: this is quite brittle. at minimium i would make a named error at the return site and use error.Is here rather than string matching

augustbleeds: sounds good, I can do that

krehermann (approved):

pkg/loop/internal/relayer/pluginprovider/ext/median/test/datasource.go
```diff
@@ -46,13 +53,34 @@ func (s staticDataSource) Observe(ctx context.Context, timestamp types.ReportTim
...
+	Got      *big.Int
+	Expected *big.Int
+}
+
+func (e *CompareError) Error() string {
+	return fmt.Sprintf("expected Value %s but got %s", e.Expected, e.Got)
+}
+
+func (e *CompareError) GotZero() bool {
+	return e.Got.Uint64() == 0
```
jmank88, line 66: nit/ Could tighten this up a bit:
```suggestion
	return e.Got.IsUint64() && e.Got.Uint64() == 0
```

PR #512, opened 2024-05-15
[KS-209] Expose all report fields to the encoder

bolekk:
Expose a list of tuples:
`[]Reports{FeedID, RawReport, Price, Timestamp}`
out of which the encoder can select needed fields.

### DIFF
--- a/pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
@@ -19,8 +19,17 @@ import (
 )
 
 const (
-	OutputFieldName = "mercury_reports"
-	addrLen         = 20
+	// Aggregator outputs reports in the following format:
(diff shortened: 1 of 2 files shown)


pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -19,8 +19,17 @@ import (
 )
 
 const (
-	OutputFieldName = "mercury_reports"
-	addrLen         = 20
+	// Aggregator outputs reports in the following format:
+	//   []Reports{FeedID []byte, RawReport []byte, Price *big.Int, Timestamp int64}
```
DeividasK, line 23:
```suggestion
	//   []Reports{FeedID [32]byte, RawReport []byte, Price *big.Int, Timestamp int64}
```
Why not?

DeividasK (approved):

PR #514, opened 2024-05-15
ks-146 standard capability framework

ettec:


pkg/capabilities/pb/capabilities.proto
```diff
@@ -66,3 +66,17 @@ service CallbackExecutable {
...
   rpc Execute(CapabilityRequest) returns (stream CapabilityResponse) {}
 }
+
+service StandardCapability {
+  rpc Initialise(InitialiseRequest) returns (google.protobuf.Empty) {}
+}
+
+message InitialiseRequest {
+  string config = 1;
+  uint32 errorLogID = 2;
```
krehermann, line 76: the proto convention is snake case. iirc, it would effect the generated code https://protobuf.dev/programming-guides/style/#message-field-names

pkg/loop/internal/core/services/capability/standard_capability.go
```diff
@@ -0,0 +1,223 @@
...
+	relayersetpb "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/relayerset"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayerset"
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/core"
+)
+
+type StandardCapability interface {
+	services.Service
+	capabilities.BaseCapability
+	Initialise(ctx context.Context, config string, telemetryService core.TelemetryService, store core.KeyValueStore,
```
krehermann, line 28: is config a string because it's up to the implementation to interpret?

pkg/loop/internal/core/services/capability/standard_capability.go
```diff
@@ -0,0 +1,223 @@
...
+func (c *StandardCapabilityClient) Initialise(ctx context.Context, config string, telemetryService core.TelemetryService,
+	keyValueStore core.KeyValueStore, capabilitiesRegistry core.CapabilitiesRegistry, errorLog core.ErrorLog,
+	pipelineRunner core.PipelineRunnerService, relayerSet core.RelayerSet) error {
+	telemetryID, telemetryRes, err := c.ServeNew("Telemetry", func(s *grpc.Server) {
+		pb.RegisterTelemetryServer(s, telemetry.NewTelemetryServer(telemetryService))
+	})
+	if err != nil {
+		return fmt.Errorf("failed to serve new telemetry: %w", err)
+	}
+	c.resources = append(c.resources, telemetryRes)
```
krehermann, line 64: are we leaking resources if any of the ServeNew fail? do we need a defer that `Close`s all the resources if there an initialization error?

pkg/loop/internal/core/services/capability/test/standard_capability.go
```diff
@@ -0,0 +1,46 @@
...
+	//TODO implement me
+	return nil
+}
+
+func (t StandardCapabilityService) HealthReport() map[string]error {
+	//TODO implement me
+	return nil
+}
+
+func (t StandardCapabilityService) Name() string { return "" }
```
krehermann, line 40: fill in `StandardCapability` or `StandardCapabilityService` for the return value

pkg/loop/internal/core/services/capability/standard_capability.go
```diff
@@ -0,0 +1,223 @@
...
+
+	errorLogID, errorLogRes, err := c.ServeNew("ErrorLog", func(s *grpc.Server) {
+		pb.RegisterErrorLogServer(s, errorlog.NewServer(errorLog))
+	})
+	if err != nil {
+		return fmt.Errorf("failed to serve error log: %w", err)
+	}
+	c.resources = append(c.resources, errorLogRes)
+
+	pipelineRunnerID, pipelineRunnerRes, err := c.ServeNew("PipelineRunner", func(s *grpc.Server) {
```
bolekk, line 90: Why do we need a pipeline runner?

pkg/loop/internal/core/services/capability/standard_capability.go
```diff
@@ -0,0 +1,223 @@
...
+	pipelineRunner core.PipelineRunnerService, relayerSet core.RelayerSet) error {
+	telemetryID, telemetryRes, err := c.ServeNew("Telemetry", func(s *grpc.Server) {
+		pb.RegisterTelemetryServer(s, telemetry.NewTelemetryServer(telemetryService))
+	})
+	if err != nil {
+		return fmt.Errorf("failed to serve new telemetry: %w", err)
+	}
+	c.resources = append(c.resources, telemetryRes)
+
+	keyValueStoreID, keyValueStoreRes, err := c.ServeNew("KeyValueStore", func(s *grpc.Server) {
```
bolekk, line 66: why do we need a KV store?

pkg/loop/internal/core/services/capability/standard_capability.go
DeividasK: If it is capability instance config, then the answer would be "yes" as on-chain it's stored as bytes. I'm not sure why it isn't `[]byte` here. A more precise name for `config` would be 👌

ettec: Correct, its up to the implementation to interpret, hence string, possible it would be json or similar specific to that standard capability type in practice.  

wrt to if it should be []byte - possibly.  One concern is that none of the existing job specs use []byte as an attribute type as I would guess using []byte would be problematic for the operator UI.

pkg/loop/internal/core/services/capability/standard_capability.go
ettec: yes, added some code to cleanup resources in the event of error.

pkg/loop/internal/core/services/capability/standard_capability.go
ettec: My understanding is that the standard capability is analogous to a generic plugin, and thus should have the same resources available to it, one of those being pipeline runner.  IIRC I did raise this specific question regarding pipeline runner with Cedric to check (same as for keyvalue store).  As I understand it the idea is that we give the capability/plugin writers as much functionality as possible to avoid them having to write custom implementations that would have to access core code directly.

pkg/loop/internal/core/services/capability/standard_capability.go
ettec: (see answer wrt pipeline runner)

pkg/loop/internal/core/services/capability/standard_capability.go
DeividasK: Do we plan to use this for any existing capabilities? If not, is it worth considering keeping the API surface small if there is a way to extend the API in a non-breaking way later?

krehermann (approved):

pkg/loop/internal/core/services/capability/standard_capability.go
DeividasK: > Do we plan to use this for any existing capabilities? If not, is it worth considering keeping the API surface small if there is a way to extend the API in a non-breaking way later?

Nvm, we actually need it for the Consensus capability.

cedric-cordenier (approved):

krehermann (approved):

PR #517, opened 2024-05-16
mathutil: add median utility

aalu1418:
downstream PR: https://github.com/smartcontractkit/chainlink-solana/pull/688

### DIFF
--- a/pkg/utils/mathutil/mathutil.go
+++ b/pkg/utils/mathutil/mathutil.go
@@ -3,6 +3,7 @@ package mathutil
 import (
 	"fmt"
 	"math"
+	"slices"
 
 	"golang.org/x/exp/constraints"
 )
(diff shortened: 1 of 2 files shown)


pkg/utils/mathutil/mathutil.go
```diff
@@ -49,3 +50,17 @@ func Avg[V constraints.Integer](arr ...V) (V, error) {
 
 	return total / V(len(arr)), nil
 }
+
+// Med mirrors implementation with generics: https://github.com/montanaflynn/stats/blob/249b5aaa10484bb7e8f3b866b0925aaebdac8170/median.go#L6
+func Med[V constraints.Integer](arr ...V) (V, error) {
```
jmank88, line 55: Is `Med` a typical shorthand for `Median`? The clarity seems worth a few more characters if this isn't already a well known name:
```suggestion
// Median mirrors implementation with generics: https://github.com/montanaflynn/stats/blob/249b5aaa10484bb7e8f3b866b0925aaebdac8170/median.go#L6
func Median[V constraints.Integer](arr ...V) (V, error) {
```

aalu1418: fair point - i was trying to be too clever with 3 char names 🤦🏼

jmank88 (approved):

PR #519, opened 2024-05-16
KS-208: move workflow yaml and graph validation

krehermann:
Move the logic added https://github.com/smartcontractkit/chainlink/commit/aeb9f4d50d6513c6b2be46e19b36b12a5de22738 from the core repo to the common repo so that it can be used by CLO

The material effect of this change is that the JSON schema root changes from
`https://github.com/smartcontractkit/chainlink/v2/core/services/workflows/workflow-spec-yaml`
to 
`https://github.com/smartcontractkit/chainlink-common/pkg/workflows/workflow-spec-yaml`

associated core PR
- https://github.com/smartcontractkit/chainlink/pull/13235


go.mod
jmank88, line 1: Weird placement for these new deps.
I ought to make another attempt at a go.mod formater

go.mod
```diff
@@ -83,6 +84,7 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20230711160842-782d3b101e98 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230711160842-782d3b101e98 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	sigs.k8s.io/yaml v1.4.0
```
jmank88, line 87: Stray `direct` down here in the `indirect` block

jmank88: Oh and `github.com/google/go-cmp`

pkg/workflows/dependency_graph.go
```diff
@@ -0,0 +1,248 @@
+package workflows
```
bolekk, line 1: These are parts of models.go + state.go, right? Let me know if you modified anything, otherwise I'm assuming it's just copy-paste.

bolekk (approved):
LGTM. It would be good to get a stamp from @HenryNguyen5

pkg/workflows/dependency_graph.go
krehermann: yes, just moved things. it's easy to see by reading the corresponding change in core https://github.com/smartcontractkit/chainlink/pull/13235

HenryNguyen5 (approved):

HenryNguyen5 (approved):

PR #522, opened 2024-05-20
[KS-196] Validate Signatures in Mercury Aggregator

bolekk:
1. Add Metadata field to TriggerEvent. Use it to pass Streams DON info ("SignersMetadata" struct) between trigger receiver and OCR aggregator.
2. Extend feeds OCR aggregator with an initial phase of agreeing on a set of valid signers and the F value.
3. Add ReportContext field to FeedReport - Streams DON signs both fields together so we need to pass them to validate sigantures.

### DIFF
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -80,8 +80,9 @@ type TriggerEvent struct {
 	TriggerType string
 	ID          string
 	Timestamp   string
-	// Trigger-specific payload
-	Payload values.Value
+	// Trigger-specific payload+metadata
+	Metadata values.Value
+	Payload  values.Value
 }
 
 type RegisterToWorkflowRequest struct {
--- a/pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
@@ -11,13 +11,17 @@ import (
 
 	ocrcommon "github.com/smartcontractkit/libocr/commontypes"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/datastreams"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
-const OutputFieldName = "mercury_reports"
+const (
+	OutputFieldName = "mercury_reports"
+	addrLen         = 20
+)
 
 type aggregatorConfig struct {
 	Feeds map[datastreams.FeedID]feedConfig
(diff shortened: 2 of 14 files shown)


pkg/capabilities/capabilities.go
```diff
@@ -80,6 +80,7 @@ type TriggerEvent struct {
 	TriggerType string
 	ID          string
 	Timestamp   string
+	Metadata    values.Value
```
DeividasK, line 83: I was thinking that Metadata is reserved for Workflow-specific data, not trigger-specific (see comment below). Having said this, if the capability response type is report, then it might make sense to include the signers as part of the metadata. This should be documented, though.

DeividasK: Also, shouldn't this be on the CapabilityResponse and not TriggerEvent?

bolekk: I want it to be trigger-specific, I will move the comment to apply to both fields.
I initially wanted to put it inside payload but didn't want to add a yet another layer of wrapping.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -37,23 +41,24 @@ type dataFeedsAggregator struct {
 
 var _ types.Aggregator = (*dataFeedsAggregator)(nil)
 
-// EncodableOutcome is a list of AggregatedPricePoints
+// This Aggregator has two phases:
+//  1. Agree on valid trigger signers by extracting them from event metadata and aggregating using MODE (i.e. F+1 copies needed).
```
krehermann, line 45: what is the comment about F+1? is that specific to MODE? if so, i don't understand how one is related to the other

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -37,23 +41,24 @@ type dataFeedsAggregator struct {
...
+//  2. For each FeedID, select latest valid report, using signers list obtained in phase 1.
+//
+// EncodableOutcome is a list of aggregated price points.
 // Metadata is a map of feedID -> (timestamp, price) representing onchain state (see DataFeedsOutcomeMetadata proto)
-func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value) (*types.AggregationOutcome, error) {
-	// find latest valid Mercury report for each feed ID
+func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value, f int) (*types.AggregationOutcome, error) {
+	allowedSigners, minRequired, payloads := a.extractSignersAndPayloads(observations, f)
+	if len(payloads) > 0 && minRequired == 0 {
```
krehermann, line 52: this is an odd check. why not return an error from the extract? or maybe rename minRequired?

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -140,6 +145,68 @@ func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcom
...
+	signers := make(map[[addrLen]byte]int)
+	mins := make(map[int]int)
+	for nodeID, nodeObservations := range observations {
+		// we only expect a single observation per node - a Streams trigger event
+		if len(nodeObservations) == 0 || nodeObservations[0] == nil {
+			a.lggr.Warnf("node %d contributed with empty observations", nodeID)
+			continue
+		}
+		if len(nodeObservations) > 1 {
+			a.lggr.Warnf("node %d contributed with more than one observation", nodeID)
```
krehermann, line 159: should there be a continue here? if not, why not (ie add a comment)

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -140,6 +145,68 @@ func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcom
 	}, nil
 }
 
+func (a *dataFeedsAggregator) extractSignersAndPayloads(observations map[ocrcommon.OracleID][]values.Value, fConsensus int) ([][]byte, int, map[ocrcommon.OracleID]values.Value) {
```
krehermann, line 148: there is an awful lot of control logic in this func. can you break it into a helper or two. specifically, the inner for loops the `breaks` are hard to read, because i have to think about which loop is being broken.

pkg/capabilities/datastreams/types.go
```diff
@@ -65,10 +66,16 @@ type FeedReport struct {
...
+	Signers               [][]byte
+	MinRequiredSignatures int
+}
+
 //go:generate mockery --quiet --name ReportCodec --output ./mocks/ --case=underscore
 type ReportCodec interface {
-	// validate each report and convert to a list of Mercury reports
-	Unwrap(raw values.Value) ([]FeedReport, error)
+	// validate each report and convert to a list of Feed reports
+	Unwrap(wrapped values.Value, allowedSigners [][]byte, minRequiredSignatures int) ([]FeedReport, error)
```
krehermann, line 78: this doesn't look like Unwrap any longer. UnwrapAndValidate? FeedsFromValue()? CreateFeeds()?

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: yeah, MODE but the value needs to appear at least F+1 times

pkg/capabilities/datastreams/types.go
bolekk: How about UnwrapValid()? :)

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: Technically not because we can just ignore extra values. But maybe better to actually treat that value as bad (misconfigured workflow spec?) - added continue.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: Moved the inner loop to a helper.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: It's a little tricky because for empty observations (from valid nodes!) we will extract zero minRequired but we still want to complete the rest of Aggregate to return meaningful outcome. I re-worded the error message to be more descriptive.

krehermann (approved):

PR #523, opened 2024-05-20
[BCI-3272] Add the ChainWriter interface

nickcorin:


pkg/types/chain_writer.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/google/uuid"
+)
+
+type ChainWriter interface {
+	// SubmitTransaction packs and broadcasts a transaction to the underlying chain.
+	//
+	// The `payload` array must contain the encoded transaction payload, and the related signatures.
+	// The `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
+	SubmitTransaction(ctx context.Context, payload []any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) (int64, error)
```
jmank88, line 15: Why is the payload a slice of `any` instead of just `any`?

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,55 @@
...
+	AssetType ChainFeeType
+}
+
+// ChainFeeType describes the asset type the underlying chain uses to estimate transaction costs.
+type ChainFeeType int
+
+const (
+	Wei ChainFeeType = iota
+	Sun
+	Octa
```
jmank88, line 54: Why are these chain-specific details leaking through?

jmank88: I think I understand the impetus here, but it's not clear to me that an enum is the right choice, since it will necessarily require updates on both client and server to support additional cases. Have we considered something more open ended, like a dynamic string field?

pkg/types/chain_writer.go
nickcorin: This has been updated after some discussions with @patrick-dowell.

pkg/types/chain_writer.go
nickcorin: Is the suggestion here to drop `ChainFeeType` altogether and have `AssetType` be a string?

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// TODO(nickcorin): Intentionally left blank for now.
+type TxMeta struct{}
+
+// TransactionStatus are the status we expect every TXM to support and that can be returned by StatusForUUID.
+type TransactionStatus int
+
+const (
+	Unconfirmed TransactionStatus = iota
```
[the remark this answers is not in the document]
jmank88: I think it makes sense to have something else before `Unconfirmed`, but specifically because to me it implies that it has been sent, so maybe something like `Unsent`? I would be OK with `Unknown` too though

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	Fatal
+)
+
+// ChainFeeComponents contains the different cost components of executing a transaction.
+type ChainFeeComponents struct {
+	// The cost of executing transaction in the chain's EVM (or the L2 environment).
+	ExecutionPrice big.Int
+
+	// The cost associated with an L2 posting a transaction's data to the L1.
+	DataAvailabilityPrice big.Int
```
jmank88, line 46: We seem to be using cost and price interchangeably - should we align on just one? "Cost" seems more intuitive to me than "Price". Perhaps "Fee" is also a candidate :shrug:

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// GetFeeComponents retrieves the associated gas costs for executing a transaction.
+	GetFeeComponents() (ChainFeeComponents, error)
+}
+
+// TxMeta contains metadata fields for a transaction.
+//
+// Eventually this will replace, or be replaced by (via a move), the `TxMeta` in core:
+// https://github.com/smartcontractkit/chainlink/blob/dfc399da715f16af1fcf6441ea5fc47b71800fa1/common/txmgr/types/tx.go#L121
+type TxMeta map[string]string
```
jmank88, line 27: Declaring this as a named type will require some explicit casts. What about using an alias?
```suggestion
type TxMeta = map[string]string
```
I don't grok the comment about future work though. Are you planning to extend this type with methods?

nickcorin: An alias makes sense. Happy to make the change.

nickcorin: There is a similar type currently being used in core (see the link). This will most likely be used in a similar manner, but extracting the struct from `core` is out of scope for this PR.

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// GetFeeComponents retrieves the associated gas costs for executing a transaction.
+	GetFeeComponents() (ChainFeeComponents, error)
+}
+
+// TxMeta contains metadata fields for a transaction.
+//
+// Eventually this will replace, or be replaced by (via a move), the `TxMeta` in core:
+// https://github.com/smartcontractkit/chainlink/blob/dfc399da715f16af1fcf6441ea5fc47b71800fa1/common/txmgr/types/tx.go#L121
+type TxMeta = map[string]string
```
patrick-dowell, line 27: There's a small issue with having this be a map of string -> string. When we decode the map to create the TxMeta, we will have to programmatically match every field name and cast each type to the appropriate types in the TxMeta struct. An alternative could be to use map of string -> any, in which case we could use JSON marshal and unmarshal to create the struct, since TxMeta is JSON encodable.

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type ChainWriter interface {
+	// SubmitSignedTransaction packs and broadcasts a transaction to the underlying chain.
+	//
+	// The `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
+	SubmitSignedTransaction(ctx context.Context, payload []byte, signature map[string]any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) (int64, error)
+
+	// StatusForUUID returns the current status of a transaction in the underlying chain's TXM.
+	StatusForUUID(ctx context.Context, transactionID uuid.UUID) (TransactionStatus, error)
```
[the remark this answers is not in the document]
jmank88: I actually think that `GetTransactionStatus` is the most straightforward. Why do we need to mention UUID? Are there other ways to look up a transaction?

pkg/types/chain_writer.go
jmank88: Why isn't this a well-defined set of fields in the first place?

patrick-dowell: @jmank88 The TxMeta struct is defined here:

https://github.com/smartcontractkit/chainlink/blob/ccb8cd85fef8e3bbe3fb5580277a7bd7f477e6bb/common/txmgr/types/tx.go#L121

We effectively want to continue to support products that want this info logged alongside their transaction, but the struct itself has a ton of generics we'd have to expose to every product using this interface. Something like I'm proposing I think lowers the burden on products that may only be using one or two fields and can pass them as map as described in the doc.

There are alternatives to this of course. We could define something that lets products know all the valid field names, but this can be gathered by importing the TxMeta struct if needed.

pkg/types/chain_writer.go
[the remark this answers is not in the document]
patrick-dowell: I'm okay with that name. Let's go ahead and change it.

pkg/types/chain_writer.go
jmank88: If this API does not specify the expected fields, how will a product know what to pass? and how will a chain relayer know what to look for?

jmank88: (If the generics are just address and hash, then we have precedent for handling that already)

patrick-dowell: @jmank88 can you show me the existing code for handling the generics? I just want to think about whether it makes sense to have the pass the TxMeta struct or not.

jmank88: We cannot support generics directly, but we have chain-agnostic types like the libocr `type Account string` for holding chain-specific encoded values, and we could constrain our generic types to be translatable to/from `Account` or similar.

pkg/types/chain_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	// SubmitSignedTransaction packs and broadcasts a transaction to the underlying chain.
+	//
+	// The `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
+	SubmitSignedTransaction(ctx context.Context, payload []byte, signature map[string]any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) (int64, error)
+
+	// GetTransactionStatus returns the current status of a transaction in the underlying chain's TXM.
+	GetTransactionStatus(ctx context.Context, transactionID uuid.UUID) (TransactionStatus, error)
+
+	// GetFeeComponents retrieves the associated gas costs for executing a transaction.
+	GetFeeComponents(ctx context.Context) (ChainFeeComponents, error)
```
nickcorin, line 20: Should we make `ChainFeeComponents` a pointer, so that we can return `nil` when there are errors instead of an empty struct?

pkg/types/chain_writer.go
nickcorin: @jmank88 the ideal situation would be for us to move `TxMeta` from `chainlink` into `chainlink-common`, which actually is the plan long term. Ideally we'd be using the typed struct now, but we can't import `chainlink` from `chainlink-common`.

I've added a `todo` indicating the plan to do this in a future PR.

jmank88: The type you are referencing cannot be used as-is. We have to create an analogous concrete, chain-agnostic type.  IMHO now is the time to do this - I don't actually see how the piece can fit together without it.

pkg/types/chain_writer.go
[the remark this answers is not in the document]
nickcorin: The current `TxMeta` in the `chainlink` repository is sitting in `common`.

Is there any reason to not just move it across as is?

pkg/types/chain_writer.go
nickcorin: Is there any reason that we cannot just move (copy for this PR) the existing `TxMeta` struct **as is** from `chainlink` into `chainlink-common`, or are changes necessary?

@patrick-dowell -- thoughts on this?

amit-momin (approved):

amit-momin:
Approved but only for the `GetTransactionStatus` method details which is the only part I've been involved with.

pkg/capabilities/datastreams/mocks/report_codec.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.38.0. DO NOT EDIT.
+// Code generated by mockery v2.43.0. DO NOT EDIT.
```
jmank88, line 1: Hmm this should have been caught by the CI check

jmank88 (approved):

PR #524, opened 2024-05-20
Add common crypto utilities used by ccip.

winder:
These utilities are used widely for CCIP security. They are also useful in general for applications that need to validate or create CCIP execution proofs.


winder:
I guess CI is probably failing because I used a fork. I'll re-open this PR once I have repo permissions.

pkg/utils/hashlib/common.go
jmank88, line 1: We've been trying to move away from the unified `utils` tree. What about `package hashutil`, up one level, alongside `package sqlutil`?

pkg/utils/hashlib/common.go
```diff
@@ -0,0 +1,41 @@
...
+
+	return keccak256Fixed(encodedArr), nil
+}
+
+// encodeBytesOfBytes encodes the nested byte arrays into a single byte array as follows
+//  1. total number of nested arrays is encoded into fix-size 8 bytes at the front of the result
+//  2. for each nested array
+//     encode the array length into fixed-size 8 bytes, append to result
+//     append the array contents to result
+func encodeBytesOfBytes(b [][]byte) ([]byte, error) {
```
jmank88, line 27: This func seems more interesting (and flexible) than `BytesOfBytesKeccak`. If this were exported, could `BytesOfBytesKeccak` be removed in favor of callers invoking the hash themselves, after calling this func?

pkg/utils/hashlib/hasher.go
```diff
@@ -0,0 +1,13 @@
+package hashlib
+
+// Hash contains all supported hash formats.
+// Add additional hash types e.g. [20]byte as needed here.
+type Hash interface {
+	[32]byte
+}
+
+type Ctx[H Hash] interface {
```
jmank88, line 9: What is the significance of `Ctx` here?
Would something simple and direct be fitting, like `Hasher`?
```suggestion
type Hasher[H Hash] interface {
```

pkg/utils/merklemulti/merkle_multi.go
jmank88, line 1: Let's move this up out of the `utils` subtree.

And is `merklemulti` too specialized, compared to just `merkle`?

winder: I believe it's because you can provide multiple indices when making a proof instead of just one.

winder:
closing in favor of #527

PR #527, opened 2024-05-21
common packages for ccip crypto

winder:
These are utilities that CCIP uses across several projects that should not depend on each other. Currently they are copy/pasted several times, so putting them here should be a nicer solution.


dimkouv (approved):

go.mod
```diff
@@ -77,7 +78,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.19.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.0.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/crypto v0.18.0 // indirect
+	golang.org/x/crypto v0.18.0
```
jmank88, line 81: Let's move this up to the direct block

winder:
@jmank88 updated with most of your suggestions:
1.  move packages from `utils` to `pkg`.
2.  make `encodeBytesOfBytes` public. I didn't remove `BytesOfBytesKeccak` because it has a nice unit test.
3. rename `hashlib.Ctx` to `hashlib.Hasher.`
4. Fix go.mod import organization.

jmank88:
> @jmank88 updated with most of your suggestions:
> 
> 1. move packages from `utils` to `pkg`.
> 2. make `encodeBytesOfBytes` public. I didn't remove `BytesOfBytesKeccak` because it has a nice unit test.
> 3. rename `hashlib.Ctx` to `hashlib.Hasher.`
> 4. Fix go.mod import organization.

LGTM! Though I still think `package hashutil` will fit better in this module. Do we have any packages named `*lib`?

winder:
> > @jmank88 updated with most of your suggestions:
> > 
> > 1. move packages from `utils` to `pkg`.
> > 2. make `encodeBytesOfBytes` public. I didn't remove `BytesOfBytesKeccak` because it has a nice unit test.
> > 3. rename `hashlib.Ctx` to `hashlib.Hasher.`
> > 4. Fix go.mod import organization.
> 
> LGTM! Though I still think `package hashutil` will fit better in this module. Do we have any packages named `*lib`?

5. renamed hashlib to hashutil :smile:

jmank88 (approved):

PR #528, opened 2024-05-21
Allow chain reader tests to run outside the testing environment to enable testing real chains as part of an application

nolag:


EasterTheBunny (approved):
LGTM

jmank88:
Why do we have to avoid the standard testing framework in order to execute against real chains? These seem like separate concerns.

nolag:
> Why do we have to avoid the standard testing framework in order to execute against real chains? These seem like separate concerns.

The tool it's being used with isn't running as a test.

jmank88 (dismissed):

jmank88 (approved):

EasterTheBunny (approved):
LGTM

PR #530, opened 2024-05-23
pkg/sqlutil: queryLogger must handle non-positive timeouts and inspect context errors; support custom timing thresholds

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3245

Handle non-positive timeouts gracefully and inspect the context error.  In the case of cancelled or unknown error, just return. Only try to calculate how much of the timeout was used for success and explicit `context.DeadlineExceeded`.

https://smartcontract-it.atlassian.net/browse/BCF-3252

Support customizable thresholds via context values.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/13358

### DIFF
--- a/pkg/sqlutil/monitor.go
+++ b/pkg/sqlutil/monitor.go
@@ -116,32 +116,84 @@ func (q *queryLogger) logError(err error) {
 	q.lggr.Errorw("SQL ERROR", "err", err, "sql", q)
 }
 
-// logTiming logs about context cancellation and timing after a query returns.
-// Queries which use their full timeout log critical level. More than 50% log error, and 10% warn.
+// logTiming logs about context errors and timing.
+// To be deferred to run after a query returns.
 func (q *queryLogger) logTiming(ctx context.Context, start time.Time) {
 	elapsed := time.Since(start)
-	if ctx.Err() != nil {
-		q.lggr.Debugw("SQL CONTEXT CANCELLED", "ms", elapsed.Milliseconds(), "err", ctx.Err(), "sql", q)
+
+	if err := ctx.Err(); err != nil {
+		if errors.Is(err, context.Canceled) {
+			q.lggr.Debugw("SQL Context Canceled", "ms", elapsed.Milliseconds(), "err", context.Cause(ctx), "sql", q)
+			return
+		} else if !errors.Is(err, context.DeadlineExceeded) {
+			q.lggr.Debugw("SQL Context Error", "ms", elapsed.Milliseconds(), "err", err, "sql", q)
+			return
+		}
+		q.lggr.Debugw("SQL Deadline Exceeded", "ms", elapsed.Milliseconds(), "err", err, "sql", q)
 	}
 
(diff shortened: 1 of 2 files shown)


krehermann (dismissed):
nit/ typo in PR name "timoeouts"

pkg/sqlutil/monitor.go
```diff
@@ -130,8 +130,14 @@ func (q *queryLogger) logTiming(ctx context.Context, start time.Time) {
 	}
 	timeout := deadline.Sub(start)
 
-	pct := float64(elapsed) / float64(timeout)
-	pct *= 100
+	var pct float64
```
patrickhuie19, line 133: /nit what does pct mean? 

`qtp` or `queryTimeoutPercentage` or even a comment would be a bit more descriptive for me

patrickhuie19 (dismissed):

pkg/sqlutil/monitor.go
jmank88: Added a comment

patrickhuie19 (approved):

patrickhuie19 (approved):

pkg/sqlutil/monitor.go
```diff
@@ -116,32 +116,82 @@ func (q *queryLogger) logError(err error) {
...
+	Error func(time.Duration) time.Duration
+}
+
+type ctxKeyLogThresholds struct{}
+
+func (t *LogThresholds) ContextWithValue(ctx context.Context) context.Context {
+	if t == nil {
+		return ctx
+	}
+	return context.WithValue(ctx, ctxKeyLogThresholds{}, *t)
```
reductionista, line 179: Cool! Glad to see us using values in contexts finally, I've wondered why we don't use them more.

Although I hope this won't interfere with any future use of them. If some component needs to have a value attached to a context, but also needs it to have a deadline with custom thresholds, is there a clean way for those to co-exist?

pkg/sqlutil/monitor.go
```diff
@@ -116,32 +116,82 @@ func (q *queryLogger) logError(err error) {
...
 		q.lggr.Errorw(slowMsg, kvs...)
-	} else if warnThreshold := timeout / 10; warnThreshold > 0 && elapsed > warnThreshold {
+	} else if warnThreshold := thresholds.Warn(timeout); warnThreshold > 0 && elapsed > warnThreshold {
 		q.lggr.Warnw(slowMsg, kvs...)
 	}
 
 	PromSQLQueryTime.Observe(pct)
 }
+
+// TODO document
```
reductionista, line 167: LGTM, just don't forget to take care of this TODO

pkg/sqlutil/monitor.go
jmank88: The context key used here is an unexported type, so nothing outside of this package can use the same key and conflict with it.

patrickhuie19 (approved):

patrickhuie19 (approved):

krehermann (approved):

PR #536, opened 2024-05-25
Validate keystone workflows + Enforce full semver on capabilities

HenryNguyen5:
This PR contains the following changes:
- Workflow definitions now enforce that a full capability semver must be specified. No version ranges are allowed.
- Validation for workflow YAMLs is now enabled.
- The `CapabilityInfo` struct now embeds the version as part of the ID itself, rather than having it as a separate field.

Related PR: https://github.com/smartcontractkit/chainlink/pull/13328

### DIFF
--- a/pkg/workflows/models_yaml.go
+++ b/pkg/workflows/models_yaml.go
@@ -200,10 +200,6 @@ type stepDefinitionYaml struct {
 	//
 	// Semver must be used to specify the version of the Capability at the end of the id field. Capability versions must be immutable.
 	//
-	// Initially, we will require major versions. This will ease upgrades early on while we develop the infrastructure.
-	//
-	// Eventually, we might support minor version and specific version pins. This will allow workflow authors to have flexibility when selecting the version, and node operators will be able to determine when they should update their capabilities.
-	//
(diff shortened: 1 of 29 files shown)


cedric-cordenier (dismissed):

pkg/workflows/models_yaml_test.go
```diff
@@ -176,7 +176,7 @@ func TestJsonSchema(t *testing.T) {
 		require.NoError(t, err)
 
 		// change this to update golden file
-		shouldUpdateSchema := false
+		shouldUpdateSchema := true
```
HenryNguyen5, line 179: revert

pkg/workflows/models_yaml.go
```diff
@@ -279,7 +304,7 @@ func (s *stepDefinitionID) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	// If the JSON data is a table, unmarshal it into a stepDefinitionTableID
+	// If the JSON data is a table, unmarshal it into a StepDefinitionTableID
```
HenryNguyen5, line 307:
```suggestion
	// If the JSON data is a table, unmarshal it into a stepDefinitionTableID
```

bolekk (approved):

PR #538, opened 2024-05-28
minor: Generate mock outside internal package

archseer:


pkg/loop/internal/types/types.go
```diff
@@ -41,7 +41,7 @@ type OCR3CapabilityProvider interface {
 
 // Relayer is like types.Relayer, but with a dynamic NewPluginProvider method.
 //
-//go:generate mockery --quiet --name Relayer --output ./mocks/ --case=underscore
```
cedric-cordenier, line 44: @archseer Do you need to delete the old mock?

cedric-cordenier (approved):

ilija42 (approved):

PR #540, opened 2024-05-28
[KS-249] Make outputs a map by default

cedric-cordenier:


bolekk:
Do we still want to merge this or abandon?

pkg/workflows/testdata/fixtures/workflows/marshalling/workflow_2_spec.json
```diff
@@ -1,33 +1,49 @@
 {
-  "name": "length_ten",
-  "owner": "0x0123456789abcdef0123456789abcdef01234567",
-  "triggers": [
+  "Name": "length_ten",
```
krehermann, line 2: why did the case change? is it important that it changes?

krehermann: i think this might be b/c of the struct change that removes the `json` tags (and `json schema`)

if that's true, then since we are not supporting json encoding of this struct in an meaningful way, i think we can delete any tests that use this file and delete it

cedric-cordenier:
@krehermann I added some comments to the ticket

pkg/workflows/testdata/fixtures/workflows/marshalling/workflow_2_spec.json
cedric-cordenier: Yes, correct :) I've removed it from StepInput as well so we're consistent now.

Note that we do need these tests since they guarantee that we convert from YAML -> JSON consistently and in a lossless way.

krehermann: we DO NOT convert from yaml to json in a lossless way.

it is one of the features of intermediate YAML representation, specifically that versions can be represented in YAML multiple ways  that result in the identical JSON.

test: https://github.com/smartcontractkit/chainlink-common/blob/97d4285d9f4871aafd3a88833c6551c850929a40/pkg/workflows/models_yaml_test.go#L167
doc: https://github.com/smartcontractkit/chainlink-common/blob/97d4285d9f4871aafd3a88833c6551c850929a40/pkg/workflows/dependency_graph.go#L30

did i misunderstand what you meant or did you think the behavior is different?

cedric-cordenier: @krehermann No, you're right -- I was referring to things such as handling conversions properly; not the id being stringified regardless of whether it's a table or not

pkg/capabilities/capabilities.go
```diff
@@ -70,7 +70,7 @@ func (c CapabilityType) IsValid() error {
 
 // CapabilityResponse is a struct for the Execute response of a capability.
 type CapabilityResponse struct {
-	Value values.Value
+	Value *values.Map
```
krehermann, line 73: why a pointer? concerned we are setting up for nil panics.

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -100,8 +105,13 @@ func CapabilityResponseFromProto(pr *CapabilityResponse) capabilities.Capability
...
 		err = errors.New(pr.Error)
 	}
-	return capabilities.CapabilityResponse{
-		Value: val,
-		Err:   err,
+
+	resp := capabilities.CapabilityResponse{
+		Err: err,
 	}
+	if mval, ok := val.(*values.Map); ok {
```
krehermann, line 112: 1. when would this ever not be the case?
2. should it be an error if !ok ? if no, why not and should it at least be logged?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -240,7 +240,7 @@ func wrapReports(reportList []datastreams.FeedReport, eventID string, timestamp
 
 	// Create a new CapabilityResponse with the MercuryTriggerEvent
 	return capabilities.CapabilityResponse{
-		Value: eventVal,
+		Value: eventVal.(*values.Map),
```
krehermann, line 243: can this panic? in other places you explicitly check the type coercion. why not here?

pkg/capabilities/capabilities.go
cedric-cordenier: Because it can be nil, for example when there's an error

pkg/capabilities/triggers/mercury_trigger.go
cedric-cordenier: No it can't panic: triggerEvent is a struct, and a struct is guaranteed to be wrapped as a map

pkg/capabilities/pb/capabilities_helpers.go
cedric-cordenier: The value can be nil which would be a different type of value type. It doesn't need to be an error if !ok because the only way to construct a CapabilityResponse proto is via the helper above: https://github.com/smartcontractkit/chainlink-common/pull/540/files/0f8e143ef729e1e50eb13e527e5b96b5b2711602#diff-14830fdbb3d14caa2427e5d7032446d63c5f6ae4a50df50414e601c2720c74ecR79

pkg/capabilities/capabilities.go
krehermann: i don't understand your response. what do you mean `it can be nil`? isn't this change in and of itself changing the type to be "nil"-able for example by either setting nil on error or not setting any value? couldn't you just as a well set an empty value of non-ptr? 

or are you saying that some other part of the system, external to anything changed herein, can set the value to nil? that doesn't make much sense to me since it ought not compile after this type change.

pkg/capabilities/pb/capabilities_helpers.go
krehermann: i read the values library and i sort of understand. i think my confusion boils down to Wrap(nil) => nil
and so `nil` is used on both side of the dynamic type system and it means different things on each side.

krehermann (dismissed):

pkg/capabilities/capabilities.go
cedric-cordenier: Ah, no -- values.Value is an interface, and you can validly set an interface to `nil` (that's what you're doing when you type assert to an interface using the notation `_ values.Value = (*values.Map)(nil)`

pkg/capabilities/pb/capabilities_helpers.go
cedric-cordenier: Yes, I agree that the handling of nil is a bit confusing: I initially had a message type called values.Nil which represented a nil value, but that was actually a terrible value since that meant we had to handle nil and values.Nil correctly.

krehermann: that makes sense. i was looking at the library and thinking about having an Nil type, and i got stuck about what to do with nil of the interface instance itself

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -69,9 +69,14 @@ func CapabilityResponseToProto(resp capabilities.CapabilityResponse) *Capability
 		errStr = resp.Err.Error()
 	}
 
+	var val values.Value
+	if resp.Value != nil {
```
bolekk, line 73: This doesn't do anything - if resp.Value is nil then val still remains nil. it looks like you could get rid of "val" althogether?

pkg/workflows/dependency_graph.go
```diff
@@ -15,29 +15,34 @@ const (
...
+type StepInputs struct {
+	OutputRef string
+	Mapping   map[string]any
+}
+
 // StepDefinition is the parsed representation of a step in a workflow.
 //
 // Within the workflow spec, they are called "Capability Properties".
 type StepDefinition struct {
-	ID     string         `json:"id" jsonschema:"required"`
```
bolekk, line 22: Why all the json tags are not needed any more?

pkg/workflows/mapping.go
```diff
@@ -0,0 +1,92 @@
+package workflows
```
bolekk, line 1: Is this file extracted from models_yaml.go without changes? If so, do we need to do that? It just makes the PR larger for no reason. Plus it looks like you created a new "inputs" type inside models_yaml.go, instead of in a separate file? Maybe just keep mapping inside there too and refactor everything in a separate PR?

pkg/workflows/models_yaml_test.go
```diff
@@ -71,7 +71,7 @@ func TestWorkflowSpecMarshalling(t *testing.T) {
 		require.NoError(t, err)
 
 		if diff := cmp.Diff(rawworkflowspecJSON, workflowspecJSON, transformJSON); diff != "" {
-			t.Errorf("ParseWorkflowWorkflowSpecFromString() mismatch (-want +got):\n%s", diff)
+			t.Errorf("ParseWorkflowSpecFromString() mismatch (-want +got):\n%s", diff)
```
bolekk, line 74: Any tests for "inputs" type maybe?

pkg/workflows/dependency_graph.go
cedric-cordenier: These weren't actually used for anything aside from tests, so I removed it. I found it confusing for example to see the jsonschema tags since that implied we enforced this somewhere, but we don't.

cedric-cordenier: I applied the same logic to the json tags: since they aren't used, let's remove them so as not to imply that this struct is serialized.

pkg/workflows/mapping.go
cedric-cordenier: Fair point!

pkg/workflows/testdata/fixtures/workflows/references/passing_1.yaml
```diff
@@ -7,11 +7,13 @@ triggers:
 consensus:
   - id: offchain_reporting@1.0.0
     ref: offchain_reporting_1
+    inputs: $(trigger.outputs)
     config: {}
 
 targets:
   - id: write_polygon_mainnet@1.0.0
     ref: write_polygon_mainnet_1 
+    inputs: $(offchain_reporting_1.outputs)
```
cedric-cordenier, line 16: @bolekk This tests the "flat" case

cedric-cordenier: And the non-flat case is tested elsewhere (in a file not covered by this PR since that's pre-existing behaviour)

bolekk (approved):

pkg/capabilities/pb/capabilities_helpers.go
cedric-cordenier: I've had to re-add a variant of this in values.Map.Proto() because the method was panicking when serializing an uninitialized *values.Map pointer.

ettec (approved):

krehermann (approved):

PR #541, opened 2024-05-28
[BCI-3339] Change signature of `SubmitSignedTransaction` in `ChainWriter`

nickcorin:
core ref: bff2850313ca3916cde85716956196d5ab620944

### DIFF
--- a/pkg/types/chain_writer.go
+++ b/pkg/types/chain_writer.go
@@ -8,10 +8,11 @@ import (
 )
 
 type ChainWriter interface {
-	// SubmitSignedTransaction packs and broadcasts a transaction to the underlying chain.
+	// SubmitTransaction packs and broadcasts a transaction to the underlying chain.
 	//
-	// The `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
-	SubmitSignedTransaction(ctx context.Context, args []any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) error
+	// - `args` specifies input parameters to the contract call.
+	// - `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
+	SubmitTransaction(ctx context.Context, contractName, method string, args []any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) error
 
 	// GetTransactionStatus returns the current status of a transaction in the underlying chain's TXM.
 	GetTransactionStatus(ctx context.Context, transactionID uuid.UUID) (TransactionStatus, error)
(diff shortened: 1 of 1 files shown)


pkg/types/chain_writer.go
```diff
@@ -11,7 +11,7 @@ type ChainWriter interface {
 	// SubmitSignedTransaction packs and broadcasts a transaction to the underlying chain.
 	//
 	// The `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
-	SubmitSignedTransaction(ctx context.Context, payload []byte, signature map[string]any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) error
+	SubmitSignedTransaction(ctx context.Context, args []any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) error
```
jmank88, line 14: Can we update the comment doc? `args []any` is not as self-descriptive

silaslenihan (dismissed):

amit-momin (dismissed):

pkg/types/chain_writer.go
archseer, line 14: Let's also rename this to `CreateTransaction`

archseer (changes requested):

pkg/types/chain_writer.go
```diff
@@ -8,10 +8,11 @@ import (
 )
 
 type ChainWriter interface {
-	// SubmitSignedTransaction packs and broadcasts a transaction to the underlying chain.
+	// SubmitTransaction packs and broadcasts a transaction to the underlying chain.
 	//
-	// The `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
-	SubmitSignedTransaction(ctx context.Context, payload []byte, signature map[string]any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) error
+	// - `args` specifies input parameters to the contract call.
```
jmank88, line 13: What are the limitations on the types that can be passed as `args`? b/c we cannot literally support `any`
We need to say something precise, just like `ChainReader` does:
https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/types/chain_reader.go#L25-L29

jmank88: If there is sufficient overlap between the expectations here (even just for serialization), then we may want to declare a new type (alias) that we can document just once, and then use in any related methods, without duplicating the same sort of information around.

@nolag WDYT?

pkg/types/chain_writer.go
nickcorin: Mh. I'm not sure about the rename. We're wrapping `CreateTransaction`, but this method only prepares and submits the transaction to the relevant TXM.

pkg/types/chain_writer.go
nickcorin: Yeah agreed. These args are likely always going to be the arguments passed into the tx to trigger some contract function.

One constraint we have, however, is that the types need to be serializable in some way to be passed over the wire.

archseer (approved):
Merging this to unblock changes in core -- typing improvements can be a follow-up

bolekk (approved):

PR #542, opened 2024-05-28
[KS-248] Fix context cancellation bug

cedric-cordenier:


pkg/capabilities/consensus/ocr3/transmitter.go
```diff
@@ -69,7 +69,10 @@ func (c *ContractTransmitter) Transmit(ctx context.Context, configDigest types.C
 		c.capability = cp.(capabilities.CallbackCapability)
 	}
 
-	_, err = c.capability.Execute(ctx, capabilities.CapabilityRequest{
+	// Execute synchronously and wait for a response: libocr's passed-in context is cancelled
+	// the moment we return, so without this we could hit a race condition where the context is cancelled
+	// before we've actually transmitted the result upstream.
+	_, err = capabilities.ExecuteSync(ctx, c.capability, capabilities.CapabilityRequest{
```
bolekk, line 75: I'm afraid this is still affected by ctx being cancelled (in fact I just tried in a local run and I'm getting a lot of "context deadline exceeded" + "timed out after 60.000000 seconds", which is not really the case - it times out immediately)

bolekk: Ignore the comment above. The fix is probably fine but I might've discovered an unrelated problem.

bolekk: OK, after dealing with some other unrelated problems I tried this fix and realized that:
1. OCR3 capability Execute() with methodSendResponse returns (nil,nil) on success and that doesn't play well with ExecuteSync(). gRPC server wraps it in a non-nil channel but ExecuteSync still waits for values.
2. I hacked passing a fake channel with an empty response in it and stuff started to work consistently!
3. I'm still seeing similar context-cancelled errors but they don't seem to prevent successful transmissions (maybe an artifact of not closing a channel in my hacky solution?).

cedric-cordenier: Yep, that's where I left it last night :) Will update the PR to ensure we return properly.

Side note: this type of thing is what I was referring to when talking about the awkwardness of using channels.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -214,9 +216,11 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 func (o *capability) queueRequestForProcessing(
 	ctx context.Context,
 	metadata capabilities.RequestMetadata,
-	i *inputs) (<-chan capabilities.CapabilityResponse, error) {
+	i *inputs,
+) (<-chan capabilities.CapabilityResponse, error) {
 	callbackCh := make(chan capabilities.CapabilityResponse, o.callbackChannelBufferSize)
 	r := &request{
+		// TODO: link this to the passed-in context
```
cedric-cordenier, line 223: This TODO got removed by someone else at some point, so putting it back to document it

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -214,9 +213,11 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
...
```
jmank88, line 220: What kind of link are you trying to achieve?

jmank88: Oh you are undoing that - what was the problem?

pkg/capabilities/consensus/ocr3/capability.go
jmank88: It was refactored deliberately, and the context was being passed through before this change.

jmank88 (changes requested):

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: I'm not objecting to the refactoring, I'm objecting to the removal of the TODO: StopCh is analogous to context.Background() in terms of how it's used, but the underlying TODO (which is a broader issue around how the caller of a capability can cancel the ongoing request) hasn't been addressed.

cedric-cordenier: I've updated the comment to provide more context on the issue

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -288,11 +279,6 @@ func (o *capability) expiryTimer(ctx context.Context, r *request) {
 			},
 		}
 		o.lggr.Debugw("expiryTimer - expired request", "workflowExecutionID", r.WorkflowExecutionID)
-		o.transmitCh <- resp
+		o.handleTransmitMsg(ctx, resp)
```
jmank88, line 282: I am unable to do a proper suggestion comment, but it looks like the context passed here is not actually used, so we could drop it and just receive from `o.stopCh` directly.
```go
func (o *capability) expiryTimer(r *request) {
	defer o.wg.Done()

	d := r.ExpiresAt.Sub(o.clock.Now())
	tr := o.clock.NewTimer(d)
	defer tr.Stop()
	select {
	case <-o.StopCh:
		return
	case <-tr.Chan():
		resp := &outputs{
			WorkflowExecutionID: r.WorkflowExecutionID,
			CapabilityResponse: capabilities.CapabilityResponse{
				Err:   fmt.Errorf("timeout exceeded: could not process request before expiry %s", r.WorkflowExecutionID),
				Value: nil,
			},
		}
		o.lggr.Debugw("expiryTimer - expired request", "workflowExecutionID", r.WorkflowExecutionID)
		o.handleTransmitMsg(resp)
``

**Review by jmank88 — APPROVED** [review]

File: pkg/capabilities/consensus/ocr3/capability.go [hunk 5]
```diff
@@ -190,8 +190,19 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 				Err:   nil,
 			},
 		}
-		err = o.transmitResponse(ctx, out)
```
bolekk, line 193: nit: you can remove transmitResponse() helper altogether

bolekk (approved):
Last series of local runs with this fix was consistently successful. I think I saw some flakes earlier but they might've been caused by KS-254.

Let's merge this and I'll report back if I see this bug re-appear.

archseer (dismissed):

pkg/loop/internal/core/services/capability/capabilities_test.go
```diff
@@ -240,7 +240,13 @@ func Test_Capabilities(t *testing.T) {
 		err = client.Close()
 		require.NoError(t, err)
 
-		_, isOpen := <-ch
+		// Closing the client will result in an error being
+		// bubbled back to the client.
+		resp := <-ch
+		assert.NotNil(t, resp.Err)
```
krehermann, line 246: i'd prefer to be using typed errors and asserting that we get the expected error.

ettec (dismissed):

bolekk (approved):

ettec (approved):

PR #543, opened 2024-05-29
BCF-3300 Add ContractStateReader to relayer interface

george-dorin:
Add new NewContractStateReader method to relayer
Core PR: https://github.com/smartcontractkit/chainlink/pull/13412


ilija42 (commented):
I am wondering if we can just reuse All of ContractReader relayer and grpc stuff and just have the thinnest layer to wrap it undet ContractStateReader

PR #545, opened 2024-05-29
Chain writer mocks

silaslenihan:


jmank88 (changes requested):
Why generate unused mocks? If importers want these, they can generate them on their own. That keeps their mockery version aligned internally as well.

silaslenihan:
> Why generate unused mocks? If importers want these, they can generate them on their own. That keeps their mockery version aligned internally as well.

@jmank88 Sorry I'm not sure what you mean, can these mocks be generated post-import in the main `chainlink` repo?

jmank88:
> > Why generate unused mocks? If importers want these, they can generate them on their own. That keeps their mockery version aligned internally as well.
> 
> @jmank88 Sorry I'm not sure what you mean, can these mocks be generated post-import in the main `chainlink` repo?   

AFAIK Yes, and I think that should be preferred

silaslenihan:
Closed in favor of generating mocks post-import

PR #549, opened 2024-05-31
[Keystone] Extend aggregator metadata with new fields

bolekk:
1. Add new fields and push them through gRPC helpers and aggregator.
2. Extra logging.


archseer (approved):

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -28,7 +28,9 @@ var info = capabilities.MustNewCapabilityInfo(
 	capabilities.CapabilityTypeConsensus,
 	"OCR3 consensus exposed as a capability.",
 	"v1.0.0",
-	nil,
+	&capabilities.DON{
+		ID: "00010203", // TODO: set real DON ID
+	},
```
cedric-cordenier, line 33: @bolekk This isn't how we should be setting the DON ID for local capabilities; this field is only intended for shimmed capabilities that are added via the syncer (at which point we'll know the the DON ID and its other settings.)

In my mind, we have two options for passing in the DON settings:
1. pass it through as part of the Execute request's metadata via the Workflow engine which already has access to the DON ID
2. pass it through when instantiating the OCR3 capability; and set it on the OCR3 capability job spec.

Let me think through which of these options is the best and implement it

cedric-cordenier: As I said on Slack, I think option 1 is best, assuming we want to pass through the WorklfowDonID, not the OCR3 Capability DON ID (I'm aware they're the same right now, but thinking ahead a bit)

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -237,8 +239,13 @@ func (o *capability) queueRequestForProcessing(
 		WorkflowExecutionID: metadata.WorkflowExecutionID,
 		WorkflowID:          metadata.WorkflowID,
 		WorkflowOwner:       metadata.WorkflowOwner,
-		Observations:        i.Observations,
-		ExpiresAt:           o.clock.Now().Add(o.requestTimeout),
+		WorkflowName:        metadata.WorkflowName,
```
krehermann, line 242: the purpose of this is so the contract can whitelist the report name, right?

we need to document that wf author probably should not ever change the value of step ref name. a note here is probably good enough for now until we have real documentation

as i'm thinking about it, it would be better to have some decoupling between the graph implementation details and the contract API. a simple solution would be some sort of tagging that lets the dev express how a report maps to the contract whitelist.

it would make it much clearer, there would be less spooky-action-at-a-distance, and and it gives more flexibility to the interface b/c more than one report can be tagged with the same value, so the contract doesn't have to change if you add a new WF or new step.

is this tagging something that would have to be done before an audit?

note: i have first hand experience of production systems breaking precisely because of this workflow-step-by-reference pattern. it's a nightmare.

pkg/values/int.go
```diff
@@ -54,6 +54,17 @@ func (i *Int64) UnwrapTo(to any) error {
 
 		*tv = uint(i.Underlying)
 		return nil
+	case *uint32:
+		if i.Underlying > math.MaxInt {
+			return fmt.Errorf("cannot unwrap int64 to int: number would overflow %d", i)
```
krehermann, line 59: nit/ "Cannot unwrap to uint32: number would overflow %d"

pkg/values/int.go
```diff
@@ -54,6 +54,17 @@ func (i *Int64) UnwrapTo(to any) error {
...
+		}
+
+		if i.Underlying < 0 {
+			return fmt.Errorf("cannot unwrap int64 to uint: number would underflow %d", i)
```
krehermann, line 63: nit/ "Cannot unwrap to uint32: number would underflow %d"

pkg/capabilities/consensus/ocr3/capability.go
cedric-cordenier: @krehermann Resolving this as we're using a report id field in the config instead.

cedric-cordenier (approved):

ettec (approved):

PR #552, opened 2024-05-31
Cached Contract Reader

EasterTheBunny:
The base implementation of `ContractReader` does not include caching. This struct wraps a `ContractReader` and provides basic caching strategies such as polling on an interval and caching the result as well as falling-through when a cached value is stale based on a configured value.

Caching is managed per contract/method where params must match exactly on previously requested values. If no cache matches occur, the request falls-through to the wrapped `ContractReader`.

The functions `QueryKey` and `Bind` always fall-through to the wrapped `ContractReader`.


pkg/types/contract_reader_cache.go
```diff
@@ -0,0 +1,441 @@
...
+var _ services.Service = &CachedContractReader{}
+
+type CachedContractReader struct {
+	UnimplementedContractReader
+
+	// injected dependencies
+	base ContractReader
+
+	// internal configuration properties
+	strategies map[string]any
```
jmank88, line 56: Why are strategies values type `any`?

pkg/types/contract_reader_cache.go
```diff
@@ -0,0 +1,441 @@
...
+}
+
+func (c *CachedContractReader) setStrategy(readIdentifier string, strategy any) {
+	c.strategyLock.Lock()
+	defer c.strategyLock.Unlock()
+
+	c.strategies[readIdentifier] = strategy
+}
+
+func (c *CachedContractReader) mustGetData(key cacheKey) *cachedValue {
```
jmank88, line 372: I think `must` will mislead readers in to thinking this "must not error". How about something like `getOrCreate`?

pkg/types/contract_reader_cache.go
```diff
@@ -0,0 +1,441 @@
...
+
+type cacheKey struct {
+	readIdentifier string
+	confidence     primitives.ConfidenceLevel
+	params         any
+	staleness      time.Duration
+}
+
+func (k cacheKey) String() string {
+	return fmt.Sprintf("%s::%s::%+v", k.readIdentifier, reflect.TypeOf(k.params), k.params)
```
jmank88, line 406: I'm worried about these keys staying aligned with how we interpret these `any` params. Ideally, we could limit ourselves to `comparable` types so that we don't have to serialize at all. However, if we must serialize, then should we use JSON instead, since we already depend on that elsewhere?

EasterTheBunny: The cache key is only used locally and is only intended to identify conditions where read parameters match. Is JSON necessary to accomplish this or can we just use Go's value stringification instead?

jmank88: > where read parameters match

How do we know that the way you are printing them is aligned with the user and the chain reader's usage? Using JSON is a trivial way to ensure this isn't a concern, because we already use it to communicate between them.

Furthermore, what if some params contain maps with a randomized iteration order? This breaks the cache key in both cases.

pkg/types/contract_reader_cache.go
EasterTheBunny: This allows multiple strategy types to be in the same key lookup. A specific read key has a cache strategy, but each could be configured differently.

jmank88: Unrestricted `any` doesn't appear to be necessary to achieve this, and makes the API very difficult to read. What about an interface type, just to contain the set of possible strategies?
```go
type CacheStrategy interface {
	cacheStrategy()
}
```

pkg/types/contract_reader_cache.go
EasterTheBunny: Fair point on the JSON.

Key order is also a good point on maps. This would need to be a special case on stringifying the parameters. A special encoder would need to exist that would put the keys in an alphanumeric order in the encoding process.

jmank88: We could use the new json v2 package. You can toggle determinism: https://pkg.go.dev/github.com/go-json-experiment/json#Deterministic
It should be much faster too.

EasterTheBunny: I was actually looking at V2 because I remembered it had encoding options. The deterministic option looks like the right solution.

pkg/types/contract_reader_cache.go
```diff
@@ -0,0 +1,441 @@
...
+	Value  any
+}
+
+type PollAndCache struct {
+	TriggerAndCache
+	Interval time.Duration
+}
+
+type FallthroughOnStale struct {
+	Staleness time.Duration
```
jmank88, line 41: I wonder if we can find a more intuitive (and more explicit?) name for this type and field. Maybe with `Expiration`/`Expires`?
Google Java caches offer `expireAfterAccess` and `expireAfterWrite`: https://github.com/google/guava/wiki/CachesExplained#timed-eviction

Also, we should consider offering this option _alongside_ the other strategies, because consider e.g. if your RPCs are unavailable for extended period of time:
- triggers are being missed
- polling is failing

but the cached value is still being reported without any errors, which is often incorrect though can be subtle and overlooked.

Fortunately, I think this could be a simplifying change overall, if we reduce the different strategy types to a single struct:
```go
type CacheStrategy struct {
	Params any
	Value any

	PollInterval time.Duration

	ExpiresAfter time.Duration
}
```
Then we can infer nocache, trigger, and poll based on which fields are set, while applying expiration to all.

pkg/types/contract_reader_cache.go
```diff
@@ -0,0 +1,441 @@
...
+	data := c.getOrCreate(key)
+
+	data.SetLastUpdate(time.Now())
+
+	retBaseType := reflect.Indirect(reflect.New(reflect.Indirect(reflect.ValueOf(returnVal)).Type()))
+	savedVal := reflect.New(retBaseType.Type())
+
+	reflect.Indirect(savedVal).Set(retBaseType)
+
+	cachedValue := savedVal.Interface()
```
jmank88, line 256: What is this doing? A copy?

jmank88: Would the overall code be simpler if we called `fallThrough` with `returnVal`, and then copied it to `cachedValue` afterwards? (e.g. only one reflect block afterwards would be necessary)

PR #554, opened 2024-06-04
add ccip ocr3 types

dimkouv:
Moving ccip ocr3 types from ccip repo.
Relevant pr: https://github.com/smartcontractkit/ccip/pull/963


pkg/types/ccipocr3/helper_types.go
jmank88, line 1: `helper` confused me at first, since we have a habit of using that term for test files that declare things to be used only in tests. What about something like `common`?

pkg/types/ccipocr3/generic_types.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (s SeqNum) String() string {
+	return strconv.FormatUint(uint64(s), 10)
+}
+
+func NewSeqNumRange(start, end SeqNum) SeqNumRange {
+	return SeqNumRange{start, end}
+}
+
+type SeqNumRange [2]SeqNum
```
jmank88, line 46: Is it worth noting whether these are inclusive or exclusive bounds?

pkg/types/ccipocr3/config.go
```diff
@@ -0,0 +1,77 @@
...
+	PricedTokens []types.Account `json:"pricedTokens"`
+
+	// TokenPricesObserver indicates that the node can observe token prices.
+	TokenPricesObserver bool `json:"tokenPricesObserver"`
+
+	// NewMsgScanBatchSize is the number of max new messages to scan, typically set to 256.
+	NewMsgScanBatchSize int `json:"newMsgScanBatchSize"`
+}
+
+func (c CommitPluginConfig) Validate() error {
```
jmank88, line 31: Where is this called from?

dimkouv: from plugin

pkg/types/ccipocr3/interfaces.go
```diff
@@ -0,0 +1,55 @@
...
+	// Messages are sorted ascending based on their timestamp and limited up to the provided limit.
+	MsgsBetweenSeqNums(ctx context.Context, chain ChainSelector, seqNumRange SeqNumRange) ([]CCIPMsg, error)
+
+	// NextSeqNum reads the destination chain.
+	// Returns the next expected sequence number for each one of the provided chains.
+	// TODO: if destination was a parameter, this could be a capability reused across plugin instances.
+	NextSeqNum(ctx context.Context, chains []ChainSelector) (seqNum []SeqNum, err error)
+
+	// GasPrices reads the provided chains gas prices.
+	GasPrices(ctx context.Context, chains []ChainSelector) ([]BigInt, error)
```
makramkd, line 31: cc @connorwstein we have this GasPrices() method defined on the CCIPReader, do we still need FeeComponents on the chain writer?

dimkouv: The implementation will use a `ChainWriter`

makramkd (dismissed):

pkg/types/ccipocr3/config.go
jmank88: Do you have a link? I couldn't find it in the PR diff.

Should it have uses in this module? At least some simple tests?

dimkouv: https://github.com/smartcontractkit/ccip/blob/46fd2d428da707bb56d39222814dedd4f027e29b/core/services/ocr3/plugins/ccip/commit/plugin.go#L178

dimkouv: added tests

dimkouv: good catch

makramkd (approved):

cedric-cordenier:
Approving to unblock @dimkouv

PR #556, opened 2024-06-04
BCF-3242 Codec Float/Int Type Loss

EasterTheBunny:
To better handle encoding and decoding with CBOR and JSON options, a new type `Number` was added to the codec that combined both encoding and decoding types. Additionally, to make the hard-coder work across a LOOP interface, all instances of `json.Number` are replaced with the new `Number` type that adds CBOR encodings. Finally, a decoding hook was added that provides mappings from the new `Number` type to primitive `int` and `float` types.


pkg/codec/config.go
```diff
@@ -250,3 +268,81 @@ func (h *modifierMarshaller[T]) MarshalJSON() ([]byte, error) {
...
+
+	var value json.Number
+
+	if err := decoder.Decode(&value); err != nil {
+		return err
+	}
+
+	*n = Number(value)
+
+	return nil
```
nolag, line 347: can't this just be 

```
var value json.Number
err := json.Unmarshal(data, &value)
*n = Number(value)
return err
```

pkg/codec/config.go
```diff
@@ -250,3 +268,81 @@ func (h *modifierMarshaller[T]) MarshalJSON() ([]byte, error) {
...
+	intVal, err := strconv.ParseInt(string(n), 10, 64)
+	if err != nil {
+		return nil, err
+	}
+
+	return cbor.Marshal(intVal)
+}
+
+func (n *Number) UnmarshalCBOR(data []byte) error {
+	fmt.Println("unmarshalling Number")
```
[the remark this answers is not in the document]
EasterTheBunny: Oooops.

pkg/codec/config.go
```diff
@@ -250,3 +268,81 @@ func (h *modifierMarshaller[T]) MarshalJSON() ([]byte, error) {
...
+
+func (n Number) MarshalJSON() ([]byte, error) {
+	buffer := bytes.NewBuffer([]byte{})
+	encoder := json.NewEncoder(buffer)
+
+	if err := encoder.Encode(json.Number(n)); err != nil {
+		return nil, err
+	}
+
+	return buffer.Bytes(), nil
```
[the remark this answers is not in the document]
EasterTheBunny: That makes sense.

pkg/codec/config.go
```diff
@@ -250,3 +268,81 @@ func (h *modifierMarshaller[T]) MarshalJSON() ([]byte, error) {
...
+}
+
+func (n *Number) UnmarshalJSON(data []byte) error {
+	fmt.Println("unmarshalling Number as JSON")
```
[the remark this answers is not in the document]
EasterTheBunny: :( I need to sleep a bit more lol

pkg/codec/config.go
EasterTheBunny: I wasn't sure if I needed the `UseNumber` bit or not. I'll try it without and see how it goes.

pkg/codec/config_test.go
```diff
@@ -1,17 +1,90 @@
...
 	"github.com/stretchr/testify/require"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/codec"
 )
 
+func TestNumberJSONMarshal(t *testing.T) {
+	type A struct {
+		B codec.Number
+	}
+
```
nolag, line 21: Nothing in the test verifies that the number is encoded as a number.  This would pass if you didn't override the marshal/unmarshal.

pkg/codec/utils.go
```diff
@@ -13,6 +13,58 @@ import (
...
+	int8Type     = reflect.TypeOf(int8(0))
+	int16Type    = reflect.TypeOf(int16(0))
+	int32Type    = reflect.TypeOf(int32(0))
+	int64Type    = reflect.TypeOf(int64(0))
+	float32Type  = reflect.TypeOf(float32(0))
+	float64Type  = reflect.TypeOf(float64(0))
+)
+
+func NumberHook(from reflect.Type, to reflect.Type, val any) (any, error) {
+	if from == codecNumType && to != codecNumType {
```
nolag, line 28: do we need to go the other way too?  Also, what is using this (or will it get used in chainlink and chainlink-solana for now?)

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -48,12 +48,12 @@ func EncodeVersionedBytes(data any, version uint32) (*pb.VersionedBytes, error)
...
+		bytes, err = json.Marshal(data)
 		if err != nil {
 			return nil, fmt.Errorf("%w: %w", types.ErrInvalidType, err)
 		}
 	case JSONEncodingVersion2:
-		bytes, err = jsonv2.Marshal(data)
+		bytes, err = json.Marshal(data)
 		if err != nil {
 			return nil, fmt.Errorf("%w: %w", types.ErrInvalidType, err)
 		}
```
nolag, line 59: This was using jsonV2 now it's using jsonV1.

To be fair, it seems odd that they have different encoding/decoding versions because only go would know the difference...

EasterTheBunny: I don't know on this one. I was considering Jordan's feedback on what we're considering `v2` for json.

pkg/codec/utils.go
EasterTheBunny: The `chainlink` repo is using this for now and was only needed on the non-loop tests.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
jmank88: In this context, we have an explicit switch for the two different versions, so why wouldn't we use v2 here inside of the v2 case?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -80,9 +80,12 @@ func DecodeVersionedBytes(res any, vData *pb.VersionedBytes) error {
...
 	switch vData.Version {
 	case JSONEncodingVersion1:
-		err = jsonv1.Unmarshal(vData.Data, res)
+		err = json.Unmarshal(vData.Data, res)
 	case JSONEncodingVersion2:
-		err = jsonv2.Unmarshal(vData.Data, res)
+		decoder := json.NewDecoder(bytes.NewBuffer(vData.Data))
+		decoder.UseNumber()
+
+		err = decoder.Decode(res)
```
jmank88, line 88: Why undo this?

jmank88 (changes requested):

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
jmank88: > To be fair, it seems odd that they have different encoding/decoding versions because only go would know the difference...

Indeed, these should be renamed from `Encoding` to `Encoder`

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
EasterTheBunny: Sounds good. I'll switch it back.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
EasterTheBunny: I switched it back and on reviewing the docs, there was an option to stringify numbers, which kinda made sense with what we were working with so I added that.

jmank88: Does that necessarily work the same way?
I think we'd need something like RawValue: https://pkg.go.dev/github.com/go-json-experiment/json@v0.0.0-20240524174822-2d9f40f7385b#example-WithUnmarshalers-RawNumber
Because even with `StringifyNumbers`, unmarshaling without a particular target type in mind results in `float64` values.

pkg/codec/utils.go
nolag: Won't you need to convert back to a Number later?  This only goes from Number => represented type, but doesn't go the other way.

nolag (approved):

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
EasterTheBunny: The float/int problem wasn't exposed in the grpc encoding/decoding. It was in the `ChainReader` config encoding/decoding due to `map[string]any` being used as types to decode into.

ex.
```
{"hardCodeProp": 3} --> decodes to map[string]any{"hardCodeProp": float64(3)}
```

Then the hard coder adds a property to a type:
```
type A struct {
    B string
}

type NewType struct {
    B string
    NewProp float64
}
```

When `GetLatestValue` is called with the following type, the error would present itself where the `NewProp` is being added by the hard-coder. The expected type is an 'int', but the hard-coder thinks it should be a `float64` because of the config decoding.
```
type Expected struct {
    B string
    NewProp int
}
```

EasterTheBunny: To be fair, anytime I switched the grpc encoding/decoding to json, I ran into more errors on when running the evm tests. If we intend to use this in the future, I think we will need to explore a lot more than just numbers.

jmank88: I'm having trouble following this scenario. But if we don't have active test coverage on these encodings, then IMHO we shouldn't be messing with them until that is added.

PR #557, opened 2024-06-04
[KS-241] Include report ID in transmitter outputs

bolekk:
1. Pass report ID.
2. Add a type for convenient unwrapping.
3. Extend Metadata with missing DONConfigVersion (set to zero for now).

### DIFF
--- a/pkg/capabilities/consensus/ocr3/types/aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/types/aggregator.go
@@ -10,14 +10,15 @@ import (
 const MetadataFieldName = "INTERNAL_METADATA"
 
 type Metadata struct {
-	Version       uint32 // 1 byte
-	ExecutionID   string // 32 hex bytes
-	Timestamp     uint32 //  4 bytes
-	DONID         string //  4 hex bytes
-	WorkflowID    string // 32 hex bytes
-	WorkflowName  string // 10 hex bytes
-	WorkflowOwner string // 20 hex bytes
-	ReportID      string //  2 hex bytes
+	Version          uint32 //  1 byte
+	ExecutionID      string // 32 hex bytes
+	Timestamp        uint32 //  4 bytes
+	DONID            string //  4 hex bytes
+	DONConfigVersion uint32 //  4 bytes
+	WorkflowID       string // 32 hex bytes
+	WorkflowName     string // 10 hex bytes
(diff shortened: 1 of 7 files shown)


pkg/capabilities/consensus/ocr3/types/encoder.go
```diff
@@ -11,3 +11,10 @@ type Encoder interface {
 }
 
 type EncoderFactory func(config *values.Map) (Encoder, error)
+
+type SignedReport struct {
+	Report     []byte
+	Context    []byte
```
krehermann, line 17: what does Context represent? please add a comment and/or change to a more meaningful name if appropriate.

should any of these be fixed length? add inline comments if there are length assumptions in other part of the code

krehermann (approved):

pkg/capabilities/consensus/ocr3/types/aggregator.go
```diff
@@ -10,14 +10,15 @@ import (
...
-	DONID         string //  4 hex bytes
-	WorkflowID    string // 32 hex bytes
-	WorkflowName  string // 10 hex bytes
-	WorkflowOwner string // 20 hex bytes
-	ReportID      string //  2 hex bytes
+	Version          uint32 //  1 byte
+	ExecutionID      string // 32 hex bytes
+	Timestamp        uint32 //  4 bytes
+	DONID            string //  4 hex bytes
+	DONConfigVersion uint32 //  4 bytes
```
bolekk, line 17: @krehermann added this field since you last reviewed. It's set to zero now anyway, until Syncer is ready.

krehermann (approved):

pkg/capabilities/consensus/ocr3/types/aggregator.go
archseer: Are we not passing it from the workflow for now, like the donID? They should be sourced from the same place in the future

bolekk: Yeah it will be passed in the same way but we won't have it available until the Syncer is ready.

bolekk: hmm do you mean that the logic to actually pass it though all the layers is missing... you're right, let me add it

archseer (approved):

PR #560, opened 2024-06-05
[fix] Handle negative big ints correctly

cedric-cordenier:


pkg/values/pb/values.proto
```diff
@@ -13,10 +13,15 @@ message Value {
     List list_value = 5;
     string decimal_value = 6;
     int64 int64_value = 7;
-    bytes bigint_value = 8;
+    BigInt bigint_value = 8;
```
bolekk, line 16: can we use 9 and mark 8 as reserved?

pkg/values/pb/values.proto
```diff
@@ -13,10 +13,15 @@ message Value {
...
   }
 }
 
+message BigInt {
+     bytes abs_val = 1;
+     int64 sign = 2;
+}
```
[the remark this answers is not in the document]
cedric-cordenier: It's totally isolated; I'm not sure the deduplication here is worth it

pkg/values/pb/values.proto
```diff
@@ -13,10 +14,15 @@ message Value {
     List list_value = 5;
     string decimal_value = 6;
     int64 int64_value = 7;
-    bytes bigint_value = 8;
```
jmank88, line 16: I think there is a formal way to reserve or deprecate this field

krehermann (approved):

bolekk (approved):

pkg/values/pb/values.proto
jmank88: Oh I just expected it to be inline here inside of the `oneof` - does it matter?

PR #562, opened 2024-06-05
Define ccip ocr3 provider interfaces

dimkouv:
Define ccip ocr3 provider interfaces.

Interfaces of the plugins are similar, (except report encoding method) I added separate definition to keep plugins decoupled, in the near future some more methods will be added.


pkg/types/provider_ccip_ocr3.go
```diff
@@ -0,0 +1,21 @@
+package types
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+type CCIPCommitProviderOCR3 interface {
```
jmank88, line 9: What about naming these as `CCIPOCR3*Provider`?

jmank88 (approved):

makramkd (approved):

PR #565, opened 2024-06-06
[KS-293] Special error value for empty reports

bolekk:


pkg/capabilities/capabilities.go
```diff
@@ -16,6 +16,8 @@ import (
 // CapabilityType is an enum for the type of capability.
 type CapabilityType int
 
+const ErrStopExecution = "__workflow_stop_execution"
```
cedric-cordenier, line 19: Err implies that it's an error, so this is a bit misleading; can we actually use an error value and return that when we want to signal interruption?

One edge case is that when the error value crosses the loop boundary it will no longer be this sentinel value, but I think we can handle this by overriding the errors Is method to check the error string and return true if they match.

cedric-cordenier (approved):

ettec (approved):

PR #566, opened 2024-06-10
feat(observability-lib): refactor filters for Core Node General dashboard

Atrax1:


observability-lib/core-node/component.go
```diff
@@ -1058,7 +1073,7 @@ func nodesRPC(p Props) []cog.Builder[dashboard.Panel] {
 		common.BigValueTextModeValueAndName,
 		common.VizOrientationHorizontal,
 		utils.PrometheusQuery{
-			Query:  `sum(multi_node_states{` + p.PlatformOpts.LabelQuery + `chainId=~"$evmChainID", state="Alive"}) by (` + p.PlatformOpts.LegendString + `, chainId)`,
+			Query:  `sum(multi_node_states{` + p.PlatformOpts.LabelQuery + `state="Alive"}) by (` + p.PlatformOpts.LegendString + `, chainId)`,
```
jmank88, line 1076: How about capturing these fields (or the whole PlatformOpts) in vars to make the queries a little more readable?
```suggestion
			Query:  `sum(multi_node_states{` + labelQuery + `state="Alive"}) by (` + legendString + `, chainId)`,
```
```suggestion
			Query:  `sum(multi_node_states{` + o.LabelQuery + `state="Alive"}) by (` + o.LegendString + `, chainId)`,
```

Atrax1: its quite a big refactor, I could do it but in another PR as its not the scope of this one

jmank88 (approved):

observability-lib/core-node/component.go
```diff
@@ -433,7 +448,7 @@ func logPoller(p Props) []cog.Builder[dashboard.Panel] {
 		"",
 		common.LegendPlacementBottom,
 		utils.PrometheusQuery{
-			Query:  `avg by (evmChainID) (rate(log_poller_logs_inserted{job=~"$instance", evmChainID=~"$evmChainID"}[$__rate_interval]))`,
+			Query:  `avg by (evmChainID) (rate(log_poller_logs_inserted{` + p.PlatformOpts.LabelQuery + `}[$__rate_interval]))`,
```
patrickhuie19, line 451: @Atrax1 for my understanding: how does `evmChainId` end up in p.PlatformOpts.LabelQuery now?

Atrax1: it does not but we are averaging by evmChainId so it shows value for each evmChainId and then we can use legend to filter

Atrax1: value of p.PlatformOpts.LabelQuery is depending on https://github.com/smartcontractkit/chainlink-common/blob/main/observability-lib/core-node/platform.go#L41 and variables at the top of dashboard where we filter on is defined here : https://github.com/smartcontractkit/chainlink-common/pull/566/files#diff-00b43326a56b37178a046bdc93b3f19a6484111b97fbf9c8af147c9aa2e50076R86

patrickhuie19 (approved):

observability-lib/core-node/component.go
patrickhuie19: My misunderstanding was that we still wanted to filter by evmChainID. We do not, you want a view that has all the chains together.

patrickhuie19 (approved):

jmank88 (approved):

PR #567, opened 2024-06-11
ks-254 - ocr3 capability late request handling

ettec:
change to address ks-254-> https://smartcontract-it.atlassian.net/browse/KS-254

### DIFF
--- a/pkg/capabilities/consensus/ocr3/capability.go
+++ b/pkg/capabilities/consensus/ocr3/capability.go
@@ -9,6 +9,7 @@ import (
 	"github.com/jonboulle/clockwork"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/requests"
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
@@ -34,12 +35,12 @@ var info = capabilities.MustNewCapabilityInfo(
 type capability struct {
 	services.StateMachine
 	capabilities.CapabilityInfo
-	capabilities.Validator[config, inputs, outputs]
+	capabilities.Validator[config, inputs, requests.Response]
 
-	store  *store
-	stopCh services.StopChan
-	wg     sync.WaitGroup
-	lggr   logger.Logger
+	reqHandler *requests.Handler
+	stopCh     services.StopChan
+	wg         sync.WaitGroup
+	lggr       logger.Logger
 
 	requestTimeout time.Duration
 	clock          clockwork.Clock
@@ -50,22 +51,19 @@ type capability struct {
 	encoderFactory types.EncoderFactory
 	encoders       map[string]types.Encoder
 
-	transmitCh chan *outputs
-	newTimerCh chan *request
-
 	callbackChannelBufferSize int
 }
 
 var _ capabilityIface = (*capability)(nil)
 var _ capabilities.ConsensusCapability = (*capability)(nil)
-var ocr3CapabilityValidator = capabilities.NewValidator[config, inputs, outputs](capabilities.ValidatorArgs{Info: info})
+var ocr3CapabilityValidator = capabilities.NewValidator[config, inputs, requests.Response](capabilities.ValidatorArgs{Info: info})
 
-func newCapability(s *store, clock clockwork.Clock, requestTimeout time.Duration, aggregatorFactory types.AggregatorFactory, encoderFactory types.EncoderFactory, lggr logger.Logger,
+func newCapability(s *requests.Store, clock clockwork.Clock, requestTimeout time.Duration, aggregatorFactory types.AggregatorFactory, encoderFactory types.EncoderFactory, lggr logger.Logger,
 	callbackChannelBufferSize int) *capability {
 	o := &capability{
 		CapabilityInfo:    info,
 		Validator:         ocr3CapabilityValidator,
-		store:             s,
+		reqHandler:        requests.NewHandler(lggr, s, clock, requestTimeout),
 		clock:             clock,
 		requestTimeout:    requestTimeout,
 		stopCh:            make(chan struct{}),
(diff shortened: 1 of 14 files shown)


pkg/capabilities/consensus/ocr3/transmitter_test.go
```diff
@@ -8,13 +8,15 @@ import (
...
-	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/requests"
+	"github.com/smartcontractkit/libocr/offchainreporting2/types"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+
```
jmank88, line 19:
```suggestion
	"github.com/smartcontractkit/libocr/offchainreporting2/types"
	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"

	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/requests"
```

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -0,0 +1,146 @@
...
+	return len(h.responseCache)
+}
+
+func (h *Handler) Start(_ context.Context) error {
+	return h.StartOnce("RequestHandler", func() error {
+		h.wg.Add(1)
+		go func() {
+			defer h.wg.Done()
+			ctx, cancel := h.stopCh.NewCtx()
+			defer cancel()
```
bolekk, line 76: You can put these 3 lines inside checkForResponseExpiry() and just run "go checkForResponseExpiry()" instead of having another function around it

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (h *Handler) SendRequest(r *Request) error {
+	resp, err := h.addRequest(r)
+	if err != nil {
+		return fmt.Errorf("failed to add request: %w", err)
+	}
+
+	if resp != nil {
+		r.sendResponse(resp)
```
bolekk, line 56: Is there a deadlock risk if we send response immediately from the same goroutine? Maybe @cedric-cordenier can confirm?

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for k, v := range h.responseCache {
+		if time.Since(v.entryTime) > h.cacheExpiryTime {
+			delete(h.responseCache, k)
+			h.lggr.Debugw("Expired response", "workflowExecutionID", k)
+		}
+	}
+}
+
+func (h *Handler) getRequestForResponse(resp *Response) *Request {
```
bolekk, line 116: Name of this function should reflect what it does. It's not only a getter, it actually does all the magic - caches the response for the future :)

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -0,0 +1,146 @@
...
+	h.mux.Lock()
+	defer h.mux.Unlock()
+
+	req, wasPresent := h.store.evict(resp.WorkflowExecutionID)
+	if !wasPresent {
+		h.responseCache[resp.WorkflowExecutionID] = &responseCacheEntry{
```
bolekk, line 122: Debug-log: "caching response without a local request"

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return req
+}
+
+func (h *Handler) addRequest(r *Request) (*Response, error) {
+	h.mux.Lock()
+	defer h.mux.Unlock()
+	existingResponse := h.responseCache[r.WorkflowExecutionID]
+	if existingResponse != nil {
+		delete(h.responseCache, r.WorkflowExecutionID)
```
bolekk, line 137: Debug-log: "serving existing response from cache"

bolekk (approved):
Looks good overall! This should solve the problem.
I left a few minor comments and we definitely need to get @cedric-cordenier's eyes on whether the coroutines here play nicely with the engine.

pkg/capabilities/consensus/ocr3/requests/handler.go
ettec: :) same reason as other PR, I think it is better to colocate the code that coordinates the routines using wait groups in the Start and Close methods as it makes the threading used by the service clearer (IMO), its a preference I am not wedded to it

pkg/capabilities/consensus/ocr3/requests/handler.go
cedric-cordenier: Yes it is possible -- all responses should go via the transmitCh, which also serialises the responses

cedric-cordenier (dismissed):

cedric-cordenier:
@ettec LGTM, thanks for following up on our offline review 👍

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func NewHandler(lggr logger.Logger, s *Store, clock clockwork.Clock, responseExpiryTime time.Duration) *Handler {
+	return &Handler{
+		lggr:            lggr,
+		store:           s,
+		pendingRequests: map[string]*Request{},
+		responseCache:   map[string]*responseCacheEntry{},
+		responseCh:      make(chan *Response),
+		requestCh:       make(chan *Request),
```
bolekk, line 49: Is there a risk that Handler will get closed first and then the Capability will hang forever on SendRequest()?

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -254,67 +257,6 @@ func (o *capability) queueRequestForProcessing(
 
 	o.lggr.Debugw("Execute - adding to store", "workflowID", r.WorkflowID, "workflowExecutionID", r.WorkflowExecutionID, "observations", r.Observations)
```
bolekk, line 258: Slightly outdated comment

pkg/capabilities/consensus/ocr3/requests/handler.go
ettec: true, it shoudl pass in the cts and select on that, sorting......

bolekk (approved):

cedric-cordenier (approved):

PR #569, opened 2024-06-11
ccip ocr3 - commit plugin protos

dimkouv:


pkg/types/ccipocr3/generic_types.go
```diff
@@ -83,9 +83,25 @@ func (c ChainSelector) String() string {
 }
 
 type CCIPMsg struct {
```
winder, line 85: Can we get rid of these and use the protobuf versions for everything?

pkg/loop/internal/pb/ccipocr3/models.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/ccipocr3;ccipocr3pb";
+
+package loop.internal.pb.ccipocr3;
+
+// CCIPMsg is a gRPC adapter to [github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3.CCIPMsg]
+message CCIPMsg {
```
makramkd, line 8: With the family-agnostic ramp work (currently in spike and ongoing) this may be affected. It may end up to be something like this:

```proto
message EVM2AnyMessage {
  // ... fields
}

message Any2EVMMessage {
  // .. fields
}

message CCIPMessage {
  oneof message_payload {
    EVM2AnyMessage evm_2_any_message = 1;
    Any2EVMMessage any_2_evm_message = 5;
  }
}
```

pkg/loop/internal/pb/ccipocr3/models.proto
```diff
@@ -0,0 +1,68 @@
...
+  BigInt fee_token_amount = 11;
+  bytes data = 12;
+  repeated TokenAmount token_amounts = 13;
+  repeated bytes source_token_data = 14;
+}
+
+// BigInt represents a [big.Int].
+message BigInt {
+  bool negative = 1;
+  bytes value = 2;
```
makramkd, line 28: Is this `bool` needed? Doesn't `BigInt`'s `Bytes()` method use 2's complement to encode negative/non-negative? I think just `bytes` is enough?

pkg/types/ccipocr3/generic_types.go
makramkd: That would be nice, though the reason I'm somewhat hesitant is that these types are kind of used for different reasons: one set of types is for the OCR plugin to express information (right now marshaling to JSON) and one is to communicate w/ gRPC implementations of various interfaces. I wonder if these may diverge at some point or not, I'm not sure right now if they will or won't.

If we're confident they will always be in sync then I'm in favor of just using the protobuf generated files and axing the duplication

makramkd: If we have good tooling for it (i.e when we log observations, we don't log the protobuf but a human-readable version) it'd be a win

pkg/loop/internal/pb/ccipocr3/msghasher.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/ccipocr3;ccipocr3pb";
+
+package loop.internal.pb.ccipocr3;
+
+import "models.proto";
+
+service MsgHasher {
```
makramkd, line 9: Lets put a comment on what this service will do and how instances of it are implemented/organized (i.e per chain, per chain family, etc.). Note that all comments you put in the proto files are automatically ported to the generated code (in this case Go) so its not wasteful.

e.g

```suggestion
// MsgHasher constructs the message hash of the provided message in a chain-family-specific
// fashion. Thus, there is a different implementation of the MsgHasher for each chain family (e.g,
// EVM, Solana, Starknet, etc.).
service MsgHasher {
```

pkg/loop/internal/pb/ccipocr3/reportcodec.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/ccipocr3;ccipocr3pb";
+
+package loop.internal.pb.ccipocr3;
+
+import "models.proto";
+
+// CommitPluginCodec is a service that can encode and decode CommitPluginReport.
```
makramkd, line 9: This is a good comment but lets also add when this is implemented (in this case I think its per family).

pkg/types/ccipocr3/generic_types.go
dimkouv: I like this idea. I have same concern with Makram but we can decouple it after we reach that point.

dimkouv: After trying to write the code though it starts getting complicated.
I added a ticket to properly investigate this approach.

pkg/loop/internal/pb/ccipocr3/models.proto
dimkouv: yeap, exactly

pkg/loop/internal/pb/ccipocr3/models.proto
dimkouv: agreed

pkg/types/ccipocr3/generic_types.go
jmank88, line 85: Pre-existing, but this type name is a bit redundant given the package: `ccipocr3.CCIPMsg`. Could be `ccipocr2.Msg`

dimkouv: I'd prefer to keep it like this because we use some alias on imports and it might become `types.Msg`

pkg/loop/internal/pb/ccipocr3/models.proto
```diff
@@ -0,0 +1,66 @@
...
+message BigInt {
+  bytes value = 1;
+}
+
+// TokenAmount is a helper type that defines a token and an amount.
+message TokenAmount {
+  string token = 1; // Token address
+  BigInt amount = 2;
+}
+
```
makramkd, line 35: nit: unnecessary newline

makramkd (approved):

krebernisak (approved):

pkg/types/ccipocr3/generic_types.go
jmank88: IMHO that is a very backwards way of addressing that problem

jmank88 (approved):

PR #573, opened 2024-06-11
 [BCF-3247] - Remove pending from boundContract and use confidence instead for CR

ilija42:
core ref: df3b03391fed8594b9bb0a6d9d625cb45c8fbdaa
solana ref: 458292c84c696c341df0164a260e191b9e8dffcf
### Description

-  Modified proto and Client Server defs to support confidence levels for `GetLatestValue`
-  Wrote tests, but geth SimulatedBackend doesn't seem to support reading anything but the latest block, which breaks the tests, so I commented them out. Still confirmed that everything works, but not sure what to do with this.

### Related PRs
[chainlink](https://github.com/smartcontractkit/chainlink/pull/13514)
[chainlink-feeds](https://github.com/smartcontractkit/chainlink-feeds/pull/72)
[chainlink-solana](https://github.com/smartcontractkit/chainlink-solana/pull/775)


pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -130,18 +130,25 @@ func DecodeVersionedBytes(res any, vData *pb.VersionedBytes) error {
 	return nil
 }
 
-func (c *Client) GetLatestValue(ctx context.Context, contractName, method string, params, retVal any) error {
+func (c *Client) GetLatestValue(ctx context.Context, contractName, method string, confidenceLevel primitives.ConfidenceLevel, params, returnVal any) error {
```
jmank88, line 133: Out of scope I guess, but why do we have a separate `package primitives`?

ilija42: Because things like `ConfidenceLevel`, `Block`, `TxHash`... don't fit under query package but they are primitives used for querying so they live under `pkg/types/query/primitives/primitives.go`. Also some of the structs aren't just used for querying, confidenceLevels are used in a bunch of other places, so having to use it with query.ConfidenceLevel would be weird.
Do you have an alternative?

jmank88: It sounds like some things could be in `package query`, but other things would not make sense there. If so, why are we grouping them together in the first place? What about moving the query-specific things to `package query`, and then doing something else with the rest?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -211,16 +218,24 @@ func (c *Server) GetLatestValue(ctx context.Context, request *pb.GetLatestValueR
...
 		return nil, err
 	}
-	err = c.impl.GetLatestValue(ctx, request.ContractName, request.Method, params, retVal)
+
+	var confidenceLevel primitives.ConfidenceLevel
+	if request.Confidence == pb.Confidence_Unconfirmed {
+		confidenceLevel = primitives.Unconfirmed
+	} else if request.Confidence == pb.Confidence_Finalized {
+		confidenceLevel = primitives.Finalized
+	}
```
jmank88, line 231: Should we package this up in a conversion helper func?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
ilija42: BCF-3295

jmank88 (approved):

krehermann (approved):

PR #574, opened 2024-06-11
Adjusting types for ocr3 exec plugin

winder:
This is likely to change further, but it's at a good point for pushing the  next round of changes in the exec plugin.


pkg/types/ccipocr3/config.go
```diff
@@ -58,13 +58,19 @@ func (c CommitPluginConfig) Validate() error {
...
 
 	// MessageVisibilityInterval is the time interval for which the messages are visible by the plugin.
-	MessageVisibilityInterval time.Duration
+	MessageVisibilityInterval time.Duration `json:"messageVisibilityInterval"`
+
+	// FChain defines the FChain value for each chain. FChain is used while forming consensus based on the observations.
+	FChain map[ChainSelector]int `json:"fChain"`
+
+	// F is the ocr3 big F value for the execute plugin.
+	F int `json:"F"`
```
0xnogo, line 73: Is it a fallback `F` if the chain selector has not value in `FChain`?

0xnogo (approved):

pkg/types/ccipocr3/plugin_execute_types.go
```diff
@@ -77,23 +84,16 @@ func DecodeExecutePluginObservation(b []byte) (ExecutePluginObservation, error)
...
 
 func NewExecutePluginOutcome(
-	nextCommits ExecutePluginCommitObservations,
-	messages ExecutePluginMessageObservations,
+	nextCommits []ExecutePluginCommitDataWithMessages,
 ) ExecutePluginOutcome {
 	return ExecutePluginOutcome{
-		NextCommits: nextCommits,
-		Messages:    messages,
+		PendingCommitReports: nextCommits,
```
dimkouv, line 96: maybe stick to one name instead of both `next` and `pending` ?

pkg/types/ccipocr3/config.go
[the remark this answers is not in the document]
winder: I named it this way since we've sometimes referred to this as big f and little f. I don't have a strong opinion either way, I suppose we've moved on to f and fChain

pkg/types/ccipocr3/config.go
```diff
@@ -58,13 +58,19 @@ func (c CommitPluginConfig) Validate() error {
...
 
 	// MessageVisibilityInterval is the time interval for which the messages are visible by the plugin.
-	MessageVisibilityInterval time.Duration
+	MessageVisibilityInterval time.Duration `json:"messageVisibilityInterval"`
+
+	// FChain defines the FChain value for each chain. FChain is used while forming consensus based on the observations.
+	FChain map[ChainSelector]int `json:"fChain"`
+
+	// F is the ocr3 big F value for the execute plugin.
+	F int `json:"f"`
```
makramkd, line 73: Does this need to be in here? This is passed into the `NewReportingPlugin` call directly. FChain isn't, however.

winder: Oh nice. I'll use the reporting cfg instead.

makramkd (approved):

jmank88 (approved):

PR #576, opened 2024-06-12
KS 318 - name owner in wf yaml

krehermann:
https://smartcontract-it.atlassian.net/browse/KS-318

Moves the name and owner fields into the YAML spec.
- adds validation via jsonschema

Adds CID (content hash), name and owner to the parse WorkflowSpec struct
- this enables reuse of validation logic in both core and clo

Supports:
- https://github.com/smartcontractkit/chainlink/pull/13522

### DIFF
--- a/pkg/capabilities/consensus/ocr3/types/aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/types/aggregator.go
@@ -1,6 +1,8 @@
 package types
 
 import (
+	"strings"
+
 	ocrcommon "github.com/smartcontractkit/libocr/commontypes"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
@@ -21,6 +23,16 @@ type Metadata struct {
 	ReportID         string //  2 hex bytes
 }
 
+// the contract requires exactly 10 bytes for the workflow name
+// the json schema allows for a variable length string <= len(10)
+// pad with trailing spaces to meet the contract requirements
+func (m *Metadata) padWorkflowName() {
+	if len(m.WorkflowName) < 10 {
+		suffix := strings.Repeat(" ", 10-len(m.WorkflowName))
+		m.WorkflowName += suffix
+	}
+}
+
 type Aggregator interface {
 	// Called by the Outcome() phase of OCR reporting.
 	// The inner array of observations corresponds to elements listed in "inputs.observations" section.
@@ -28,6 +40,7 @@ type Aggregator interface {
 }
 
 func AppendMetadata(outcome *AggregationOutcome, meta *Metadata) (*AggregationOutcome, error) {
+	meta.padWorkflowName()
 	metaWrapped, err := values.Wrap(meta)
 	if err != nil {
 		return nil, err
--- a/pkg/workflows/models_yaml_test.go
+++ b/pkg/workflows/models_yaml_test.go
@@ -27,7 +27,6 @@ func yamlFixtureReaderObj(t *testing.T, testCase string) func(name string) any {
 		var testFileYaml any
 		err := yaml.Unmarshal(testFileBytes, &testFileYaml)
 		require.NoError(t, err)
-		fmt.Sprintf("%s", &testFileBytes)
 		return testFileYaml
 	}
 }
@@ -163,11 +162,9 @@ func TestWorkflowSpecMarshalling(t *testing.T) {
 			t.FailNow()
 		}
 	})
-
 }
 
 func TestJsonSchema(t *testing.T) {
-
 	t.Parallel()
 	t.Run("GenerateJsonSchema", func(t *testing.T) {
 		expectedSchemaPath := fixtureDir + "workflow_schema.json"
@@ -300,7 +297,6 @@ func TestJsonSchema(t *testing.T) {
 			})
 		}
 	})
-
 }
 
 func TestMappingCustomType(t *testing.T) {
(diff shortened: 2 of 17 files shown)


pkg/workflows/testutils.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"bytes"
+	"testing"
+	"text/template"
+
+	"github.com/stretchr/testify/require"
+)
+
+// WFYamlSpec generates a validate yaml spec for a workflow for the given name and owner
+func WFYamlSpec(t *testing.T, name, owner string) string {
```
krehermann, line 12: this is a standalone utility because it's useful in core tests that need to create job specs

pkg/workflows/testdata/fixtures/workflows/workflow_schema.json
```diff
@@ -64,6 +64,15 @@
     },
     "workflowSpecYaml": {
       "properties": {
+        "name": {
+          "type": "string",
+          "maxLength": 10,
+          "minLength": 10
```
DeividasK, line 70: Could this not be less than 10?

pkg/workflows/models_yaml_test.go
```diff
@@ -222,15 +218,84 @@ func TestJsonSchema(t *testing.T) {
...
+				name:     "ten_digits",
+				owner:    "0x0123456789abcdef0123456789abcdef01234567",
+			},
+			{
+				testName: "valid: no name",
+				// the helper function will omit the name field, which not the same as an empty string for a name
+				owner: "0x0123456789abcdef0123456789abcdef01234567",
+			},
+			{
+				testName: "invalid: name to long",
```
DeividasK, line 248:
```suggestion
				testName: "invalid: name too long",
```

DeividasK: there's more below

DeividasK (approved):

pkg/workflows/models_yaml_test.go
```diff
@@ -176,7 +172,7 @@ func TestJsonSchema(t *testing.T) {
 		require.NoError(t, err)
 
 		// change this to update golden file
-		shouldUpdateSchema := false
+		shouldUpdateSchema := true
```
cedric-cordenier, line 175: @krehermann You forgot to flip this back to `false`

pkg/workflows/testutils.go
cedric-cordenier: Could you rename it to utils_test.go? That way it will only be included in tests?

cedric-cordenier: I'm not a huge fan of this refactoring, this feels like a case of "a little duplication is better than a little coupling" to me and we could add this to whatever package needs it.

krehermann: hmm, i started by copying then moved it here

 the packages are strongly coupled though. you can't do *anything useful* with a TOML spec without this package because the workflow job TOML spec depends on this pkg. given that, it seemed like this package ought to be responsible for providing some helpers.

said differently, if we copy this, and we change the yaml and/or validation in this repo, then we will not only need to bump the core repo, but also fix the embedded string literal to match the change.

wdyt @cedric-cordenier

pkg/workflows/testdata/fixtures/workflows/workflow_schema.json
krehermann: how should this be encoded for the contract? the contract forces exactly ten bytes.

i agree that this is pretty gross, but i'm concerned something will go wrong in the serialization.

pkg/workflows/testdata/fixtures/workflows/workflow_schema.json
```diff
@@ -64,6 +64,15 @@
...
+        },
+        "owner": {
+          "type": "string",
+          "pattern": "^0x[0-9a-f]{40}$"
```
HenryNguyen5, line 74: Should we allow for uppercase addresses too? They may be copy/pasting in checksummed addresses

krehermann: sure, will add.

pkg/workflows/testdata/fixtures/workflows/workflow_schema.json
krehermann: we decided to add padding. see aggregator.Metadata.padWorkflowName

pkg/capabilities/consensus/ocr3/types/aggregator.go
```diff
@@ -21,13 +23,24 @@ type Metadata struct {
 	ReportID         string //  2 hex bytes
 }
 
+// the contract requires exactly 10 bytes for the workflow name
+// the json schema allows for a variable length string <= len(10)
+// pad with trailing spaces to meet the contract requirements
+func (m *Metadata) padWorkflowName() {
+	if len(m.WorkflowName) < 10 {
+		suffix := strings.Repeat(" ", 10-len(m.WorkflowName))
```
DeividasK, line 31: Actually, I think it might make sense to allow spaces in the name 🙈

krehermann: we don't allow spaces in any of the other identifiers (like ref). it seems easier and consistent to keep it simple here

cedric-cordenier (approved):

pkg/workflows/testutils.go
cedric-cordenier: Discussed offline, I'm happy to go with this in the end

DeividasK (approved):

PR #578, opened 2024-06-13
KS-318: Hide derived field

krehermann:
This is a quick follow up to https://github.com/smartcontractkit/chainlink-common/pull/576

When refactoring CLO, I realized that they need access 1. parsing validation 2. the original YAML so it can be embedded in the job TOML.

The CLO implementation had been re-marshaling the WorkflowSpec after parsing validation to achieve #2. This worked by accident and was broken by adding the CID.

Here we hide the derived CID field, and an accessor for the original YAML spec, and add tests to show that logically equivalent YAMLs produce different CIDs.

Supports:
- https://github.com/smartcontractkit/feeds-manager/pull/3359


jmank88 (dismissed):

pkg/workflows/dependency_graph.go
```diff
@@ -28,14 +28,29 @@ type StepDefinition struct {
...
-	Owner     string           `json:"owner,omitempty"`           // empty owner allowed for anonymous workflows
+	Name      string           `json:"name,omitempty"`
+	Owner     string           `json:"owner,omitempty"`
 	Triggers  []StepDefinition `json:"triggers" jsonschema:"required"`
 	Actions   []StepDefinition `json:"actions,omitempty"`
 	Consensus []StepDefinition `json:"consensus" jsonschema:"required"`
 	Targets   []StepDefinition `json:"targets" jsonschema:"required"`
+
+	cid  string // content hash of the original yaml spec
+	yaml string // original yaml spec
```
cedric-cordenier, line 43: @krehermann Note: CID `json:"-"` achieves the same thing without needing to add accessors to private attributes

cedric-cordenier (approved):

bolekk (approved):

PR #579, opened 2024-06-13
Prune previous outcomes for deleted workflows

vyzaldysanchez:


pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -216,6 +218,16 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
 		o.Outcomes[weid.WorkflowId] = outcome
 	}
 
+	// Prune previous outcomes for deleted workflows
+	var prunedWorkflows []string
+	for _, wfID := range r.r.getDeregisteredWorkflows() {
```
bolekk, line 223: I'm afraid this is not going to be that simple. Outcome() method needs to be pure, not related to any local state. We need to receive 2F+1 identical outcomes, otherwise we risk deadlocking OCR. 

One idea I had is to ask nodes to send a full list of workflow IDs that they think are still alive inside their observations. Then in the Outcome we could weed out the ones that have fewer than F+1 votes. @cedric-cordenier thoughts?

vyzaldysanchez: Understood

cedric-cordenier: @bolekk 🤦  So we're effectively getting consensus on the workflows we want to prune. That makes sense I think, but doesn't that mean we would effectively be propagating the deregistration? Is that desirable? I think that could have weird side effects for running executions, but maybe that's OK

bolekk: Execution would fail the consensus step anyway, if enough nodes already deleted that workflow. I'm open to other approaches, I just don't see any good alternatives atm.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -52,6 +52,9 @@ type capability struct {
 	encoders       map[string]types.Encoder
 
 	callbackChannelBufferSize int
+
+	registeredWorkflows   map[string]struct{}
```
cedric-cordenier, line 56: Nit: I often find it easier to use a map[string]bool since you can then use the following shortform to test presence:

`if registeredWorkflows["aKey"] ...` rather than having to check the presence operator via `_, ok := registeredWorkflows["aKey"]`

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -109,6 +114,9 @@ func (o *capability) HealthReport() map[string]error {
 }
 
 func (o *capability) RegisterToWorkflow(ctx context.Context, request capabilities.RegisterToWorkflowRequest) error {
+	o.registeredWorkflowsMu.Lock()
+	defer o.registeredWorkflowsMu.Unlock()
```
cedric-cordenier, line 118: For more cohesion you could move these two lines just above where it's used; that also minimises the amount of time the lock is held to the minimum necessary

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -216,9 +226,23 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
...
+			// Update the last seen round for this outcome. But this should only happen if the workflow is seen by F+1 nodes
+			outcome.LastSeenAt = outctx.SeqNr
+		}
+
 		o.Outcomes[weid.WorkflowId] = outcome
 	}
 
+	// We need to prune outcomes from previous workflows that are no longer relevant.
+	for workflowID, outcome := range o.Outcomes {
+		// TODO: 3,600 is the amount of rounds we allow as threshold. This should be configurable.
```
cedric-cordenier, line 241: We should document somewhere that we are making the assumption that 1 round == 1 sec

vyzaldysanchez: agreed

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -216,9 +228,23 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
 		o.CurrentReports = append(o.CurrentReports, report)
 		allExecutionIDs = append(allExecutionIDs, weid.WorkflowExecutionId)
 
+		if seenWorkflows[weid.WorkflowId] >= (r.config.F + 1) {
```
cedric-cordenier, line 231: This is incorrect since it will only update LastSeenAt if the id was part of the query (q.Ids); we want to update this for all registered workflows regardless of whether they were in the query or not (of course assuming F+1)

cedric-cordenier: So you probably want to loop over seenWorkflows

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -74,6 +77,8 @@ func newCapability(s *requests.Store, clock clockwork.Clock, requestTimeout time
 		encoders:          map[string]types.Encoder{},
 
 		callbackChannelBufferSize: callbackChannelBufferSize,
+
```
bolekk, line 80: nit: I'd get rid of this empty line and the one above too

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -135,6 +137,7 @@ func (r *reportingPlugin) ObservationQuorum(outctx ocr3types.OutcomeContext, que
 func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Query, aos []types.AttributedObservation) (ocr3types.Outcome, error) {
 	// execution ID -> oracle ID -> list of observations
 	m := map[string]map[ocrcommon.OracleID][]values.Value{}
+	seenWorkflows := map[string]int{}
```
bolekk, line 140: seenWorkflows -> seenWorkflowIDs?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -143,6 +146,15 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
 			continue
 		}
 
+		for _, id := range obs.RegisteredWorkflowIds {
+			// Count how many times a workflow ID is seen from Observations
+			if _, ok := seenWorkflows[id]; ok {
```
bolekk, line 151: We need protection from malicious nodes sending a list with one ID repeated multiple times.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -178,7 +190,7 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
 	// every time since we only want to transmit reports that
 	// are part of the current Query.
 	o.CurrentReports = []*pbtypes.Report{}
-	allExecutionIDs := []string{}
+	var allExecutionIDs []string
```
bolekk, line 193: Is this better or cleaner in any way?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
bolekk: I'd rather simply state that this is affected by OCR round time, without referencing the 1sec value.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -216,9 +228,23 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
...
+			outcome.LastSeenAt = outctx.SeqNr
+		}
+
 		o.Outcomes[weid.WorkflowId] = outcome
 	}
 
+	// We need to prune outcomes from previous workflows that are no longer relevant.
+	for workflowID, outcome := range o.Outcomes {
+		// TODO: 3,600 is the amount of rounds we allow as threshold. This should be configurable.
+		if outctx.SeqNr-outcome.LastSeenAt > 3600 {
```
bolekk, line 242: Put 3600 into a const or, better yet, just add this config somewhere.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
vyzaldysanchez: should be a cleaner way, suggested by the IDE. Though I have no preference.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
vyzaldysanchez: Any examples of config setup that I could use here, or just set it up as a param when constructing the struct?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
vyzaldysanchez: What if we just make sure the workflows are not repeated on the Observation phase, or you would rather give the cleanup responsibility to Outcome instead @bolekk ?

bolekk: It has to happen here because a malicious node can send a bad observation regardless.

vyzaldysanchez: I agree.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
bolekk: It probably can sit in the onchain config that we extract in newReportingPlugin - BTW I see there is a TODO there that we might need to address seprately. For now you can use a const there and later we will replace it with a configured value, if present.

vyzaldysanchez:
This should now only be missing tests.

pkg/capabilities/consensus/ocr3/types/ocr3_types.proto
```diff
@@ -36,6 +37,8 @@ message Observation {
 message Observations {
   // batched observations for multiple workflow execution IDs
   repeated Observation observations = 1;
+  // the workflow IDs that are registered in the node
+  repeated string registeredWorkflowIds = 2;
```
krehermann, line 41: why does this need to be in the proto now? specifically, why is a field in Observations? it seems out of place. i would have expected either an api to list all the workflows or the observation itself has a workflow id

krehermann (commented):
i don't understand this change.

i expected to see that it was trimming or cleaning up an existing leak or unbounded resource.

but i don't see that, instead i see some new bookkeeping for registeredWorkflows and some adding and delete to that newly introduced struct.

what is it accomplishing relative to the current state?

cedric-cordenier (approved):

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -146,7 +153,27 @@ func (o *capability) getEncoder(workflowID string) (types.Encoder, error) {
 	return enc, nil
 }
 
+func (o *capability) getRegisteredWorkflowsIDs() []string {
+	o.registeredWorkflowsMu.RLock()
+	defer o.registeredWorkflowsMu.RUnlock()
+
+	workflows := make([]string, 0, len(o.registeredWorkflowsIDs))
+	for wf := range o.registeredWorkflowsIDs {
```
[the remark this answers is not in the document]
cedric-cordenier: Yes this should be OK @krehermann : it's only used in the Observation step; only Outcome needs to be pure.

cedric-cordenier:
@krehermann This change effectively means we get consensus on which workflows are active, and prune those workflows that aren't active from the Outcome struct. This ensures the Outcome stored by OCR doesn't grow in size too much

pkg/capabilities/consensus/ocr3/types/ocr3_types.proto
cedric-cordenier: See my answer below: this is so that we get consensus from all nodes on which workflows they think are registered to the capability

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -143,6 +151,19 @@ func (r *reportingPlugin) Outcome(outctx ocr3types.OutcomeContext, query types.Q
 			continue
 		}
 
+		countedWorkflowIds := map[string]bool{}
```
krehermann, line 154: this `countedWorkflowIds` is redundant. it's hard to see, but the the RegisteredWorkflowIds are derived from `getRegisteredWorkflowsIDs` which only contains unique ids becuase it's derived from a map.

this would be see if `getRegisteredWorkflows` -> `listRegisteredWorkflows` and a comment on that func to say that it's the elements are unique.

pkg/capabilities/consensus/ocr3/capability.go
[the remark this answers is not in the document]
krehermann: that's confusing.

you said the point of this change is to get consensus about the workflows that are registered. why should that consensus work if each node can report a non deterministic value of the list of workflows?

pkg/capabilities/consensus/ocr3/reporting_plugin.go
vyzaldysanchez: `countedWorkflowIds` exists for the reason that we need protection from malicious nodes sending a list with one ID repeated multiple times, as @bolekk stated in a previous review. We cannot rely on `Outcome` that what was received is in fact always what `getRegisteredWorkflowsIDs` contains as part of `RegisteredWorkflowIds`.

krehermann: i misread the diff and thought this `obs` was the same as the one on line 128. i was wrong.

krehermann:
> @krehermann This change effectively means we get consensus on which workflows are active, and prune those workflows that aren't active from the Outcome struct. This ensures the Outcome stored by OCR doesn't grow in size too much

thanks. there was so much extra machinery that i missed the `delete` call on the Outcome

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -146,7 +153,27 @@ func (o *capability) getEncoder(workflowID string) (types.Encoder, error) {
...
+func (o *capability) unregisterWorkflowID(workflowID string) {
+	o.registeredWorkflowsMu.Lock()
+	defer o.registeredWorkflowsMu.Unlock()
+	delete(o.registeredWorkflowsIDs, workflowID)
+}
+
 func (o *capability) UnregisterFromWorkflow(ctx context.Context, request capabilities.UnregisterFromWorkflowRequest) error {
+	o.registeredWorkflowsMu.Lock()
+	defer o.registeredWorkflowsMu.Unlock()
+	delete(o.registeredWorkflowsIDs, request.Metadata.WorkflowID)
```
bolekk, line 176: nit: we could probably just call unregisterWorkflowID() here

bolekk (approved):

PR #582, opened 2024-06-14
[BCI-3500] Modify transactionID to be of type string

nickcorin:
We're no longer enforcing the `uuid.UUID` type on `transactionID` as some products would prefer to generate a key in their own way. A `uuid` will still be valid by using `uuidVal.String()`.

### DIFF
--- a/pkg/types/chain_writer.go
+++ b/pkg/types/chain_writer.go
@@ -10,9 +10,9 @@ import (
 type ChainWriter interface {
 	// SubmitTransaction packs and broadcasts a transaction to the underlying chain.
 	//
-	// - `args` specifies input parameters to the contract call.
+	// - `args` should be any object which maps a set of method param into the contract and method specific method params.
 	// - `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
-	SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID uuid.UUID, toAddress string, meta *TxMeta, value big.Int) error
+	SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, meta *TxMeta, value *big.Int) error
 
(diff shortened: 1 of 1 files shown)


pkg/types/chain_writer.go
```diff
@@ -10,9 +10,9 @@ import (
 type ChainWriter interface {
 	// SubmitTransaction packs and broadcasts a transaction to the underlying chain.
 	//
-	// - `args` specifies input parameters to the contract call.
+	// - `args` should be any object which maps a set of method param into the contract and method specific method params.
 	// - `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
```
ilija42, line 14: can you add to the description of the PR why UUID doesn't work?

nickcorin: UUID does work, and it's fine to send a UUID. It's just that it doesn't have to be enforced.

pkg/types/chain_writer.go
```diff
@@ -10,9 +10,9 @@ import (
 type ChainWriter interface {
 	// SubmitTransaction packs and broadcasts a transaction to the underlying chain.
 	//
-	// - `args` specifies input parameters to the contract call.
+	// - `args` should be any object which maps a set of method param into the contract and method specific method params.
```
jmank88, line 13: Did we lose some work here? ChainReader has a much more comprehensive comment, and ChainWriter should either have the same, or we should be using a shared `type SomeAlias any` with the doc in one place.

jmank88 (approved):

pkg/types/chain_writer.go
nickcorin: Yeah, the `args` param changed and the comment didn't get updated accordingly. Probably my fault.

nickcorin: I'll update the docs in a follow up PR.

ilija42 (approved):

silaslenihan (approved):

PR #583, opened 2024-06-14
Enable govet linter printf check for Logger and SugaredLogger.

pavel-raykov:


jmank88 (approved):

.golangci.yml
```diff
@@ -33,6 +33,28 @@ linters-settings:
   govet:
     # report about shadowed variables
     check-shadowing: true
+    settings:
+      printf:
+        # We are overgenerous here to accomodate potential not yet existing functions.
```
jmank88, line 38: Oh we could probably trim this back. The plain Logger is a very limited subset by design, so that things like the official Zap logger implement it. All of the extra stuff was kept in SugaredLogger.

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -167,7 +167,7 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 	}
 	err := r.Inputs.UnwrapTo(&m)
 	if err != nil {
-		o.lggr.Warnf("could not unwrap method from CapabilityRequest, using default: %w", err)
+		o.lggr.Warnf("could not unwrap method from CapabilityRequest, using default: %v", err)
```
patrickhuie19, line 170: If this is getting signaled because `err` isn't of type `error`, then a variable name change might be useful to differentiate.

patrickhuie19 (approved):

pkg/capabilities/consensus/ocr3/capability.go
jmank88: Nah because %w is just for constructing new errors, not formatting for a log message.

krehermann (approved):

pkg/capabilities/consensus/ocr3/capability.go
pavel-raykov: For posterity, here is the discussion why %w should not be used in the logging formatting functions https://github.com/golang/go/issues/47641

patrickhuie19 (dismissed):

pkg/capabilities/consensus/ocr3/capability.go
jmank88: go vet will complain about some std lib cases, but not even all of those are caught until runtime:
https://go.dev/play/p/KQCLIA7ZBM3

.golangci.yml
jmank88: Oh I was actually thinking fewer `Logger` methods, since some are only on `SugaredLogger`, but I may not be up to date with the API....

It looks like for `Logger` we currently have: 
- Debug
- Info
- Warn
- Error
- Panic
- Fatal

And for `SugaredLogger` we have:
- AssumptionViolation
- Critical
- Trace
Plus, it embeds `Logger`, so do we need to explicitly list out those method again in order to catch cases where someone calls a simple `Logger` method through a `SugaredLogger`?

.golangci.yml
```diff
@@ -34,6 +35,22 @@ linters-settings:
   govet:
     # report about shadowed variables
     check-shadowing: true
+    settings:
+      printf:
+        # Additionally check chainlink custom loggers
+        funcs:
+          - (github.com/smartcontractkit/chainlink-common/pkg/logger.Logger).AssumptionViolationf
+          - (github.com/smartcontractkit/chainlink-common/pkg/logger.Logger).Tracef
```
jmank88, line 43: From my other comment:
- I think we can drop these two lines (from below too)
- But I wonder if we need to repeat all of these Logger methods for SugaredLogger too, since it extends that interface

pavel-raykov: Fortunately, there is no need to repeat. Here is a minimal example:

```
package main

import "github.com/smartcontractkit/chainlink-common/pkg/logger"

func main() {
	ll, _ := logger.New()
	ll.Infof("Type = %d", "zzz")
        ll.Infow("Type", "zzz")
	sl := logger.Sugared(ll)
	sl.Infof("Type = %d", "zzZ")
        sl.Infow("Type", "zzz") 
}
```

If you run `golangci-lint run` on this file with a single rule (for loggercheck)

- (github.com/smartcontractkit/chainlink-common/pkg/logger.Logger).Infow

or (for printf)
 
- (github.com/smartcontractkit/chainlink-common/pkg/logger.Logger).Infof

you will catch both violations.

.golangci.yml
pavel-raykov: Fortunately, there is no need to repeat - see my other comment.

george-dorin (approved):

PR #585, opened 2024-06-14
ChainReader Loop Interface Test Matrix

EasterTheBunny:
Client and Server ChainReader LOOP interfaces can accept options that configure the encoding version for client-server communication. A test matrix helper function is included for running tests as a test matrix.

### DIFF
--- a/pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
+++ b/pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
@@ -66,7 +66,7 @@ func TestVersionedBytesFunctions(t *testing.T) {
 func TestChainReaderInterfaceTests(t *testing.T) {
 	t.Parallel()
 
-	chainreadertest.LoopEncodingTestMatrix(t, func(version chainreader.EncodingVersion) func(t *testing.T) {
+	chainreadertest.TestAllEncodings(t, func(version chainreader.EncodingVersion) func(t *testing.T) {
 		return func(t *testing.T) {
 			t.Parallel()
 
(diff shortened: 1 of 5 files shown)


pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -64,99 +64,148 @@ func TestVersionedBytesFunctions(t *testing.T) {
...
+			crTester.Setup(t)
+			cr := crTester.GetChainReader(t)
+
+			es := &errChainReader{}
+			errTester := chainreadertest.WrapChainReaderTesterForLoop(&fakeChainReaderInterfaceTester{impl: es})
+			errTester.Setup(t)
+			chainReader := errTester.GetChainReader(t)
+
+			t.Run("nil reader should return unimplemented", func(t *testing.T) {
+				t.Parallel()
```
jmank88, line 180: This parallel call seems out of place since the other sub-tests under this parent test do not have parallel. IIUC this test would just defer until the non-parallel siblings are complete.

pkg/loop/internal/relayer/pluginprovider/chainreader/test/chain_reader_loop_tester.go
```diff
@@ -8,33 +8,73 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer/pluginprovider/chainreader"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
+
 	. "github.com/smartcontractkit/chainlink-common/pkg/types/interfacetests" //nolint common practice to import test mods with .
 )
 
+func LoopEncodingTestMatrix(t *testing.T, test func(chainreader.EncodingVersion) func(t *testing.T)) {
```
jmank88, line 15: Probably no need to say LOOP given the package scope:
```suggestion
func EncodingTestMatrix(t *testing.T, test func(chainreader.EncodingVersion) func(t *testing.T)) {
```

nolag (approved):

PR #589, opened 2024-06-17
ccip multiramps

dimkouv:
Update CCIP generic msg type to support multiramps.


dimkouv (commented):
I did not update `.proto` because we'll probably remove it.

pkg/types/ccipocr3/generic_types.go
```diff
@@ -108,7 +116,16 @@ func (c CCIPMsg) String() string {
...
 	SourceChain ChainSelector `json:"sourceChain,string"`
-	SeqNum      SeqNum        `json:"seqNum,string"`
+	// SeqNum is an auto-incrementing sequence number for the message.
+	// NOTE: Sequence numbers are unique per chain. Meaning that the same sequence number can exist on multiple chains.
+	SeqNum SeqNum `json:"seqNum,string"`
+
+	// MsgHash is the hash of all the message fields. The hash is computed on the destination chain.
+	// This means that on the source chain we cannot compute the msg hash and in the off-chain side we need MsgHasher
+	// implementation for the destination chain.
+	MsgHash Bytes32 `json:"msgHash"` // computed
```
jmank88, line 130: The `// computed` comment reads to be like we should not be expecting to receive this field over the wire from encoded JSON, but that is not actually the case, right? Can we update this to be clear about when this is computed? The long comment reads like it happens on-chain and cannot happen off-chain, but that doesn't make sense to me either. Is it a particular kind of hash?

dimkouv: > comment reads to me like we should not be expecting to receive this field over the wire from encoded JSON, but that is not actually the case, right?

that's the case, i'll update the comment

jmank88: Oh, but then do we actually need the json tag? Or could it be:
```go
MsgHash Bytes32 `json:"-"`
```

pkg/types/ccipocr3/generic_types.go
```diff
@@ -83,18 +83,26 @@ func (c ChainSelector) String() string {
 }
 
 type CCIPMsg struct {
```
makramkd, line 85: This is going to be the `Any2Any` message type right? The type that holds the fields that are present on all `Any2{Chain}` and `{Chain}2Any` messages? If so lets make it clear that this type will hold all fields and is not chain specific?

pkg/types/ccipocr3/generic_types.go
```diff
@@ -83,18 +83,26 @@ func (c ChainSelector) String() string {
...
+	SourceTokenData [][]byte        `json:"sourceTokenData"`
+	Metadata        CCIPMsgMetadata `json:"metadata"`
+}
+
+type CCIPMsgMetadata struct {
+	// Version of the message metadata. Required in order to be able to parse the metadata
+	// by the underlying implementation.
+	Version uint64 `json:"version"`
+	// Data is the metadata payload. The underlying implementation should know how to parse this data.
+	Data []byte `json:"data"`
```
makramkd, line 105: What do we mean by "metadata payload" here? Where is this fetched from? Is this part of the onchain emitted message?

pkg/types/ccipocr3/generic_types.go
```diff
@@ -83,18 +83,26 @@ func (c ChainSelector) String() string {
...
+	Data            []byte          `json:"data"`
+	TokenAmounts    []TokenAmount   `json:"tokenAmounts"`
+	SourceTokenData [][]byte        `json:"sourceTokenData"`
+	Metadata        CCIPMsgMetadata `json:"metadata"`
+}
+
+type CCIPMsgMetadata struct {
+	// Version of the message metadata. Required in order to be able to parse the metadata
+	// by the underlying implementation.
+	Version uint64 `json:"version"`
```
makramkd, line 103: How is this version set / managed? Is this the CCIP capability semantic version?

pkg/types/ccipocr3/generic_types.go
```diff
@@ -108,7 +116,15 @@ func (c CCIPMsg) String() string {
 }
 
 type CCIPMsgBaseDetails struct {
-	ID          Bytes32       `json:"id"`
+	// ID is a unique identifier for the message, it should be unique across all chains.
+	ID string `json:"id"`
```
makramkd, line 120: Just a disambiguating detail
```suggestion
	// ID is a unique identifier for the message, it should be unique across all chains.
	// It is generated on the chain in which the CCIP send is requested (i.e the source chain of a message).
	ID string `json:"id"`
```

pkg/types/ccipocr3/generic_types.go
dimkouv: > This is going to be the Any2Any message type right?
Yeah, I wanted to avoid `Any2Any` prefix. It's just THE `CCIPMsg`.

We can name the others `CCIPMsgToChainX` or ``CCIPMsgFromChainX` to make them chain specific

pkg/types/ccipocr3/generic_types.go
dimkouv: It's populated by the underlying chain specific implementation e.g.
```
solMsg = CCIPMsgToSolana{
   sender
   recipient
   someSolSpecificField1
   someSolSpecificField2
}

ccipMsg = CCIPMsg{
   sender
   recipient
   []byte{ "someSolSpecificField1", "someSolSpecificField2" }
}
}

```

pkg/types/ccipocr3/generic_types.go
makramkd: We shouldn't have different versions from the ccip capability version. I think its fine to use that here but we shouldn't create another set of versions

pkg/types/ccipocr3/generic_types.go
makramkd: I'm wondering if we should make CCIPMsgBaseDetails the only thing that is required by all chains and push the rest down to this `Data []byte` field

makramkd: That way the plugin is still able to perform its business logic; MsgHasher implementations will correctly decode the `Data` into the required structure to end up hashing

jmank88 (approved):

pkg/types/ccipocr3/generic_types.go
makramkd: Just synced offline. Will keep this as-is for now; suspect that we'll add to the `Data` field if nothing fits in the CCIPMsg explicit fields anyway

makramkd (approved):

jmank88 (approved):

PR #591, opened 2024-06-18
types: Add call options to submit transaction to simplify API

nickcorin:
This is just a draft proposal intended to facilitate a conversation around API simplification across chains.

For example, most transctions submitted through the `ChainWriter` will use a zero `value` parameter. The implications of this is that most products will be passing in `nil`. We may as well try and simplify the call for the majority of use cases. while still providing the functionality for those who want to use it.

I've made options for `TxMeta` and the `TransactionID`. This is mostly for PoC sake, and is not a proposal to make those specific params optional.

### DIFF
--- a/pkg/types/chain_writer.go
+++ b/pkg/types/chain_writer.go
@@ -7,12 +7,94 @@ import (
 	"github.com/google/uuid"
 )
 
+// CallOption can be used to configure a single call to the ChainWriter.
+type CallOption interface {
+	apply(*CallOptions)
+}
+
+// CallOptionFunc is a functional type that implements the CallOption interface.
+type callOptionFunc func(*CallOptions)
+
+func (fn callOptionFunc) apply(o *CallOptions) {
+	fn(o)
+}
+
+// CallOptions contains the values which are configurable per-call to the ChainWriter.
+//
+// ChainWriter implementations should hold an instance of this struct and apply the options to the struct before making the call.
+type CallOptions struct {
+	// TransactionID is submitted to the particular chain's TXM as an idempotency key.
+	// It should be unique across transaction.
+	TransactionID   string
+	TransactionMeta *TxMeta
+	Value           *big.Int
+}
+
+// Apply applies the given options to the CallOptions.
+func (o *CallOptions) Apply(opts ...CallOption) {
+	for _, opt := range opts {
+		opt.apply(o)
+	}
+}
+
+// WithTransactionID sets the transactionID (idempotency key) for a call to the ChainWriter.
+//
+// Default: ""
+func WithTransactionID(transactionID string) CallOption {
+	return callOptionFunc(func(o *CallOptions) {
+		o.TransactionID = transactionID
+	})
+}
+
+// WithTransactionMeta sets the transaction metadata for a call to the ChainWriter.
+//
+// Default: <nil>
+func WithTransactionMeta(meta *TxMeta) CallOption {
+	return callOptionFunc(func(o *CallOptions) {
+		o.TransactionMeta = meta
+	})
+}
+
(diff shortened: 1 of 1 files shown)


pkg/types/chain_writer.go
```diff
@@ -7,12 +7,94 @@ import (
...
+//
+//	 type ReportParams struct {
+//	 		Receiver	string
+//			RawReport 	[]byte
+//			Signatures	[][]byte
+//	 }
+//
+//	 func submitReport(ctx context.Context, cw ChainWriter) error {
+//			params := ReportParams{ ... }
+//			return cw.SubmitTransaction(ctx, "ReportContract", "Report", &params, "0x1234")
```
patrick-dowell, line 82: what does this look like when you specify the optional args?

pkg/types/chain_writer.go
```diff
@@ -7,12 +7,94 @@ import (
...
+// Default: <nil>
+func WithTransactionMeta(meta *TxMeta) CallOption {
+	return callOptionFunc(func(o *CallOptions) {
+		o.TransactionMeta = meta
+	})
+}
+
+// WithValue sets the value to be sent with a transaction.
+//
+// Default: <nil>
```
patrick-dowell, line 60: If Value is now an optional argument, does it make sense for it to be a pointer? We could just have it be an int with default value 0, right?

pkg/types/chain_writer.go
```diff
@@ -7,12 +7,94 @@ import (
...
+		opt.apply(o)
+	}
+}
+
+// WithTransactionID sets the transactionID (idempotency key) for a call to the ChainWriter.
+//
+// Default: ""
+func WithTransactionID(transactionID string) CallOption {
+	return callOptionFunc(func(o *CallOptions) {
+		o.TransactionID = transactionID
```
patrick-dowell, line 45: If we do make this optional, we should probably create a new uuid and convert it to a string to ensure uniqueness.

patrick-dowell: Actually, I guess the TXM can just take empty string and deal with it appropriately, so maybe it shouldn't have an ID assigned. It just means that you can never check the Status() of it, even if you wanted to later.

nickcorin: I think the `TXM` already handles an empty string, but we can create a random UUID as the default too.

pkg/types/chain_writer.go
nickcorin: It could be, but the canonical way to use `big.Int` types is as a pointer and we've had multiple people requesting the change to a pointer.

pkg/types/chain_writer.go
nickcorin:
```
return cw.SubmitTransaction(ctx, "ReportContract", "Report", &params, "0x1234", WithTransactionID(uuid.NewString), WithValue(big.NewInt(100))
```

pkg/types/chain_writer.go
```diff
@@ -7,12 +7,94 @@ import (
...
+//
+//	 func submitReport(ctx context.Context, cw ChainWriter) error {
+//			params := ReportParams{ ... }
+//			return cw.SubmitTransaction(ctx, "ReportContract", "Report", &params, "0x1234")
+//	 }
+//
+// Note that implementations should ignore extra fields in params that are not expected in the call to allow easier
+// use across chains and contract versions.
+// Similarly, when using a struct for returnVal, fields in the return value that are not on-chain will not be set.
+type MethodArguments any
```
nickcorin, line 88: Btw -- this type was due to a suggestion from Jordan that we have a common type between chain reader and chain writer for the inputs which are expected to have a particular format for `codec`.

PR #594, opened 2024-06-18
golangci-lint: add gomoddirectives

jmank88:


jmank88:
This does not work how we want. It is flagging all `replace`s statements not just local.

PR #598, opened 2024-06-19
add BatchingStrategyID field to ExecOffchainConfig

0xnogo:
Adding `BatchingStrategyID` to `ExecOffchainConfig`


dimkouv (dismissed):

winder (approved):

pkg/loop/internal/pb/ccip/offramp.proto
```diff
@@ -182,6 +182,7 @@ message ExecOffchainConfig {
     google.protobuf.Duration inflight_cache_expiry = 4; // config.Duration
     google.protobuf.Duration root_snooze_time = 5; // config.Duration
     google.protobuf.Duration message_visibility_interval = 6; // config.Duration
+    uint32 batching_strategy_id = 7; // uint32
```
jmank88, line 185: nit/ comment seems unecessary
```suggestion
    uint32 batching_strategy_id = 7;
```

jmank88:
Don't the conversion funcs need to be updated to? https://github.com/smartcontractkit/chainlink-common/blob/97d4285d9f4871aafd3a88833c6551c850929a40/pkg/loop/internal/relayer/pluginprovider/ext/ccip/offramp.go#L663
It seems like they are missing tests. We should have round-trip encoding/decoding tests for types like this.

jmank88 (approved):

krehermann (approved):

makramkd (approved):

PR #599, opened 2024-06-19
Home chain config syncer types

asoliman92:
## Motivation

Plugins need to refresh HomeChain config regularly to know which oracles support which chains and the latest F for each chain.

## Solution

Add required types (configs, interfaces)

CCIP [PR](https://github.com/smartcontractkit/ccip/pull/978)
Jira [Ticket](https://smartcontract-it.atlassian.net/browse/CCIP-2466)


pkg/types/ccipocr3/config.go
```diff
@@ -12,12 +16,6 @@ type CommitPluginConfig struct {
...
 	DestChain ChainSelector `json:"destChain"`
 
-	// FChain defines the FChain value for each chain. FChain is used while forming consensus based on the observations.
-	FChain map[ChainSelector]int `json:"fChain"`
-
-	// ObserverInfo is a map of oracle IDs to ObserverInfo.
-	ObserverInfo map[commontypes.OracleID]ObserverInfo `json:"observerInfo"`
-
 	// PricedTokens is a list of tokens that we want to submit price updates for.
 	PricedTokens []types.Account `json:"pricedTokens"`
```
asoliman92, line 20: I think this should be part of ObservationConsensus struct not the plugin config. It seems that it will be changing and we need to reach consensus on it between observers in DON.
CC @dimkouv @makramkd

pkg/types/ccipocr3/common_types.go
```diff
@@ -85,3 +86,29 @@ func (b *BigInt) UnmarshalJSON(p []byte) error {
 func (b BigInt) IsEmpty() bool {
 	return b.Int == nil
 }
+
+type P2PID [32]byte
+
+func (b P2PID) MarshalJSON() ([]byte, error) {
+	return json.Marshal(hex.EncodeToString(b[:]))
```
makramkd, line 93: P2P ID's should not be marshaled as hex, they have their own string encoding scheme please see libocr's PeerID type in ragep2p/types/types.go

Do we need to redefine the P2PID here? What's the win over using what we have in libocr?

pkg/types/ccipocr3/common_types.go
```diff
@@ -85,3 +86,29 @@ func (b *BigInt) UnmarshalJSON(p []byte) error {
...
+	}
+	decoded, err := hex.DecodeString(s)
+	if err != nil {
+		return err
+	}
+	copy(b[:], decoded)
+	return nil
+}
+
+func GetP2pID(id int) P2PID {
```
makramkd, line 109: What's this function for?

pkg/types/ccipocr3/config.go
```diff
@@ -1,9 +1,13 @@
 package ccipocr3
 
+import "C"
```
makramkd, line 3: Why is this import here? This is only imported for cgo files

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+
+func (cm *P2PToSupportedChains) UnmarshalJSON(data []byte) error {
+	temp := make(map[string]SupportedChains)
+	if err := json.Unmarshal(data, &temp); err != nil {
+		return err
+	}
+	//if temp == nil {
+	//	*cm = nil
+	//	return nil
+	//}
```
makramkd, line 78: Can delete this if not used

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+	*cm = make(P2PToSupportedChains)
+	for k, v := range temp {
+		var key P2PID
+		decoded, err := hex.DecodeString(k)
```
makramkd, line 82: See my comment above about p2p id encoding

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+
+func (cm P2PToSupportedChains) MarshalJSON() ([]byte, error) {
+	temp := make(map[string]SupportedChains)
+	for k, v := range cm {
+		temp[hex.EncodeToString(k[:])] = v
+	}
+	return json.Marshal(temp)
+}
+
+type ConsensusObservation struct {
```
makramkd, line 100: This name is too general, I suggest calling it something more specific since we come to consensus on a lot of things

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+
+	return nil
+}
+
+type HomeChainConfig struct {
+	// FChain defines the FChain value for each chain. FChain is used while forming consensus based on the observations.
+	FChain map[ChainSelector]int `json:"fChain"`
+	// NodeSupportedChains is a map of PeerIDs to SupportedChains.
+	NodeSupportedChains P2PToSupportedChains `json:"nodeSupportedChains"`
+}
```
makramkd, line 129: Why not embed this in the struct above?

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+		}
+	}
+	if len(c.PricedTokens) == 0 {
+		return fmt.Errorf("priced tokens not set, at least one priced token is required")
+	}
+
+	return nil
+}
+
+type HomeChainConfig struct {
```
makramkd, line 124: Lets add a comment to this and other structures to make it clear what they represent and what they're used for

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+	supportedSlice := make([]ChainSelector, 0, sc.Supported.Cardinality())
+	for ch := range sc.Supported.Iter() {
+		supportedSlice = append(supportedSlice, ch)
+	}
+
+	// Define a temporary struct with a slice for Supported
+	temp := struct {
+		Supported []ChainSelector `json:"supported"`
+	}{
+		Supported: supportedSlice,
```
makramkd, line 196: Is this deterministically generated in the same way all the time? If this is using a `map` under the hood iteration order is undefined so you might get different lists on different nodes and hence consensus failure.

If there is a risk of this we should `sort` the slice first before plugging it into this struct.

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+
+	return json.Marshal(temp)
+}
+
+type OnChainConfig struct {
+	Readers []P2PID `json:"readers"`
+	FChain  uint8   `json:"fChain"`
+	Config  []byte  `json:"config"`
+}
+type OnChainCapabilityConfig struct {
```
makramkd, line 207: This is not a capability configuration but a chain configuration

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +64,148 @@ type ObserverInfo struct {
...
+	temp := struct {
+		Supported []ChainSelector `json:"supported"`
+	}{
+		Supported: supportedSlice,
+	}
+
+	return json.Marshal(temp)
+}
+
+type OnChainConfig struct {
```
makramkd, line 202: Lets just have one struct called `ChainConfig` that has `Readers`, `FChain`, `Config`, `ChainSelector`, etc. The structure of offchain code does not have to mirror on-chain code

pkg/types/ccipocr3/interfaces.go
```diff
@@ -53,3 +53,9 @@ type ExecutePluginCodec interface {
 type MessageHasher interface {
 	Hash(context.Context, CCIPMsg) (Bytes32, error)
 }
+
+type HomeChainPoller interface {
+	GetConfig() HomeChainConfig
```
makramkd, line 58: This needs to be either scoped to a particular chain selector or we get all the chain configs in one call. I think having both is more flexible at this stage.

```suggestion
	GetChainConfig(ctx context.Context, chainSelector uint64) HomeChainConfig
	GetAllChainConfigs(ctx context.Context) map[uint64]HomeChainConfig
```

pkg/types/ccipocr3/interfaces.go
```diff
@@ -53,3 +53,9 @@ type ExecutePluginCodec interface {
...
+	Start(ctx context.Context) error
+	Close() error
```
makramkd, line 60: Probably better to inherit the `services.ServiceCtx` interface instead of doing this to make it clear what we want to implement

makramkd (changes requested):

pkg/types/ccipocr3/interfaces.go
asoliman92: I tried but it's a cyclic dependency unfortunately :/

pkg/types/ccipocr3/common_types.go
asoliman92: Might not be a good place. I'm using it in tests. Will remove and add to tesethelpers in ccip.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.21
 
 require (
 	github.com/confluentinc/confluent-kafka-go/v2 v2.3.0
+	github.com/deckarep/golang-set/v2 v2.1.0
```
jmank88, line 7: Does this library do anything sophisticated? Or could we use `map[T]struct{}` with `maps.Contains`/`maps.Keys`?

pkg/types/ccipocr3/common_types.go
asoliman92: I remember trying the libocr's PeerID first and then changing my mind for some reason. Let me check again why.

asoliman92: I was using `PeerID` from `github.com/smartcontractkit/chainlink/v2/core/services/p2p/types` and this seems like a cyclic dependency.

makramkd: Yeah don't import from core but you can import from libocr

makramkd: Bonus is that libocr already implements the required Marshal* methods so we don't need that boilerplate again

pkg/types/ccipocr3/interfaces.go
makramkd: Why is it cyclic? That is part of chainlink-common

pkg/types/ccipocr3/common_types.go
```diff
@@ -85,3 +86,29 @@ func (b *BigInt) UnmarshalJSON(p []byte) error {
 func (b BigInt) IsEmpty() bool {
 	return b.Int == nil
 }
+
+type P2PID [32]byte
```
dimkouv, line 90: There is also a `Bytes32` type

pkg/types/ccipocr3/config.go
asoliman92: I wanted them to be separate as they might grow differently in the future.

pkg/types/ccipocr3/config.go
asoliman92: How are we going to se/deserialize the actual on chain config that we get from the contract in this case?

pkg/types/ccipocr3/config.go
asoliman92: very good observation

go.mod
asoliman92: I'm using it for thread safe `Set` not for `map`

pkg/types/ccipocr3/config.go
makramkd: Deserialization of the contract data is done by the chain reader

pkg/types/ccipocr3/interfaces.go
makramkd: You can just use `services.Service` I think, its in chainlink-common.

pkg/types/ccipocr3/config.go
asoliman92: We talked about this, the plugin is already using a Set as well, comparing sets is easy using the library and doesn't need to be in a specific order or anything.

pkg/types/ccipocr3/common_types.go
asoliman92: I removed this completely and using libocr PeerID 👍

makramkd (dismissed):

makramkd (dismissed):

pkg/types/ccipocr3/interfaces.go
```diff
@@ -53,3 +58,15 @@ type ExecutePluginCodec interface {
 type MessageHasher interface {
 	Hash(context.Context, CCIPMsg) (Bytes32, error)
 }
+
+type HomeChainPoller interface {
+	GetChainConfig(chainSelector ChainSelector) (ChainConfig, error)
+	GetAllChainConfigs() (map[ChainSelector]ChainConfig, error)
+	// GetSupportedChainsForPeer Gets all chain selectors that the peerID can read/write from/to
+	GetSupportedChainsForPeer(id libocrtypes.PeerID) (mapset.Set[ChainSelector], error)
```
winder, line 66: How widely adopted is mapset? Generally I would avoid dependency types as part of a shared interface. Go uses `map[ChainSelector]bool` (or `map[ChainSelector]struct{}`) for sets.

0xnogo (approved):

pkg/types/ccipocr3/interfaces.go
jmank88: Right, this does not make sense to send over the wire. We could even return a plain `[]ChainSelector` and let the caller construct a set only if necessary.

jmank88 (changes requested):

pkg/types/ccipocr3/interfaces.go
```diff
@@ -53,3 +58,15 @@ type ExecutePluginCodec interface {
 type MessageHasher interface {
 	Hash(context.Context, CCIPMsg) (Bytes32, error)
 }
+
+type HomeChainPoller interface {
```
jmank88, line 62: Where is this used?

go.mod
jmank88: The idiomatic representation for a set in Go is a map with either `struct{}` or `bool` values.
You can see the implementation of the module imported here is doing the same thing: https://github.com/deckarep/golang-set/blob/main/threadunsafe.go#L34

There are interesting set operations in this library, but we don't seem to be using them, so the import may not be worth it. Although more importantly, it doesn't typically make sense to send types like this over grpc, since we want to pass data, not provide shared access to the raw data structures that we are holding on to.

jmank88:
Does this work have a ticket and/or corresponding core/ccip PR?

pkg/types/ccipocr3/interfaces.go
asoliman92: https://github.com/smartcontractkit/ccip/actions/runs/9614422103/job/26519178069?pr=978

asoliman92: It will be used in all CCIP plugins to get the dynamic configs from Home Chain

pkg/types/ccipocr3/interfaces.go
asoliman92: good question, I saw it used in a couple of places in CCIP. Didn't notice that it was that little though.

pkg/types/ccipocr3/config.go
```diff
@@ -78,3 +62,13 @@ type ObserverInfo struct {
 	// Reads define the chains that the current node can read from.
 	Reads []ChainSelector `json:"reads"`
 }
+
+// ChainConfig will live on the home chain and will be used to update chain configuration like F value and supported nodes dynamically.
+type ChainConfig struct {
```
jmank88, line 67: Is this type also unused from this module?

asoliman92: Yes. Will move it now 👍

jmank88 (approved):

0xnogo (approved):

PR #601, opened 2024-06-19
add more description to codec type errors

EasterTheBunny:


pkg/codec/modifier_base.go
```diff
@@ -242,16 +252,27 @@ func doForMapElements[T any](valueMapping map[string]any, fields map[string]T, f
 		path := strings.Split(key, ".")
 		name := path[len(path)-1]
 		path = path[:len(path)-1]
+
 		extractMaps, err := getMapsFromPath(valueMapping, path)
 		if err != nil {
-			return err
+			return PathMappingError{Err: err, Path: key}
```
nolag, line 258: This no longer returns a common error so it won't unwrap when sent over a LOOP

EasterTheBunny: Ah ok. If I wrap the type error within `Error()` would that unwrap over LOOP?

jmank88: I think you can just add a `Cause() error` method so it can match the wrapped error

pkg/codec/modifier_base.go
```diff
@@ -242,16 +252,27 @@ func doForMapElements[T any](valueMapping map[string]any, fields map[string]T, f
...
 	return nil
 }
+
+type PathMappingError struct {
+	Err  error
+	Path string
+}
+
+func (e PathMappingError) Error() string {
+	return fmt.Sprintf("mapping error for path (%s): %s", e.Path, e.Err.Error())
```
jmank88, line 277: `Error()` should be called automatically for `error`s:
```suggestion
	return fmt.Sprintf("mapping error for path (%s): %s", e.Path, e.Err)
```
> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

https://pkg.go.dev/fmt

EasterTheBunny: Haha. This is just a pattern I have gotten into due to a linter I have activated elsewhere. ¯\_(ツ)_/¯

pkg/codec/modifier_base.go
jmank88: @nolag Does this behave the way that we want? Or are the outer wrapped errors lost? (if so, I think we could change that easily)

nolag (approved):

pkg/codec/modifier_base.go
jmank88: I think we are good:
https://github.com/grpc/grpc-go/blob/v1.64.0/status/status.go#L123
Original `err` message is used even after doing the `As` conversion, so outer wrapped message will pass through.

jmank88 (approved):

PR #603, opened 2024-06-19
ContractReader Multiple Read Addresses

EasterTheBunny:
`ContractReader` needs the ability to read events and methods from the same contract configuration at multiple addresses.

*Restrictions*
- `Bind` is required due to needing a contract address for event processing _before_ `GetLatestValue` or `QueryKey` is called

*Other Considerations*
- A request has been made to simplify `contract` and `method` into a single identifier
- Supplying an address for a binding makes it unique and not able to be 'overridden'. `Unbind` is necessary to remove a binding.
- Concatenating multiple values into a single 'key' might convolute the key. This might not be the optimal solution, favoring a UUID or other key generation method instead.


pkg/types/chain_reader.go
```diff
@@ -44,16 +44,29 @@ type ChainReader interface {
 	// Note that implementations should ignore extra fields in params that are not expected in the call to allow easier
 	// use across chains and contract versions.
 	// Similarly, when using a struct for returnVal, fields in the return value that are not on-chain will not be set.
-	GetLatestValue(ctx context.Context, contractName, method string, params, returnVal any) error
+	GetLatestValue(ctx context.Context, contractName, method string, params, returnVal any, opts ...ReadOpt) error
```
ilija42, line 47: I'd rather go for something simpler like 
```suggestion
	GetLatestValue(ctx context.Context, singleIdentifier string, address string, params, returnVal any) error
```

jmank88: We need to consider the proto changes, at least alongside this API, but IMHO the protos should actually be what we think about first. Variadic options are just syntax sugar.

EasterTheBunny: So in this case, would it be better to alter the function signature in favor of explicit types instead of variadic options for the sake of keeping the proto API clean?

jmank88: I think they are separate concerns because protos cannot express functional options, they can only be used to transmit some resulting underlying config. IMHO it will be easier to think of that underlying proto first, and then after we can consider this kind of external API change. There are other options though, such as exposing (or matching) the Request/Request types generated by GRPC, which don't have the issues with ever-expanding method args.

EasterTheBunny: I'll make a new proposal.

pkg/types/chain_reader.go
```diff
@@ -44,16 +44,17 @@ type ChainReader interface {
 	// Note that implementations should ignore extra fields in params that are not expected in the call to allow easier
 	// use across chains and contract versions.
 	// Similarly, when using a struct for returnVal, fields in the return value that are not on-chain will not be set.
-	GetLatestValue(ctx context.Context, contractName, method string, params, returnVal any) error
+	GetLatestValue(ctx context.Context, bindingKey string, params, returnVal any) error
```
ilija42, line 47: LGTM, I am just wondering if its better UX for `GetLatestValue` to just accept `boundContract BoundContract`, I assume that products will store `BoundContract` anyway, so what is the point of the caller having to do `boundContract.Key()` instead of just `boundContract`

pkg/types/chain_reader.go
```diff
@@ -44,16 +44,17 @@ type ChainReader interface {
...
 
-	// Bind will override current bindings for the same contract, if one has been set and will return an error if the
-	// contract is not known by the ChainReader, or if the Address is invalid
-	// TODO: (BCF-3139) can't handle log poller filters,explore handling this through chain reader config.
+	// Bind will add provided bindings and will return an error if the contract is not known by the ChainReader, or if
+	// the Address is invalid. Any provided binding that already exists should result in a noop.
 	Bind(ctx context.Context, bindings []BoundContract) error
 
+	// Unbind will remove a binding identified by the binding key and will return an error if the binding is not found.
+	Unbind(ctx context.Context, bindingKey string) error
```
nolag, line 54: nit: If bind is taking multiple contracts, shouldn't unbind allow you to unbind multiple at a time too?

pkg/types/chain_reader.go
EasterTheBunny: I was thinking that too, but `BoundContract` also has `Pending`, which doesn't really make sense in this context. Maybe that's ok, but it felt kinda weird to me.

pkg/types/chain_reader.go
EasterTheBunny: Good callout. I'll add that.

pkg/types/chain_reader.go
nolag: Do we need to worry about pending?  Eg: Could someone subscribe to events twice, once to know it's on chain with but hasn't hit finality than again when it has finality?  Not sure the use case, but if so, then the pending flag could be ok to leave here...

EasterTheBunny: What would be the expected behavior of `Pending` then in the context of `GetLatestValue`? I'm just hesitant to include inputs that don't do anything. It makes for a confusing UX.

jmank88: Didn't we decide to remove pending? IIRC it was only added due to a miscommunication and is not actually used.

ilija42: I have an open PR that removes Pending, its just blocked atm

EasterTheBunny: Ahhh. Ok so in that case we don't have to consider `Pending` because we are eliminating it anyway.

pkg/types/chain_reader.go
```diff
@@ -70,6 +74,7 @@ type Sequence struct {
 type BoundContract struct {
 	Address string
 	Name    string
+	// Deprecated: Pending is being removed.
```
ilija42, line 77: Not sure if we should flag it as Deprecated before we have an alternative?

ilija42 (commented):
lgtm

pkg/types/chain_reader.go
```diff
@@ -70,6 +74,7 @@ type Sequence struct {
 type BoundContract struct {
```
ilija42, line 74: I think a method on BoundContract that returns a single identifier is very useful since codec contract types are identified by contractName + methodName

EasterTheBunny: Is that not what `Name` is? The single identifier?

ilija42: Ah right, the types would be the same for every address, so name is fine

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -11,23 +11,23 @@ service ChainReader {
...
 message QueryKeyRequest {
-  string contract_name = 1;
+  BoundContract contract = 1;
   QueryKeyFilter filter = 2;
   LimitAndSort limit_and_sort = 3;
 }
 
-// BindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind].
+// BindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind/Unbind].
 message BindRequest {
```
ilija42, line 31: Probably better to have BindRequest and UnBindRequest separately

EasterTheBunny: I started doing that but left it as the same to keep the interface simpler. I can make them separate to ensure extensibility instead.

pkg/types/chain_reader.go
```diff
@@ -73,6 +78,31 @@ type BoundContract struct {
 	Pending bool
 }
 
+func NewBoundContract(address string, contract string, method string) BoundContract {
```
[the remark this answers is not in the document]
EasterTheBunny: Switched to splitting contract and method into properties in the `BoundContract` struct.

ilija42 (commented):
Looks good to me now, just needs the single identifier PR to be merged first

pkg/types/chain_reader.go
```diff
@@ -105,10 +109,19 @@ type Sequence struct {
 }
 
 type BoundContract struct {
-	Address string
-	Name    string
+	Address  string
```
ilija42, line 112: Related to the comment above, maybe we don't need a method field, but can instead have a method on BoundContract for creating a key, so something like 
`GetLatestValue(ctx, bc.ReadID("key"), primitives.Finalized, params, returnVal)`

bc.ReadID("key")

pkg/types/chain_reader.go
```diff
@@ -45,20 +46,23 @@ type ChainReader interface {
...
 	// BatchGetLatestValuesRequest params and returnVal follow same rules as GetLatestValue params and returnVal arguments, with difference in how response is returned.
 	// BatchGetLatestValuesResult response is grouped by contract names, which contain read results that maintain the order from the request.
 	// Contract call errors are returned in the Err field of BatchGetLatestValuesResult.
 	BatchGetLatestValues(ctx context.Context, request BatchGetLatestValuesRequest) (BatchGetLatestValuesResult, error)
 
-	// Bind will override current bindings for the same contract, if one has been set and will return an error if the
-	// contract is not known by the ChainReader, or if the Address is invalid
+	// Bind will add provided bindings and will return an error if the contract is not known by the ContractReader, or if
+	// the Address is invalid. Any provided binding that already exists should result in a noop.
 	Bind(ctx context.Context, bindings []BoundContract) error
```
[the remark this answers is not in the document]
ilija42: left a comment below with alternative approaches

pkg/types/chain_reader.go
ilija42: Alternatively if its easier or better to pass in `BoundContract` instead of a string identifiet, method field can be Unexported and hidden from the constructor and  `bc.ReadID("key")` could return a BoundContract with populated method field

ilija42 (changes requested):
Left some suggestions

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -95,7 +100,7 @@ message Sequence {
 // BoundContract represents a [github.com/smartcontractkit/chainlink-common/pkg/types.BoundContract].
 message BoundContract {
   string address = 1;
-  string name = 2;
+  string contract = 2;
```
ilija42, line 103: Why? Maybe at least contract_name?

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -12,14 +12,14 @@ service ChainReader {
...
+  rpc Unbind(UnbindRequest) returns (google.protobuf.Empty) {}
 }
 
 // GetLatestValueRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.GetLatestValue].
 message GetLatestValueRequest {
-  string contract_name = 1;
-  string method = 2;
-  Confidence confidence = 3;
-  VersionedBytes params = 4;
+  string read_name = 1;
```
ilija42, line 20: read_name is used in some places to signify a specific contact read (method or event on evm). Maybe read_identifier?

pkg/types/chain_reader.go
```diff
@@ -105,10 +109,10 @@ type Sequence struct {
...
 type BoundContract struct {
-	Address string
-	Name    string
+	Address  string
+	Contract string
 }
 
-func (bc BoundContract) Key() string {
-	return bc.Address + "-" + bc.Name
+func (bc BoundContract) ReadKey(method string) string {
```
ilija42, line 116: Name here should match with GetLatestValue param, so either ReadName or ReadIdentifier per my suggestiin.

pkg/loop/internal/pb/chain_reader.proto
EasterTheBunny: It was left this way because I added `method`, but then removed it.

pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -70,16 +70,25 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
...
+				bindings := tester.GetBindings(t)
+
+				var bound types.BoundContract
+				for idx := range bindings {
+					if bindings[idx].Name == AnyContractName {
+						bound = bindings[idx]
+					}
+				}
+
+				require.NoError(t, cr.Bind(ctx, bindings))
```
ilija42, line 82: This code is repeated a lot, I am not sure why you don't just Bind everything like it used to do. Alternatively there should be a GetBinding() method that returns a specific bound contract

pkg/types/chain_reader.go
```diff
@@ -45,37 +46,40 @@ type ChainReader interface {
...
 }
 
 // BatchGetLatestValuesRequest string is contract name.
 type BatchGetLatestValuesRequest map[string]ContractBatch
 type ContractBatch []BatchRead
 type BatchRead struct {
-	ReadName  string
-	Params    any
-	ReturnVal any
+	ReadIdentifier string
```
ilija42, line 72: Comment should explain what ReadIdentifier is, something like " ReadIdentifier should come from `types.BoundContract.ReadIdentifier()`"

pkg/types/chain_reader.go
```diff
@@ -45,37 +46,40 @@ type ChainReader interface {
...
+	Params         any
+	ReturnVal      any
 }
```
ilija42, line 75: I am scared of this being uninutitive, WDYT about changing  `type BatchGetLatestValuesRequest map[string]ContractBatch` to type `BatchGetLatestValuesRequest map[BoundContract]` and keep BatchRead unchanged. This should keep the api querying more uniform and maybe more intuitive?

pkg/types/chain_reader.go
```diff
@@ -45,37 +46,40 @@ type ChainReader interface {
...
 	// Contract call errors are returned in the Err field of BatchGetLatestValuesResult.
 	BatchGetLatestValues(ctx context.Context, request BatchGetLatestValuesRequest) (BatchGetLatestValuesResult, error)
 
-	// Bind will override current bindings for the same contract, if one has been set and will return an error if the
-	// contract is not known by the ChainReader, or if the Address is invalid
+	// Bind will add provided bindings and will return an error if the contract is not known by the ContractReader, or if
+	// the Address is invalid. Any provided binding that already exists should result in a noop.
 	Bind(ctx context.Context, bindings []BoundContract) error
 
+	// Unbind will remove all bindings provided.
```
ilija42, line 61: nit, does "Unbind will remove all provided bindings." sound better?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -104,7 +105,7 @@ func TestBind(t *testing.T) {
 				es.err = errorType
 				t.Run("Bind unwraps errors from server "+errorType.Error(), func(t *testing.T) {
```
ilija42, line 106: Add similar test for Bind

pkg/types/interfacetests/chain_reader_interface_tests.go
EasterTheBunny: I need to get a specific binding, but yes it is repeated a lot. I had thought about adding a new testing method to get specific bindings.

pkg/types/chain_reader.go
EasterTheBunny: As long as I add a `String()` method, that should work fine. Seems a bit cleaner that way.

EasterTheBunny: The map structure of the GRPC interface has a problem with `map<BoundContract, ContractBatch>`. What if we just went with a simple array in the GRPC interface and the map in the Go type?

pkg/types/chain_reader.go
EasterTheBunny: 6 of one, half dozen of another. Every english/literature teacher in my past would be appalled at my writing. lol

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -12,41 +12,46 @@ service ChainReader {
...
 // QueryKeyRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.QueryKey].
 message QueryKeyRequest {
-  string contract_name = 1;
+  BoundContract contract = 1;
   QueryKeyFilter filter = 2;
   LimitAndSort limit_and_sort = 3;
 }
 
-// BindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind].
+// BindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind/Unbind].
```
ilija42, line 37:
```suggestion
// BindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind].
```

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -12,41 +12,46 @@ service ChainReader {
...
 message BindRequest {
   repeated BoundContract bindings = 1;
 }
 
+// UnbindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Bind/Unbind].
```
ilija42, line 42:
```suggestion
// UnbindRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.Unbind].
```

pkg/types/chain_reader.go
```diff
@@ -109,6 +113,10 @@ type BoundContract struct {
 	Name    string
 }
 
-func (bc BoundContract) Key() string {
+func (bc BoundContract) ReadIdentifier(method string) string {
+	return bc.Address + "-" + bc.Name + "-" + method
+}
```
ilija42, line 118: nit
```suggestion
func (bc BoundContract) ReadIdentifier(readName string) string {
	return bc.Address + "-" + bc.Name + "-" + readName
}
```

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -300,25 +337,54 @@ func (c *Server) QueryKey(ctx context.Context, request *pb.QueryKeyRequest) (*pb
...
+
+	return &emptypb.Empty{}, c.impl.Unbind(ctx, tBindings)
+}
+
+func getContractEncodedType(readIdentifier string, possibleTypeProvider any, forEncoding bool) (any, error) {
 	if ctp, ok := possibleTypeProvider.(types.ContractTypeProvider); ok {
-		return ctp.CreateContractType(contractName, itemType, forEncoding)
+		val, err := ctp.CreateContractType(readIdentifier, forEncoding)
+
+		return val, err
```
ilija42, line 365: nit, why change this
```suggestion
		return ctp.CreateContractType(readIdentifier, forEncoding)
```

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -330,37 +396,53 @@ func newPbBatchGetLatestValuesReply(result types.BatchGetLatestValuesResult, enc
...
 }
 
 func convertBatchGetLatestValuesRequestToProto(request types.BatchGetLatestValuesRequest, encodeWith EncodingVersion) (*pb.BatchGetLatestValuesRequest, error) {
-	pbRequest := &pb.BatchGetLatestValuesRequest{Requests: make(map[string]*pb.ContractBatch)}
-	for contractName, contractBatch := range request {
-		if _, ok := pbRequest.Requests[contractName]; !ok {
-			pbContractBatch := &pb.ContractBatch{Reads: []*pb.BatchRead{}}
-			pbRequest.Requests[contractName] = pbContractBatch
-		}
+	requests := make([]*pb.ContractBatch, len(request))
```
ilija42, line 411: why not simplify this function a bit

`	for binding, nextBatch := range request {
		batch := &pb.ContractBatch{Contract: convertBoundContractToProto(binding)}
		
		for _, batchCall := range nextBatch {
			versionedParams, err := EncodeVersionedBytes(batchCall.Params, encodeWith)
			if err != nil {
				return nil, err
			}

			batch.Reads = append(batch.Reads, &pb.BatchRead{
				ReadName: batchCall.ReadName,
				Params:   versionedParams,
			})
		}

		requests = append(requests)
	}
`

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -520,16 +602,18 @@ func parseBatchGetLatestValuesReply(request types.BatchGetLatestValuesRequest, r
...
+	for _, contractBatch := range reply.Results {
+		binding := convertBoundContractFromProto(contractBatch.Contract)
+		result[binding] = make([]types.BatchReadResult, len(contractBatch.Results))
 		resultsContractBatch := contractBatch.Results
-		requestContractBatch, ok := request[contractName]
+
+		requestContractBatch, ok := request[binding]
 		if !ok {
-			return nil, fmt.Errorf("received unexpected contract name %s from grpc BatchGetLatestValues reply", contractName)
+			return nil, fmt.Errorf("received unexpected contract name %s from grpc BatchGetLatestValues reply", binding)
```
ilija42, line 612:
```suggestion
			return nil, fmt.Errorf("received unexpected contract binding %s from grpc BatchGetLatestValues reply", binding)
```

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
ilija42, line 411: This will render this check `if len(requestContractBatch) != len(resultsContractBatch) ` in ` parseBatchGetLatestValuesReply()` useless

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -557,35 +643,45 @@ func convertBatchGetLatestValuesRequestFromProto(pbRequest *pb.BatchGetLatestVal
...
+		binding := convertBoundContractFromProto(batch.Contract)
+
+		if _, ok := request[binding]; !ok {
+			request[binding] = make([]types.BatchRead, len(batch.Reads))
 		}
 
-		for _, pbCall := range pbContractBatch.Reads {
-			call := types.BatchRead{ReadName: pbCall.ReadName}
-			params, err := getContractEncodedType(pbContractName, pbCall.ReadName, impl, true)
+		for idx, readCall := range batch.Reads {
```
ilija42, line 653: I find these conversion funcs easier to read if pb is prepended to protobuf types

ilija42 (commented):
left some easily addressable nits

ilija42 (approved):

ilija42 (approved):

jmank88 (approved):

PR #605, opened 2024-06-20
Expose capability registry from relayer

archseer:
core ref: 1aaac0375d8c17b9309eecfce1e4704186d95425


pkg/loop/internal/core/services/capability/standard_capabilities.go
```diff
@@ -33,7 +34,7 @@ type StandardCapabilities interface {
 type StandardCapabilitiesClient struct {
 	*goplugin.PluginClient
 	capabilitiespb.StandardCapabilitiesClient
-	*baseCapabilityClient
+	// *baseCapabilityClient
```
archseer, line 37: Not sure if this is used or not, would need to make the types public otherwise

cedric-cordenier: I think this is probably safe to remove, @ettec would you agree?

pkg/loop/internal/core/services/reportingplugin/ocr3/reporting_plugin_service.go
```diff
@@ -96,9 +96,6 @@ func (o *ReportingPluginServiceClient) NewReportingPluginFactory(
 		deps.Add(keyValueStoreRes)
 
 		relayerSetServer, relayerSetServerRes := relayerset.NewRelayerSetServer(o.Logger, relayerSet, o.BrokerExt)
-		if err != nil {
```
archseer, line 99: This was comparing an error that's already previously checked, `NewRelayerSetServer` doesn't return an err

pkg/loop/internal/relayer/relayer.go
```diff
@@ -88,14 +98,23 @@ func (p *pluginRelayerServer) NewRelayer(ctx context.Context, request *pb.NewRel
...
+	if err != nil {
+		return nil, net.ErrConnDial{Name: "CapabilityRegistry", ID: request.KeystoreID, Err: err}
+	}
+	crRes := net.Resource{Closer: capRegistryConn, Name: "CapabilityRegistry"}
+	capRegistry := capability.NewCapabilitiesRegistryClient(capRegistryConn, p.BrokerExt)
+
+	r, err := p.impl.NewRelayer(ctx, request.Config, newKeystoreClient(ksConn), capRegistry)
 	if err != nil {
 		p.CloseAll(ksRes)
+		p.CloseAll(crRes)
```
cedric-cordenier, line 111: @archseer Note, you can just `p.CloseAll(ksRes, crRes)`

cedric-cordenier (dismissed):

pkg/loop/internal/relayer/relayer.go
```diff
@@ -88,14 +98,23 @@ func (p *pluginRelayerServer) NewRelayer(ctx context.Context, request *pb.NewRel
 		return nil, net.ErrConnDial{Name: "Keystore", ID: request.KeystoreID, Err: err}
 	}
 	ksRes := net.Resource{Closer: ksConn, Name: "Keystore"}
-	r, err := p.impl.NewRelayer(ctx, request.Config, newKeystoreClient(ksConn))
+	capRegistryConn, err := p.Dial(request.KeystoreID)
+	if err != nil {
+		return nil, net.ErrConnDial{Name: "CapabilityRegistry", ID: request.KeystoreID, Err: err}
```
jmank88, line 103:
```suggestion
	capRegistryConn, err := p.Dial(request.CapabilityRegistryID)
	if err != nil {
		return nil, net.ErrConnDial{Name: "CapabilityRegistry", ID: request.CapabilityRegistryID, Err: err}
```

archseer: Good catch 🤦🏻

pkg/loop/internal/core/services/capability/capabilities.go
```diff
@@ -120,10 +120,10 @@ func (c *baseCapabilityServer) Info(ctx context.Context, request *emptypb.Empty)
 		return nil, err
 	}
 
-	return capabilityInfoToCapabilityInfoReply(info), nil
+	return InfoToReply(info), nil
 }
 
-func capabilityInfoToCapabilityInfoReply(info capabilities.CapabilityInfo) *capabilitiespb.CapabilityInfoReply {
+func InfoToReply(info capabilities.CapabilityInfo) *capabilitiespb.CapabilityInfoReply {
```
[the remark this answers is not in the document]
archseer: I've kept them as is since they're symmetric and only do a conversion to protobuf and back. I can rename if there's a strong preference

jmank88 (approved):

bolekk (approved):

PR #606, opened 2024-06-20
[BCF-3269] - Contract Reader BatchGetLatestValues

ilija42:
core ref: 14ade151fb28fbd072b8570786ccf8514b649eb7
solana ref: e56d2f19a66028c3b64115e923e61c5aad5e3578

Related:
[Ticket](https://smartcontract-it.atlassian.net/browse/BCF-3269)
[Core](https://github.com/smartcontractkit/chainlink/pull/13635)
[Solana](https://github.com/smartcontractkit/chainlink-solana/pull/764)


pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -213,6 +215,228 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
...
+				require.NoError(t, err)
+
+				anyContractBatch := result[AnyContractName]
+				require.NoError(t, anyContractBatch[0].Err)
+				assert.Equal(t, MethodReturningUint64, anyContractBatch[0].ReadName)
+				assert.Equal(t, AnyValueToReadWithoutAnArgument, *anyContractBatch[0].ReturnValue.(*uint64))
+			},
+		},
+		{
+			name: "BatchGetLatestValue allows a contract name to resolve different contracts internally",
```
patrickhuie19, line 269: does `TestContract` resolve to different contracts?

pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -46,6 +47,7 @@ const AnyExtraValue = 3
 
 func RunChainReaderInterfaceTests[T TestingT[T]](t T, tester ChainReaderInterfaceTester[T]) {
 	t.Run("GetLatestValue for "+tester.Name(), func(t T) { runChainReaderGetLatestValueInterfaceTests(t, tester) })
+	t.Run("BatchGetLatestValue for "+tester.Name(), func(t T) { runChainReaderBatchGetLatestValueInterfaceTests(t, tester) })
```
patrickhuie19, line 50: /nit `"BatchGetLatestValue for " + tester.Name(),`

ilija42: this is how Goland lints it and the linter doesn't complain 🤷

jmank88: Strings might get special treatment for readability since they often have leading or tailing whitespace :shrug:

pkg/types/chain_reader.go
```diff
@@ -46,6 +46,12 @@ type ChainReader interface {
 	// Similarly, when using a struct for returnVal, fields in the return value that are not on-chain will not be set.
 	GetLatestValue(ctx context.Context, contractName, method string, params, returnVal any) error
 
+	// BatchGetLatestValue batches get latest value calls based on request, which is grouped by contract names that each have a slice of BatchRead.
+	// BatchRead params and returnVal follow same rules as GetLatestValue params and returnVal arguments, with difference in how response is returned.
+	// BatchGetLatestValueResult response is grouped by contract names, which contain read results that maintain the order from the request.
+	// Contract call errors are returned in the Err field of BatchReadResult.
+	BatchGetLatestValue(ctx context.Context, request BatchGetLatestValueRequest) (BatchGetLatestValueResult, error)
```
jmank88, line 53: The name feels a little clunky to me. Does any of these read more intuitively?
```suggestion
	GetBatchLatestValue(ctx context.Context, request GetBatchLatestValueRequest) (GetBatchLatestValueResult, error)
```
```suggestion
	GetLatestValueBatch(ctx context.Context, request GetLatestValueBatchRequest) (GetLatestValueBatchResult, error)
```
```suggestion
	GetLatestValues(ctx context.Context, request GetLatestValuesRequest) (GetLatestValuesResult, error)
```

pkg/types/interfacetests/chain_reader_interface_tests.go
ilija42: hmm, good catch, I copied some tests from GetLatestValue, the title is a bit misleading

pkg/types/chain_reader.go
ilija42: `BatchGetLatestValue` and `GetLatestValue` mimic `BatchCall` and `Call`, so the current naming makes sense to me because of that. If you have a strong opinion I would at least keep Batch in the name, `GetLatestValueBatch` could work

jmank88: Are those the eth client names? We need to be careful not to let that influence the agnostic API too much.

@krehermann where is that google naming guide? Does it cover list kind of multivalue request? because 'List' doesn't really fit.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -228,6 +243,20 @@ func (c *Server) GetLatestValue(ctx context.Context, request *pb.GetLatestValueR
 	return &pb.GetLatestValueReply{RetVal: encodedRetVal}, nil
 }
 
+func (c *Server) BatchGetLatestValue(ctx context.Context, pbRequest *pb.BatchGetLatestValueRequest) (*pb.BatchGetLatestValueReply, error) {
```
EasterTheBunny, line 246: Are there any limits to batch size? Should we consider them or leave that up to the implementations?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -327,10 +392,6 @@ func (f *fakeChainReader) GetLatestValue(_ context.Context, contractName, method
 			}
 		}
 		return types.ErrNotFound
-	} else if method == DifferentMethodReturningUint64 {
```
EasterTheBunny, line 330: What was the purpose of removing this? Is it not used in the tests?

pkg/loop/internal/relayer/pluginprovider/chainreader/helper_test.go
```diff
@@ -77,7 +77,7 @@ func (fakeTypeProvider) CreateContractType(_, itemType string, isEncode bool) (a
 			return &LatestParams{}, nil
 		}
 		return &TestStruct{}, nil
-	case MethodReturningUint64, DifferentMethodReturningUint64:
```
EasterTheBunny, line 80: Same comment here about removal. I was thinking this was a method with a different contract name. Maybe it's not necessary?

pkg/types/chain_reader.go
EasterTheBunny: I tend to like the plural version and it works well with code completion: `GetLatestValue_`.

pkg/types/codec.go
```diff
@@ -11,6 +11,7 @@ const (
 	ErrNotASlice       = InvalidArgumentError("element is not a slice")
 )
 
+//go:generate mockery --quiet --name Encoder --output ./mocks/ --case=underscore
```
EasterTheBunny, line 14: We really should consider using a mockery config file. I think @jmank88 played with this a bit already?

pkg/types/interfacetests/chain_reader_interface_tests.go
EasterTheBunny: String concatenation gets condensed for me too. Seems like a common pattern.

pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -88,20 +90,7 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			},
 		},
 		{
-			name: "Get latest value allows a contract name to resolve different contracts internally",
```
EasterTheBunny, line 91: Why eliminate this test?

pkg/types/codec.go
jmank88: https://github.com/smartcontractkit/chainlink-common/pull/607

jmank88: Oh wait, are these new mocks being used in this repo? If not, they should be created from the repos that use them instead.

pkg/types/interfacetests/chain_reader_interface_tests.go
ilija42: It doesn't test what it says and the thing that it actually tests already has a test

pkg/types/codec.go
ilija42: @jmank88  wdym by 
> they should be created from the repos that use them instead.
How does that work?

pkg/loop/internal/relayer/pluginprovider/chainreader/helper_test.go
ilija42: answered in another comment

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
ilija42: answered in another comment

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
ilija42: I would leave that to the implementation. Right now the evm impl. uses defaults, but it if needed it can be added to interface level or to config

pkg/types/chain_reader.go
ilija42: Im okay with `GetLatestValues`

jmank88: IMHO `Values` is the most ambiguous option, and something we might want to use later for a different purpose.

ilija42: tbh `GetLatestValueBatch` and `GetBatchLatestValue` don't feel less clunky than `BatchGetLatestValue`, but I am not too defensive

jmank88: I'm worried about presuming we will only need one plural form, but maybe that isn't a realistic concern :shrug:

ilija42: Changed to `BatchGetLatestValues`

pkg/types/interfacetests/chain_reader_interface_tests.go
ilija42: my bad, I returned the test, it makes sense in the evm impl.

pkg/types/interfacetests/chain_reader_interface_tests.go
ilija42: Actually my bad, this test does resolve different contracts in chain impl. tests

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader.go
```diff
@@ -228,6 +243,20 @@ func (c *Server) GetLatestValue(ctx context.Context, request *pb.GetLatestValueR
...
+	if err != nil {
+		return nil, err
+	}
+
+	reply, err := c.impl.BatchGetLatestValues(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return newPbBatchGetLatestValuesReply(reply, c.encodeWith)
```
krehermann, line 257: so if one of the embedded calls results in an error, this returns no error. is that what user would expect? i kind of doubt it.

every caller will have to loop thru all the of result and check every string Err for a value. 

would it be easier to use if this returns an errors.Join of all the accumulated inner errors or something similar? 

my main concern is whether or not callers will correctly check for individual errors. perhaps really good documentation is sufficient, but that seems filmsy

ilija42: This is how our current implementation of batch call works [here](https://github.com/smartcontractkit/ccip/blob/54ee4f13143d3e414627b6a0b9f71d5dfade76c5/core/services/ocr2/plugins/ccip/internal/rpclib/evm.go#L25). Although seems like outputs are always parsed with [ParseOutputs()](https://github.com/smartcontractkit/ccip/blob/dacf09fbba3fa20e1fc596640b74f8b15d3d2b58/core/services/ocr2/plugins/ccip/internal/rpclib/evm.go#L297), which returns an error as soon as it encounters one, but it does leave room for handling this differently if you don't care if only some results error.

Probably best to ask people that will use this what is the best approach. CC @patrickhuie19 @dimkouv 

Do you want `BatchGetLatesetValues()` to return an `errors.Join` of all the accumulated errors if there are any? Or is the current approach okay where you have to loop through responses to check for errors ( which you have to do anyway to collect the results)? Or do you have something else in mind?

ilija42: Also the same parse function can be used, just without the parse part since codec does that now, which would just make it an error checking loop

dimkouv: hmm.. so It's going to be used like this:

```
myBatchCalls = [...]

results, err := reader.BatchCall(myBatchCalls)
if err -> return err

for res in results:
   if res.err -> return err
   handle res.data
```

Someone has to forget the `if res.err -> return err` which might happen...

Maybe we can somehow enforce this check by doing:
```
myBatchCalls = [...]

results, err := reader.BatchCall(myBatchCalls)
if err -> return err

for res in results:
   data, err = res.data()
```

ilija42: Nice, I like the approach in the second example, I'll hide data behind a getter that checks for error

ilija42: resolved

krehermann (approved):

dimkouv (approved):

jmank88 (approved):

PR #607, opened 2024-06-20
mockery: use config file instead of generate statements

jmank88:
This PR proposes converting to a mockery config file.
https://vektra.github.io/mockery/latest/features/#packages-configuration
https://vektra.github.io/mockery/latest/migrating_to_packages/


pkg/codec/encodings/binary/gen/main.go
```diff
@@ -33,22 +39,9 @@ func main() {
 			panic(err)
 		}
 
-		_, location, _, ok := runtime.Caller(0)
-		if !ok {
-			panic("cannot find location of file")
-		}
-		// We need to use strings.LastIndex instead of strings.Index because the
-		// file path in CI can contain the string "chainlink-common" multiple
```
jmank88, line 41: I still have an old 	`chainlink-relay/` dir, so this was failing for me. The new approach should be equivalent, without caring about directory names.

mockery.yaml
```diff
@@ -0,0 +1,39 @@
...
+      mockname: "{{ .InterfaceName }}Mock"
+    interfaces:
+      Metrics:
+      Source:
+      SourceFactory:
+      Exporter:
+      ExporterFactory:
+  github.com/smartcontractkit/chainlink-common/pkg/services:
+    interfaces:
+      Foo:
```
krehermann, line 28: Foo? that not a real thing :) what's going on?

mockery.yaml
```diff
@@ -0,0 +1,39 @@
...
+  github.com/smartcontractkit/chainlink-common/pkg/loop/internal/types:
+    interfaces:
+      Relayer:
+        config:
+          dir: "{{ .InterfaceDir }}/../../mocks"
+  github.com/smartcontractkit/chainlink-common/pkg/monitoring:
+    config:
+      dir: "{{ .InterfaceDir }}"
+      filename: "{{ .InterfaceName | snakecase }}_mock.go"
+      inpackage: true
```
krehermann, line 18: do you know why monitoring is different than other packages?

mockery.yaml
jmank88: CI should be failing :cry:

mockery.yaml
jmank88: I thought there used to be a whole separate module under `ops` that used it but now I'm not so sure :shrug:

pkg/services/types.go
```diff
@@ -11,7 +11,6 @@ import "context"
 //
 // Mockable Foo service with a run loop
 //
-//	//go:generate mockery --quiet --name Foo --output ../internal/mocks/ --case=underscore
```
jmank88, line 14: @krehermann oh I didn't notice it was a double comment :yodawg:

mockery.yaml
jmank88: Fixed but I don't understand why it doesn't fail :thinking:

krehermann: why would ci fail? iiuc this change would generate `Foo` mocks that simple aren't used anywhere. why would that additive, unused code fail CI?

jmank88: It is referencing a type which doesn't exist. If we typo something like this, we want it to fail. Otherwise we could accidentally disable things without realizing it.

krehermann (approved):

momentmaker:
fixed an issue here related to gha:
https://github.com/smartcontractkit/chainlink-common/pull/625

samsondav (approved):

krehermann (dismissed):

krehermann (approved):

samsondav (approved):

PR #609, opened 2024-06-25
[KS-356] Support for ID remapping in feeds aggregator

bolekk:


pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -145,12 +149,17 @@ func (a *dataFeedsAggregator) Aggregate(previousOutcome *types.AggregationOutcom
 	toWrap := []any{}
 	for _, report := range reportsNeedingUpdate {
 		feedID := datastreams.FeedID(report.FeedID).Bytes()
+		remappedID := a.config.Feeds[datastreams.FeedID(report.FeedID)].RemappedID
```
krehermann, line 152: this is odd. why not use the canonical 
x, exists := mymap[y]
if !exists { x = ...}

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -312,6 +321,14 @@ func ParseConfig(config values.Map) (aggregatorConfig, error) {
 			feedCfg.Deviation = dec
 			parsedConfig.Feeds[feedID] = feedCfg
 		}
+		if len(feedCfg.RemappedIDHex) > 2 {
```
krehermann, line 324: what is magical about 2? are you trying to check for a `0x` prefix? if so, string trim it explicitly.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk: This is not a check for existence in a map - it has to exist at this point. RemappedID field can simply be empty and then we fall back to original ID (I'll add a comment)

jmank88 (approved):

krehermann (approved):

PR #613, opened 2024-06-25
add QueryKey helper functions

EasterTheBunny:


pkg/utils/query/query.go
```diff
@@ -0,0 +1,338 @@
...
+- EventSigComparator
+- TopicComparator
+*/
+package query
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
```
jmank88, line 16: Why do we need two separate `package query`s?

EasterTheBunny: Moved to `pkg/types/query`

pkg/types/query/helpers.go
jmank88, line 1: So these are all functionally `KeyFilter` constructors, and the package docs will organize them that way.
- Would it make sense to name the file more specifically?
- Did you consider naming these funcs in a more idiomatic way that includes `KeyFilter`?

pkg/types/query/helpers.go
```diff
@@ -0,0 +1,328 @@
...
+
+const (
+	AddressComparator  ComparatorType = "ADDRESS"
+	EventSigComparator ComparatorType = "EVENT"
+	TopicComparator    ComparatorType = "TOPIC"
+)
+
+// SelectIndexedLogs creates a KeyFilter that filters logs for the provided topic values at the specified
+// topic index. Topic value filters are 'OR'ed together.
+func SelectIndexedLogs(
```
jmank88, line 20: As an example from my other comment:
```suggestion
func IndexedLogsKeyFilter(
```
Something like this would be more self explanatory, and not mistaken for performing an operation like `Select*` might be.

pkg/types/query/helpers.go
EasterTheBunny: I named them the same as the `LogPoller` equivalent to ensure there is no confusion on the expected results. I could rename the file to `constructors.go` or something like that to offer a better description.

pkg/types/query/helpers.go
EasterTheBunny: This could work fine. I just don't want to deviate too much from the original name yet. These are intended to be pre-rolled filters that help in the transition from using `LogPoller` directly.

pkg/types/query/helpers.go
jmank88: IMHO something like `key_filter.go` would make more sense than `constructors.go`.

pkg/types/query/helpers.go
jmank88: Another option would be `query/keyfilter/keyfilter.go`, so `package keyfilter`, with users calling `keyfilter.IndexedLogs(...)` instead of `query.IndexedLogsKeyfilter(...)`

EasterTheBunny: I'm not sure if we even want to keep these as long-term names anyway. We've been trying to move away from `Logs` as a term in favor of `Keys`. But for now, naming them as close to the original can be beneficial for ease of transition.

jmank88: Everything exported from this module needs to be thinking long-term, because we will have to stamp v1.0.0 and commit to a stable API.

ilija42: Right now all of these are EMV impl. keywords: logs, events, topics, indexed, words. Its good UX for devs to have naming similar to Log Poller, but as Jordan said we should be thinking long-term here. Not sure how to make best use of these helper functions. In Q3 we will work on improving CR docs and adding some dev guides, one approach that comes to mind is to keep these helper funcs completely chain agnostic and to document how they map to chain specific concepts. Something similar will have to exist anyway and will have to be maintained by BIX later down the line.

Thoughts? @mateusz-sekara @dimkouv

pkg/types/query/constructors.go
```diff
@@ -0,0 +1,328 @@
...
+			Block(start, primitives.Gte),
+			Block(end, primitives.Lte),
+		},
+	}
+}
+
+// IndexedLogsTopicGreaterThanKeyFilter creates a KeyFilter that filters logs for the provided topic value and index
+// at or above the specified confidence level.
+func IndexedLogsTopicGreaterThanKeyFilter(
+	addr, eventSig string,
```
dimkouv, line 60: I believe this helpers violate the chain agnostic aspect of `ContractReader`.
Because with this changes cl-common now exposes evm specific details, such as `event sig, logs, log idx, ...`. This means that the caller should be aware of this information, for example CCIP logic should have a reference to an event sig, but this event sig will not exist on two different chain fams.

tl;dr I like the idea of having this helpers but I believe it will lead to chain-specific product implementations and violation of the whole purpose of `ContractReader`.

On the other hand though, If this makes sense and this is something that products will have to do, maybe should at least be under a package that indicates it's `evm` related helpers.

pkg/types/query/constructors.go
```diff
@@ -0,0 +1,328 @@
...
+
+const (
+	AddressComparator  ComparatorType = "ADDRESS"
+	EventSigComparator ComparatorType = "EVENT"
+	TopicComparator    ComparatorType = "TOPIC"
+)
+
+// IndexedLogsKeyFilter creates a KeyFilter that filters logs for the provided topic values at the specified
+// topic index. Topic value filters are 'OR'ed together.
+func IndexedLogsKeyFilter(
```
dimkouv, line 20: I am thinking whether it makes sense to have unit tests for this functions.

pkg/types/query/constructors.go
EasterTheBunny: I changed from event sig to sequence signature (which identifies the sequence type), topic index to sequence value name (which defines the indexed property name), and sequence value which is the filtered value in the defined property to search. I made some changes to display this in the followup commit. Let me know your thoughts.

EasterTheBunny: I do see some issues in the data word index filters as these appear to be doing filtering on raw log data, which would not be consistent between chains.

pkg/types/query/key_filter.go
```diff
@@ -0,0 +1,164 @@
...
+	confidence primitives.ConfidenceLevel,
+) KeyFilter {
+	return KeyFilter{
+		Key: readName,
+		Expressions: []Expression{
+			filtersForValues(propertyName, values),
+			Confidence(confidence),
+		},
+	}
+}
```
jmank88, line 26: What do you think about making these methods on `KeyFilter`?
```suggestion
func (f *KeyFilter) IndexedSequences(
	propertyName string,
	values []string,
	confidence primitives.ConfidenceLevel,
) {
	f.Expressions = []Expression{
		filtersForValues(propertyName, values),
		Confidence(confidence),
	}
}
```
This leads to shorter func names names and fewer arguments (and keeps the package namespace cleaner)
```go
filter := IndexedSequencesKeyFilter(name, property, values, confidence)
```
```go
filter := KeyFilter{ReadName: name}
filter.IndexedSequences(property, values, confidence)
```
Perhaps the method name could be refined further :shrug:

pkg/types/query/key_filter.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
+)
+
+// IndexedSequencesKeyFilter creates a KeyFilter that filters logs for the provided sequence property values at the
+// specified property name. Sequence value filters are 'OR'ed together. A sequence read name is the value that
+// identifies the sequence type. The signature value name is the sequence property to apply the filter to and the
+// sequence values are the individual values to search for in the provided property.
+func IndexedSequencesKeyFilter(
+	readName string,
+	propertyName string,
```
ilija42, line 15: lets not introduce new terms, what about just callng this comparatorName

pkg/types/query/key_filter.go
EasterTheBunny: These are intended to only be helper functions that should produce the same queries as the ones in log poller. I would expect the usage as such:

```
contract := types.BoundContract{Address: "0x", Name: "contract"}
seq, err := relay.QueryKey(ctx, contract, IndexedSequencesKeyFilter(readName, property, values, confidence), limiter, retVal)
```

The method name is long, but the original was something like `SelectIndexedLogs` so we keep with the same pattern for the sake of familiarity.

jmank88: I don't think that context changes anything except perhaps to prefer the fluent version.

samsondav (approved):

reductionista (approved):
LGTM

pkg/types/query/key_filter.go
reductionista: I think I agree with @jmank88 on this, including KeyFilter in every function name is a little confusing so making it a method seems like it might be slightly preferable. But I don't have a strong preference so fine with it either way.

PR #614, opened 2024-06-25
build & test: dump data races to separate file

jmank88:


.github/workflows/pkg.yml
```diff
@@ -18,7 +18,15 @@ jobs:
         run: go build -v ./...
 
       - name: Unit Tests
-        run: go test -race ./... -coverpkg=./... -coverprofile=pkg_coverage.out
+        run: GORACE="log_path=$PWD/race" go test -race ./... -coverpkg=./... -coverprofile=pkg_coverage.out
+
+      - name: Print Races
+        id: print-races
+        run: |
+          find race.* | xargs cat > race.txt
```
HenryNguyen5, line 26: more than one race file gets generated?

.github/workflows/pkg.yml
```diff
@@ -18,7 +18,15 @@ jobs:
...
+          if [[ -s race.txt ]]; then
+            cat race.txt
+          fi
```
HenryNguyen5, line 29:
```suggestion
        run: cat race.*
```

Could we do this instead?

jmank88: Possibly, but I copied core: https://github.com/smartcontractkit/chainlink/blame/develop/.github/workflows/ci-core.yml#L182-L189
cc @momentmaker

momentmaker: I think the issue/bug before was when there is no `race.txt` that was output and this will fail

momentmaker: chatgpt says something like this?
```
  if ls race.* 1> /dev/null 2>&1; then
    cat race.* > race.txt
    cat race.txt
  fi
  ```

jmank88: Hmm is that an improvement?

momentmaker: good question - I think it's more a preference thing?

HenryNguyen5: Feel free to resolve this, if you copied it from core its fine as-is, i didnt realize we had an edge case to cover. If anything, a comment would be nice

PR #616, opened 2024-06-27
[KS-358] Custom UnwrapTo for FeedReport list

bolekk:
It's much faster than mapstructure/reflection.
Also adding a test for values.Wrap(), which doesn't show any excessive slowness.


DeividasK (approved):

DeividasK:
Isn't there a way to improve/replace the generic unwrap function?

bolekk:
> Isn't there a way to improve/replace the generic unwrap function?

Yes, @cedric-cordenier is working on improving it separately. However, it won't get close to performance of a custom one and this particular structure is in the hot path of aggregations so I think it makes sense to have a special one.

ettec (approved):

PR #620, opened 2024-06-27
henry/keystone capability registry get local node

HenryNguyen5:
- **Remove stale mock usage**
- **Update CR interface to include GetLocalNode**
- **Implement GetLocalNode GRPC Method**


pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -21,6 +22,44 @@ type capabilitiesRegistryClient struct {
 	grpc pb.CapabilitiesRegistryClient
 }
 
+func toPeerID(peerID []byte) p2ptypes.PeerID {
+	return p2ptypes.PeerID(peerID)
+}
+
+func toDON(wf *pb.DON) capabilities.DON {
```
krehermann, line 29: /nit `wf` seems misleading. is there anything workflow specific about the input arg?

pkg/loop/internal/core/services/capability/capabilities_registry_test.go
```diff
@@ -153,6 +169,65 @@ func TestCapabilitiesRegistry(t *testing.T) {
 	_, err = rc.Get(tests.Context(t), "some-id")
 	require.ErrorContains(t, err, "capability not found")
 
+	pid := p2ptypes.PeerID(randomWord())
+	node := capabilities.Node{
+		PeerID: &pid,
+		WorkflowDON: capabilities.DON{
+			ID: "workflow-don-id",
+			Members: []p2ptypes.PeerID{
+				randomWord(),
```
krehermann, line 178: /nit if there is no explicit utility in the randomness of these values then i would prefer determenisitic tests.

note: you can easily allocate bytes by assigning values at indices
[]byte{0:1, 31:2} //sets byte 0 to `1` and byte 31 to `2`
https://go.dev/play/p/pCSesxuK8GS

pkg/loop/internal/pb/capabilities_registry.proto
```diff
@@ -22,6 +23,22 @@ enum ExecuteAPIType {
   EXECUTE_API_TYPE_CALLBACK = 2;
 }
 
+
+// DON has arguments for [github.com/smartcontractkit/chainlink-common/pkg/capabilities.DON].
+message DON {
+  string id = 1;
+  repeated bytes members = 2; // [membern][32]byte
```
krehermann, line 30: typo in `membern`?  [][32]byte?

pkg/loop/internal/pb/capabilities_registry.proto
```diff
@@ -22,6 +23,22 @@ enum ExecuteAPIType {
...
+  uint32 f = 3; // uint8
+  bytes config  = 4;
+}
+
+// GetLocalNode has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.CapabilitiesRegistry.GetLocalNode].
+message GetLocalNodeReply {
+  bytes peerID = 1; // [32]byte
+  DON workflowDON = 2;
```
krehermann, line 38: it's worth a comment here that a node is asymmetric in membership: it only allowed to be a member of 1 workflow don but any number of capabilities dons, and IIUC that is enforced by the contrace

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -202,6 +241,44 @@ func (c *capabilitiesRegistryServer) Get(ctx context.Context, request *pb.GetReq
...
+			Id:      don.ID,
+			Members: membersBytes,
+			F:       uint32(don.F),
+			Config:  don.Config,
+		}
+	}
+
+	membersBytes := make([][]byte, len(node.WorkflowDON.Members))
+	for i, m := range node.WorkflowDON.Members {
+		membersBytes[i] = m[:]
```
bolekk, line 266: You need to add m := m here

pkg/loop/internal/pb/capabilities_registry.proto
HenryNguyen5: not a typo, but if its not obvious i should change it to `n` instead, thoughts? its supposed to mean the nth member

pkg/loop/internal/core/services/capability/capabilities_registry_test.go
HenryNguyen5: there isnt, i just copied from tests in the chainlink repo that do the same thing

pkg/loop/internal/core/services/capability/capabilities_registry.go
HenryNguyen5: That's an artifact when i was working with just the workflow don, fixed

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -202,6 +237,50 @@ func (c *capabilitiesRegistryServer) Get(ctx context.Context, request *pb.GetReq
...
+	}
+
+	// convert workflowDON to pb.DON
+	membersBytes := make([][]byte, len(node.WorkflowDON.Members))
+	for i, m := range node.WorkflowDON.Members {
+		m := m
+		membersBytes[i] = m[:]
+	}
+	workflowDONpb := &pb.DON{
+		Id:      node.WorkflowDON.ID,
```
patrickhuie19, line 253: If a node doesn't belong to a workflow DON, what do these values look like? I might have missed it, but is this case covered by unit testing?

HenryNguyen5: Good question, i dont know either lol

It should be zero values but i should check the (de)serialization logic via test 

https://github.com/smartcontractkit/chainlink-common/pull/620/files#diff-8c726a1dbd12cccdfdc7bd212b79d5bcc20cf100a7298f933e05e14dbbcd4288R196

krehermann (approved):

krehermann (approved):

pkg/capabilities/capabilities.go
```diff
@@ -174,6 +174,10 @@ type TargetCapability interface {
 	CallbackCapability
 }
 
+// DON represents a network of connected nodes.
+//
+// For an example of an empty DON check, see the following link:
+// https://github.com/smartcontractkit/chainlink/blob/2a1b0c4eecb9cc23a364d62a080cc9ae0d08c518/core/capabilities/transmission/local_target_capability.go#L31
```
bolekk, line 180: This check is actually incorrect because we changed the type from string to int. Maybe just say here that ID==0 means invalid/empty DON?

pkg/capabilities/capabilities.go
```diff
@@ -183,6 +187,20 @@ type DON struct {
...
+// There is a 1:0..1 relationship between this node and a workflow DON.
+// This means that this node can be part at most one workflow DON at a time.
+// As a side note, a workflow DON can have multiple nodes.
+//
+// There is a 1:N relationship between this node and capability DONs, where N is the number of capability DONs.
+// This means that this node can be part of multiple capability DONs at a time.
+//
+// Although WorkflowDON is a value rather than a pointer, a node can be part of no workflow DON but 0 or more capability DONs.
+// You can assert this by checking for zero values in the WorkflowDON field.
+// See https://github.com/smartcontractkit/chainlink/blob/2a1b0c4eecb9cc23a364d62a080cc9ae0d08c518/core/capabilities/transmission/local_target_capability.go#L31 for an example.
```
bolekk, line 203: same here

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -21,6 +23,44 @@ type capabilitiesRegistryClient struct {
 	grpc pb.CapabilitiesRegistryClient
 }
 
+func toDON(don *pb.DON) capabilities.DON {
```
bolekk, line 26: nit: Maybe call it "toCapDON"?

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -202,6 +242,54 @@ func (c *capabilitiesRegistryServer) Get(ctx context.Context, request *pb.GetReq
 	}, nil
 }
 
+func (c *capabilitiesRegistryServer) GetLocalNode(ctx context.Context, _ *emptypb.Empty) (*pb.GetLocalNodeReply, error) {
+	node, err := c.impl.GetLocalNode(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// convert workflowDON to pb.DON
```
bolekk, line 251: another helper like the one above maybe?

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -202,6 +242,54 @@ func (c *capabilitiesRegistryServer) Get(ctx context.Context, request *pb.GetReq
...
+		membersBytes[i] = m[:]
+	}
+	workflowDONpb := &pb.DON{
+		Id:            node.WorkflowDON.ID,
+		Members:       membersBytes,
+		F:             uint32(node.WorkflowDON.F),
+		ConfigVersion: node.WorkflowDON.ConfigVersion,
+	}
+
+	// convert capabilityDONs to pb.DON
```
bolekk, line 264: could be re-used here

bolekk (dismissed):

pkg/loop/internal/core/services/capability/capabilities_registry.go
HenryNguyen5: I'm using it for the workflow DON struct too

pkg/capabilities/capabilities.go
```diff
@@ -174,6 +174,10 @@ type TargetCapability interface {
 	CallbackCapability
 }
 
+// DON represents a network of connected nodes.
+//
+// For an example of an empty DON check, see the following link:
+// https://github.com/smartcontractkit/chainlink/blob/develop/core/capabilities/transmission/local_target_capability.go#L31
```
patrickhuie19, line 180: /nit the commit hash links age better :)

patrickhuie19 (approved):
lgtm

pkg/capabilities/capabilities.go
HenryNguyen5: I originally had that, but then ran into https://github.com/smartcontractkit/chainlink-common/pull/620#discussion_r1664885470 heh. I'd imagine it might be better if it breaks TBH. Then we'd know this needs to be updated anyway

bolekk (approved):

PR #621, opened 2024-06-28
[BCI-3438] Add chain writer support to the relayer

nickcorin:
Depends on:

- https://github.com/smartcontractkit/chainlink/pull/13718
- https://github.com/smartcontractkit/chainlink-solana/pull/767
- https://github.com/smartcontractkit/chainlink-solana/pull/780
- https://github.com/smartcontractkit/chainlink-starknet/pull/475
- https://github.com/smartcontractkit/chainlink/pull/13777
- https://github.com/smartcontractkit/chainlink/pull/13789


pkg/loop/internal/pb/chain_writer.proto
```diff
@@ -0,0 +1,45 @@
...
+    string contract_name = 1;
+    string method = 2;
+    VersionedBytes params = 3;
+    string transaction_id = 4;
+    string to_address = 5;
+    TransactionMeta meta = 6;
+    BigInt value = 7;
+}
+
+message TransactionMeta {
```
krehermann, line 27: why is this it's own message type?

if it's going to be extended and generic, did you using a map key-value pairs, like an http header and letting the application interpret as needed?

if it's not going to be extended and generic, why not embed the wf id field in the request

pkg/loop/internal/pb/chain_writer.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// GetTransactionStatusRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainWriter.GetTransactionStatus].
+message GetTransactionStatusRequest {
+    string transaction_id = 1;
+}
+
+// GetTransactionStatusReply has return arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainWriter.GetTransactionStatus].
+message GetTransactionStatusReply {
+    int64 transaction_status = 1;
```
krehermann, line 38: this would be better as an enum applications know how to interpret and set the value

pkg/loop/internal/relayer/pluginprovider/chainwriter/chain_writer.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	for _, opt := range opts {
+		opt(server)
+	}
+
+	return server
+}
+
+func WithServerEncoding(version chainreader.EncodingVersion) ServerOpt {
```
krehermann, line 117: why is this option on both the client and server?

is the server has it set can the client really choose independently and expect the server to respect it?

pkg/loop/internal/pb/chain_writer.proto
nickcorin: Sure. It's an enum in the Go code, I can add the types here too.

pkg/loop/internal/pb/chain_writer.proto
nickcorin: This is just a proto representation of the `TxMeta` struct that the `ChainWriter` uses. It might very well be extended over time. If we use KV pairs we'll probably have to use reflection to map the keys to struct fields, which seems like more work than maintaining the message.

pkg/loop/internal/relayer/pluginprovider/chainwriter/chain_writer.go
nickcorin: The encoding is not passed over the wire, so the client and server need to be configured independently.

pkg/loop/internal/relayer/test/relayer.go
```diff
@@ -135,6 +135,10 @@ func (s staticPluginRelayer) Name() string { panic("unimplemented") }
 
 func (s staticPluginRelayer) HealthReport() map[string]error { panic("unimplemented") }
 
+func (s staticPluginRelayer) NewChainWriter(_ context.Context, chainWriterConfig []byte) (types.ChainWriter, error) {
+	return nil, fmt.Errorf("not implemented")
```
jmank88, line 139:
```suggestion
	return nil, errors.New("not implemented")
```

jmank88 (approved):

krehermann (dismissed):

krehermann (approved):

PR #626, opened 2024-07-02
[KS-285] Updated DON struct; pass through DON Config Version

cedric-cordenier:
* Update DON struct to have a uint32 ID and uint32 config version -- this brings it inline with the capabilities registry contract.
* Use the config version throughout the OCR3 capability and add it to the consensus metadata.

Companion PR:
- https://github.com/smartcontractkit/chainlink/pull/13739


cedric-cordenier:
Note: I'm intentionally breaking the various protos, rather than migrating from one field to another to save some time. I think this is acceptable since it will only affect staging, but lmk if you disagree

DeividasK (approved):

ettec (approved):

PR #636, opened 2024-07-03
pkg/loop: swap out deprecated grpc API

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3290

### DIFF
--- a/pkg/loop/internal/net/test/simple_broker.go
+++ b/pkg/loop/internal/net/test/simple_broker.go
@@ -75,8 +75,7 @@ func (v *Broker) DialWithOptions(id uint32, opts ...grpc.DialOption) (conn *grpc
 		return nil, fmt.Errorf("listener with id %d does not exist", id)
 	}
 	opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
-	//TODO https://smartcontract-it.atlassian.net/browse/BCF-3290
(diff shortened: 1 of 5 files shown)


jmank88:
Updated but still failing. Something about this difference matters:
```go
// One subtle difference between NewClient and Dial and DialContext is that the
// former uses "dns" as the default name resolver, while the latter use
// "passthrough" for backward compatibility.  This distinction should not matter
// to most users, but could matter to legacy users that specify a custom dialer
// and expect it to receive the target string directly.
//
// Deprecated: use NewClient instead.  Will be supported throughout 1.x.
func DialContext(ctx context.Context, target string, opts ...DialOption) (conn *ClientConn, err error) {
	// At the end of this method, we kick the channel out of idle, rather than
	// waiting for the first rpc.
	//
	// WithLocalDNSResolution dial option in `grpc.Dial` ensures that it
	// preserves behavior: when default scheme passthrough is used, skip
	// hostname resolution, when "dns" is used for resolution, perform
	// resolution on the client.
	opts = append([]DialOption{withDefaultScheme("passthrough"), WithLocalDNSResolution()}, opts...)
	cc, err := NewClient(target, opts...)
```
Because of what go-plugin is doing here with a custom dialer:
https://github.com/hashicorp/go-plugin/blame/main/grpc_client.go#L24-L48

PR #639, opened 2024-07-05
pkg/types/ccipocr3: message type changes

makramkd:
Related PRs:
* https://github.com/smartcontractkit/ccip/pull/1140/
* https://github.com/smartcontractkit/ccip/pull/1117

Recent family-agnostic ramp changes requires a re-think of the offchain message types.

It seems most appropriate to mirror at the very least the family-agnostic ramp header which will be the same data across all chain families.

When it comes to representing the CCIP message offchain, we will use `ccipocr3.Message` to represent the "any to any" CCIP message. This refers to a common representation of a CCIP message across all chain families.

In order to correctly hash a message, we will need to transform the `ccipocr3.Message` into a `Any2Family` message where `Family` is the destination chain family.

This doesn't need to be represented in the types here because these types will not be sent "over the wire" via gRPC. Instead, the translation of `Family2Any` to `Any2Family` will occur within the message hasher implementation itself.


pkg/types/ccipocr3/generic_types.go
```diff
@@ -82,51 +82,85 @@ func (c ChainSelector) String() string {
...
-
-func (c CCIPMsg) String() string {
+// Message is the generic Any2Any message type for CCIP messages.
+// It represents, in particular, a message emitted by a CCIP onramp.
+// The message is expected to be consumed by a CCIP offramp after
+// translating it into the appropriate format for the destination chain.
+type Message struct {
+	// Header is the family-agnostic header for OnRamp and OffRamp messages.
+	// This is always set on all CCIP messages.
+	Header RampMessageHeader `json:"header"`
```
dimkouv, line 92: Is the `Header.Hash` field populated on OnRamp or it's computed offChain/offRamp as it was the case so far?

pkg/types/ccipocr3/generic_types.go
```diff
@@ -82,51 +82,85 @@ func (c ChainSelector) String() string {
...
+
+	// EVM2AnyRampMessage is populated if the CCIP message originated from an EVM chain.
+	EVM2AnyRampMessage *EVM2AnyRampMessage `json:"evm2AnyRampMessage,omitempty"`
```
dimkouv, line 95: Maybe include another field e.g. `Origin="evm"` so we can do:
```
switch Origin:
   case "evm":
      use msg.EVM2AnyRampMessage
```

otherwise we have to:
```
if msg.EVM2AnyRampMessage != nil {
  ...
}
```

pkg/types/ccipocr3/generic_types.go
makramkd: I think like before, `Hash` would be populated by the plugin. `MessageID` however is computed onchain by the onramp.

pkg/types/ccipocr3/generic_types.go
makramkd: I was imagining there'd be an API in `chain-selectors` that will do this:

```
switch chainsel.GetChainFamily(msg.DestChainSelector) {
  case chainsel.EVMFamily:
    // ...
  case chainsel.SolanaFamily:
    // ...
}
```

pkg/loop/internal/pb/ccipocr3/models.proto
```diff
@@ -4,22 +4,76 @@ option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/inter
...
+  BigInt fee_token_amount = 6;
+  // TokenAmounts is the array of tokens and amounts to transfer.
+  repeated RampTokenAmount token_amounts = 7;
+}
+
+// Message is a gRPC adapter to [github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3.Message].
+message Message {
+  // Header is the family-agnostic header for OnRamp and OffRamp messages.
+	// This is always set on all CCIP messages.
+  RampMessageHeader header = 1;
```
asoliman92, line 72: Missing Metadata field for the message?

pkg/types/ccipocr3/generic_types.go
asoliman92: I have another suggestion.

```
// Base struct
type IMessage interface {
	GetHeaders() RampMessageHeader
}

type struct WEVM2AnyRampMessage {
   message EVM2AnyRampMessage
}

func (msg * WEVM2AnyRampMessage)  GetHeaders() {
 ....
}

func (msg * WEVM2AnyRampMessage)  SomeOtherFunc() {
 ....
}

switch msg.(type) {
case WEVM2AnyRampMessage:
  msg.SomeOtherFunc()
}
```

Should be more resilient than a type string like "evm". 


> I was imagining there'd be an API in chain-selectors that will do this:
```
switch chainsel.GetChainFamily(msg.DestChainSelector) {
  case chainsel.EVMFamily:
    // use EVM2Any
  case chainsel.SolanaFamily:
    // use Sol2Any
}
```
Is this the only place that it will be called from?

pkg/types/ccipocr3/common_types.go
```diff
@@ -4,18 +4,73 @@ import (
...
+	return []byte(fmt.Sprintf(`"%s"`, b.String())), nil
+}
+
+func (b *Bytes) UnmarshalJSON(data []byte) error {
+	v := string(data)
+	if len(v) < 2 {
+		return fmt.Errorf("Bytes must be of at least length 2 (i.e, '0x' prefix): %s", v)
+	}
+
+	// trim the start and end double quotes
```
elatoskinas, line 43: q: do we need to handle edge-cases where there's leading / trailing whitespace?

pkg/types/ccipocr3/generic_types.go
```diff
@@ -82,51 +82,85 @@ func (c ChainSelector) String() string {
...
+	DestChainSelector ChainSelector `json:"destChainSelector,string"`
+	// SequenceNumber is an auto-incrementing sequence number for the message.
+	// Not unique across lanes.
+	SequenceNumber SeqNum `json:"seqNum,string"`
+	// Nonce is the nonce for this lane for this sender, not unique across senders/lanes
+	Nonce uint64 `json:"nonce"`
 
 	// MsgHash is the hash of all the message fields.
 	// NOTE: The field is expected to be empty, and will be populated by the plugin using the MsgHasher interface.
 	MsgHash Bytes32 `json:"msgHash"` // populated
```
winder, line 122: Is the destination's commit-root computed based on these hashes? If so, the on-chain code should recompute them and use the commit-root as a CRC on the parameters. This would ensure the message fields haven't been tampered with since the commit report was created.

```suggestion
	MsgHash Bytes32 `json:"-"` // populated
```

pkg/loop/internal/pb/ccipocr3/msghasher.pb.go
```diff
@@ -124,7 +124,7 @@ var file_msghasher_proto_rawDesc = []byte{
 	0x73, 0x68, 0x4d, 0x73, 0x67, 0x49, 0x6e, 0x70, 0x75, 0x74, 0x12, 0x34, 0x0a, 0x03, 0x6d, 0x73,
 	0x67, 0x18, 0x01, 0x20, 0x01, 0x28, 0x0b, 0x32, 0x22, 0x2e, 0x6c, 0x6f, 0x6f, 0x70, 0x2e, 0x69,
 	0x6e, 0x74, 0x65, 0x72, 0x6e, 0x61, 0x6c, 0x2e, 0x70, 0x62, 0x2e, 0x63, 0x63, 0x69, 0x70, 0x6f,
-	0x63, 0x72, 0x33, 0x2e, 0x43, 0x43, 0x49, 0x50, 0x4d, 0x73, 0x67, 0x52, 0x03, 0x6d, 0x73, 0x67,
+	0x63, 0x72, 0x33, 0x2e, 0x4d, 0x65, 0x73, 0x73, 0x61, 0x67, 0x65, 0x52, 0x03, 0x6d, 0x73, 0x67,
```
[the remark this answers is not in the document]
elatoskinas: Auto-generated protobuf model right? Just strange that there's hardcoded bytes :D

pkg/types/ccipocr3/generic_types.go
elatoskinas: ^ agreed on above, on-chain I'm currently thinking:

* `chainFamilyTag` gets assigned to a dest chain config, e.g. for ETH, ARB, etc. assign `EVM_FAMILY_TAG`
* Each family tag is generated deterministically, e.g. right now I use `bytes4(keccak256("CCIP EVM Family"))` (32-bit family ids should be more than enough I recon)

pkg/loop/internal/pb/ccipocr3/msghasher.pb.go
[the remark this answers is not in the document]
makramkd: I believe this is a [protobuf descriptor](https://protobuf.com/docs/descriptors#descriptor-production). Basically a serialized representation of the proto file. I don't think we make use of it directly though.

pkg/loop/internal/pb/ccipocr3/models.proto
```diff
@@ -4,22 +4,76 @@ option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/inter
...
+  uint64 source_chain_selector = 2;
+  // DestChainSelector is the chain selector of the chain that the message is destined for.
+  uint64 dest_chain_selector = 3;
+  // SequenceNumber is an auto-incrementing sequence number for the message.
+	// Not unique across lanes.
+  uint64 sequence_number = 4;
+  // Nonce is the nonce for this lane for this sender, not unique across senders/lanes
+  uint64 nonce = 5;
+  // MsgHash is the hash of all the message fields.
+  repeated bytes message_hash = 6;
```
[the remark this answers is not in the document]
makramkd: I believe each plugin node calculates the message hash of a message offchain and compares it with the hashes produced by other nodes in order to get some consensus on it.

pkg/loop/internal/pb/ccipocr3/models.proto
```diff
@@ -4,22 +4,76 @@ option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/inter
...
+  // OnRamp is the address of the onramp that sent the message.
+  repeated bytes on_ramp = 7;
```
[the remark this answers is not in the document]
makramkd: This is an "optimization" of sorts because in order to hash the message we need the onramp address. This way the message hasher interface stays simple `Hash(ctx context.Context, msg cciptypes.Message)` instead of `Hash(ctx context.Context, msg cciptypes.Message, onRamp []byte)`.

The CCIP reader will create the Message structure offchain after reading from the chain directly, so it can fill this out because it should know the ramp its reading from.

pkg/types/ccipocr3/common_types.go
makramkd: I only expect this to be used in tests, when we do actual chain reading we don't go from string -> bytes but read bytes directly and then deserialize appropriately.

pkg/types/ccipocr3/generic_types.go
```diff
@@ -82,51 +82,85 @@ func (c ChainSelector) String() string {
...
+
+	// OnRamp is the address of the onramp that sent the message.
+	// NOTE: This is populated by the ccip reader. Not emitted explicitly onchain.
+	OnRamp Bytes `json:"onRamp"`
+}
+
+// EVM2AnyRampMessage is the family-agnostic message sent from an EVM onramp
+// to all other chain families.
+type EVM2AnyRampMessage struct {
```
winder, line 131: Some thoughts on type-per-family approach.

1. You're using bytes for everything, thats already universal so why bother with custom types? I'd recommend moving all of these into the header and either (a) putting chain specific fields in `Data` or having an `Extensions` field for chain specific fields to be added in later, `ExtraArgs` could also serve this purpose.

2. This approach would lead to a lot of nearly duplicate code because you'd have to select the type and then perform the same business logic. Are you thinking that we'll have some parallel type that swaps in some sort of multi-addr type?

3. I'm not sure this even makes sense. The Sender and Receiver addresses are exactly (one of) the problems we need to solve for  different families. They're both included in the same data structure so why bother splitting out a type like this?

4. IMO this data structure should contain at most the fields specific to the destination, in other words `Any2EVM` rather than `EVM2Any`.

winder (changes requested):
Some questions about the approach, I don't really understand why this is helpful but it definitely looks like it will add maintenance overhead

pkg/types/ccipocr3/generic_types.go
makramkd: > Is the destination's commit-root computed based on these hashes?

Yes, the dest commit root will be computed by first "translating" the Any2Any message into Any2DestFamily and then hashing it after encoding it appropriately in the dest family specific encoding scheme + hash function.

> If so, the on-chain code should recompute them and use the commit-root as a CRC on the parameters. This would ensure the message fields haven't been tampered with since the commit report was created.

The onchain code will receive a Any2DestFamily message, re-hash it, and use the provided proof to verify that it was part of the merkle root committed to in the commit phase. If message data was tampered with the proof would not verify onchain.

winder: Thanks, that was my understanding. In that case this field isn't needed onchain except for caching or potentially failing the transaction earlier than when validating against the recomputed merkle root.

pkg/types/ccipocr3/generic_types.go
makramkd: Updated based on your feedback and offline discussion.

winder: Changes look good. The only nitpick I have left is that maybe some of these other fields could go to the header. i.e. Sender/Receiver and maybe fee info.

winder (commented):
Looks good, just some remaining nits for you to consider.

pkg/types/ccipocr3/generic_types.go
makramkd: I was keeping Header this way to try and remain consistent with the onchain header, but not sure if there's much value in that (we're already divergent given the fact that we have the message hash and the onramp address in the header, which is not in the onchain header)

elatoskinas: Fee info should probably not go to the header, since it's an OnRamp detail (not needed for OffRamp)

For the rest I think it might be easier to stay consistent w/ OnChain? We do not include sender / receiver in the header because we use specific types OnChain (i.e. one of them is not always bytes - `address sender` for `EVM2Any` and `address receiver` for `Any2EVM`)

makramkd: Yeah I guess the `Bytes` type for both sender and receiver here are pretty opaque compared w/ onchain. I think its fine to keep it this way if only for consistency and maintaining the same mental model.

winder: I pretty much agree. There was a comment `// This is always set on all CCIP messages.` which had me thinking that these are the fields which are always set.

pkg/loop/internal/pb/ccipocr3/models.proto
makramkd: Metadata no longer needed

elatoskinas (approved):

winder (approved):

asoliman92 (approved):

asoliman92 (approved):

dimkouv (approved):

PR #644, opened 2024-07-08
PoC/WIP: Workflow Builder

DeividasK:
Do not review. There are many changes that need to be removed as I made them when experimenting.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/smartcontractkit/chainlink-common
 
-go 1.21
+go 1.22
+
+toolchain go1.22.5
```
DeividasK, line 5: Undo

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,134 @@
...
+	fmt.Print("consensusOutput", consensusOutput)
+
+	return workflow.Spec(), nil
+}
+
+func TestBuilder_ValidSpec(t *testing.T) {
+	testWorkflowSpec, err := NewWorkflowSpec()
+	require.NoError(t, err)
+
+	expectedSpec := workflows.WorkflowSpec{
```
DeividasK, line 57: Expected output - this can be converted to YAML or used by workflow engine. Incomplete.

pkg/workflows/dependency_graph_test.go
```diff
@@ -363,6 +364,15 @@ targets:
...
 
+				if tc.name == "passthrough ref interpolation" {
+
+					fmt.Println("wf.Spec", wf.Spec)
+					fmt.Println("wf.Spec", wf.Spec.Consensus[0].Inputs)
+
+					// 	fmt.Println("got", got)
+					// 	fmt.Println("adjacencies", adjacencies)
+				}
+
```
DeividasK, line 375: To remove

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,134 @@
...
+		},
+		Actions: []workflows.StepDefinition{
+			{
+				ID:  "offchain_reporting@1.0.0",
+				Ref: "consensus-0",
+				Inputs: workflows.StepInputs{
+					OutputRef: "trigger-0",
+					Mapping: map[string]any{
+						"observations": "$(trigger-0.outputs)",
+					},
```
DeividasK, line 83: Is this a correct expected outcome? Couldn't quite figure this out from tests.

pkg/workflows/builder.go
```diff
@@ -0,0 +1,73 @@
...
+
+// 1. Capability defines JSON schema for inputs and outputs of a capability.
+// Trigger: triggerOutputType := workflowBuilder.addTrigger(DataStreamsTrigger.Config{})
+// Adds metadata to the builder. Returns output type.
+// 2. Consensus: consensusOutputType := workflowBuilder.addConsensus(ConsensusConfig{
+// 	Inputs: triggerOutputType,
+// })
+
+type Workflow struct {
+	spec *workflowSpecYaml
```
nolag, line 16: I would build the underlying WorkflowSpec here, and make that into a proto.  You can move the validation from workflowSpecYaml to WorkflowSpec as well.  This would make it more flexible.

pkg/workflows/builder.go
```diff
@@ -0,0 +1,73 @@
...
+	consensus.Definition.Ref = fmt.Sprintf("consensus-%s", strconv.Itoa((len(w.spec.Consensus))))
+
+	return CapabilityDefinition[O]{
+		Output: consensus.Output,
+		Ref:    consensus.Definition.Ref,
+	}
+}
+
+func (w Workflow) Spec() WorkflowSpec {
+	return w.spec.toWorkflowSpec()
```
nolag, line 72: I would validate here too, it'll make things easier to debug quicker.

nolag (commented):
I don't want to mess up any thoughts you're having, so I may fork and play around a bit.

pkg/workflows/builder.go
```diff
@@ -0,0 +1,73 @@
...
+		spec: &workflowSpecYaml{
+			Owner: params.Owner,
+			Name:  params.Name,
+		},
+	}
+}
+
+func AddTrigger[O any](w *Workflow, trigger Trigger[O]) CapabilityDefinition[O] {
+	// Add ref to trigger.Definition
+	trigger.Definition.Ref = fmt.Sprintf("trigger-%s", strconv.Itoa((len(w.spec.Triggers))))
```
nolag, line 52: The generation of the ref is confusing.  It disallows users from having names that make sense to them.  If I ever add/remove a step I can't easily consolidate logs with prior runs.

pkg/workflows/builder.go
```diff
@@ -0,0 +1,73 @@
...
+	w.spec.Triggers = append(w.spec.Triggers, trigger.Definition)
+
+	return CapabilityDefinition[O]{
+		Output: trigger.Output,
+		Ref:    trigger.Definition.Ref,
+	}
+}
+
+func AddConsensus[O any](w *Workflow, consensus Consensus[O]) CapabilityDefinition[O] {
```
nolag, line 61: Why did you kill the typesafety?  You should be adding input from a CapabilityDefinition.

pkg/workflows/builder.go
```diff
@@ -0,0 +1,73 @@
...
+	// Add ref to trigger.Definition
```
nolag, line 62: That's also a good reason to pass the capability definition as input instead of the workflow only.

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,134 @@
...
+				},
+				CapabilityType: capabilities.CapabilityTypeTrigger,
+			},
+		},
+		Actions: []workflows.StepDefinition{
+			{
+				ID:  "offchain_reporting@1.0.0",
+				Ref: "consensus-0",
+				Inputs: workflows.StepInputs{
+					OutputRef: "trigger-0",
```
DeividasK, line 80:
```suggestion
					OutputRef: nil,
```

pkg/workflows/builder.go
```diff
@@ -0,0 +1,74 @@
...
+// 	Inputs: triggerOutputType,
+// })
+
+type Workflow struct {
+	spec *WorkflowSpec
+}
+
+type Trigger[O any] struct {
+	Definition StepDefinition
+	Output     O
```
nolag, line 21: Why do you need an instance of O for any of the structs?

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -32,6 +34,74 @@ var info = capabilities.MustNewCapabilityInfo(
...
+// 		FeedIDs: []string{
+// 			"0x0003fbba4fce42f65d6032b18aee53efdf526cc734ad296cb57565979d883bdd",
+// 			"0x0003c317fec7fad514c67aacc6366bf2f007ce37100e3cddcacd0ccaa1f3746d",
+// 			"0x0003da6ab44ea9296674d80fe2b041738189103d6b4ea9a4d34e2f891fa93d12",
+// 		},
+// 		MaxFrequencyMs: 100,
+// 	},
+// },
+
+func NewOCR3Consensus(params NewOCR3ConsensusParams) workflows.Consensus[Output] {
```
nolag, line 73: I feel like we'll need to separate built-in consensus algorithms (if we keep them) from each other and from custom.  I'm not sure exactly how to do it, but we would probably use different schemas for them.  Also, it's odd because it ends up with a report that's bytes, so it's not like you can use it as output.  Writers will need to be flexible on their input...

Also, we'll eventually need to think about how to tell it what function to use for aggregation when it's custom, but we can leave that out for now....

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -32,6 +34,74 @@ var info = capabilities.MustNewCapabilityInfo(
...
+	Encoder           string `json:"encoder" jsonschema:"required"`
+	EncoderConfig     any    `json:"encoderConfig" jsonschema:"required"`
+	ReportID          string `json:"reportId" jsonschema:"required"`
+}
+
+type CapabilityInputs struct {
+	// This Input is coupled to the Datastreams Trigger. This was an
+	// intentional short-term trade-off that we'll fix in the future.
+	Observations workflows.CapabilityDefinition[[]datastreams.FeedReport]
+}
```
nolag, line 51: I like this a lot because there's no need for merge steps anymore, it's free and more native feeling!  Good work!

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -32,6 +34,74 @@ var info = capabilities.MustNewCapabilityInfo(
...
+	// 		FeedID: feedID,
+	// 	})
+	// }
+
+	return workflows.Consensus[Output]{
+		Definition: workflows.StepDefinition{
+			ID: ocrCapabilityID,
+			Inputs: workflows.StepInputs{
+				Mapping: map[string]any{
+					"observations": fmt.Sprintf("$(%s.outputs)", params.Inputs.Observations.Ref),
```
nolag, line 88: As an FYI, in either my attempt at a builder or with this attempt, we can allow trigger normalization by using the %s.metadata (assuming that contains the trigger's id somehow).  The step can simply return the input trigger if it's not the matching one.

Ideally, we just allow trigger references as either "trigger" or "trigger.reference" to say "any trigger" vs "a specific trigger"

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -32,6 +34,74 @@ var info = capabilities.MustNewCapabilityInfo(
...
+
+func NewOCR3Consensus(params NewOCR3ConsensusParams) workflows.Consensus[Output] {
+	// // TODO: Call .ValidateConfig to check for more complex JSON Schema validation
+	// outputs := []datastreams.FeedReport{}
+
+	// for _, feedID := range params.Config.FeedIDs {
+	// 	outputs = append(outputs, datastreams.FeedReport{
+	// 		FeedID: feedID,
+	// 	})
+	// }
```
nolag, line 81: This is a hack because we don't have parallel composition, right?  As far as I can tell, there's no way to know to generate it.

pkg/workflows/builder.go
```diff
@@ -0,0 +1,66 @@
...
+
+func AddTrigger[O any](w *Workflow, trigger CapabilityDefinition[O]) CapabilityDefinition[O] {
+	return trigger
+}
+
+func AddConsensus[O any](w *Workflow, consensus CapabilityDefinition[O]) CapabilityDefinition[O] {
+	w.spec.Consensus = append(w.spec.Consensus, consensus.Definition())
+	return consensus
+}
+
```
nolag, line 63: reminder to remove these in favour of only using AddStep :)

PR #646, opened 2024-07-09
Support v0.3 schema report format in LLO (common)

samsondav:


pkg/types/llo/types_test.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		b := ChannelDefinition{
+			ReportFormat: ReportFormatEVMPremiumLegacy,
+			StreamIDs:    []StreamID{0, 1},
+			Aggregators:  []Aggregator{AggregatorMedian, AggregatorMode},
+			Opts:         nil,
+		}
+		assert.False(t, a.Equals(b))
+	})
+	t.Run("different StreamIDs", func(t *testing.T) {
```
msuchacz-cll, line 73: is this array always sorted? I noticed your equals function compares them by checking a corresponding index

msuchacz-cll (approved):
if StreamIDs array is always sorted then the rest LGTM

pkg/types/llo/types_test.go
samsondav: Yes ordering is considered to be a critical part of the channel definition. Different order is a different channel

pkg/types/llo/types.go
```diff
@@ -95,14 +191,82 @@ type Transmitter ocr3types.ContractTransmitter[ReportInfo]
...
-	// We assume that StreamIDs is always non-empty and that the 0-th stream
-	// contains the verification price in LINK and the 1-st stream contains the
-	// verification price in the native coin.
-	StreamIDs []StreamID
+	ReportFormat ReportFormat `json:"reportFormat"`
+	// StreamIDs contains a list of streams to be observed for this channel.
+	StreamIDs []StreamID `json:"streamIDs"`
+	// Aggregators contains a list of aggregator methods used for each stream
+	// e.g. "median", "mode" or other more exotic methods.
+	// StreamIDs and Aggregators must have the same length.
```
akuzni2, line 199: is there any limitation to why StreamIDs and Aggregators are split up? If they are a 1:1 pair wouldn't it just be easier + less error prone to define them together?
```
Things []Thing json:"things"`

type Thing struct {
   s StringId
   a Aggregator 
}

```

samsondav: Original reason is that it maps more closely to how you'd represent that data on-chain, and it keeps all data types that may need to be interpreted on-chain simple.

samsondav: It's a fair point though, it might be safer to use a tuple/composite type here.

akuzni2 (approved):

austinborn (approved):

akuzni2 (approved):

austinborn (dismissed):

msuchacz-cll (approved):

Atrax1 (approved):

PR #647, opened 2024-07-10
Ryan/Deividas - capability types generator

DeividasK:


pkg/capabilities/cli/cmd/testdata/streams.capability.json
```diff
@@ -0,0 +1,82 @@
+{
```
DeividasK, line 1: Will move it to an actual streams folder in a follow-up. This should be the real deal (today; duplicating for testing purposes as it might change later).

pkg/capabilities/cli/cmd/testdata/streams.generated.go
```diff
@@ -0,0 +1,128 @@
+// Code generated by pkg/capabilities/cli, DO NOT EDIT.
```
DeividasK, line 1: This is what will be generated for each capability. Will need work to include capability type, maybe figure out the encoded string => []byte conversion, too.

pkg/capabilities/cli/cmd/testdata/streams.generated.go
```diff
@@ -0,0 +1,128 @@
...
+	// to validate the signatures.
+	ReportContext string
+
+	// Signature over full report and report context represented as bytes encoded as
+	// base64 string.
+	Signatures []string
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
+func (j *StreamsCapabilityOutputsElem) UnmarshalJSON(b []byte) error {
```
DeividasK, line 74: I wonder if we could swap this to Proto/Values generation.

go.mod
```diff
@@ -35,21 +39,30 @@ require (
...
+require (
+	github.com/goccy/go-yaml v1.11.3 // indirect
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
+	github.com/sanity-io/litter v1.5.5 // indirect
+	github.com/spf13/pflag v1.0.5 // indirect
+	golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028 // indirect
+)
+
 require (
```
jmank88, line 58: We don't know why the tooling sometimes creates these extra `indirect` blocks, but if you combine them, then `tidy` will sort them for you.
```suggestion
```

DeividasK:
PR is failing because of failures in downstream dependencies, namely https://github.com/smartcontractkit/chainlink/pull/13820

PR #648, opened 2024-07-10
adding reusable benchmarking action

patrickhuie19:
@jmank88  added a benchmark for loop performance here: https://github.com/smartcontractkit/chainlink-common/pull/628

This PR adds the reusable benchmarking action we created in core. Once this PR is merged, the benchmarking action will comment on merge commits to `main` with benchmarking information, comparing the performance observed on the merge to historical values.


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Run Go Benchmark Action
+        uses: patrickhuie19/benchmark-action@b46d0ce6c3b4bd423380f7d3c6abed389e38ea4a
```
Atrax1, line 19: Why not putting it in here : https://github.com/smartcontractkit/.github/tree/main/actions ?

patrickhuie19: Easier this way until we hit v1

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          benchmarks-pr: 'BenchmarkKeystore_Sign'  # Use benchmarks specified in PR, or default benchmarks
+          benchmarks-merge: 'BenchmarkKeystore_Sign' # Default list of benchmarks to run on merges
```
jmank88, line 22: Can we default to all benchmarks to avoid a maintenance burden?

patrickhuie19: If someone writes a benchmark will they be comfortable with the performance data being stored and reporting on all subsequent commits that include that benchmark? 

It's not a one way door, the backing action can be updated whenever we like.

Atrax1 (approved):

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,27 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
```
jmank88, line 13: Do we need to be deliberate about hardware and OS for consistent results?

patrickhuie19: Yea, great question. Reproducible env's was one of my follow-up tasks: https://github.com/patrickhuie19/benchmark-action?tab=readme-ov-file#open-questions

We can pin the OS with this top level yaml field:
```
runs-on: ubuntu-22.04
```

This gives us:
4 CPU + 16gb RAM

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

jmank88 (approved):

Atrax1 (approved):

PR #649, opened 2024-07-11
feature/ccip-1274: added necessary commit factory server wiring

patrickhuie19:
Sets up a PR that will do something similar to [this exec work](https://github.com/smartcontractkit/chainlink-common/pull/641), but for commit.


pkg/loop/internal/reportingplugin/ccip/commit.go
```diff
@@ -0,0 +1,139 @@
...
+	cc := c.NewClientConn("CommitFactory", newCommitClientFn)
+	return ocr2.NewReportingPluginFactoryClient(c.BrokerExt, cc), nil
+}
+
+/*
+func RegisterCommitProviderServices(s *grpc.Server, provider types.CCIPCommitProvider, brokerExt *net.BrokerExt) {
+	// register the handler for the custom methods of the provider eg NewOffRampReader
+	ccippb.RegisterCommitCustomHandlersServer(s, NewCommitProviderServer(provider, brokerExt))
+}
+*/
```
jmank88, line 89: Remove?
```suggestion
```

pkg/loop/ccip_commit.go
```diff
@@ -0,0 +1,92 @@
...
+	newService := func(ctx context.Context, instance any) (types.ReportingPluginFactory, error) {
+		plug, ok := instance.(types.CCIPCommitFactoryGenerator)
+		if !ok {
+			return nil, fmt.Errorf("expected CCIPCommitFactoryGenerator but got %T", instance)
+		}
+		return plug.NewCommitFactory(ctx, provider)
+	}
+	stopCh := make(chan struct{})
+	lggr = logger.Named(lggr, "CCIPCommitService")
+	var efs CommitFactoryService
```
krehermann, line 81: /nit var name copy pasta from exec?

pkg/loop/ccip_commit.go
```diff
@@ -0,0 +1,92 @@
...
+	broker := BrokerConfig{StopCh: stopCh, Logger: lggr, GRPCOpts: grpcOpts}
+	efs.Init(CCIPCommitLOOPName, &CommitLoop{BrokerConfig: broker}, newService, lggr, cmd, stopCh)
```
krehermann, line 83: this odd that you can Init by value rather than a pointer. i would have guessed Init is mutating the struct. can you double check want Init does?

cfs := &CommitFactoryService{}
...
cfs.Init(..)
...
return cfs

as an alternative

patrickhuie19: It is mutating the struct, good call. I think it's fine because it's a pointer receiver. I added 

```
	t.Run("service not nil", func(t *testing.T) {
		require.NotPanics(t, func() { commit.Name() })
	})
```

to the unit test suite to demonstrate.

jmank88 (approved):

krehermann (approved):

PR #654, opened 2024-07-12
CCIP-2829 Supporting And/Or cases when there is only one expression passed as an argument

mateusz-sekara:
Main goal is to simplify the client's code. As a user of the DSL I don't want to handle on my end edge cases when joining only a single condition under `or` or `and` clauses

Before
```go
        var seqNumsFilter query.Expression
        var seqNumRanges []query.Expression

	if len(seqNumRanges) == 1 {
		seqNumsFilter = seqNumRanges[0]
	} else {
		seqNumsFilter = query.Or(seqNumRanges...)
	}

```

After
```go
       var seqNumRanges []query.Expression
       seqNumsFilter = query.Or(seqNumRanges...)
```

Example in CCIP repo https://github.com/smartcontractkit/ccip/pull/1135/files#r1675529655

### DIFF
--- a/pkg/types/query/query.go
+++ b/pkg/types/query/query.go
@@ -82,12 +82,18 @@ func TxHash(txHash string) Expression {
 }
 
 func And(expressions ...Expression) Expression {
+	if len(expressions) == 1 {
+		return expressions[0]
+	}
 	return Expression{
 		BoolExpression: BoolExpression{Expressions: expressions, BoolOperator: AND},
 	}
 }
 
 func Or(expressions ...Expression) Expression {
+	if len(expressions) == 1 {
+		return expressions[0]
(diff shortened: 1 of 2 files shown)


mateusz-sekara:
@ilija42 I wanted to write a simple unit test for that but I'm not sure where it fits in this structure, I don't see any existing tests for `query.go`, should I just create a `query_test.go` or put it somewhere else?

ilija42:
> @ilija42 I wanted to write a simple unit test for that but I'm not sure where it fits in this structure, I don't see any existing tests for `query.go`, should I just create a `query_test.go` or put it somewhere else?

You can create query_test.go

ilija42 (approved):

PR #658, opened 2024-07-15
[BCF] -3288 Contract Reader Replay

ilija42:


pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -34,6 +35,13 @@ message QueryKeyRequest {
   LimitAndSort limit_and_sort = 3;
 }
 
+// ReplaySequenceRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ChainReader.ReplaySequence].
+message ReplaySequenceRequest {
+  string contract_name = 1;
+  string key = 2;
+  string blockID = 3;
```
jmank88, line 42:
```suggestion
  string block_id = 3;
```

ilija42: Resolved

jmank88: Oh interesting. I guess it can't detect acronyms. Is there a way to override?

ilija42: looks like its possible by using  protoc-gen-gogoslick

pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -12,6 +12,7 @@ service ChainReader {
   rpc BatchGetLatestValues (BatchGetLatestValuesRequest) returns (BatchGetLatestValuesReply) {}
   rpc QueryKey(QueryKeyRequest) returns (QueryKeyReply) {}
   rpc Bind(BindRequest) returns (google.protobuf.Empty) {}
+  rpc ReplaySequence(ReplaySequenceRequest) returns (google.protobuf.Empty) {}
```
krehermann, line 15: what's the control flow and data flow here? it doesn't return anything. so something else is going to consume the result? i would have expected some sort of handle or stream as the return value so the caller can consume it.

as a corollary - what happens if this called concurrently? it doesn't appear to be safe to do so because there is no explicit handle to the result of the operation.

/nit, /dumbquestion: is there something other than a sequence that can be replayed?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -318,10 +320,14 @@ func (it *fakeChainReaderInterfaceTester) SetBatchLatestValues(t *testing.T, bat
 	fake.SetBatchLatestValues(batchCallEntry)
 }
 
-func (it *fakeChainReaderInterfaceTester) TriggerEvent(t *testing.T, testStruct *TestStruct) {
+func (it *fakeChainReaderInterfaceTester) TriggerEvent(t *testing.T, testStruct *TestStruct) string {
 	fake, ok := it.impl.(*fakeChainReader)
 	assert.True(t, ok)
 	fake.SetTrigger(testStruct)
+	if testStruct.Field == nil {
```
krehermann, line 327: what's happening here? add a comment?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -502,7 +492,7 @@ func (f *fakeChainReader) BatchGetLatestValues(_ context.Context, request types.
 	return result, nil
 }
 
-func (f *fakeChainReader) QueryKey(_ context.Context, _ string, filter query.KeyFilter, limitAndSort query.LimitAndSort, _ any) ([]types.Sequence, error) {
+func (f *fakeChainReader) QueryKey(_ context.Context, contractName string, filter query.KeyFilter, limitAndSort query.LimitAndSort, _ any) ([]types.Sequence, error) {
```
krehermann, line 495: out of scope for now -- this API needs pagination if the resulting sequence slice can be large

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -524,11 +514,57 @@ func (f *fakeChainReader) QueryKey(_ context.Context, _ string, filter query.Key
...
+	if err != nil {
+		return err
+	}
+
+	if blockNumber > len(f.triggers) {
+		return fmt.Errorf("blockID: %d requested is in the future: %d", blockNumber, len(f.triggers))
+	}
+
+	// we only have a single event type used in tests for querying, so we can ignore key and just use contract name.
+	f.contractBoundAt[contractName] = blockNumber
```
krehermann, line 543: why no write lock?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -524,11 +514,57 @@ func (f *fakeChainReader) QueryKey(_ context.Context, _ string, filter query.Key
...
 	return nil, nil
 }
 
+func (f *fakeChainReader) ReplaySequence(_ context.Context, contractName, _ string, blockID string) error {
+	blockNumber, err := strconv.Atoi(blockID)
+	if err != nil {
+		return err
+	}
+
+	if blockNumber > len(f.triggers) {
```
krehermann, line 538: why no read lock?

pkg/types/chain_reader.go
```diff
@@ -53,12 +58,10 @@ type ChainReader interface {
 	// Contract call errors are returned in the Err field of BatchGetLatestValuesResult.
 	BatchGetLatestValues(ctx context.Context, request BatchGetLatestValuesRequest) (BatchGetLatestValuesResult, error)
 
-	// Bind will override current bindings for the same contract, if one has been set and will return an error if the
-	// contract is not known by the ChainReader, or if the Address is invalid
-	Bind(ctx context.Context, bindings []BoundContract) error
-
 	// QueryKey provides fetching chain agnostic events (Sequence) with general querying capability.
 	QueryKey(ctx context.Context, contractName string, filter query.KeyFilter, limitAndSort query.LimitAndSort, sequenceDataType any) ([]Sequence, error)
+	// ReplaySequence makes sure that sequence data that belongs to contractName/key pair is available for querying from blockID.
```
krehermann, line 63: given that this is not actually replaying, a better name would ~ `SetReadBlock` or `SetBlockCursor` or something that more explicit about what is happening

Add an example usage, too. IE is seems like the caller must call QueryKeys after this method for something useful to occur(?).

pkg/loop/internal/pb/chain_reader.proto
ilija42: It returns an error, this is how Bind was handled also. Is there another way to define error return value through proto?
Yes, only sequences can be replayed

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
ilija42: It can be paginated through limit and sort since it returns a sequence cursor and allows limits

pkg/types/chain_reader.go
ilija42: It is "replaying" the underlying chain component so that it allows to `QueryKey` a sequence that happened before `Bind`

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
ilija42: Tests are not parallel, only encoding tests are, but they don't reuse the tester instance

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
ilija42: Tests are not parallel, only encoding tests are, but they don't reuse the tester instance

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
ilija42: Resolved

pkg/types/chain_reader.go
ilija42: Should be better now

krehermann: >It is "replaying" the underlying chain component so that it allows to QueryKey a sequence that happened before Bind

This is leaky. the caller isn't supposed to think or know about the underlying chain.

the name should mean something WRT to this api, not the chain. Setting the starting point for Reads is the scope for the caller.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
krehermann: why do all the other methods lock then? it seems inconsistent. should they be removed?

are you saying that the tests must, by some inherent characteristic, be serial or that the current test happen to be written that way?

pkg/types/chain_reader.go
ilija42: Maybe something like `EnableQueryingFrom()` ?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
ilija42: Initially this was testable in parallel, but the tests got a lot more complicated and share a lot of state. Also I am not sure if this can even work in parallel since then we are enforcing all of relayer implementations to also adhere to that. Anwyay I created a ticket to revisit this to see if it makes sense to work in parallel and if it does to make fix this by fixing the shared state and potential race conditions.[ BCF-3327](https://smartcontract-it.atlassian.net/browse/BCF-3327)

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -524,11 +515,57 @@ func (f *fakeChainReader) QueryKey(_ context.Context, _ string, filter query.Key
...
+		boundAt, ok := f.contractBoundAt[contractName]
+		if !ok {
+			return nil, fmt.Errorf("contract %s not bound", contractName)
+		}
+
+		if boundAt == 0 {
+			return sequences, nil
+		}
+
+		return sequences[:boundAt], nil
```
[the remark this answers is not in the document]
ilija42: boundAt shows from which block this data should be available. Here we are simulating this by saying that each sequence belongs to one block, so if we trim this sorted slice we are simulating "give me data from block xy to latest".

PR #663, opened 2024-07-17
remove grpc-proxy replace; use directly instead

jmank88:
If possible, we should pair up with other replace updates.

Requires:
- https://github.com/smartcontractkit/grpc-proxy/pull/2


jmank88:
Done in #745

PR #664, opened 2024-07-17
ChainComponents Generalized Interface Tests

silaslenihan:
Description
Updated ChainReader interface tests to be comprehensive Chain components tests, which test ChainWriter and ChainReader together. 

Requires Dependencies

Resolves Dependencies
- https://github.com/smartcontractkit/chainlink-solana/pull/827
- https://github.com/smartcontractkit/chainlink/pull/13735


pkg/loop/internal/relayer/pluginprovider/chainreader/chain_components_test.go
```diff
@@ -347,6 +332,52 @@ type fakeChainReader struct {
 	lock        sync.Mutex
 }
 
+type fakeChainWriter struct {
+	types.ChainWriter
+	cr *fakeChainReader
+}
+
+func (f *fakeChainWriter) SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, meta *types.TxMeta, value *big.Int) error {
```
Farber98, line 340: really like this! Makes testing much easier

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_components_test.go
```diff
@@ -65,20 +65,21 @@ func TestVersionedBytesFunctions(t *testing.T) {
 	})
 }
 
-func TestChainReaderInterfaceTests(t *testing.T) {
+func TestChainComponentsInterfaceTests(t *testing.T) {
```
ilija42, line 68: I assume that CW will still need its own set of tests, in that case it would make more sense to keep chainReader everywhere instead of chain components

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_components_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb"
 	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer/pluginprovider/chainreader"
-	chainreadertest "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer/pluginprovider/chainreader/test"
+	chaincomponentstest "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayer/pluginprovider/chainreader/test"
```
ilija42, line 22: Why rename this import when its just chain reader stuff?

ilija42 (commented):
Looks good, I just have some concerns around package naming. A lot of the things that were renamed to chaincomponents are only really related to chain reader. I asuume that in the future Chain Writer will have some tests of its own, so I don't really see a reason to rename these tests to chaincomponents

silaslenihan:
> Looks good, I just have some concerns around package naming. A lot of the things that were renamed to chaincomponents are only really related to chain reader. I asuume that in the future Chain Writer will have some tests of its own, so I don't really see a reason to rename these tests to chaincomponents

I see your point on the second one, that package only uses `ChainWriter` superficially so can be changed back. I do think overall it makes sense to change the name of the main test suite to be `ChainComponenents` rather than `ChainReader` though. It was my understanding that we agreed these tests would be end-to-end Chain components tests and at the moment they test most of the functionality of both components. It was my plan to add the `ChainWriter` specific tests to this suite so there could be a one-stop compatibility / interface test for both `ChainWriter` and `ChainReader`.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_components_test.go
silaslenihan: Reverted this part of the refactor

gotest.log
```diff
@@ -0,0 +1,312 @@
...
+  [0;32m✅ TestBatchGetLatestValues/CBOR/BatchGetLatestValues_unwraps_errors_from_server_field_not_found[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/CBOR/BatchGetLatestValues_unwraps_errors_from_server_invalid_encoding[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/CBOR/BatchGetLatestValues_unwraps_errors_from_server_invalid_type[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/CBOR/BatchGetLatestValues_unwraps_errors_from_server_not_found[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/CBOR/BatchGetLatestValues_unwraps_errors_from_server_slice_is_wrong_length[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/CBOR/nil_reader_should_return_unimplemented[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/JSONv1[0;37m (10ms)[0m
+  [0;32m✅ TestBatchGetLatestValues/JSONv1/BatchGetLatestValues_returns_error_if_type_cannot_be_encoded_in_the_wire_format[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/JSONv1/BatchGetLatestValues_unwraps_errors_from_server_element_is_not_a_slice[0;37m (0s)[0m
+  [0;32m✅ TestBatchGetLatestValues/JSONv1/BatchGetLatestValues_unwraps_errors_from_server_field_not_found[0;37m (0s)[0m
```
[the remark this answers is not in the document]
silaslenihan: Yes, removed. Core has a `.gitignore` for this

pkg/loop/internal/relayer/pluginprovider/chainreader/test/chain_reader_loop_tester.go
```diff
@@ -33,11 +33,11 @@ func TestAllEncodings(t *testing.T, test func(chainreader.EncodingVersion) func(
 
 type LoopTesterOpt func(*contractReaderLoopTester)
 
-// WrapChainReaderTesterForLoop allows you to test a [types.ContractReader] implementation behind a LOOP server
-func WrapChainReaderTesterForLoop(wrapped ChainReaderInterfaceTester[*testing.T], opts ...LoopTesterOpt) ChainReaderInterfaceTester[*testing.T] {
+// WrapChainComponentsTesterForLoop allows you to test a [types.ContractReader] and [types.ChainWriter] implementation behind a LOOP server
+func WrapChainComponentsTesterForLoop(wrapped ChainComponentsInterfaceTester[*testing.T], opts ...LoopTesterOpt) ChainComponentsInterfaceTester[*testing.T] {
```
ilija42, line 37: This should be called `WrapContractReaderTesterForLoop` since it only wraps contract reader server

pkg/loop/testutils/utils.go
```diff
@@ -11,9 +11,9 @@ import (
 // the duplication of the function is required so that the test of the LOOP servers themselves
 // can dog food the same testers without creating a circular dependency.
 
-// WrapChainReaderTesterForLoop allows you to test a [types.ContractReader] implementation behind a LOOP server
-func WrapChainReaderTesterForLoop(wrapped interfacetests.ChainReaderInterfaceTester[*testing.T]) interfacetests.ChainReaderInterfaceTester[*testing.T] {
-	return test.WrapChainReaderTesterForLoop(wrapped)
+// WrapChainComponentsTesterForLoop allows you to test a [types.ContractReader] and [types.ChainWriter] implementation behind a LOOP server
+func WrapChainComponentsTesterForLoop(wrapped interfacetests.ChainComponentsInterfaceTester[*testing.T]) interfacetests.ChainComponentsInterfaceTester[*testing.T] {
```
ilija42, line 15: same as above, should be WrapContractReaderTesterForLoop

pkg/types/interfacetests/chain_components_interface_tests.go
```diff
@@ -52,22 +44,26 @@ var AnySliceToReadWithoutAnArgument = []uint64{3, 4}
...
 const AnyExtraValue = 3
 
-func RunChainReaderInterfaceTests[T TestingT[T]](t T, tester ChainReaderInterfaceTester[T]) {
-	t.Run("GetLatestValue for "+tester.Name(), func(t T) { runChainReaderGetLatestValueInterfaceTests(t, tester) })
-	t.Run("BatchGetLatestValues for "+tester.Name(), func(t T) { runChainReaderBatchGetLatestValuesInterfaceTests(t, tester) })
+func RunChainComponentsInterfaceTests[T TestingT[T]](t T, tester ChainComponentsInterfaceTester[T], mockRun bool) {
+	t.Run("GetLatestValue for "+tester.Name(), func(t T) { runChainComponentsGetLatestValueInterfaceTests(t, tester, mockRun) })
+	t.Run("BatchGetLatestValues for "+tester.Name(), func(t T) { runChainComponentsBatchGetLatestValuesInterfaceTests(t, tester, mockRun) })
 	t.Run("QueryKey for "+tester.Name(), func(t T) { runQueryKeyInterfaceTests(t, tester) })
 }
```
ilija42, line 51: These test cases are all for ContractReader, so the naming should be:
runChainComponentsGetLatestValueInterfaceTests -> `runContractReaderGetLatestValueInterfaceTests`.

I would keep also keep the RunContractReaderInterfaceTests and add same for ChainWriter later, its better to keep this granular, especially because in the relayers we will test these two components in separate places.

ilija42 (changes requested):
Mostly naming stuff again, I'm okay with having a united testing interface, but some things like test cases should have a more granular naming

pkg/types/interfacetests/chain_components_interface_tests.go
silaslenihan: @ilija42 I disagree in this case actually. Most of these tests are inherently testing the ChainWriter by design, allowing these to be end-to-end tests. Perhaps we should extract the `ChainReader` specific tests (i.e. ones that don't use `ChainWriter`) like [this one](https://github.com/smartcontractkit/chainlink-common/blob/b10e2708ab783ff269fac0d98a0dfc1624649c8c/pkg/types/interfacetests/chain_components_interface_tests.go#L82-L94) into their own test function. Similarly, once `ChainWriter` specific tests have been implemented, they can have their own test function. Then we have `ChainComponents` tests, which are these end-to-end tests I've created in this PR.

ilija42: These are all very distinct Chain Reading test cases , there is no functionality called `ChainComponentsGetLatestValue` its `ChainReaderGetLatestValue`. The purpose of these tests is for a developer that is doing a chain integration to be able to track the progress of implementing Chain Reader features. The testing of CW is a byproduct and we should never measure if CW works by these tests since CW should have its own tests that cover all of these use cases. This means that if any of these tests break it has to be obvious that its because of ChainReader not ChainWriter.

silaslenihan: > These are all very distinct Chain Reading test cases , there is no functionality called ChainComponentsGetLatestValue its ChainReaderGetLatestValue.

This is a very valid point. The main purpose of these tests are providing coverage to all `ChainReader` cases while also using an end-to-end flow via the `ChainWriter`. The only issue is that they are also testing `ChainWriter` inherently, making it inaccurate to say they are isolating the `ChainReader`'s functionality.

silaslenihan: I reverted the test names to be `ContractReader` specific, with the idea that we will add additional `ChainWriter` specific tests later.

patrick-dowell: Talking to Silas about this earlier, I agree that it is possible that ChainWriter changes could break these changes, but I thought we were all okay with that to allow these tests to fully run E2E (so the Chain/Contract Reader actually reads data written by the Writer on chain). I think we should keep the tests functionally as they are in this PR, but revert the test names back to Contract/Chain Reader specific. In future PRs, we can add additional ChainWriter specific tests as we come up with ideas for what parts of the interface we want to specifically test.

patrick-dowell (commented):
A few small comments, but overall this looks pretty reasonable.

pkg/types/interfacetests/utils.go
```diff
@@ -36,12 +41,88 @@ func runTests[T TestingT[T]](t T, tester BasicTester[T], tests []testcase[T]) {
 	}
 }
 
+func batchChainWrite[T TestingT[T]](t T, tester ChainComponentsInterfaceTester[T], batchCallEntry BatchCallEntry, mockRun bool) {
+	if mockRun {
+		cw := tester.GetChainWriter(t)
+		err := cw.SubmitTransaction(tests.Context(t), AnyContractName, "batchChainWrite", batchCallEntry, "", "", nil, big.NewInt(0))
+		require.NoError(t, err)
+		return
+	}
```
nolag, line 50: Why is this needed?  The mock is a good test for the relay so it's nice when realish data goes over it.  It makes it more likely to catch serialization errors earlier on.

pkg/types/interfacetests/utils.go
```diff
@@ -36,12 +41,88 @@ func runTests[T TestingT[T]](t T, tester BasicTester[T], tests []testcase[T]) {
...
+}
+
+func WaitForTransactionStatus[T TestingT[T]](t T, tester ChainComponentsInterfaceTester[T], txID string, status types.TransactionStatus, mockRun bool) error {
+	ctx, cancel := context.WithTimeout(tests.Context(t), 15*time.Minute)
+	defer cancel()
+
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
+	fmt.Printf("Waiting for transaction %s to reach status %d\n", txID, status)
```
nolag, line 89: please remove

pkg/types/interfacetests/utils.go
```diff
@@ -36,12 +41,88 @@ func runTests[T TestingT[T]](t T, tester BasicTester[T], tests []testcase[T]) {
...
+			}
+			current, err := tester.GetChainWriter(t).GetTransactionStatus(ctx, txID)
+			if err != nil {
+				return fmt.Errorf("failed to get transaction status: %w", err)
+			}
+
+			if current == types.Failed || current == types.Fatal {
+				return fmt.Errorf("transaction %s has failed or is fatal", txID)
+			} else if current >= status {
+				fmt.Printf("Transaction %s reached status: %d\n", txID, current)
```
nolag, line 107: please remove

pkg/types/interfacetests/utils.go
```diff
@@ -36,12 +41,88 @@ func runTests[T TestingT[T]](t T, tester BasicTester[T], tests []testcase[T]) {
...
+	}
+
+	for contractName, contractBatch := range batchCallEntry {
+		require.Contains(t, nameToAddress, contractName)
+		for _, readEntry := range contractBatch {
+			val, isOk := readEntry.ReturnValue.(*TestStruct)
+			if !isOk {
+				require.Fail(t, "expected *TestStruct for contract: %s read: %s, but received %T", contractName, readEntry.Name, readEntry.ReturnValue)
+			}
+			SubmitTransactionToCW(t, tester, "addTestStruct", val, types.BoundContract{Name: contractName, Address: nameToAddress[contractName]}, types.Unconfirmed)
```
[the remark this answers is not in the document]
silaslenihan: Yeah, this function was created specifically for the `BatchGetLatestValues` tests on the `ChainReader`, which only use test structs. I don't see a reason why we would need to use this function to the `ChainWriter` in isolation though, so my opinion is it's fine the way it is right now.

pkg/types/interfacetests/utils.go
silaslenihan: This is needed because in the mocked tests, we have to use the helper function [SetBatchLatestValues](https://github.com/smartcontractkit/chainlink-common/blob/b10e2708ab783ff269fac0d98a0dfc1624649c8c/pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go#L407-L414) which accepts a `batchCallEntry`. When we aren't doing a mockRun, we have to send individual transactions sequentially since `ChainWriter` has no batch send functionality atm.

pkg/types/interfacetests/utils.go
silaslenihan: I assume you mean just the Printf statement, just removed.

ilija42 (commented):
there is an empty gotest.log file included in the PR

silaslenihan:
> there is an empty gotest.log file included in the PR

removed

pkg/types/interfacetests/utils.go
```diff
@@ -36,12 +41,91 @@ func runTests[T TestingT[T]](t T, tester BasicTester[T], tests []testcase[T]) {
...
+		}
+	}
+}
+
+// SubmitTransactionToCW submits a transaction to the ChainWriter and waits for it to reach the given status.
+func SubmitTransactionToCW[T TestingT[T]](t T, tester ChainComponentsInterfaceTester[T], method string, args any, contract types.BoundContract, status types.TransactionStatus) string {
+	tester.DirtyContracts()
```
ilija42, line 74: Since each test case sets up new contracts now can we remove this?

silaslenihan: I actually added this back in to ensure not every tests created new contracts. i.e. the ones that don't send any transactions don't need new contracts to be deployed after them. This allows for a slight optimization in runtime / token spend on testnet.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -347,6 +332,52 @@ type fakeChainReader struct {
...
 }
 
+type fakeChainWriter struct {
+	types.ChainWriter
+	cr *fakeChainReader
+}
+
+func (f *fakeChainWriter) SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, meta *types.TxMeta, value *big.Int) error {
+	switch method {
+	case "addTestStruct":
```
ilija42, line 342: Can we move these into constants so that they can be reused in chain reader configs for the tests in the relayers. If you think its out of scope then just a TODO with a ticket number

pkg/types/interfacetests/utils.go
```diff
@@ -36,12 +41,91 @@ func runTests[T TestingT[T]](t T, tester BasicTester[T], tests []testcase[T]) {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("transaction %s not finalized within timeout period", txID)
+		case <-ticker.C:
+			if mockRun {
+				tester.GenerateBlocksTillConfidenceLevel(t, "", "", primitives.Finalized)
+				return nil
+			}
```
ilija42, line 102: I don't see the EVM interface tester having an impl. for `GenerateBlocksTillConfidenceLevel()`. Does this have an implementation anywhere? If not we should just remove mockRun completely

silaslenihan: I only have this so that the `fakeChainReader` tests can run [GenerateBlocksTillConfidenceLevel](https://github.com/smartcontractkit/chainlink-common/blob/c7d0f8681a84627d94180fabfec95470d7fca7f5/pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go#L569-L579). You're right, this isn't used by the actual implementations but it was necessary for these tests so I wasn't sure how to remove the method without breaking this test.

ilija42: Yeah, I don't see a trivial solution, lets create a ticket just for this

ilija42 (approved):

pkg/types/interfacetests/utils.go
silaslenihan: [Done](https://smartcontract-it.atlassian.net/jira/software/c/projects/BCI/boards/23?assignee=61a6f090ef18ca00717f84db&quickFilter=398&selectedIssue=BCI-4134)

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -347,6 +332,52 @@ type fakeChainReader struct {
...
+			return fmt.Errorf("unexpected type %T", args)
+		}
+		f.cr.SetUintLatestValue(v.Value, ExpectedGetLatestValueArgs{})
+	case MethodTriggeringEvent:
+		v, ok := args.(TestStruct)
+		if !ok {
+			return fmt.Errorf("unexpected type %T", args)
+		}
+		f.cr.SetTrigger(&v)
+	case "batchChainWrite":
```
patrick-dowell, line 360: shouldn't this be a const as well?

pkg/types/interfacetests/utils.go
patrick-dowell: Is this resolved now? What Silas said seems reasonable to me.

patrick-dowell (approved):
1 nit and 1 unresolved comment but otherwise we're good!

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
silaslenihan: Addressed that in [this comment](https://github.com/smartcontractkit/chainlink-common/pull/664#discussion_r1742110028)

PR #665, opened 2024-07-18
Add pkg/sqlutils/models and pkg/utils/tests/json.go from core repo

reductionista:
This moves some chain agnostic types out of chainlink into here, so they can be used by more than just that repo.  See linked PR

### Resolves Dependencies

- https://github.com/smartcontractkit/chainlink/pull/13881


jmank88:
Let's find a better place than `package morels`: https://go.dev/blog/package-names

What about the existing sqlutil package?

Atrax1 (approved):
👍

PR #666, opened 2024-07-18
feat(observability-lib): add delete cmd

Atrax1:
Add delete command

Can be used as followed : 

```
go run main.go delete \
  --dashboard-name "${DASHBOARD_NAME}" \
  --grafana-url $GRAFANA_URL \
  --grafana-token $GRAFANA_TOKEN
```

### DIFF
--- a/observability-lib/cmd/dashboard.go
+++ b/observability-lib/cmd/dashboard.go
@@ -87,6 +87,32 @@ func (d *Dashboard) Deploy() error {
 	return nil
 }
 
+func (d *Dashboard) Delete() error {
+	ctx := context.Background()
+	client := grabana.NewClient(
+		&http.Client{},
+		d.url,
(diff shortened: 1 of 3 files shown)


observability-lib/cmd/dashboard.go
```diff
@@ -87,6 +87,32 @@ func (d *Dashboard) Deploy() error {
 	return nil
 }
 
+func (d *Dashboard) Delete() error {
+	ctx := context.Background()
+	client := grabana.NewClient(
+		&http.Client{},
```
jmank88, line 93: Is a zeroed `http.Client` different than `http.DefaultClient`?

jmank88: Looks like no

jmank88 (approved):

PR #675, opened 2024-07-29
changes to support remote target deterministic message hash - fixes KS-379

ettec:
Fixes https://smartcontract-it.atlassian.net/browse/KS-379

Changes to support deterministic creation of message hash in remote targets


pkg/capabilities/capabilities.go
```diff
@@ -398,6 +398,14 @@ type RemoteTriggerConfig struct {
 	MessageExpiry           time.Duration
 }
 
+type RemoteTargetConfig struct {
+	RequestHashExcludedAttributes []string
+}
+
+func (c *RemoteTargetConfig) ApplyDefaults() {
```
cedric-cordenier, line 405: @ettec Please remove this if we're not using it :)

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -296,14 +304,25 @@ func (c *capabilitiesRegistryServer) ConfigForCapability(ctx context.Context, re
 
 	ccp := &capabilitiespb.CapabilityConfig{
 		DefaultConfig: ecm,
-		RemoteConfig: &capabilitiespb.CapabilityConfig_RemoteTriggerConfig{
+	}
+
+	if cc.RemoteTriggerConfig != nil {
```
cedric-cordenier, line 309: This feels like we could move it into a switch statement: I'd expect one of these to apply for a given capability, not both

pkg/values/map.go
```diff
@@ -69,6 +70,33 @@ func (m *Map) UnwrapTo(to any) error {
 	return d.Decode(m.Underlying)
 }
 
+// DeleteAtPath deletes a value from a map at a given dot separated path.  Returns true if an element at the given
+// path was found and deleted, false otherwise.
+func (m *Map) DeleteAtPath(path string) bool {
```
krehermann, line 75: this is only used in a test AFAICT. what's going on?

pkg/capabilities/pb/registry.proto
```diff
@@ -14,11 +14,16 @@ message RemoteTriggerConfig {
   google.protobuf.Duration messageExpiry = 4;
 }
 
+message RemoteTargetConfig {
+  repeated string requestHashExcludedAttributes = 1;
```
krehermann, line 18: how is this supposed to be used? document please.

moreover, why is it a parameter? does the user need to remember to set it on all calls?

why is it an exclusion list rather than inclusion list? the former implies that the we have have not determinism by default, which seems wrong

pkg/values/map.go
ettec: its used in core, seemed like the best place for the logic to sit

pkg/capabilities/pb/registry.proto
ettec: its set in the registry, so config, not per call.  Its true that there is not determinisim by default if you hash the request message due to the behaviour of libocr callbacks where the signature set is different per node, this is expected.  Its exclusion rather than inclusion as  any new added attributes should be hashed by default, it was felt that the exception to that will be rare and explicitly declared as exclude to avoid accidentally not including new attributes on the request.

I'll add some doc.

ettec: doc added

cedric-cordenier (approved):

pkg/capabilities/pb/registry.proto
krehermann: thank you

krehermann (dismissed):

bolekk (approved):
Can't say I'm a huge fan of this approach but it looks like it solves the problem.

ettec:
> Can't say I'm a huge fan of this approach but it looks like it solves the problem.

agree, but seemed liked the best solution we could come up with that would keep the remote target from having to know about the internals of the target it was acting as the remote shim for.

cedric-cordenier (approved):

jmank88 (approved):

PR #680, opened 2024-08-01
embeddable implementation of contract reader

EasterTheBunny:
When updates are made to a core interface and committed to common, builds break on core or relay repos due to not matching interfaces. This change proposes using 'unimplemented' versions of an interface to embed in downstream implementations to allow interface additions to be added without breaking downstream builds.

Usage:
```
type ImplementedContractReader struct {
    types.UnimplementedContractReader
}

// ... actual interface implementation below
```

### DIFF
--- a/pkg/types/contract_reader.go
+++ b/pkg/types/contract_reader.go
@@ -60,6 +60,8 @@ type ContractReader interface {
 
 	// QueryKey provides fetching chain agnostic events (Sequence) with general querying capability.
 	QueryKey(ctx context.Context, contract BoundContract, filter query.KeyFilter, limitAndSort query.LimitAndSort, sequenceDataType any) ([]Sequence, error)
+
+	mustEmbedUnimplementedContractReaderServer()
 }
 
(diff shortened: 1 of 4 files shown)


pkg/types/contract_reader.go
```diff
@@ -118,3 +118,47 @@ func (bc BoundContract) ReadIdentifier(readName string) string {
 func (bc BoundContract) String() string {
 	return bc.Address + "-" + bc.Name
 }
+
+type UnimplementedContractReader struct{}
+
+var _ ContractReader = UnimplementedContractReader{}
```
jmank88, line 124: The strict version of this includes an unexported method so that impls _must_ embed `UnimplementedContractReader` to satisfy the interface. Should we add that now?

EasterTheBunny: done

jmank88 (approved):

Farber98 (approved):
this is good!

PR #681, opened 2024-08-02
[DRAFT] Add Beholder sdk v0.1

cll-gg:


pkg/beholder/message.go
```diff
@@ -0,0 +1,259 @@
...
+	// Hash of the workflow spec.
+	WorkflowSpecId string
+	// The unique execution of a workflow.
+	WorkflowExecutionId string
+	// The address for the contract.
+	CapabilityContractAddress string
+	CapabilityId              string
+	CapabilityVersion         string
+	CapabilityName            string
+	NetworkChainId            string
```
jmank88, line 45:
```suggestion
	WorkflowId           string
	WorkflowName         string
	WorkflowOwnerAddress string
	// Hash of the workflow spec.
	WorkflowSpecID string
	// The unique execution of a workflow.
	WorkflowExecutionID string
	// The address for the contract.
	CapabilityContractAddress string
	CapabilityID              string
	CapabilityVersion         string
	CapabilityName            string
	NetworkChainID            string
```

pkg/beholder/logger/logger.go
```diff
@@ -0,0 +1,66 @@
...
+func New() Logger {
+	return newZapLogger().Sugar()
+}
+
+// NewOtelzapLogger records Zap log messages as messages on the existing span that must be passed in a context.Context as a first argument.
+// It does not record anything if the context does not contain a span.
+// See more details [here](https://github.com/uptrace/opentelemetry-go-extra/tree/main/otelzap)
+func NewOtelzapLogger() *otelzap.Logger {
+	zl := newZapLogger()
+	return otelzap.New(newZapLogger(), otelzap.WithMinLevel(zl.Level()))
```
jmank88, line 50: I don't think this approach will work well with our existing loggers. We would have to switch over, rather than just hooking in and forwarding from the existing API. Is there a way to get an otel zap core that we just plug in to our existing logger?

jmank88: https://pkg.go.dev/github.com/agoda-com/opentelemetry-go/otelzap#NewOtelCore

pkg/beholder/client.go
```diff
@@ -0,0 +1,312 @@
...
+
+func newOtelClient(cfg Config, errorHandler errorHandlerFunc, otlploggrpcNew otlploggrpcFactory) (Client, error) {
+	ctx := context.Background()
+	baseResource, err := newOtelResource()
+	if err != nil {
+		return nil, err
+	}
+	sharedLogExporter, err := otlploggrpcNew(
+		ctx,
+		otlploggrpc.WithInsecure(),
```
4of9, line 99: should this option be configurable? 

(also applies to the `Meter` and `Tracer` configs later on in this file)

pkg/beholder/client.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return nil, err
+	}
+	meter := meterProvider.Meter(cfg.PackageName)
+	otel.SetMeterProvider(meterProvider)
+
+	// MessageEmitter
+	messageLogProcessor := sdklog.NewBatchProcessor(
+		sharedLogExporter,
+		sdklog.WithExportTimeout(1*time.Second), // Default is 30s
```
4of9, line 150: we should probably make this configurable as well (also applies to the meter and tracer providers in this file)

pkcll: There is a separate PR where this was addressed, will move it here.

pkg/beholder/client.go
pkcll: Addressed [here](https://github.com/smartcontractkit/chainlink-common/pull/681/commits/0bdc6d2b4de8da9acc5ce6c3950060f26a09d061)

pkg/beholder/logger/logger.go
pkcll: @jmank88 
https://pkg.go.dev/github.com/agoda-com/opentelemetry-go/otelzap
package depends on [github.com/agoda-com/opentelemetry-logs-go](https://github.com/agoda-com/opentelemetry-logs-go) which is a custom implementation of https://github.com/open-telemetry/opentelemetry-go Logs [sdk](https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/log).
Because of that it's incompatible with current Beholder implementation which relies on open-telemetry/opentelemetry-go [sdk](https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/log).  

I'm going to remove beholder/logger based on otelzap from the scope of this PR.

pkcll:
Copied to https://github.com/smartcontractkit/chainlink-common/pull/694

PR #682, opened 2024-08-02
clarify execute sync error messages

ettec:
when using the utility execute as sync function misleading 60s 'timeout' errors seen in the keystone e2e tests when the cause is test shutdown not timeout

### DIFF
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -345,8 +345,8 @@ func ExecuteSync(ctx context.Context, c CallbackExecutable, request CapabilityRe
 }
 
 // ExecuteSyncWithTimeout allows explicitly passing in a timeout to customise the desired duration.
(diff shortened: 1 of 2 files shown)


pkg/capabilities/capabilities.go
```diff
@@ -370,6 +370,8 @@ outerLoop:
 			}
 
 			vs = append(vs, response.Value)
+		case <-parentCtx.Done():
+			return nil, fmt.Errorf("parent context cancelled: %w", parentCtx.Err())
```
ettec, line 374: actually would be better to return 'capabilities.ErrStopExecution' here I think, @cedric-cordenier agree?

ettec:
This is no longer relevent, synchronous execute no longer uses a channel

ettec:
This is no longer relevent, synchronous execute no longer uses a channel

PR #683, opened 2024-08-02
[BCI-3862][chainlink-common] - Change DSL Block primitive to string instead of int

Farber98:
### Task Description:
[Change ChainReader block DSL](https://github.com/smartcontractkit/chainlink-common/blob/4ed45812784df8f781beab8748a5a1b504527990/pkg/loop/internal/pb/chain_reader.proto#L154) uint64 type to string and update EVM implementation to reflect changes

### This PR:
Changed the [block proto field](https://github.com/smartcontractkit/chainlink-common/blob/4ed45812784df8f781beab8748a5a1b504527990/pkg/loop/internal/pb/chain_reader.proto#L154) and fixed all the reference failures within `chainlink-common`

### Ticket:
- [BCI-3862](https://smartcontract-it.atlassian.net/browse/BCI-3862)

### Unblocks:
- https://github.com/smartcontractkit/chainlink/pull/14033

[BCI-3862]: https://smartcontract-it.atlassian.net/browse/BCI-3862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/loop/internal/pb/chain_reader.proto
```diff
@@ -151,7 +151,7 @@ message Comparator  {
 }
 
 message Block{
-  uint64 block_number = 1;
```
Farber98, line 154: Should I instead deprecate the previous field and add a new one because of [backward compatibility](https://protobuf.dev/programming-guides/dos-donts/#change-type)?

```
message Block {
  uint64 block_number_deprecated = 1 [deprecated = true]; // rename prev field and mark as deprecated
  string block_number = 3; // new field with string type that has the old name
  ComparisonOperator operator = 2;
}
```

Farber98:
"update EVM implementation to reflect changes" means updating [this file](https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/chain_reader.go#L1)? I don't see we are directly referencing the changed field

pkg/loop/internal/pb/chain_reader.proto
ilija42: Usually that is the best practice, but I don't see anyone using block querying so as long as you update tests in core evm it is fine to make a breaking change.

ilija42: You'll have to update the PGDSLParser  [here](https://github.com/smartcontractkit/chainlink/blob/05ef7fdbb115f55a85bcbbc5402350818501e1f5/core/chains/evm/logpoller/parser.go#L45), and check if any of the LP tests need to be fixed after this change

Farber98: Fixed in this PR: https://github.com/smartcontractkit/chainlink/pull/14033

jmank88 (dismissed):

ilija42 (approved):

Farber98:
reopened here with signed commits https://github.com/smartcontractkit/chainlink-common/pull/688

PR #684, opened 2024-08-05
[KS-408] Handle nils in Unwrap/UnwrapTo

cedric-cordenier:


pkg/values/big_int.go
```diff
@@ -23,10 +24,15 @@ func (b *BigInt) proto() *pb.Value {
 }
 
 func (b *BigInt) Unwrap() (any, error) {
-	return b.Underlying, nil
+	bi := &big.Int{}
+	return bi, b.UnwrapTo(bi)
 }
 
 func (b *BigInt) UnwrapTo(to any) error {
+	if b == nil {
```
bolekk, line 32: Should you also check if b.Underlying is nil? (same for other types below)

cedric-cordenier: For bigInt yes, good catch :) For the others, I've added tests to show that it isn't necessary

krehermann (approved):

bolekk (approved):

PR #686, opened 2024-08-06
use gomods

jmank88:
This repo has multiple modules, but the `Makefile` commands were only written for one. This PR swaps in [gomods](https://github.com/jmank88/gomods) for tidy and generate commands.


Makefile
```diff
@@ -22,11 +26,9 @@ rm-mocked:
 	grep -rl "^// Code generated by mockery" | grep .go$ | xargs -r rm
 
 .PHONY: generate
-generate: mockery install-protoc
-	# add our installed protoc to the head of the PATH
-	# maybe there is a cleaner way to do this
-	PATH=$$HOME/.local/bin:$$PATH go generate -x ./...; \
```
jmank88, line 28: @krehermann you touched this last - are you the original author? I think it seems reasonable to drop this and expect `GOBIN` to already be on the path, no?

Atrax1 (approved):

vyzaldysanchez (approved):

PR #692, opened 2024-08-12
capabilities: triggers: Expose meta so it can be set for mock-trigger

archseer:


pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -49,6 +49,7 @@ type MercuryTriggerService struct {
 	stopCh             services.StopChan
 	wg                 sync.WaitGroup
 	lggr               logger.Logger
+	Meta               datastreams.SignersMetadata // usually empty, but set to a value in mock trigger
```
bolekk, line 52: To make is slightly cleaner, let's call it metaOverride (private) and set via NewMercuryTriggerService().

bolekk (changes requested):

pkg/capabilities/triggers/mercury_trigger.go
archseer: I added a separate setter method for it so existing code still compiles with this change

bolekk (approved):

cedric-cordenier (approved):

bolekk (approved):

cedric-cordenier (approved):

PR #694, opened 2024-08-12
Add Beholder sdk

pkcll:
Add Beholder sdk.

Related PR's 
- https://github.com/smartcontractkit/chainlink-common/pull/696
- https://github.com/smartcontractkit/chainlink/pull/14110


pkg/beholder/pb/example.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/beholder/pb";
+
+package pb;
+
+// Used for testing 
+message CustomMessage {
```
pavel-raykov, line 8: If it is only used for testing, does it make sense to also call it "TestOnlyMessage"?

pkg/beholder/client.go
```diff
@@ -0,0 +1,332 @@
...
+	"go.opentelemetry.io/otel/sdk/trace"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	oteltrace "go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type Emitter interface {
+	// Sends message with bytes and attributes to OTel Collector
+	Emit(ctx context.Context, body []byte, attrs map[string]any) error
```
krehermann, line 28: these 2 appear to doing the exact same thing. prefer having only one:

`Emit(ctx context.Context, m Message)`
this lets us evolve the Message struct without breaking the api

jmank88: I mostly agree, although I could see a case being made for a variadic arg version for brevity:
```suggestion
	Emit(ctx context.Context, body []byte, attrKVs ...[]any) error
```
This version loses some compile time type safety just like a logger `With` API, but it may simplify callsites significantly:
```go
emitter.Emit(ctx, data, map[string]any{
	"key": "value",
	"field": 10,
})
```
```go
emitter.Emit(ctx, data, "key": "value", "field": 10)
```

pkg/beholder/client.go
```diff
@@ -0,0 +1,332 @@
...
+	otelglobal.SetLoggerProvider(loggerProvider)
+
+	// Tracer
+	tracerProvider, err := newTracerProvider(cfg, baseResource, creds)
+	if err != nil {
+		return nil, err
+	}
+	tracer := tracerProvider.Tracer(cfg.PackageName)
+	otel.SetTracerProvider(tracerProvider)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
```
jmank88, line 143: Auto-setting globals prevents us from using this in core, which is already setting its own globals. Can we provide a more flexible API here? What if this returned without calling any global funcs? Can we give the caller the option of doing the global calls themselves, or otherwise opting in explicitly for them?

jmank88: IIUC our situation requires a single provider with multiple exporters.

pkg/beholder/client.go
pkcll: > emitter.Emit(ctx, data, "key", "value", "field", 10)

I can call it `SugarEmit` :-)

pkg/beholder/client.go
pkcll: Makes sense, will provide a separate way to set globals

jmank88: It looks like we can call `RegisterSpanProcessor` on a global `TracerProvider`, so we don't technically need to constructor the provider with both exporters on hand, but we still need to ultimately have some control of this.

pkg/beholder/client.go
pkcll: That sugary syntax is supported for beholder message Attributes see here https://github.com/smartcontractkit/chainlink-common/pull/694/files#diff-d5bbc8388dc6b5c9b3b4011eb5cd2448839eb67b352eae919a8a68fb3c3fdb12R46-R47

pkg/beholder/client.go
```diff
@@ -0,0 +1,332 @@
...
+) Client {
+	return &otelClient{
+		config:    config,
+		logger:    logger,
+		tracer:    tracer,
+		meter:     meter,
+		emitter:   emitter,
+		closeFunc: onClose,
+	}
+}
```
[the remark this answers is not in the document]
pkcll: Addressed [here](https://github.com/smartcontractkit/chainlink-common/pull/694/commits/656550bf232be5293e161fed201c3059a9e8192f)

pkg/beholder/pb/example.proto
pkcll: @pavel-raykov  addressed [here](https://github.com/smartcontractkit/chainlink-common/pull/694/commits/43b59453daa04efffbbecd830856e921dd50b066)

pkg/beholder/client.go
pkcll: @jmank88, addressed [this](https://github.com/smartcontractkit/chainlink-common/pull/694#discussion_r1714389524) comment [here](https://github.com/smartcontractkit/chainlink-common/pull/694/commits/14b67f1738ef1a7c193fcb1367f3d041e61ec336)

pkcll: @jmank88 What is the use case for RegisterSpanProcessor?
I made providers to be exported (see [here](https://github.com/smartcontractkit/chainlink-common/blob/14b67f1738ef1a7c193fcb1367f3d041e61ec336/pkg/beholder/client.go#L53-L55)) so it's possible to call it like this `OtelClient.TracerProvider.RegisterSpanProcessor`.

pkg/beholder/client.go
pkcll: `Emit` variadic version https://github.com/smartcontractkit/chainlink-common/pull/694/commits/ec50c05c7b341745d81426767ff0a6e0f339a3d8

pkg/beholder/client.go
jmank88: Core and LOOP Plugins both call in to the existing `SetupTracing` helper which creates and registers its own provider: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/loop/telem.go#L124-L134

IIUC We will have to configure just one provider/resource/set-of-attributes, but the provider can have multiple batchers/exporters/processors.

pkg/beholder/message.go
```diff
@@ -0,0 +1,275 @@
...
+	case []byte:
+		return otellog.Bytes(key, v)
+	case nil:
+		return otellog.Empty(key)
+	case otellog.Value:
+		return otellog.KeyValue{Key: key, Value: v}
+	case attribute.Value:
+		return OtelAttr(key, v.AsInterface())
+	default:
+		return otellog.String(key, fmt.Sprintf("<unhandled beholder attribute value type: %T>", v))
```
jmank88, line 206: Should we try to print the value too? e.g. with `%v`?

pkg/beholder/config.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+type Config struct {
+	InsecureConnection       bool
+	CACertFile               string
+	OtelExporterGRPCEndpoint string
+
+	PackageName string
+	// OTel Resource
+	ResourceAttributes map[string]string
```
jmank88, line 16: How about using the eventual otel type here? We already have values of this type to pass in from the existing config.
```suggestion
	ResourceAttributes []otelattr.KeyValue
```

jmank88:
Here is a WIP PR attempting to integrate beholder in to the existing loop setup helpers: #696

pkg/beholder/config.go
jmank88: https://github.com/smartcontractkit/chainlink-common/pull/696/files#diff-8ae7d9439c8e585e8376ebac5e4f399dbf8309df188de9882a8925be6c208159R96

pkg/beholder/client.go
pkcll: Most of the examples I have seen for OTel logger,tracer,meter have one provider configured per combination of single batcher/exporter/processor. 

You can have multiple providers but only one global for logger,tracer,meter. 
OTel resource/set-of-attributes could and should be shared between providers. 
OTel Resource support is merging so you can have a base resource and extend it if needed with custom attributes per provider.

jmank88: > You can have multiple providers but only one global per logger,tracer,meter.

But AFAIK in our case we do not want two providers because we want to send the same data to both places. The goal is for existing trace usages to be picked up and sent through to beholder as well.

jmank88: > @jmank88, I see that it's used for Loop GRPC endpoint instrumentation via [otelgrpc](https://github.com/smartcontractkit/chainlink-common/blob/a4bce717e58335b066c65ab01f0dd3977215c6c2/pkg/loop/telem.go#L152) library. Is that OTel tracing enabled on production for Loops?

It is opt-in per node op via the `Tracing` section of their config.toml file: https://github.com/smartcontractkit/chainlink/blob/develop/docs/CONFIG.md#tracing

pkg/beholder/config.go
pkcll: @jmank88 addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/b8f6c6f310051c056d18f9ba9dd20b98812996ac

pkg/beholder/client.go
```diff
@@ -0,0 +1,289 @@
...
+	// Meter
+	otel.SetMeterProvider(c.MeterProvider)
+}
+
+func (c OtelClient) Close() (err error) {
+	for _, provider := range []any{c.LoggerProvider, c.TracerProvider, c.MeterProvider} {
+		if p, ok := provider.(otelProvider); ok {
+			err = errors.Join(err, p.Shutdown(context.Background()))
+		} else {
+			err = errors.Join(err, errors.New("provider does not implement Shutdown method"))
```
patrickhuie19, line 171: can we do this at build time rather than run time?

pkg/beholder/client.go
```diff
@@ -0,0 +1,289 @@
...
+
+	setOtelErrorHandler(errorHandler)
+
+	client := OtelClient{cfg, logger, tracer, meter, messageEmitter, loggerProvider, tracerProvider, meterProvider}
+
+	return client, nil
+}
+
+func (c OtelClient) SetGlobals() {
+	// Logger
```
patrickhuie19, line 157: these comments are somewhat helpful (the name of the provider is straightforward thankfully), but I think a reader would benefit more from a top level comment explaining what providers `SetGlobals` is expected to set and a link out to relevant docs

pkg/beholder/client_test.go
```diff
@@ -0,0 +1,345 @@
...
+		}, {
+			name:                   "Test EmitMessage",
+			makeCustomAttributes:   defaultCustomAttributes,
+			messageBody:            defaultMessageBody,
+			messageCount:           10,
+			exporterMockErrorCount: 0,
+			exporterOutputExpected: true,
+			messageGenerator: func(client OtelClient, messageBody []byte, customAttributes map[string]any) {
+				message := NewMessage(messageBody, customAttributes)
+				err := client.Emitter.EmitMessage(context.Background(), message)
```
patrickhuie19, line 89: could we use a context that inherit/derives from the test's natural deadline? i.e. `tests.Context(t)`

pkg/beholder/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+	otelattr "go.opentelemetry.io/otel/attribute"
+)
+
+type Config struct {
+	InsecureConnection       bool
+	CACertFile               string
+	OtelExporterGRPCEndpoint string
+
+	PackageName string
```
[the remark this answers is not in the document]
pkcll: @krebernisak 
> How is PackageName expected to be used? Can we document it here?

The package name is a high-level name for an app or a sub-module that is instrumented.
Is intended to be used for routing in the Beholder/OpenTelemetry pipeline. 
Once that telemetry data is persisted in respective sink/storage it will be used for querying/filtering I would imagine.

> if this attribute will need to change often?

No, it should be fairly static per app/sub-module

pkg/beholder/client_test.go
patrickhuie19, line 89: From these tests its hard for me to see the value of both an `Emit` and `EmitMessage` API - would you be able to provide some details on that or perhaps more illustrative testing?

pkg/beholder/config.go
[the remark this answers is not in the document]
pkcll: I can certainly add support for `client.ForPackage(packageName)` builder

pkg/beholder/message.go
```diff
@@ -0,0 +1,275 @@
...
+	attrs["workflow_owner_address"] = m.WorkflowOwnerAddress
+	attrs["workflow_spec_id"] = m.WorkflowSpecID
+	attrs["workflow_execution_id"] = m.WorkflowExecutionID
+	attrs["beholder_data_schema"] = m.BeholderDataSchema
+	attrs["capability_contract_address"] = m.CapabilityContractAddress
+	attrs["capability_id"] = m.CapabilityID
+	attrs["capability_version"] = m.CapabilityVersion
+	attrs["capability_name"] = m.CapabilityName
+	attrs["network_chain_id"] = m.NetworkChainID
+	return attrs
```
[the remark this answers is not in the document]
pkcll: @jmank88 addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/15d4bebb34706b81fa7b69157cce812157e6a20d

pkg/beholder/client.go
pkcll: @patrickhuie19 https://github.com/smartcontractkit/chainlink-common/pull/694/commits/b80e230cd57a43ccf512f67c615aa5913148270b

pkg/beholder/client.go
pkcll: @patrickhuie19, added comment [here](https://github.com/smartcontractkit/chainlink-common/pull/694/commits/b80e230cd57a43ccf512f67c615aa5913148270b#diff-74ce7f3d5474f67956a5d9ba5779b478b715235a0d33e8826eb99b146664ae4dR168-R175)

pkg/beholder/client_test.go
pkcll: @patrickhuie19, 

Please see the related conversion https://github.com/smartcontractkit/chainlink-common/pull/694#discussion_r1714219755

I'm ok with reducing Message Emitter API to a single method `EmitMessage(context.Context, Message) error`


See usage examples:

https://github.com/smartcontractkit/chainlink-common/pull/694/files#diff-d5bbc8388dc6b5c9b3b4011eb5cd2448839eb67b352eae919a8a68fb3c3fdb12R117-R120

https://github.com/smartcontractkit/chainlink-common/pull/694/files#diff-d5bbc8388dc6b5c9b3b4011eb5cd2448839eb67b352eae919a8a68fb3c3fdb12R144-R147

https://github.com/smartcontractkit/atlas/blob/master/beholder/cmd/demo/main.go#L31-L46

pkcll: @patrickhuie19 addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/96027a3e2b2379bb38f5302876c0afc08a95bc97

pkg/beholder/client.go
pkcll: > But AFAIK in our case we do not want two providers because we want to send the same data to both places. The goal is for existing trace usages to be picked up and sent through to beholder as well.

yes, that's correct

pkg/beholder/message.go
pkcll: @jmank88 Addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/84357e5185442a17dd870226c45b83cd983dbd7b

pkg/beholder/client_test.go
patrickhuie19: I'm in favor of starting with the lowest surface area API. If we need to add more later on, we can, but I'd like to err on the side of supporting as few exported methods as possible for backwards compatibility. 

I believe an `Emit` API that takes in key values and will construct the message for you is the cleanest developer experience

pkg/beholder/client.go
patrickhuie19: IFIUC the consumer has to construct a `CustomMessage` for `EmitMessage`. Is there any value to consumers constructing a `CustomMessage` from the key value pairs they care about? If not, they may prefer to just put the key-value pairs into the API directly, which to me is a vote in favor for removing `EmitMessage` and keeping `Emit`

pavel-raykov (dismissed):

pkg/beholder/client.go
pkcll: Thats ok with me.
@jmank88 any thoughts on that ?

pkg/beholder/config.go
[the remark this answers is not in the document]
pkcll: @krebernisak added `client.ForPackage(packageName)` [here](https://github.com/smartcontractkit/chainlink-common/pull/694/commits/b69b21e9e30864a913aed32713de6ec0d4255d4e)

pkg/beholder/client_test.go
pkcll: Makes sense

pkg/beholder/client.go
pkcll: @patrickhuie19, removed `EmitMessage` method, see [here](https://github.com/smartcontractkit/chainlink-common/pull/694/commits/1392fd20dad391512f255acf8a2c827d74b3d66c)

jmank88: I actually thought that `EmitMessage` was the more important method, since it would be able to evolve compatibly, and since a Message could be cumulatively constructed before being sent :shrug: 

What do the call sites for these methods actually look like?

pkg/beholder/config.go
```diff
@@ -0,0 +1,52 @@
...
+	InsecureConnection       bool
+	CACertFile               string
+	OtelExporterGRPCEndpoint string
+
+	// OTel Resource
+	ResourceAttributes []otelattr.KeyValue
+	// Message Emitter
+	EmitterExportTimeout time.Duration
+	// OTel Trace
+	TraceSampleRate   float64
```
jmank88, line 19: This is called `Ratio` in the otel api and our other existing options:
```suggestion
	TraceSampleRatio   float64
```

pkg/beholder/client_test.go
```diff
@@ -0,0 +1,353 @@
...
+				} else {
+					exporterMock.AssertNotCalled(t, "Export")
+				}
+			}
+
+			waitUntilSent := func(done <-chan struct{}) {
+				for range tc.exporterCalledTimes {
+					select {
+					case <-done:
+					case <-time.After(10 * time.Second):
```
patrickhuie19, line 230: is there a way to instantly flush messages?

pkg/beholder/client_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+	client := NewStdoutClient()
+
+	err := client.Close()
+	assert.NoError(t, err)
+
+	exporterMock.AssertExpectations(t)
+}
+
+func TestClient_SetGlobas(t *testing.T) {
```
patrickhuie19, line 273: /nit TestClient_SetGlobals

pkg/beholder/client_test.go
```diff
@@ -0,0 +1,353 @@
...
+		return exporterMock, nil
+	}
+	client, err := newOtelClient(DefaultConfig(), otelErrorHandler, exporterFactory)
+	if err != nil {
+		t.Fatalf("Error creating beholder client: %v", err)
+	}
+	assert.NoError(t, err)
+
+	globals := getGlobals()
+	defer restoreGlobas(t, globals)
```
patrickhuie19, line 291: /nit `restoreGlobals`

pkg/beholder/client_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+	client.SetGlobals()
+
+	assert.Equal(t, client.LoggerProvider, otelglobal.GetLoggerProvider())
```
patrickhuie19, line 295: Does `GetLoggerProvider` also exist in `otel`? Noticing that the providers are split between two packages

pkg/beholder/client_test.go
```diff
@@ -0,0 +1,353 @@
...
+	client, err := newOtelClient(DefaultConfig(), otelErrorHandler, exporterFactory)
+	if err != nil {
+		t.Fatalf("Error creating beholder client: %v", err)
+	}
+	assert.NoError(t, err)
+
+	globals := getGlobals()
+	defer restoreGlobas(t, globals)
+
+	client.SetGlobals()
```
patrickhuie19, line 293: This feels odd to me. The flow I would expect is that you set your global providers in some type of init, and then inherit those providers in your clients. The globals are "read-only" with respect to the later initialized clients that need to configure those providers.

thoughts @jmank88 ?

pkg/beholder/config_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func ExampleConfig() {
+	config := beholder.Config{
+		InsecureConnection:       true,
+		CACertFile:               "",
+		OtelExporterGRPCEndpoint: "localhost:4317",
+		// Resource
+		ResourceAttributes: []otelattr.KeyValue{
+			otelattr.String("package_name", packageName),
+			otelattr.String("sender", "beholdeclient"),
```
patrickhuie19, line 24: /nit `beholderclient`

pkg/beholder/example_test.go
```diff
@@ -0,0 +1,89 @@
...
+	payloadBytes, err := proto.Marshal(payload)
+	if err != nil {
+		log.Fatalf("Failed to marshal protobuf")
+	}
+
+	// Emit the custom message anywhere from application logic
+	for range 10 {
+		err := beholder.Emit(context.Background(), payloadBytes,
+			"beholder_data_schema", "/custom-message/versions/1", // required
+			"beholder_data_type", "custom_message",
```
patrickhuie19, line 42: Is `beholder_data_type` just an example (i.e. it's just as meaningful as `foo bar`)?

pkg/beholder/example_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		log.Fatalf("failed to create new gauge")
+	}
+
+	// Use the counter and gauge for metrics within application logic
+	counter.Add(context.Background(), 1)
+	gauge.Record(context.Background(), rand.Int63n(101))
+
+	// Create a new trace span
+	_, rootSpan := beholder.Tracer().Start(context.Background(), "foo", trace.WithAttributes(
```
patrickhuie19, line 78: for parity of system behavior, I think it's worth it to drill context through `ExampleBeholderMetricTraces()` and use that inherited context here.

pkg/beholder/message.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func NewMessage(body []byte, attrKVs ...any) Message {
+	return Message{
+		Body:  body,
+		Attrs: NewAttributes(attrKVs...),
+	}
+}
+
+func (e *Message) AddAttributes(attrs Attributes) {
```
patrickhuie19, line 112: `NewMessage` takes `attrKVs ...any`, but then `AddAttributes` takes `attrs Attributes` which feels inconsistent.

pkg/beholder/message.go
```diff
@@ -0,0 +1,274 @@
...
+func (e *Message) AddOtelAttributes(attrs ...attribute.KeyValue) {
+	if e.Attrs == nil {
+		e.Attrs = make(map[string]any, len(attrs))
+	}
+	for _, v := range attrs {
+		e.Attrs[string(v.Key)] = v.Value
+	}
+}
+
+func (e *Message) OtelRecord() otellog.Record {
```
patrickhuie19, line 130: documentation on why OtelRecord vs SdkOtelRecord are needed would be helpful

pkg/beholder/example_test.go
pkcll: Yes its just an example

pkg/beholder/message.go
pkcll: SdkOtelRecord was used in previous implementations, removed it https://github.com/smartcontractkit/chainlink-common/pull/694/commits/cd6f5a8399c982e16d2a32a8e0fc9cf9c7901f5a

pkg/beholder/message.go
pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/adcbefac63c645216cd5913a05558d6da7894612

pkg/beholder/example_test.go
pkcll: https://github.com/smartcontractkit/chainlink-common/pull/694/commits/d0a857af3419bae08cbc01c1fbb6c6f9d26db407

pkg/beholder/example_test.go
pkcll: yes, its just an example

pkg/beholder/client_test.go
pkcll: `GetLoggerProvider` exists in `go.opentelemetry.io/otel/log/global`
while meter and tracer providers are from `go.opentelemetry.io/otel`.

pkg/beholder/client_test.go
jmank88: I think it depends how we want to coordinate the resource, but it does seems more composable to separate the concern of initing the global provider and registering the exporters etc.

pkg/beholder/client_test.go
pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/c8e8291c0792cd0d2f6f4cca0478361477481dd8

pkg/beholder/client_test.go
pkcll: @patrickhuie19 @jmank88 see https://github.com/smartcontractkit/chainlink-common/pull/694/commits/faa61c20702f1c1ccc1bebd6ca2e64c450f63df1

pkg/beholder/global/global.go
jmank88, line 1: I think that this package is too opinionated. Globals are for main programs to manage, and we should aim to provide an API that composes with existing globals.

pkg/beholder/message.go
```diff
@@ -0,0 +1,257 @@
...
+
+type Attributes map[string]any
+
+func NewAttributes(attrKVs ...any) Attributes {
+	attrs := make(Attributes, len(attrKVs)/2)
+	attrs.Add(attrKVs...)
+	return attrs
+}
+
+func (a Attributes) Add(attrKVs ...any) Attributes {
```
patrickhuie19, line 75: I think some direct unit testing of this method would be helpful for the reader.

pkg/beholder/global/global.go
patrickhuie19: > I think that this package is too opinionated
What's the difference between this and a non-global exported API? 

> provide an API that composes with existing globals.
WDYM?

jmank88: > What's the difference between this and a non-global exported API?

This approach is less flexible and incompatible with other libraries and programs that expect to set globals. We have our core node as an example of this.

> WDYM?

There are separate concerns here of constructing things and registering them globally. Keeping them separate is more flexible and can be integrated along with an existing solution.

pkg/beholder/message.go
```diff
@@ -0,0 +1,257 @@
...
+	attrs := NewAttributes(attrKVs...)
+	if e.Attrs == nil {
+		e.Attrs = make(map[string]any, len(attrs))
+	}
+	for k, v := range attrs {
+		e.Attrs[k] = v
+	}
+}
+
+func (e *Message) AddOtelAttributes(attrs ...attribute.KeyValue) {
```
patrickhuie19, line 121: Is it going to be clear to the consumer when they should use `AddAttributes` vs `AddOtelAttributes`?  You've put in a lot of work to add functionality around a `map[string]any`, so you could leverage the existing `AddAttributes` and add this otel case to the type switch case statement in `Add`.

pkg/beholder/message.go
```diff
@@ -0,0 +1,257 @@
...
+		"beholder_data_schema":        m.BeholderDataSchema,
+		"capability_contract_address": m.CapabilityContractAddress,
+		"capability_id":               m.CapabilityID,
+		"capability_version":          m.CapabilityVersion,
+		"capability_name":             m.CapabilityName,
+		"network_chain_id":            m.NetworkChainID,
+	}
+}
+
+type Attributes map[string]any
```
patrickhuie19, line 67: I think I understand why you ended up creating `Attributes`. Just to explore a bit - is this over-engineering? I don't see how this makes anything simpler for the consumer. Is there another way to accomplish this? 

cc @jmank88

jmank88: Using an alias would disallow methods but also drop any required conversions to/from the basic map type :shrug: 
```suggestion
type Attributes = map[string]any
```

pkcll: > I don't see how this makes anything simpler for the consumer.

If all attributes are constructed next to the `Emit` call then Attributes and sugar methods will not be of any use.

These are use cases I had in mind originally, see here https://github.com/smartcontractkit/chainlink-common/pull/694/files#diff-0ebe30d99f28fddcfef8e717720321077f917a7fe18aa1207d49c3b989812e79R16

After the code in `message.go` has evolved I can see that having it as a `map[string]any` alias makes more sense. 

Removing sugar methods, see here https://github.com/smartcontractkit/chainlink-common/commit/4e80921db29d6f144890789656570af35a17f528

pkcll: https://github.com/smartcontractkit/chainlink-common/commit/4e80921db29d6f144890789656570af35a17f528

pkg/beholder/message.go
pkcll: Removed it https://github.com/smartcontractkit/chainlink-common/commit/4e80921db29d6f144890789656570af35a17f528

pkg/beholder/global/global.go
pkcll: > This approach is less flexible and incompatible with other libraries and programs that expect to set globals. We have our core node as an example of this.

I intended to use a similar patter as OTel does with global providers getters/setters. 

The flow I had in mind:
1. Setup all OTel components logger,tracer,meter, emitter (uses separate logger provider under the hood) somewhere in the main package.
2. Set globals for logger,tracer,meter, and emitter via global setters (e.g otel.SetTracerProvider etc)
which will make them available for third parties libraries which rely on OTel
3. Set global reference to OtelClient wich references otel globals to make them available via sugar getters
e.g beholder.Tracer(), beholder.Meter()   
4. Use globals in various parts of the code via global otel getters or beholder getters

pkcll: I think that this package is too opinionated. Globals are for main programs to manage, and we should aim to provide an API that composes with existing globals.

@jmank88 addressed here https://github.com/smartcontractkit/chainlink-common/commit/99140c81d0b5c0a394cfb09a718b65889045bef8
Let me know if that works

go.mod
```diff
@@ -33,19 +34,33 @@ require (
...
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
+	github.com/go-playground/locales v0.13.0 // indirect
+	github.com/go-playground/universal-translator v0.17.0 // indirect
+	github.com/leodido/go-urn v1.2.0 // indirect
+)
+
 require (
```
jmank88, line 64:
```suggestion
```

pkg/beholder/client.go
```diff
@@ -0,0 +1,306 @@
...
+		return otlploggrpc.New(ctx, options...)
+	}
+	return newOtelClient(cfg, errorHandler, factory)
+}
+
+// Used for testing to override the default exporter
+type otlploggrpcFactory func(ctx context.Context, options ...otlploggrpc.Option) (sdklog.Exporter, error)
+
+func newOtelClient(cfg Config, errorHandler errorHandlerFunc, otlploggrpcNew otlploggrpcFactory) (OtelClient, error) {
+	ctx := context.Background()
```
jmank88, line 69: It's unfortunately that a context is required for construction, but let's include it as an argument so we can still short circuit start-up if necessary:
```suggestion
func newOtelClient(ctx context.Context, cfg Config, errorHandler errorHandlerFunc, otlploggrpcNew otlploggrpcFactory) (OtelClient, error) {
```

pkg/beholder/client.go
pkcll: @jmank88, @patrickhuie19  I thought initially that we might want to do that. Now I'm not sure. Let's clarify this.

pkg/beholder/global/global.go
jmank88: I think that is a better scope of responsibility for the global helpers. Do we even need a separate package? I think it would be OK to have this reduced functionality embedded directly in the beholder package.

pkg/beholder/client.go
pkcll: https://github.com/smartcontractkit/chainlink-common/commit/11c86c564cb226cc3b226c4e488b28aa63fa691a

patrickhuie19 (dismissed):

pkg/beholder/client.go
pkcll: Answered here https://github.com/smartcontractkit/chainlink-common/pull/696#discussion_r1727514619

pkg/beholder/global/global.go
pkcll: @jmank88 moved global into beholder https://github.com/smartcontractkit/chainlink-common/commit/555b582978b4936916442ff5eb92bb394728474b

pkg/beholder/client.go
```diff
@@ -0,0 +1,305 @@
...
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	sdkresource "go.opentelemetry.io/otel/sdk/resource"
+	"go.opentelemetry.io/otel/sdk/trace"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	oteltrace "go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type MessageEmitter interface {
```
jmank88, line 24: nit/ Does the `Message` prefix add value?
```suggestion
type Emitter interface {
```

pkg/beholder/client.go
```diff
@@ -0,0 +1,305 @@
...
+	// Sends message with bytes and attributes to OTel Collector
+	Emit(ctx context.Context, body []byte, attrKVs ...any) error
+}
+type Client interface {
```
jmank88, line 28: What is this interface for? And is it the only reason that `OtelClient` is not named just `Client`? Seems like we could simplify a lot by dropping it.

pkg/beholder/client.go
```diff
@@ -0,0 +1,305 @@
...
+	TracerProvider        oteltrace.TracerProvider
+	MeterProvider         otelmetric.MeterProvider
+	MessageLoggerProvider otellog.LoggerProvider
+
+	// OnClose
+	OnClose func() error
+}
+
+// NewOtelClient creates a new Client with OTel exporter
+func NewOtelClient(ctx context.Context, cfg Config, errorHandler errorHandlerFunc) (OtelClient, error) {
```
jmank88, line 58: We should either export this error handler type:
```suggestion
func NewOtelClient(ctx context.Context, cfg Config, errorHandler ErrorHandlerFunc) (OtelClient, error) {
```
Or just inline the literal since it is so simple:
```suggestion
func NewOtelClient(ctx context.Context, cfg Config, errorHandler func(error)) (OtelClient, error) {
```

jmank88: Oh, actually this is only used to set the global otel error handler. I think that is a separate concern like the other global setters that we dropped. Let's remove it from here. We could doc that it is important to set, but there is no need to entangle it with otel client construction.

pkg/beholder/client.go
```diff
@@ -0,0 +1,305 @@
...
+	factory := func(ctx context.Context, options ...otlploggrpc.Option) (sdklog.Exporter, error) {
+		return otlploggrpc.New(ctx, options...)
+	}
+	return newOtelClient(ctx, cfg, errorHandler, factory)
+}
+
+// Used for testing to override the default exporter
+type otlploggrpcFactory func(ctx context.Context, options ...otlploggrpc.Option) (sdklog.Exporter, error)
+
+func newOtelClient(ctx context.Context, cfg Config, errorHandler errorHandlerFunc, otlploggrpcNew otlploggrpcFactory) (OtelClient, error) {
```
jmank88, line 68: The `OtelClient` type is simple and stateless right now, but that might change in the future, and other parts of the API already use a pointer type, so let's return a pointer here too:
```suggestion
func newOtelClient(ctx context.Context, cfg Config, errorHandler errorHandlerFunc, otlploggrpcNew otlploggrpcFactory) (*OtelClient, error) {
```

pkg/beholder/noop.go
```diff
@@ -0,0 +1,129 @@
...
+// NewStdoutClient creates a new Client with stdout exporters
+// Use for testing and debugging
+// Also this client is used as a noop client when otel exporter is not initialized properly
+func NewStdoutClient(opts ...StddutClientOption) OtelClient {
+	cfg := DefaultStdoutClientConfig()
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+	// Logger
+	loggerExporter, _ := stdoutlog.New(
```
jmank88, line 52: Unchecked error

jmank88: Oh there are a few more below. I see the implementations do not actually return errors right now, but this seems like an unnecessary risk. Let's propagate any errors up through this func.

pkg/beholder/client.go
pkcll: I renamed it to resolve a name conflict with global getter `beholder.Emitter()`.
Renamed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/e749bb1f5386ca4f06a3eb239e53a812c7182cba

pkg/beholder/client.go
pkcll: > What is this interface for?
Initially, it was planned to be used for testing/mocking. 
After refactoring otel client constructors do not return the interface but OtelClient struct, so it can be removed. 

Removed, renamed here https://github.com/smartcontractkit/chainlink-common/pull/694/commits/1582cc2709b6da497f708e62971162128c9942d4

pkg/beholder/noop.go
pkcll: https://github.com/smartcontractkit/chainlink-common/pull/694/commits/4ebb5fc4fcd38e805ac6290d3d04ff7c44e9d729

pkg/beholder/client.go
pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/694/commits/26c09d7b0c62e7fdf69559393440bb076ac84864

go.mod
```diff
@@ -47,6 +56,12 @@ require (
...
```
jmank88, line 65:
```suggestion
```

pkg/beholder/global.go
```diff
@@ -0,0 +1,64 @@
...
+
+func Tracer() oteltrace.Tracer {
+	return GetClient().Tracer
+}
+
+func Meter() otelmetric.Meter {
+	return GetClient().Meter
+}
+
+func MessageEmitter() Emitter {
```
jmank88, line 40:
```suggestion
func Emitter() Emitter {
```

pkg/beholder/client.go
```diff
@@ -0,0 +1,299 @@
...
+	TracerProvider        oteltrace.TracerProvider
+	MeterProvider         otelmetric.MeterProvider
+	MessageLoggerProvider otellog.LoggerProvider
+
+	// OnClose
+	OnClose func() error
+}
+
+// NewClient creates a new Client with initialized OpenTelemetry components
+// To handle OpenTelemetry errors use SetOtelErrorHandler
```
jmank88, line 55: I think it would be OK to just link to the otel func here and drop the `SetOtelErrorHandler` helper, in order to keep this package API simpler.
```suggestion
// To handle OpenTelemetry errors use [otel.SetErrorHandler]
```

pkg/beholder/global.go
jmank88: Hmmm that is an unfortunately inconsistency. Should they all be aligned as `Get*()`ers?

pkcll: > Should they all be aligned as Get*()ers?

Sure, see https://github.com/smartcontractkit/chainlink-common/pull/694/commits/fa6cec4a1dd1a5d01f48dfc36220bc66086771db

pkg/beholder/noop.go
```diff
@@ -0,0 +1,139 @@
...
+	MetricOptions []stdoutmetric.Option
+}
+
+func DefaultStdoutClientConfig() StdoutClientConfig {
+	return StdoutClientConfig{
+		Config: DefaultConfig(),
+	}
+}
+
+func WithWriter(w io.Writer) StddutClientOption {
```
jmank88, line 133: For consistency, we should rename `NewStdoutClient` since it can be configured to write to other places.
I also wonder if functional options are the right fit for this. What about starting with something simple, like:
```go
func NewStdoutClient() (*Client, error) {
	return NewWriterClient(os.Stdout)
}
func NewWriterClient(w io.Writer) (*Client, error) {
	...
}
```

jmank88: And then `StdoutClientConfig` could be `WriterClientConfig` but also doesn't seem to need to be exported, does it?

pkcll: > And then StdoutClientConfig could be WriterClientConfig but also doesn't seem to need to be exported, does it?

It could be very handy for testing: use StdoutClient instead of OtlpClient to verify that otel instrumentation emits correct data from the code.

jmank88: But the API in its current form doesn't really make that config struct very usable, because a caller would have to write their own functional options to access it. If we want to expose that config, why not do it directly like the standard constructor?

pkcll: I agree with this point `And then StdoutClientConfig could be WriterClientConfig` 

See https://github.com/smartcontractkit/chainlink-common/pull/694/commits/3504324de7b444664cd40e5ed1ab70f42a6c4480

pkcll: > But the API in its current form doesn't really make that config struct very usable, because a caller would have to write their own functional options to access it. If we want to expose that config, why not do it directly like the standard constructor?

Right, makes sense, I misunderstood original suggestion 

https://github.com/smartcontractkit/chainlink-common/pull/694/commits/2187c017d32052632b16fe25f40034fa0d5a4ae9

jmank88 (approved):

patrickhuie19 (approved):

PR #695, opened 2024-08-12
Allow workflow specs to be created in code with the assistance of a CLI that creates helpers for building them.

nolag:
Note: No need to review the generated code in the same package as it's using an existing generator and just replacing a comment.

To come in later PRs for the builder:

Multiple functions in a capability (eg: ocr will have DataFeeds and custom, chain reader will have a bunch of methods).

For now, they will just have different json schemas with something like "type" defined as a fixed string for each of them.

Custom compute (in builder, making it testable and plugging into the engine will be it's own PRs also to keep things smaller)


pkg/capabilities/triggers/streams/streams_trigger-schema.json
```diff
@@ -0,0 +1,92 @@
...
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": ["feedIds", "maxFrequencyMs"]
+    },
+    "outputs": {
+      "items": {
+        "$ref": "#/$defs/Feed"
+      },
```
nolag, line 84: Looking for someone from Keystone to confirm, but I think this is actually an array of feeds that streams emits not a feed.  Then consensus changes to an array of array of feed.

When @DeividasK and I wrote this it felt much more natural this way, so I don't want to change it without some confirmation.

pkg/capabilities/triggers/streams/streams_trigger-schema.json
```diff
@@ -0,0 +1,92 @@
...
+    },
+    "Feed": {
+      "type": "object",
+      "properties": {
+        "feedId": {
+          "$ref": "#/$defs/feedId"
+        },
+        "fullReport": {
+          "type": "string",
+          "contentEncoding": "base64",
```
nolag, line 19: Calling out this base64 encoding hack.  Not sure what we want to do about it...

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	workflow := workflows.NewWorkflow(conf.Workflow)
+	streamsTrigger := streamscap.NewTrigger(workflow, *conf.Streams)
+
+	ocrInput := ocr3cap.ConsensusInput{Observations: workflows.ListOf[streams.Feed](streamsTrigger)}
+	consensus := ocr3cap.NewConsensus(workflow, "ccip_feeds", ocrInput, *conf.Ocr)
+
+	input := chainwritercap.TargetInput{SignedReport: consensus}
+	chainwritercap.NewTarget(workflow, conf.TargetChain, input, *conf.ChainWriter)
```
jmank88, line 49: If we are able to co-locate some of the generated types, then we could define these constructor/adders as methods on the config types in order to reduce verbosity:
```suggestion
	workflow := workflows.NewWorkflow(conf.Workflow)
	streamsTrigger := conf.Streams.New(workflow)

	ocrInput := ocr3cap.ConsensusInput{Observations: workflows.ListOf[streams.Feed](streamsTrigger)}
	consensus := conf.Ocr.New(workflow, "ccip_feeds", ocrInput)

	input := chainwritercap.TargetInput{SignedReport: consensus}
	conf.ChainWriter.New(workflow, conf.TargetChain, input)
```

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	return workflow.Spec()
```
jmank88, line 51: Could the signature be changed in order to allow returning the workflow more directly?
```suggestion
	return workflow, nil
```
Either with `(*workflows.Workflow, error)` or e.g. `WorkflowSpecFactory, error` capturing the `Spec()` method.

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	workflow := workflows.NewWorkflow(conf.Workflow)
+	streamsTrigger := conf.Streams.New(workflow)
+
+	ocrInput := ocr3.ConsensusInput{Observations: workflows.ListOf[streams.Feed](streamsTrigger)}
+	consensus := conf.Ocr.New(workflow, "ccip_feeds", ocrInput)
+
+	input := chainwriter.TargetInput{SignedReport: consensus}
+	conf.ChainWriter.New(workflow, conf.TargetChain, input)
```
jmank88, line 45: Some users might prefer this style of omitting the intermediate vars :shrug: 
```suggestion
	consensus := conf.Ocr.New(workflow, "ccip_feeds", ocr3.ConsensusInput{
		Observations: workflows.ListOf[streams.Feed](streamsTrigger),
	})

	conf.ChainWriter.New(workflow, conf.TargetChain, chainwriter.TargetInput{
		SignedReport: consensus,
	})
```

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,296 @@
...
+	Ocr         *ocr3.ConsensusConfig
+	ChainWriter *chainwriter.TargetConfig
+	TargetChain string
+}
+
+func NewWorkflowSpec(rawConfig []byte) (workflows.WorkflowSpec, error) {
+	conf := Config{}
+	if err := yaml.Unmarshal(rawConfig, &conf); err != nil {
+		return workflows.WorkflowSpec{}, err
+	}
```
jmank88, line 36: I'm not familiar with this yaml lib, but it looks like we could wrap it with generic helpers. Also, changing the signature to return `*Workflow` or an interface would pay off here too:
```suggestion
	conf, err := UnmarshalYaml[Config](rawConfig)
	if err != nil {
		return nil, err
	}
```
```go
func UnmarshalYaml[T any](raw []byte) (T, error) {
	var v T
	err := yaml.Unmarshal(raw, &v)
	return v, err 
}
```

pkg/workflows/builder.go
```diff
@@ -0,0 +1,175 @@
...
+			Targets:   make([]StepDefinition, 0),
+		},
+		names:          map[string]bool{},
+		duplicateNames: map[string]bool{},
+		emptyNames:     false,
+	}
+}
+
+// AddStep is meant to be called by generated code
+func AddStep[O any](w *Workflow, step Step[O]) CapDefinition[O] {
```
jmank88, line 117: If we declare a method on `Step`, then we can simply some of the generated code:
```suggestion
func (step Step[O]) AddTo(w *Workflow) CapDefinition[O] {
	return AddStep(w, step)
}

// AddStep is meant to be called by generated code
func AddStep[O any](w *Workflow, step Step[O]) CapDefinition[O] {
```

pkg/capabilities/consensus/ocr3/consensus_builders_generated.go
```diff
@@ -0,0 +1,194 @@
...
+           "aggregation_method": cfg.AggregationMethod,
+           "encoder": cfg.Encoder,
+           "encoder_config": cfg.EncoderConfig,
+           "report_id": cfg.ReportId,
+       },
+       CapabilityType: capabilities.CapabilityTypeConsensus,
+   }
+    step := workflows.Step[SignedReport]{Definition: def}
+     raw := workflows.AddStep(w, step)
+    return &consensus{CapDefinition: raw}
```
jmank88, line 30: Given a method on `Step`, these can be reduced:
```suggestion
    return &consensus{CapDefinition: step.AddTo(w)}
```

go.mod
```diff
@@ -35,21 +40,32 @@ require (
...
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
+	github.com/sanity-io/litter v1.5.5 // indirect
+	github.com/spf13/pflag v1.0.5 // indirect
+	golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028 // indirect
+)
+
+require github.com/iancoleman/strcase v0.3.0
+
 require (
```
jmank88, line 61:
```suggestion
	github.com/iancoleman/strcase v0.3.0
)

require (
	github.com/goccy/go-yaml v1.12.0 // indirect
	github.com/inconshreveable/mousetrap v1.1.0 // indirect
	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
	github.com/sanity-io/litter v1.5.5 // indirect
	github.com/spf13/pflag v1.0.5 // indirect
	golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028 // indirect
```

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,307 @@
...
+func NewWorkflowSpec(rawConfig []byte) (*workflows.WorkflowSpecFactory, error) {
+	conf, err := UnmarshalYaml[Config](rawConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	workflow := workflows.NewWorkflow(conf.Workflow)
+	streamsTrigger := conf.Streams.New(workflow)
+	consensus := conf.Ocr.New(workflow, "ccip_feeds", ocr3.ConsensusInput{
+		Observations: workflows.ListOf[streams.Feed](streamsTrigger)},
```
jmank88, line 41: Would it be possible and worthwhile to generate or define alternate names for these types?
```go
package streams
type FeedList workflows.ListOf[Feed]
```
```suggestion
		Observations: streams.FeedList(streamsTrigger)},
```
If there is a generic alias problem, those should be possible in go 1.23:
```go
package streams
type FeedList = workflows.ListOf[Feed]
```

nolag: ListOf isn't a type it's a function call.

pkg/workflows/builder_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+type Config struct {
+	Workflow    workflows.NewWorkflowParams
+	Streams     *streams.TriggerConfig
+	Ocr         *ocr3.ConsensusConfig
+	ChainWriter *chainwriter.TargetConfig
+	TargetChain string
+}
+
+func NewWorkflowSpec(rawConfig []byte) (*workflows.WorkflowSpecFactory, error) {
```
jmank88, line 32: nit/ I was thinking `WorkflowSpecFactory` would be an interface that just captures the `Spec() (WorkflowSpec, error)` method. That would be slightly more flexible, if possible. I think the concrete type as just `Workflow` is fine :shrug:

nolag: I'll rename the function, but I was trying to force usage of the SDK so that users can't shoot themselves in the foot as easily.

Thoughts?

jmank88: Yeah that makes sense

PR #696, opened 2024-08-13
pkg/loop: use beholder

jmank88:
#694 introduced the beholder package. This PR is a follow up to initialize the beholder client from LOOP plugin servers.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/14110


pkg/beholder/config.go
```diff
@@ -17,8 +18,9 @@ type Config struct {
 	// Message Emitter
 	EmitterExportTimeout time.Duration
 	// OTel Trace
-	TraceSampleRate   float64
-	TraceBatchTimeout time.Duration
+	TraceSampleRate    float64
+	TraceBatchTimeout  time.Duration
+	TracerProviderOpts []sdktrace.TracerProviderOption
```
jmank88, line 23: Supporting configured `TracerProviderOption`s let's us pass in our legacy trace exporter.

pkg/loop/config.go
jmank88, line 1: Here we forward the config from core as environment vars.
- [ ] TODO other fields

pkg/loop/telem.go
```diff
@@ -57,18 +57,69 @@ func NewGRPCOpts(registerer prometheus.Registerer) GRPCOpts {
 
 // SetupTracing initializes open telemetry with the provided config.
 // It sets the global trace provider and opens a connection to the configured collector.
-func SetupTracing(config TracingConfig) (err error) {
+func SetupTracing(config TracingConfig) error {
```
jmank88, line 60: This method is preserved since it is exported.

pkg/loop/server.go
```diff
@@ -68,18 +72,41 @@ func (s *Server) start() error {
...
+	tracingConfig := TracingConfig{
 		Enabled:         envCfg.TracingEnabled,
 		CollectorTarget: envCfg.TracingCollectorTarget,
 		SamplingRatio:   envCfg.TracingSamplingRatio,
 		TLSCertPath:     envCfg.TracingTLSCertPath,
 		NodeAttributes:  envCfg.TracingAttributes,
 		OnDialError:     func(err error) { s.Logger.Errorw("Failed to dial", "err", err) },
-	}); err != nil {
+	}
+
```
jmank88, line 83: I need to add back the old path here, for cases where Beholder is not configured.

pkg/beholder/client.go
```diff
@@ -256,17 +257,16 @@ func newTracerProvider(config Config, resource *sdkresource.Resource, creds cred
 	if err != nil {
 		return nil, err
 	}
-	tp := sdktrace.NewTracerProvider(
-		sdktrace.WithBatcher(exporter,
-			trace.WithBatchTimeout(config.TraceBatchTimeout)), // Default is 5s
+	opts := slices.Clone(config.TracerProviderOpts)
+	opts = append(opts, sdktrace.WithBatcher(exporter,
```
pkcll, line 261: @jmank88 Probably not safe to just append it like this. Some of these options are mutually exclusive. 
Probably better to not mix them up and use the default options from Beholder or use overrides from the config

patrickhuie19 (approved):

pkcll (approved):

pkg/beholder/client.go
```diff
@@ -269,17 +269,20 @@ func newTracerProvider(config Config, resource *sdkresource.Resource, creds cred
...
 		sdktrace.WithSampler(
 			sdktrace.ParentBased(
 				sdktrace.TraceIDRatioBased(config.TraceSampleRatio),
 			),
 		),
-	)
-	return tp, nil
+	}
+	if config.TraceSpanExporter != nil {
+		opts = append(opts, sdktrace.WithBatcher(config.TraceSpanExporter))
```
jmank88, line 283: @pkcll @patrickhuie19 Does this optional config for an additional exporter make sense? Or should we consider other alternatives?

pkcll: Passing the configured exporter is more of a dependency injection rather than part of the config. Passing an exporter as part of the config seems like mixing different types of code aspects.

pkcll: I see 4 options:
- loop.telem.SetupTracing calling beholder.newTraceProvider
- beholder.newTraceProvider calling loop.telem.SetupTracing
- loop.telem.SetupTracing and beholder.newTraceProvider calling some "common denominator" function e.g telemetry.NewTraceProvider
- both loop.telem.SetupTracing and beholder.newTraceProvider using shared configuration

patrickhuie19: What does the TraceSpanExporter do on top of everything else configured in `opts`?

jmank88: This is one approach to bridging the two tracing systems.

pkcll: > Passing the configured exporter is more of a dependency injection rather than part of the config. Passing an exporter as part of the config seems like mixing different types of code aspects.

Otel actually use config as an "umbrella" struct for both pure configuration (collection of option fields) and dependencies/sub-components 

See here: https://github.com/open-telemetry/opentelemetry-go/blob/fe6c67e7e9b408d7f1e05356e4ccafadf0475b10/sdk/trace/provider.go#L26-L45

```
type tracerProviderConfig struct {
	processors []SpanProcessor
	sampler Sampler
	idGenerator IDGenerator
	spanLimits SpanLimits
	resource *resource.Resource
}
```

The question is do we want to follow chainlink/chainlink-common config semantics or use it the way Otel does it.

pkcll: > What does the TraceSpanExporter do on top of everything else configured in opts?  
@patrickhuie19 
It allows to override the default otel.Exporter configured in in NewOtelClient/newTracerProvider

jmank88: It adds an additional exporter. It does not replace the normal one. They are accumulated: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/provider.go#L338

pkg/loop/telem.go
pkcll: There are not that many usages of it thought, see https://github.com/search?q=%22loop.SetupTracing%22&type=code
So it should be feasible if we decide to deprecate, move or change `SetupTracing`

pkg/beholder/client.go
pkcll: You are right. 
Do you see use case for having two separate trace processors/exporters for loop/server otel instrumentation and for custom traces produced from code?

pkcll: @jmank88 @patrickhuie19, ^^^ WDYT

jmank88: We could add a check for duplicate endpoints. But otherwise using both at the same time is not inherently problematic, and having the flexibility will make it easier to transition as well as support non-production cases.

pkcll: > having the flexibility will make it easier to transition 

Makes sense

pkcll (dismissed):

pkg/loop/config.go
```diff
@@ -11,13 +11,20 @@ import (
...
 	envTracingCollectorTarget = "CL_TRACING_COLLECTOR_TARGET"
 	envTracingSamplingRatio   = "CL_TRACING_SAMPLING_RATIO"
 	envTracingAttribute       = "CL_TRACING_ATTRIBUTE_"
 	envTracingTLSCertPath     = "CL_TRACING_TLS_CERT_PATH"
+
+	envBeholderEndpoint         = "CL_BEHOLDER_ENDPOINT"
+	envBeholderInsecureConn     = "CL_BEHOLDER_INSECURE_CONNECTION"
+	envBeholderCACertFile       = "CL_BEHOLDER_CA_CERT_FILE"
+	envBeholderAttribute        = "CL_BEHOLDER_ATTRIBUTE_"
+	envBeholderTraceSampleRatio = "CL_BEHOLDER_TRACE_SAMPLE_RATIO"
```
pkcll, line 27: Should other beholder config [options](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/beholder/config.go#L10-L30) be configurable ?

pkg/beholder/global.go
```diff
@@ -12,7 +12,9 @@ import (
 )
 
 // Pointer to the global Beholder Client
-var globalClient = defaultClient()
+var globalClient atomic.Pointer[Client]
```
patrickhuie19, line 15: This is to sync access within the `beholder` pkg? What consumers interact with `globalClient`?

jmank88: There is an exported `SetClient` function.

pkg/loop/config.go
```diff
@@ -164,3 +205,12 @@ func getDatabaseURL() (*url.URL, error) {
 	}
 	return u, nil
 }
+
+// getTelemetryInsecureConnection parses and validates the TRACING_ENABLED environment variable.
+func getTelemetryInsecureConnection() (bool, error) {
```
patrickhuie19, line 210: /nit `getTelemetryEnabled` ?

patrickhuie19 (approved):

pkg/loop/config.go
jmank88: Updated, but I just simplified things by removing helpers and in-lining instead for the simple ones.

pkg/loop/config.go
```diff
@@ -11,13 +11,20 @@ import (
...
 	envTracingCollectorTarget = "CL_TRACING_COLLECTOR_TARGET"
 	envTracingSamplingRatio   = "CL_TRACING_SAMPLING_RATIO"
 	envTracingAttribute       = "CL_TRACING_ATTRIBUTE_"
 	envTracingTLSCertPath     = "CL_TRACING_TLS_CERT_PATH"
+
+	envTelemetryEndpoint         = "CL_TELEMETRY_ENDPOINT"
+	envTelemetryInsecureConn     = "CL_TELEMETRY_INSECURE_CONNECTION"
+	envTelemetryCACertFile       = "CL_TELEMETRY_CA_CERT_FILE"
+	envTelemetryAttribute        = "CL_TELEMETRY_ATTRIBUTE_"
+	envTelemetryTraceSampleRatio = "CL_TELEMETRY_TRACE_SAMPLE_RATIO"
```
pkcll, line 27: #nit to have the same spelling as `CL_TRACING_SAMPLING_RATIO`

```suggestion
	envTelemetryTraceSamplingRatio = "CL_TELEMETRY_TRACE_SAMPLING_RATIO"
```

pkg/loop/config.go
```diff
@@ -11,13 +11,20 @@ import (
...
+
 	envTracingEnabled         = "CL_TRACING_ENABLED"
 	envTracingCollectorTarget = "CL_TRACING_COLLECTOR_TARGET"
 	envTracingSamplingRatio   = "CL_TRACING_SAMPLING_RATIO"
 	envTracingAttribute       = "CL_TRACING_ATTRIBUTE_"
 	envTracingTLSCertPath     = "CL_TRACING_TLS_CERT_PATH"
+
+	envTelemetryEndpoint         = "CL_TELEMETRY_ENDPOINT"
+	envTelemetryInsecureConn     = "CL_TELEMETRY_INSECURE_CONNECTION"
+	envTelemetryCACertFile       = "CL_TELEMETRY_CA_CERT_FILE"
```
pkcll, line 25: #nit naming unification
```suggestion
	envTelemetryTLSCertPath      = "CL_TELEMETRY_TLS_CERT_PATH"
```

pkg/loop/config.go
```diff
@@ -11,13 +11,20 @@ import (
...
+	envDatabaseURL = "CL_DATABASE_URL"
+	envPromPort    = "CL_PROMETHEUS_PORT"
+
 	envTracingEnabled         = "CL_TRACING_ENABLED"
 	envTracingCollectorTarget = "CL_TRACING_COLLECTOR_TARGET"
 	envTracingSamplingRatio   = "CL_TRACING_SAMPLING_RATIO"
 	envTracingAttribute       = "CL_TRACING_ATTRIBUTE_"
 	envTracingTLSCertPath     = "CL_TRACING_TLS_CERT_PATH"
+
+	envTelemetryEndpoint         = "CL_TELEMETRY_ENDPOINT"
```
pkcll, line 23: #nit naming unification

```suggestion
	envTelemetryCollectorTarget         = "CL_TELEMETRY_COLLECTOR_TARGET"
```

pkcll, line 23: Should beholder config also have an explicit `CL_TELEMETRY_ENABLED` flag ?

pkg/loop/config.go
jmank88: Why path instead of file? I was following the `beholder.Config` field names.

pkg/loop/config.go
jmank88: I think we need to prioritize aligning the field end to end, rather than with other sections. Do we want to rename this everywhere?

pkg/loop/config.go
jmank88: This is also aligned with the `beholder.Config`

jmank88: Yes, this field was added later, so it was missing from this set.

jmank88: Updated

pkg/loop/config.go
pkcll: Maybe we need to use the same naming for beholder config and existing tracing config because it will be part of the public config doc. But we can keep it as is

pkg/loop/config.go
pkcll: Yes, if you rename env vars here then renaming beholder config fields as will be necessary as well

pkg/loop/server.go
```diff
@@ -68,16 +69,42 @@ func (s *Server) start() error {
...
+				return fmt.Errorf("failed to setup tracing exporter: %w", err)
+			}
+			beholderCfg.TraceSpanExporter = exporter
+		}
+
+		beholderClient, err := beholder.NewClient(beholderCfg)
+		if err != nil {
+			return fmt.Errorf("failed to create beholder client: %w", err)
+		}
+		beholder.SetClient(beholderClient)
```
pkcll, line 107: @jmank88,  `SetupTracing` sets global otel TracerProvider in the end [here](https://github.com/smartcontractkit/chainlink-common/blob/14a5c7af361f61afb7e02360e286012770edf8f0/pkg/loop/telem.go#L134-L135). 
Its reflected in README in chainlink repo see [here](https://github.com/smartcontractkit/chainlink/blob/72f4cc8aaa128202fd5974c6dbfb29c6beb1be12/.github/tracing/README.md?plain=1#L78) 

To achieve parity in behavior we need to do the same after beholder client is created.
At minimum set global tracer
```
  otel.SetTracerProvider(tracerProvider)
  otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
```
or set all globals
 
```
beholder.SetGlobalOtelProviders
```


See `otel.GetTracerProvider` calls in chainlink core:
- https://github.com/smartcontractkit/chainlink/blob/72f4cc8aaa128202fd5974c6dbfb29c6beb1be12/core/services/pg/connection.go#L56
- https://github.com/smartcontractkit/chainlink/blob/72f4cc8aaa128202fd5974c6dbfb29c6beb1be12/core/web/router.go#L66

```suggestion
		beholder.SetClient(beholderClient)
		
		// Set global providers including Tracer to have parity with 'SetupTracing'
		beholder.SetGlobalOtelProviders()
```

patrickhuie19: We can have parity in the tracing systems and how they function, but there's no real/immediate need for it. Tracing will be used and IFIUC gradually phased out internally in support of Beholder. Externally no NOPs are currently using the tracing feature. 

IFIUC the only thing that won't function as intended is if both tracing and beholder are enabled as the globally registered tracing provider could conflict.

patrickhuie19 (approved):

pkcll (approved):

pkg/loop/server.go
jmank88: Updated to call `beholder.SetGlobalOtelProviders`

pkcll (approved):

patrickhuie19 (approved):

PR #698, opened 2024-08-13
Add json schemas for existing capabilities

nolag:
This is part of a larger effort to allow workflow specs to be created in code with the assistance of a CLI that creates helpers for building them.

To see in its full context, visit [this PR](https://github.com/smartcontractkit/chainlink-common/pull/695/files)

### DIFF
--- a/pkg/capabilities/consensus/ocr3/ocr3_consensus-schema.json
+++ b/pkg/capabilities/consensus/ocr3/ocr3_consensus-schema.json
@@ -1,0 +1,134 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/offchain_reporting@1.0.0",
+  "$defs": {
+    "SignedReport": {
+      "type": "object",
+      "properties": {
+        "WorkflowExecutionID": {
+          "type": "string"
+        },
+        "Value": {
+          "type": "object",
+          "properties": {
+            "Underlying": {
+              "type": "object"
+            }
+          },
+          "additionalProperties": false,
+          "required": ["Underlying"]
+        },
+        "Err": {
+          "type": "boolean"
+        }
+      },
+      "additionalProperties": false,
+      "required": ["WorkflowExecutionID", "Value", "Err"]
+    },
+    "FeedValue": {
+      "type": "object",
+      "properties": {
+        "deviation": {
+          "type": "string",
+          "description": "The deviation that is required to generate a new report. Expressed as a percentage. For example, 0.01 is 1% deviation."
+        },
+        "heartbeat": {
+          "type": "integer",
+          "minimum": 1,
+          "description": "The interval in seconds after which a new report is generated, regardless of whether any deviations have occurred. New reports reset the timer."
+        },
+        "remappedID": {
+          "type": ["string", "null"],
+          "description": "An optional remapped ID for the feed."
+        }
+      },
+      "additionalProperties": false,
+      "required": ["deviation", "heartbeat"]
+    }
+  },
+  "properties": {
+    "config": {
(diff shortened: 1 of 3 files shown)


pkg/capabilities/consensus/ocr3/ocr3_consensus-schema.json
```diff
@@ -0,0 +1,131 @@
...
+      ]
+    },
+    "inputs": {
+      "properties": {
+        "observations": {
+          "type": "array",
+          "items": {
+            "$ref": "../../triggers/streams/streams_trigger-schema.json#/$defs/Feed"
+          }
+        }
```
nolag, line 117: Can someone confirm if this is right?  When @DeividasK and I where working on it, this felt natural, but looking closer at the code I think it accepts an array of arrays of feed, not just an array of feeds

pkg/capabilities/triggers/streams/streams_trigger-schema.json
```diff
@@ -0,0 +1,92 @@
...
+    },
+    "Feed": {
+      "type": "object",
+      "properties": {
+        "feedId": {
+          "$ref": "#/$defs/feedId"
+        },
+        "fullReport": {
+          "type": "string",
+          "contentEncoding": "base64",
```
nolag, line 19: Calling out this base64 encoding hack (also on other lines). Not sure what we want to do about it...

pkg/capabilities/consensus/ocr3/ocr3_consensus-schema.json
```diff
@@ -0,0 +1,134 @@
...
+          "enum": ["data_feeds"]
+        },
+        "aggregation_config": {
+          "type": "object",
+          "properties": {
+            "allowedPartialStaleness": {
+              "type": "string",
+              "description": "Allowed partial staleness as a number between 0 and 1."
+            },
+            "feeds": {
```
bolekk, line 63: This is specific to the "data_feeds" aggregator and won't work with other ones. Why do you want to put it in the schema?

pkg/capabilities/targets/chainwriter/chainwriter_target-schema.json
```diff
@@ -0,0 +1,45 @@
...
+          "description": "The address to write to."
+        },
+        "deltaStage": {
+          "type": "string",
+          "pattern": "^[0-9]+[smhd]$",
+          "description": "The delta stage which must be a number followed by a time symbol (s for seconds, m for minutes, h for hours, d for days)."
+        },
+        "schedule": {
+          "type": "string",
+          "enum": ["oneAtATime"],
```
bolekk, line 19: There is also the "allAtOnce" option.

pkg/capabilities/triggers/streams/streams_trigger-schema.json
bolekk: Do we need to specify contentEncoding at all?

pkg/capabilities/consensus/ocr3/ocr3_consensus-schema.json
nolag: Yeah, this is meant to represent the capability as it is today.  I'll add support for having different methods in a follow-up PR.

pkg/capabilities/triggers/streams/streams_trigger-schema.json
nolag: It's not really a string, it's bytes under the hood.  Having this would at least allow the engine to potentially do the conversion or something...

We really do need something better long-term, I'm open to a fix now but I don't think it's clear what to do.  Otherwise, maybe some follow-up PR.

bolekk (approved):
Approving but not a huge fan of explicitly listing specific aggregators there.

cedric-cordenier (approved):

PR #701, opened 2024-08-14
[KS-425] Fix parallel use of pointers

cedric-cordenier:
### What

Store items are used in different threads, so to guard against potential race conditions we perform a copy of the request in each read method.


pkg/capabilities/consensus/ocr3/requests/store.go
```diff
@@ -30,16 +30,16 @@ func (s *Store) GetByIDs(ctx context.Context, requestIDs []string) []*Request {
 	for _, r := range requestIDs {
 		gr, ok := s.requests[r]
 		if ok {
-			o = append(o, gr)
+			o = append(o, gr.Copy())
```
krehermann, line 33: i find this unintuitive and likely to lead to subtle bugs if more hands are in this code base.

how can you make it more obvious when Copy is needed?

at minimum a short comment around at the usage sites and func comment. 

a defensive alternative might be to type def something like  
```
type copiedRequests []*request
 func (c *copiedRequests) append(r *request) {
 c = append(c, r.Copy)
 }
 ```
 
 something like that leverage the compiler to help defend intermixing and confusing append(,r) vs append(, r.Copy())

cedric-cordenier: I added some documentation -- I preferred this to something more defensive which feels odd to me (more complicated and I don't feel like it would reduce bugs; new methods would explicitly need to opt in to using this)

DeividasK (approved):

vyzaldysanchez (approved):

PR #704, opened 2024-08-14
Add code generators to help build workflows

nolag:
Last part taken from [this PR](https://github.com/smartcontractkit/chainlink-common/pull/695/files)


pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -27,7 +27,7 @@ const defaultTickerResolutionMs = 1000
 // TODO pending capabilities configuration implementation - this should be configurable with a sensible default
 const defaultSendChannelBufferSize = 1000
 
-type config struct {
+type Config struct {
 	// strings should be hex-encoded 32-byte values, prefixed with "0x", all lowercase, minimum 1 item
 	FeedIDs []string `json:"feedIds" jsonschema:"pattern=^0x[0-9a-f]{64}$,minItems=1"`
 	// must be greater than 0
```
DeividasK, line 33: Shouldn't this be replaced with the generated structs? This tests if what is generated matches what is needed.

DeividasK:
It might make sense to have tests for the generators so the edge cases are covered (I'm not sure if you plan to do a builder for every possible combination).

pkg/capabilities/triggers/mercury_trigger.go
nolag: I'll verify they are the same, then do the replacement :D.  If they aren't, I'll update them to be

pkg/capabilities/pb/capabilities.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.31.0
-// 	protoc        v4.25.1
+// 	protoc        v4.24.4
```
krehermann, line 4: regression here and elsewhere

nolag: yeah, I used make generated, I'm surprised it doesn't enforce the tool version.  Looking into it

go.mod
```diff
@@ -26,6 +30,7 @@ require (
 	github.com/riferrei/srclient v0.5.4
 	github.com/shopspring/decimal v1.4.0
 	github.com/smartcontractkit/libocr v0.0.0-20240419185742-fd3cab206b2c
+	github.com/spf13/cobra v1.8.1
```
krehermann, line 33: we are using `github.com/urfave/cli` in core. let's stay unified.

pkg/capabilities/cli/cmd/template_workflow_generator_helper.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	for file, t := range t.Templates {
+		content, err := genFromTemplate(file, t, info)
+		if err != nil {
+			return nil, err
+		}
+
+		// can use a template, but it's simple for now
+		fileName, err := genFromTemplate("file name for "+file, file, info)
```
krehermann, line 34: this is bizarre. genFromTemplate is both creating content and creating file names?

pkg/capabilities/cli/cmd/generate_types.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+)
+
+var Dir string
+
+// CapabilitySchemaFilePattern is used to extract the package name from the file path.
+// This is used as the package name for the generated Go types.
+var CapabilitySchemaFilePattern = regexp.MustCompile(`([^/]+)_(action|trigger|consensus|target)-schema\.json$`)
+
+// reg := regexp.MustCompile(`([^/]+)_(trigger|action)\.json$`)
```
krehermann, line 24: rm?

pkg/capabilities/cli/cmd/generate_types.go
```diff
@@ -0,0 +1,219 @@
...
+	configInfo := ConfigInfo{
+		Config: generator.Config{
+			Tags:   []string{"json", "yaml", "mapstructure"},
+			Warner: func(message string) { fmt.Printf("Warning: %s\n", message) },
+		},
+		SchemaToTypeInfo: map[string]TypeInfo{},
+	}
+
+	for _, schemaFilePath := range schemaFilePaths {
+		jsonSchema, err := schemas.FromJSONFile(schemaFilePath)
```
krehermann, line 169: validate the schemaFilePath against the regex first and return clear error msg. iiuc, failure will happen on 176, but user won't know what the pattern is supposed to be

pkg/capabilities/cli/cmd/generate_types.go
```diff
@@ -0,0 +1,219 @@
...
+			RootType:    rootType,
+			OutputName:  outputName,
+		})
+	}
+	return configInfo, nil
+}
+
+// TypesFromJSONSchema generates Go types from a JSON schema file.
+func TypesFromJSONSchema(schemaFilePath string, cfgInfo ConfigInfo) (outputFilePath, outputContents string, err error) {
+	typeInfo := cfgInfo.SchemaToTypeInfo[schemaFilePath]
```
krehermann, line 200: check not exist condition

pkg/capabilities/cli/cmd/generated_info.go
```diff
@@ -0,0 +1,173 @@
...
+	"unicode"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+)
+
+type GeneratedInfo struct {
+	Package        string
+	Config         Struct
+	Input          *Struct
+	Types          map[string]Struct
```
krehermann, line 18: `Struct`	is a confusing and non obvious name. can we pick something better?

pkg/capabilities/cli/cmd/generated_info.go
```diff
@@ -0,0 +1,173 @@
...
+				start := fset.Position(field.Type.Pos()).Offset
+				end := fset.Position(field.Type.End()).Offset
+				typeStr := src[start:end]
+				if typeStr == "interface{}" {
+					typeStr = "any"
+				}
+				f := Field{}
+
+				if field.Tag != nil {
+					tag := reflect.StructTag(field.Tag.Value[1 : len(field.Tag.Value)-1])
```
krehermann, line 124: this looks like it can panic. why is structTag guaranteed to have at least 1 elem?

go.mod
nolag: not sure what pulled this in.

pkg/values/value.go
```diff
@@ -251,7 +251,13 @@ func unwrapTo[T any](underlying T, to any) error {
 		}
 		*tb = underlying
 	default:
-		return fmt.Errorf("cannot unwrap to value of type: %T", to)
+		rTo := reflect.ValueOf(to)
```
krehermann, line 254: why is this changing? a comment about intention would be useful

pkg/capabilities/cli/cmd/generated_info.go
nolag: The way we called the generator for the other code will guarantee it.  I can add a comment to that effect.

pkg/values/value.go
nolag: type FeedId string

then decode to FeedId shouldn't break.

go.mod
nolag: It's not a drop in replacement and we use it in the tools (see cmd for example).  Is this a blocker for you or can it be done later?

pkg/capabilities/cli/cmd/template_workflow_generator_helper.go
nolag: genFromTemplate just runs template.  I'm using a template to get the file name.

pkg/capabilities/cli/cmd/generated_info.go
nolag: It's pretty common among generators to represent the info about the struct that you are creating in a type called Struct.  Do you have a better name in mind?

go.mod
jmank88: The problem is that this repo is imported everywhere else, so these kinds of deps have major consequences. What features are we using? Could we even use pure std lib instead?

pkg/capabilities/cli/cmd/field.go
```diff
@@ -0,0 +1,8 @@
+package cmd
+
+type Field struct {
```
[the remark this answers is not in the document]
nolag: It's literally a field on the generated structs.  I'm open to a better name if you have one.  Many code generators I've seen use Struct and Field though.

go.mod
nolag: What does the standard provide?  Can this be done in a follow-up PR instead (lower priority), I really want to unblock the rest of my work.

jmank88: If you just need basic [flags](https://pkg.go.dev/flag) and [env vars](https://pkg.go.dev/os), then there is no need for a third party library. Are you using any other features?

nolag: @DeividasK you set up the CLI, mind looking at ^

go.mod
```diff
@@ -78,15 +88,19 @@ require (
...
 	golang.org/x/text v0.16.0 // indirect
+	golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20240711142825-46eb208f015d // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240711142825-46eb208f015d // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
 replace (
+	// until merged upstream:
+	github.com/atombender/go-jsonschema => github.com/nolag/go-jsonschema v0.16.0-rtinianov
```
jmank88, line 102: What do we need this for?
We are hoping to shed these in order to be able to use gorelease, and to un-complicate importers.

nolag: I forgot to add the PR, but I added support for a lot of things we need

https://github.com/omissis/go-jsonschema/pull/264

pkg/capabilities/cli/cmd/generator_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/basicconsensus"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/basictarget"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/basictrigger"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/externalreferenceaction"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/nestedaction"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/referenceaction"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows"
+)
+
+//go:generate go run github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli generate-types --dir $GOFILE
```
jmank88, line 17: Why does this generate call depend on the environment? Ideally these can be invoked directly, without extra assumptions.

nolag: By the environment, you mean the $GOFILE right?  I'm using it to know where to generate from.  This tool will eventually work for third party devs too.

jmank88: How does `$GOFILE` get set?

jmank88: Oh I see it comes from `go generate` itself, so no problem :+1:

jmank88 (approved):

jmank88 (approved):

pkg/values/value.go
DeividasK: This should have a corresponding test in the values package.

pkg/values/value.go
```diff
@@ -251,7 +251,15 @@ func unwrapTo[T any](underlying T, to any) error {
 		}
 		*tb = underlying
 	default:
-		return fmt.Errorf("cannot unwrap to value of type: %T", to)
+		// Don't break for custom types that are the same underlying type
+		// eg: type FeedId string allows verification of FeedId's shape while unmarshalling
```
DeividasK, line 255: This should have a corresponding test in the values package.

DeividasK (approved):
Approving on the assumption that you'll add a test in a follow-up PR.

PR #705, opened 2024-08-14
check cr service is started

Farber98:

### DIFF
--- a/pkg/types/interfacetests/chain_reader_interface_tests.go
+++ b/pkg/types/interfacetests/chain_reader_interface_tests.go
@@ -48,6 +48,7 @@ const (
 	AnySecondContractName                       = "Not" + AnyContractName
 )
 
+var ErrServiceNotStarted = errors.New("service not started")
 var AnySliceToReadWithoutAnArgument = []uint64{3, 4}
 
 const AnyExtraValue = 3
@@ -70,6 +71,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 				tester.SetTestStructLatestValue(t, &secondItem)
 
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				actual := &TestStruct{}
@@ -88,6 +91,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				var prim uint64
@@ -101,6 +106,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				var returnVal1 uint64
@@ -140,6 +147,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				bindings := tester.GetBindings(t)
 				seenAddrs := map[string]bool{}
 				for _, binding := range bindings {
@@ -160,6 +169,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				var slice []uint64
@@ -176,6 +187,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 				testStruct.BigField = nil
 				testStruct.Account = nil
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				actual := &TestStructWithExtraField{}
@@ -194,6 +207,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 				ts := CreateTestStruct[T](0, tester)
 				tester.TriggerEvent(t, &ts)
@@ -212,6 +227,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 				ts1 := CreateTestStruct[T](2, tester)
 				tester.TriggerEvent(t, &ts1)
@@ -242,6 +259,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				result := &TestStruct{}
@@ -254,6 +273,8 @@ func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester Chain
 			test: func(t T) {
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 				ts0 := CreateTestStruct(0, tester)
 				tester.TriggerEvent(t, &ts0)
@@ -296,6 +317,8 @@ func runChainReaderBatchGetLatestValuesInterfaceTests[T TestingT[T]](t T, tester
 
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 				result, err := cr.BatchGetLatestValues(ctx, batchGetLatestValueRequest)
@@ -318,6 +341,8 @@ func runChainReaderBatchGetLatestValuesInterfaceTests[T TestingT[T]](t T, tester
 
 				ctx := tests.Context(t)
 				cr := tester.GetChainReader(t)
+				require.Error(t, ErrServiceNotStarted)
+				require.NoError(t, cr.Start(ctx))
 				require.NoError(t, cr.Bind(ctx, tester.GetBindings(t)))
 
 				result, err := cr.BatchGetLatestValues(ctx, batchGetLatestValuesRequest)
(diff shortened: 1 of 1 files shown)


pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -16,6 +16,8 @@ import (
 
 type ChainReaderInterfaceTester[T TestingT[T]] interface {
 	BasicTester[T]
+	StartChainReader(t T)
```
Farber98, line 19: Didn't find a way to make the tests pass without doing this breaking change. Open to ideas on how to prevent the multi repo PR!

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -340,16 +352,25 @@ type eventConfidencePair struct {
...
-	vals     []valConfidencePair
-	triggers []eventConfidencePair
-	stored   []TestStruct
+	vals        []valConfidencePair
+	triggers    []eventConfidencePair
+	stored      []TestStruct
 	batchStored BatchCallEntry
-	lock     sync.Mutex
+	lock        sync.Mutex
+	isStarted   bool
```
Farber98, line 360: this flag will be accessed within the methods to check if the service `isStarted`

pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -60,6 +62,15 @@ func RunChainReaderInterfaceTests[T TestingT[T]](t T, tester ChainReaderInterfac
 
 func runChainReaderGetLatestValueInterfaceTests[T TestingT[T]](t T, tester ChainReaderInterfaceTester[T]) {
 	tests := []testcase[T]{
+		{
```
Farber98, line 65: Highlights of the file:
- Call `StartChainReader` and `CloseChainReader` on every test that needs it. 
- Add error case for each of `GetLatestValue`, `BatchGetLatestValues`, and `QueryKey` when calling without starting the service previously.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -287,6 +287,18 @@ func (it *fakeChainReaderInterfaceTester) GetChainReader(_ *testing.T) types.Con
 	return it.impl
 }
 
+func (it *fakeChainReaderInterfaceTester) StartChainReader(t *testing.T) {
```
Farber98, line 290: Highlights of the file:

- Breaking change within `ChainReaderInterfaceTester` to add `StartChainReader` and `CloseChainReader`. 
- Add `isStarted` flag to check if the service `isStarted`.
- Add `fakeChainReader` implementations of `Start` and `Close`. Add `IsStarted` checks within `GetLatestValue`, `BatchGetLatestValues`, and `QueryKey`.

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -287,6 +287,18 @@ func (it *fakeChainReaderInterfaceTester) GetChainReader(_ *testing.T) types.Con
...
+	fake, ok := it.impl.(*fakeChainReader)
+	assert.True(t, ok)
+	require.NoError(t, fake.Start(context.Background()))
+}
+
+func (it *fakeChainReaderInterfaceTester) CloseChainReader(t *testing.T) {
```
jmank88, line 296: Could these be called just `Start` and `Close`? The type already says `ChainReader`
```suggestion
func (it *fakeChainReaderInterfaceTester) Start(t *testing.T) {
	fake, ok := it.impl.(*fakeChainReader)
	assert.True(t, ok)
	require.NoError(t, fake.Start(context.Background()))
}

func (it *fakeChainReaderInterfaceTester) Close(t *testing.T) {
```

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
jmank88: If it was possible for a method to be called before `Start`, then it must also be possible for a method to be called _during_ `Start`, so doesn't this field need to be safe for concurrent use?

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
Farber98: Thanks for this! Renamed here: [`8e1388b` (#705)](https://github.com/smartcontractkit/chainlink-common/pull/705/commits/8e1388bc2819bbd99eed8831e18286e9ad49860f)

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
Farber98: Thanks! addressed here: [`215fb70` (#705)](https://github.com/smartcontractkit/chainlink-common/pull/705/commits/215fb70b764ba3ff8fd0eee82682291ffb4cac5e)

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -383,6 +408,10 @@ func (f *fakeChainReader) SetBatchLatestValues(batchCallEntry BatchCallEntry) {
 }
 
 func (f *fakeChainReader) GetLatestValue(_ context.Context, contractName, method string, confidenceLevel primitives.ConfidenceLevel, params, returnVal any) error {
+	if !f.isStarted {
```
jmank88, line 411: These reads of `f.isStarted` need to be synchronized with the writes in `Start`/`Close`. How about using an atomic variable instead?

jmank88: https://pkg.go.dev/sync/atomic#Bool

Farber98: Thanks! Didn't know I could do this 😄

pkg/types/interfacetests/chain_reader_interface_tests.go
```diff
@@ -16,6 +16,8 @@ import (
 
 type ChainReaderInterfaceTester[T TestingT[T]] interface {
 	BasicTester[T]
+	Start(t T)
+	Close(t T)
```
silaslenihan, line 20: Why do we need this?

Farber98: Thanks for pointing this out. We just need `Close()`. 

The idea is to be able to `Close` the service before calling its methods and check for the error

ilija42 (dismissed):

jmank88 (dismissed):

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
```diff
@@ -287,6 +288,18 @@ func (it *fakeChainReaderInterfaceTester) GetChainReader(_ *testing.T) types.Con
...
+	fake, ok := it.impl.(*fakeChainReader)
+	assert.True(t, ok)
+	require.NoError(t, fake.Start(context.Background()))
+}
+
+func (it *fakeChainReaderInterfaceTester) Close(t *testing.T) {
+	fake, ok := it.impl.(*fakeChainReader)
+	assert.True(t, ok)
+	require.NoError(t, fake.Close())
+}
```
ilija42, line 301: Why do we need these, why not directly call Start and Close on the ChainReader returned from GetChainReader

Farber98: It fails when we run the LOOPs tests. It tries to execute [this](https://github.com/smartcontractkit/chainlink-common/blob/v0.2.1/pkg/loop/internal/goplugin/service.go#L34-L35) `Start()` func which is a no-op. Exposing these 2 funcs in the `InterfaceTester` was a workaround on this issue.

Let me know your thoughts if there's other way of achieving this without exposing these methods. I'm not very familiar with LOOPs yet and might be missing some context

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
```diff
@@ -311,6 +312,18 @@ func (it *fakeContractReaderInterfaceTester) GetChainWriter(_ *testing.T) types.
...
+	fake, ok := it.impl.(*fakeContractReader)
+	assert.True(t, ok)
+	require.NoError(t, fake.Start(context.Background()))
+}
+
+func (it *fakeContractReaderInterfaceTester) CloseContractReader(t *testing.T) {
+	fake, ok := it.impl.(*fakeContractReader)
+	assert.True(t, ok)
+	require.NoError(t, fake.Close())
+}
```
jmank88, line 325: I'm having trouble seeing why we need these helper methods. `impl` is already a `ContractReader`, so we don't need to cast it to a specific fake type:
```suggestion
func (it *fakeContractReaderInterfaceTester) StartContractReader(t *testing.T) {
	require.NoError(t, it.impl.Start(context.Background()))
}

func (it *fakeContractReaderInterfaceTester) CloseContractReader(t *testing.T) {
	require.NoError(t, it.impl.Close())
}
```
Additionally, there is already a `GetContractReader` method, so callers that are doing this:
```go
	tester.StartContractReader(t)
	defer tester.CloseContractReader(t)
```
Could instead do this:
```go
	tester.GetContractReader(t).Start(ctx)
	defer tester.GetContractReader(t).Close()
```
Or better yet:
```go
	servicetest.Run(t, tester.GetContractReader(t))
```

jmank88 (changes requested):
It looks like there are still some simplifications available

pkg/loop/internal/relayer/pluginprovider/chainreader/chain_reader_test.go
jmank88: Casting the interface to a specific struct type does not affect which method is called by name via `Start`/`Close`. Please see my other comment: https://github.com/smartcontractkit/chainlink-common/pull/705#discussion_r1758735194

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
Farber98: @jmank88 @silaslenihan  I couldn't do it this way. Tests fail for the loop relay client. 

Pushed some changes with a new idea:
- Modify `testCases` and `Setup` to receive a `startCR` flag and control whether we should start CR or not
- With this, `Start` and `Close` will be handled by `Setup` with `Cleanup`

All tests are passing, including new ones where we test that the service does not start. Let me know your thoughts

jmank88: Could you please update the description? I'm having trouble understanding the fundamental problem we are trying to solve.

jmank88: Since the `BasicTester` interface is used more generally than for just chain reader, I don't think it makes sense to add a chain reader only argument to `BasicTester.Setup`.

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
```diff
@@ -291,12 +292,16 @@ type fakeContractReaderInterfaceTester struct {
...
-func (it *fakeContractReaderInterfaceTester) Setup(_ *testing.T) {
+func (it *fakeContractReaderInterfaceTester) Setup(t *testing.T, started bool) {
 	fake, ok := it.impl.(*fakeContractReader)
 	if ok {
 		fake.vals = []valConfidencePair{}
 		fake.triggers = []eventConfidencePair{}
 		fake.stored = []TestStruct{}
+		if started {
+			require.NoError(t, it.impl.Start(context.Background()))
+			t.Cleanup(func() { require.NoError(t, it.impl.Close()) })
```
jmank88, line 303: Please use `servicetest.Run`. Notice it uses `assert` during cleanup.

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
jmank88: > We can't just call Start and Close directly on the return of GetContractReader to achieve this. This is because we also run relay client loop tests here. We are calling Start and Close on the client and not on the server as would be required.

> The two solutions I proposed were aimed at solving this problem, the one with the helper methods and now the one with the flag.

Something isn't adding up here. The modification that I proposed to your helper methods was intended to execute _exactly_ equivalently. It was not proposing a change in behavior - definitely not one that switched from client to server or vice versa. If the helper method solution was valid, then we should revisit it, but in the most simplified form.

pkg/types/interfacetests/chain_components_interface_tests.go
```diff
@@ -15,8 +15,28 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
 )
 
-type ChainComponentsInterfaceTester[T TestingT[T]] interface {
+type ChainComponentsTester[T any] interface {
```
Farber98, line 18: - defined `ChainComponentsTester` so we use the `Setup(t T, startCR bool)`  only for `chainComponents` tests
- defined `chainComponentsTestcase` so we use the `startCR` testcase flag only for `chainComponents` testcases
- defined `runChainComponentsTests` to execute `chainComponents` specific tests using the `ChainComponentsTester` interface with `Setup(t T, startCR bool)`

pkg/types/interfacetests/codec_interface_tests.go
```diff
@@ -19,8 +19,27 @@ type EncodeRequest struct {
 	TestOn       string
 }
 
+type codecTestcase[T any] struct {
```
Farber98, line 22: - defined CodecTester so we use the `Setup(t T)` without the `startCR` flag for the `codec`
- defined `codecTestcase` so we don't use the `startCR` testcase flag for `codec` testcases
- defined `runCodecTests` to execute `codec` specific tests using the `CodecTester` interface with `Setup(t T)`

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
```diff
@@ -291,12 +293,16 @@ type fakeContractReaderInterfaceTester struct {
...
-func (it *fakeContractReaderInterfaceTester) Setup(_ *testing.T) {
+func (it *fakeContractReaderInterfaceTester) Setup(t *testing.T, startCR bool) {
 	fake, ok := it.impl.(*fakeContractReader)
 	if ok {
 		fake.vals = []valConfidencePair{}
 		fake.triggers = []eventConfidencePair{}
 		fake.stored = []TestStruct{}
+
+		if startCR {
+			servicetest.Run(t, it.impl)
```
Farber98, line 304: added `servicetest.Run(t, it.impl)`

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
Farber98: Pushed a new solution with some comments on changes. If do you think it's okay we can go with it. Else, I'll revisit the one with the helper methods

ilija42 (changes requested):
Wdyt about this solution

https://github.com/smartcontractkit/chainlink-common/pull/777
https://github.com/smartcontractkit/chainlink/pull/14480

PR #711, opened 2024-08-19
Use lint action instead of our workflow

nolag:


pkg/loop/internal/relayer/pluginprovider/ext/ccip/offramp_test.go
```diff
@@ -24,8 +24,8 @@ func Test_OnchainConfig(t *testing.T) {
 
 	tests := []struct {
 		name          string
-		onchainConfig ccip.ExecOnchainConfig
-		want          ccippb.ExecOnchainConfig
+		onchainConfig ccip.ExecOnchainConfig   // nolint
+		want          ccippb.ExecOnchainConfig // nolint
```
patrickhuie19, line 28: what is the linter unhappy with here?

pkg/loop/internal/relayer/pluginprovider/ext/ccip/offramp_test.go
```diff
@@ -57,8 +57,8 @@ func Test_OnchainConfig(t *testing.T) {
 		},
 	}
 
-	for _, tt := range tests {
-		tt := tt
+	for _, tt := range tests { // nolint
+		tt := tt // nolint
```
patrickhuie19, line 61: what's going on here?

patrickhuie19 (approved):
LGTM, had a couple of non-blocking questions when you get a moment

.github/workflows/golangci_lint.yml
```diff
@@ -6,28 +6,22 @@ jobs:
...
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # v6.1.0
         with:
-          name: golangci-lint-report
-          path: |
-            ./golangci-lint-report.xml
\ No newline at end of file
+          version: v1.60.1
+          # only-new-issues is only applicable to PRs, otherwise it is always set to false
+          only-new-issues: false # disabled for PRs due to unreliability
```
jmank88, line 22: We should be able to remove this now

pkg/loop/internal/relayer/pluginprovider/ext/ccip/offramp_test.go
nolag: lint doesn't like that the proto is copied because protos have a lock.

pkg/loop/internal/relayer/pluginprovider/ext/ccip/offramp_test.go
nolag: lint doesn't like that the proto is copied because protos have a lock.

jmank88 (dismissed):

patrickhuie19 (dismissed):

patrickhuie19 (approved):

jmank88 (approved):

PR #712, opened 2024-08-19
[CAPPL-6] Formalize trigger API

cedric-cordenier:
### What

* Update the Trigger API so it doesn't rely on magic fields in a values.Map object.
* Merge Payload and Metadata into a single Outputs field and make it a map. This is important for compatibility with other parts of the API.
* Update the Mercury Trigger and the feeds aggregator so it understands the new message format.


pkg/capabilities/capabilities.go
```diff
@@ -127,9 +127,20 @@ type BaseCapability interface {
 	Info(ctx context.Context) (CapabilityInfo, error)
 }
 
+type TriggerRegistrationRequest struct {
+	TriggerID string
+	Metadata  RequestMetadata
+	Config    *values.Map
+}
+
+type TriggerRegistrationResponse struct {
```
justinkaseman, line 136: The name `TriggerRegistrationRequest` make sense, because it only registers once per trigger, but would response be `TriggerRegistrationResponse` or `TriggerResponse`? Since it can respond multiple times.

pkg/capabilities/capabilities.go
```diff
@@ -127,9 +127,20 @@ type BaseCapability interface {
 	Info(ctx context.Context) (CapabilityInfo, error)
 }
 
+type TriggerRegistrationRequest struct {
+	TriggerID string
```
justinkaseman, line 131: If this is made unique per workflow by the engine, could we indicate it with a doc comment?

pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -99,12 +99,7 @@ func (o *MercuryTriggerService) RegisterTrigger(ctx context.Context, req capabil
 		return nil, err
 	}
 
-	inputs, err := o.ValidateInputs(req.Inputs)
-	if err != nil {
-		return nil, err
-	}
-
-	triggerID := o.getTriggerID(inputs.TriggerID, wid)
+	triggerID := o.getTriggerID(req.TriggerID, wid)
```
justinkaseman, line 102: If this is already unique by workflow, could this just be
```suggestion
	triggerID := req.TriggerID
```

pkg/capabilities/capabilities.go
cedric-cordenier: Good point, I think TriggerResponse makes sense 👍

pkg/capabilities/triggers/mercury_remote_aggregator.go
```diff
@@ -34,21 +34,16 @@ func NewMercuryRemoteAggregator(codec datastreams.ReportCodec, allowedSigners []
...
 			a.lggr.Errorw("could not unmarshal one of capability responses (faulty sender?)", "error", err)
 			continue
 		}
-		triggerEvent := &capabilities.TriggerEvent{}
-		if err = unmarshaled.Value.UnwrapTo(triggerEvent); err != nil {
-			a.lggr.Errorw("could not unwrap one of trigger events", "error", err)
-			continue
-		}
-		feedReports, err := a.codec.Unwrap(triggerEvent.Payload)
+		feedReports, err := a.codec.Unwrap(unmarshaled.Event.Outputs)
```
cedric-cordenier, line 46: TODO

pkg/capabilities/capabilities.go
```diff
@@ -128,9 +128,23 @@ type BaseCapability interface {
 	Info(ctx context.Context) (CapabilityInfo, error)
 }
 
+type TriggerRegistrationRequest struct {
+	// TriggerID uniquely identifies the trigger by concatenating
+	// the workflow ID and the trigger's index in the spec.
+	TriggerID string
```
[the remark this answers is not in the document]
cedric-cordenier: Are you proposing to replace TriggerID with CapabilityIdx? They aren't totally interchangeable: the TriggerID also contains the workflowID for example. I'm a little reluctant to add CapabilityIdx generally to the RequestMetadata since I don't really understand what it would be used for. The fact that it's used to generate the TriggerID is an implementation detail, not something we should be reverse engineering.

pkg/capabilities/datastreams/types.go
```diff
@@ -73,6 +73,12 @@ type SignersMetadata struct {
 	MinRequiredSignatures int
 }
 
+type StreamsTriggerPayload struct {
```
bolekk, line 76: "Payload" in the name is a little confusing. How about "StreamsTriggerEvent"? The line of thinking being Event=Payload+Metadata.

pkg/capabilities/capabilities.go
```diff
@@ -83,12 +83,12 @@ type CapabilityRequest struct {
...
 	TriggerType string
-	ID          string
-	Timestamp   string
-	// Trigger-specific payload+metadata
-	Metadata values.Value
-	Payload  values.Value
+	// The ID of the trigger event
+	ID string
+	// Trigger-specific payload
+	Outputs *values.Map
```
bolekk, line 91: What's the benefit of this change? You're adding one more layer of nesting: previously it was Event->Payload and now it is Event->Outputs->Payload. In the PR description you say "This is important for compatibility with other parts of the API." - can you explain it please?

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -252,7 +247,7 @@ func (a *dataFeedsAggregator) extractSignersAndPayloads(observations map[ocrcomm
 			signers[signer]++
 		}
 		mins[meta.MinRequiredSignatures]++
-		payloads[nodeID] = triggerEvent.Payload
+		payloads[nodeID] = nodeObservations[0]
```
bolekk, line 250: nit: again the name "payload" is a little misleading to me now (see my other comment). After your change we're not saving payloads any more, we are saving whole events.

pkg/capabilities/datastreams/types.go
```diff
@@ -73,6 +73,12 @@ type SignersMetadata struct {
...
+	Payload   []FeedReport
+	Metadata  SignersMetadata
+	Timestamp int64
+}
+
 type ReportCodec interface {
 	// unwrap reports and convert to a list of FeedReport
 	Unwrap(wrapped values.Value) ([]FeedReport, error)
```
bolekk, line 84: This threw me off initially when reading the rest of the code. You are silently changing the expectation on the Codec to Unwrap the whole Event, instead of just the payload. This is OK but I wonder if we can document it somehow. Maybe change the "wrapped" param name to "wrappedTriggerEvent"? Or technically "wrappedTriggerEventOutputs", which doesn't sound great :/

pkg/capabilities/capabilities.go
cedric-cordenier: Sure: for interpolation to work correctly we need to be passing a single values.Map around from step to step; we previously worked around this by wrapping this TriggerEvent in a map and then shipping this to the next step.

However, the TriggerType + ID should really be part of the typed API, which would have meant that we'd have to combine the Metadata and Payload before passing it to the next step (and handling cases where they aren't maps too).

I don't think we've actually changed the amount of nesting with this change:
* previously we had CapabilityResponse -> Value -> (unwrap TriggerEvent) -> (unwrap Payload) -> []FeedReport 
* now we have TriggerResponse -> Event -> Outputs -> (unwrap StreamsTriggerPayload) -> []FeedReport

This will largely not matter for capabilities -- they'll receive response.Event.Outputs directly. BTW we could reduce the nesting by one level by making the TriggerResponse:

```
type TriggerResponse struct {
    TriggerType string
     ID string
     Outputs *values.Map
     Err error
}
```

but that feels a bit less clean to me? Not opposed to it though

pkg/capabilities/datastreams/types.go
cedric-cordenier: I'll add some comments

pkg/capabilities/capabilities.go
cedric-cordenier: Yes can do 👍 @bolekk Are you happy with that?

bolekk: sorry, what's the proposal?

bolekk: After chatting offline I'm OK with keeping stuff as-is.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
bolekk, line 250: nit: you could still rename this "payload" map

bolekk (dismissed):

DeividasK (approved):

ettec (approved):

PR #713, opened 2024-08-19
Add reference id to capability request

nolag:


pkg/capabilities/pb/capabilities.proto
```diff
@@ -35,6 +35,7 @@ message RequestMetadata {
   string workflow_name = 4;
   uint32 workflow_don_id = 6;
   uint32 workflow_don_config_version = 7;
+  optional string reference_id = 8;
```
cedric-cordenier, line 38: @nolag Can we leave this as non-optional? This should still maintain backwards compatiblity if not provided (it'll default to ""), and in any case I assume this will be followed up shortly with a PR in the engine to pass the value in all the time

nolag: Ah, I didn't realize the change from proto 2.  Will change this now.

DeividasK (approved):

PR #715, opened 2024-08-20
(feat): Cron Trigger Service Capability

justinkaseman:
### What
Adds a new trigger capability for scheduling workflows via a Cron schedule.


pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	// actually uses cronv3 internally for its
+	// parser to turn cron strings into cron schedules
+	"github.com/go-co-op/gocron/v2"
```
justinkaseman, line 14: TODO: narrow down to a single cron package

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+func (cts *CronTriggerService) RegisterTrigger(ctx context.Context, req capabilities.CapabilityRequest) (<-chan capabilities.CapabilityResponse, error) {
+	cts.mu.Lock()
+	defer cts.mu.Unlock()
+
+	config, err := cts.ValidateConfig(req.Config)
+	if err != nil {
+		return nil, err
+	}
+
+	triggerID := getTriggerID(config.TriggerID, req.Metadata.WorkflowID)
```
cedric-cordenier, line 88: Reminder that this has already been done by the engine

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+	if _, ok := cts.jobs[triggerID]; ok {
+		return nil, fmt.Errorf("triggerId %s already registered", triggerID)
+	}
+
+	job := gocron.CronJob(config.Schedule, true)
+
+	cronParser := cron.NewParser(
+		cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow,
+	)
```
cedric-cordenier, line 97: I would just use cron.NewParser(cron.WithSeconds) here so that seconds are required all the time. This should make it simpler to validate schedules upstream if we need to and will accommodate smaller granularities

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	cts.lggr.Debugw("RegisterTrigger", "triggerId", triggerID, "jobId", j.ID())
+	return ch, nil
+}
+
+func (cts *CronTriggerService) makeProcess(triggerId string, schedule cron.Schedule) func() {
+	return func() { cts.process(triggerId, schedule) }
+}
+
+func (cts *CronTriggerService) process(triggerId string, schedule cron.Schedule) {
```
cedric-cordenier, line 129: You could pass the callback channel here which should remove any need for logic around locking or checking that the channel exists (this isn't actually done in the implementation below but should be)

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+		} else {
+			break
+		}
+	}
+
+	// timestamp in ns precision, to show the discrepancy between the scheduled time and the actual time
+	// we need scheduled time to generate a deterministic triggerEventId
+	timestampNs := time.Now().UTC().UnixNano()
+	// format to ISO 8601
+	timestampMsIso8601 := formatUnixNanoToISO8601(timestampNs)
```
cedric-cordenier, line 149: This feels overly cumbersome: can we just do `time.Now().Format(time.RFC3339Nano)`?

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+	return wid + "|" + triggerID
+}
+
+func wrapCronTriggerEvent(eventID string, timestamp string, scheduledExecutionTime string) (capabilities.CapabilityResponse, error) {
+	payload, err := values.Wrap(CronTriggerPayload{ScheduledExecutionTime: scheduledExecutionTime})
+	if err != nil {
+		return capabilities.CapabilityResponse{}, err
+	}
+
+	metadata, err := values.Wrap(nil)
```
cedric-cordenier, line 259: Nit this will just return nil, so no need to do anything here I think

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+
+	triggerEvent := capabilities.TriggerEvent{
+		TriggerType: cronTriggerID,
+		ID:          eventID,
+		Timestamp:   timestamp,
+		Metadata:    metadata,
+		Payload:     payload,
+	}
+
+	eventVal, err := values.Wrap(triggerEvent)
```
cedric-cordenier, line 272: WrapMap will return a map or error, this would avoid the need to coerce on l 278

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+	cts.wg.Add(1)
+
+	if cts.scheduler == nil {
+		return errors.New("no scheduler initialized")
+	}
+	cts.scheduler.Start()
+
+	cts.lggr.Info(cts.Name() + " started")
+
+	// block until ready to shut down
```
cedric-cordenier, line 207: This is very confusing: Start() shouldn't block, and it's not clear what this goroutine is doing; I'm not even sure it's needed, so I'd suggest removing

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+// again, new Service will need to be built to do so.
+func (cts *CronTriggerService) Close() error {
+	err := cts.scheduler.Shutdown()
+	if err != nil {
+		return fmt.Errorf("scheduler shutdown encountered a problem: %s", err)
+	}
+
+	close(cts.stopCh)
+
+	cts.wg.Wait()
```
cedric-cordenier, line 227: This two bits seem unnecessary: there are no goroutines that are managed directly by this service, and all we do is start the scheduler in Start(); it feels like all we should do here is stop it.

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+
+	cts.lggr.Info(cts.Name() + " closed")
+
+	return nil
+}
+
+func (cts *CronTriggerService) Ready() error {
+	if cts.scheduler == nil {
```
cedric-cordenier, line 235: This also seems unnecessary: the scheduler is instantiated in the constructor, so this would always be the case.

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+func (cts *CronTriggerService) makeProcess(triggerId string, schedule cron.Schedule) func() {
+	return func() { cts.process(triggerId, schedule) }
+}
+
+func (cts *CronTriggerService) process(triggerId string, schedule cron.Schedule) {
+	cts.mu.Lock()
+	defer cts.mu.Unlock()
+
+	now := time.Now()
+	scheduledExecutionTime := schedule.Next(now)
```
cedric-cordenier, line 134: You might be able to simplify this log by passing the job into the task closure and then calling j.LastRun()

cedric-cordenier: See here for an example: https://go.dev/play/p/EdyoEzZBihj

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,287 @@
...
+	ch := make(chan capabilities.CapabilityResponse, defaultSendChannelBufferSize)
+	cts.jobs[triggerID] = cronJob{
+		ch:  ch,
+		job: j,
+	}
+	cts.lggr.Debugw("RegisterTrigger", "triggerId", triggerID, "jobId", j.ID())
+	return ch, nil
+}
+
+func (cts *CronTriggerService) makeProcess(triggerId string, schedule cron.Schedule) func() {
```
cedric-cordenier, line 125: This can be simplified to just passing in cts.process to NewTask

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Good tip!

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Ah, NewTask has parameters, yes way cleaner.

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Gets simplified with the Trigger interface using a direct timestamp

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: I like this. Good suggestion.

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Not sure that I follow, because `cron.NewParser(cron.WithSeconds)` doesn't seem to be valid syntax --- but it doesn't matter because with all the cleanup this has gone away.

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Way nicer!

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,221 @@
...
+	mu        sync.Mutex
+	scheduler gocron.Scheduler
+	triggers  map[string]cronTrigger
+}
+
+var _ capabilities.TriggerCapability = (*CronTriggerService)(nil)
+var _ services.Service = &CronTriggerService{}
+
+func NewCronTriggerService(lggr logger.Logger) *CronTriggerService {
+	location, _ := time.LoadLocation("UTC")
```
cedric-cordenier, line 66: You can use time.UTC rather than loading it, this will avoid needing to handle the error

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,221 @@
...
+func (cts *CronTriggerService) HealthReport() map[string]error {
+	return map[string]error{cts.Name(): nil}
+}
+
+func (cts *CronTriggerService) Name() string {
+	return "CronTriggerService"
+}
+
+func wrapCronTriggerEvent(eventID string, timestamp time.Time, scheduledExecutionTime string) (capabilities.TriggerRegistrationResponse, error) {
+	payload, err := values.Wrap(CronTriggerPayload{ScheduledExecutionTime: scheduledExecutionTime})
```
cedric-cordenier, line 206: @justinkaseman Have you tested that this works? I don't think we have a time.Time type in the values library

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,221 @@
...
+
+	var job gocron.Job
+	callbackCh := make(chan capabilities.TriggerRegistrationResponse, defaultSendChannelBufferSize)
+
+	allowSeconds := true
+	jobDef := gocron.CronJob(config.Schedule, allowSeconds)
+
+	task := gocron.NewTask(func() {
+		// use scheduled time to generate a deterministic triggerEventId
+		scheduledExecutionTime, errr := job.LastRun()
```
cedric-cordenier, line 100: @justinkaseman I don't know if this will quite work for larger granularities of cron schedules; this is probably worth a test

For example: we need to verify that if we have a CRON schedule of every 5 minutes that the scheduled execution time will be 00:05:00/00:10:00/00:15:00 for all nodes, and not:

00:05:01 for one node
00:05:02 for another
and so on

Lmk if that makes sense?

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,221 @@
...
+			cts.lggr.Errorw("process could not get last job run", "err", errr)
+		}
+		// format to use smallest time unit of a second
+		scheduledExecutionTimeFormatted := scheduledExecutionTime.Format(time.RFC3339)
+		hash := sha256.Sum256([]byte(scheduledExecutionTimeFormatted))
+		triggerEventID := hex.EncodeToString(hash[:])
+
+		currentTime := time.Now()
+
+		// cts.lggr.Debugw
```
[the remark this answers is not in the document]
justinkaseman: Eagle eyes. Thanks.

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Makes sense. I'll add a long running test

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: `scheduledExecutionTime` probably should have been named `scheduledExecutionTimeString`

Collapsed `wrapCronTriggerEvent` away since it's fairly small now. Hopefully clears makes it more clear.

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: Added some tests, see if you think they sufficiently cover what you are talking about

pkg/capabilities/triggers/cron_trigger.go
justinkaseman: TODO

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,226 @@
...
+	callbackCh := make(chan capabilities.TriggerRegistrationResponse, defaultSendChannelBufferSize)
+
+	allowSeconds := true
+	jobDef := gocron.CronJob(config.Schedule, allowSeconds)
+
+	task := gocron.NewTask(func() {
+		// Use the scheduled execution time as a deterministic identifier
+		// .LastRun() is the run that triggered this task run
+		scheduledExecutionTime, errr := job.LastRun()
+		if errr != nil {
```
justinkaseman, line 112: Return error here as a capability reponse

pkg/capabilities/triggers/cron_trigger.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+var _ capabilities.TriggerCapability = (*CronTriggerService)(nil)
+var _ services.Service = &CronTriggerService{}
+
+func NewCronTriggerService(lggr logger.Logger, clock clockwork.Clock) *CronTriggerService {
+	var options []gocron.SchedulerOption
+	// Set scheduler location to UTC for consistency across nodes
+	options = append(options, gocron.WithLocation(time.UTC))
+	// options = append(options, gocron.WithLogger(gocron.NewLogger(gocron.LogLevelInfo))) // TODO: do we want this?
```
justinkaseman, line 71: Turn this on

go.mod
```diff
@@ -47,6 +48,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require github.com/robfig/cron/v3 v3.0.1 // indirect
+
 require (
```
jmank88, line 53:
```suggestion
require (
	github.com/robfig/cron/v3 v3.0.1 // indirect
```

pkg/workflows/utils.go
```diff
@@ -0,0 +1,21 @@
+package workflows
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+)
+
+func GenerateExecutionID(workflowID, eventID string) (string, error) {
```
jmank88, line 8: nit/ `Generate` feels like a strong word for hash + hex encode. What about something like `Encode`?
```suggestion
func EncodeExecutionID(workflowID, eventID string) (string, error) {
```
Also, you can probably eliminate the error from the signature if you wanted to, since the hash impl used does not return errors and is just satisfying the interface.

jmank88: For comparison:
```suggestion
func NewExecutionID(workflowID, eventID string) string {
```
```suggestion
func EncodeExecutionID(workflowID, eventID string) string {
```

jmank88:
```suggestion
func DeriveExecutionID(workflowID, eventID string) string {
```

justinkaseman:
Superceded by Capabilities PR

PR #717, opened 2024-08-20
compute helpers

nolag:


pkg/workflows/compute_generated.go
jmank88, line 1: The file name says generated - is this missing a header?

pkg/workflows/sdk.go
```diff
@@ -0,0 +1,5 @@
+package workflows
+
+type Sdk interface {
```
jmank88, line 3: Should this be upper-case per go naming?
```suggestion
type SDK interface {
```

pkg/utils/codegeneration/utils.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"golang.org/x/tools/imports"
+)
+
+func PrintFiles(dir string, files map[string]string) error {
```
jmank88, line 12: How about `Write` instead of `Print`?
```suggestion
// WriteFiles writes the map of files to dir and runs goimports on all *.go files.
func WriteFiles(dir string, files map[string]string) error {
```

pkg/utils/codegeneration/utils.go
```diff
@@ -0,0 +1,39 @@
...
+		if !strings.HasPrefix(file, dir) {
+			file = dir + "/" + file
+		}
+
+		if strings.HasSuffix(file, ".go") {
+			imports.LocalPrefix = "github.com/smartcontractkit"
+			rawContent, err := imports.Process(file, []byte(content), nil)
+			if err != nil {
+				// print an error, but also write the file so debugging the generator isn't a pain.
+				fmt.Printf("Error formatting file %s: %s\n", file, err)
```
jmank88, line 23: Always writing makes sense, but is just printing the error sufficient? Or should it also be returned?

pkg/utils/codegeneration/utils.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func PrintFiles(dir string, files map[string]string) error {
+	for file, content := range files {
+		if !strings.HasPrefix(file, dir) {
+			file = dir + "/" + file
+		}
+
+		if strings.HasSuffix(file, ".go") {
+			imports.LocalPrefix = "github.com/smartcontractkit"
```
jmank88, line 19: Is this embedded value too opinionated for an exported API?

pkg/utils/codegeneration/utils.go
nolag: I didn't want to return errors there because it might prevent other files from being written.

nolag: fair, I was thinking that the generator probably shouldn't stop other work, but this is the last step anyways.  Also, callers can do what they want with the errors anyway.

pkg/utils/codegen/utils.go
```diff
@@ -0,0 +1,44 @@
...
+		} else if err = os.WriteFile(file, []byte(content), 0600); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if len(errs) == 0 {
+		return nil
+	}
+
+	return errors.Join(errs...)
```
jmank88, line 43: `errors.Join` is smart enough that you can just accumulate a regular `error` as you go:
```go
var errs error
```
```suggestion
			errs = errors.Join(errs, err)
		}
	}

	return errs
```

pkg/workflows/gen/main.go
```diff
@@ -0,0 +1,40 @@
...
+	"bytes"
+	_ "embed"
+	"log"
+	"text/template"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/codegen"
+)
+
+//go:embed compute.go.templ
+var computeGo string
```
bolekk, line 13: Would you mind writing down instructions on how to use this generator? (either in a comment here or in a README file). This file embeds compute.go.templ but the comment on top of compute.go.templ says it was generated by this script..? I'm confused.

nolag: The script has that line so that the generated code will all have it.  I did that for the other code generators templates as well.

pkg/workflows/compute_test.go
```diff
@@ -0,0 +1,204 @@
...
+					ID:             "notstreams@1.0.0",
+					Ref:            "trigger",
+					Inputs:         workflows.StepInputs{},
+					Config:         map[string]any{"maxFrequencyMs": 5000},
+					CapabilityType: capabilities.CapabilityTypeTrigger,
+				},
+			},
+			Actions: []workflows.StepDefinition{
+				{
+					ID:  "internal!!custom_compute@1.0.0",
```
nolag, line 51: Making this more visible, to keystone team members, what's a good name to take that won't be in the real registry?  Is the prefix ok to use?

bolekk: Maybe `__internal__`?

pkg/workflows/compute_test.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func createWorkflow(fn func(_ workflows.SDK, inputFeed notstreams.Feed) ([][]streams.Feed, error)) *workflows.WorkflowSpecFactory {
+	workflow := workflows.NewWorkflowSpecFactory(workflows.NewWorkflowParams{
+		Owner: "owner",
+		Name:  "name",
+	})
+
+	trigger := notstreams.TriggerConfig{MaxFrequencyMs: 5000}.New(workflow)
+	computed := workflows.Compute1(workflow, "Compute", workflows.Compute1Inputs[notstreams.Feed]{
```
bolekk, line 158: Why have ComputeN as a generated name instead of having a single Compute with N as a parameter?  Won't that cause problem for the Engine? Can it build a function name in runtime?

nolag: A parameter where? The engine won't care it just maps the inputs form what the workflow spec has as inputs.  It has no idea if

Compute { Arg0: b, Arg1: d} is two objects or one inside the compute.

bolekk (approved):

pkg/workflows/compute_test.go
bolekk: I misunderstood - thought that N references the N-th compute step in a workflow.

bolekk (dismissed):

pkg/workflows/gen/main.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"bytes"
+	_ "embed"
+	"log"
+	"text/template"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/codegen"
+)
+
+//go:embed compute.go.templ
```
jmank88, line 12: I've never seen `.templ` before :shrug: We typically use `.tmpl`:
```suggestion
//go:embed compute.go.tmpl
```

bolekk (approved):

jmank88 (approved):

PR #719, opened 2024-08-20
Fix a bug with big.Int and values.Value, and allow same backing type to be encoded and decoded

nolag:
Same backing types are used to ensure marshalling, for example json shcema can generate code like

type FeedId string

`func (f *FeedId) UnmarshalJson(...)... `

that verifies the feed id has the right length etc.

### DIFF
--- a/pkg/values/big_int.go
+++ b/pkg/values/big_int.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"reflect"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/values/pb"
 )
@@ -45,6 +46,10 @@ func (b *BigInt) UnwrapTo(to any) error {
 		}
 		*tb = b.Underlying
 	default:
+		rto := reflect.ValueOf(to)
+		if rto.CanConvert(reflect.TypeOf(new(big.Int))) {
+			return b.UnwrapTo(rto.Convert(reflect.TypeOf(new(big.Int))).Interface())
+		}
 		return fmt.Errorf("cannot unwrap to value of type: %T", to)
 	}
 
--- a/pkg/values/int.go
+++ b/pkg/values/int.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"reflect"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/values/pb"
 )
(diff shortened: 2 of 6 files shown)


pkg/values/big_int.go
```diff
@@ -45,6 +46,10 @@ func (b *BigInt) UnwrapTo(to any) error {
 		}
 		*tb = b.Underlying
 	default:
+		rto := reflect.ValueOf(to)
+		if rto.CanConvert(reflect.TypeOf(new(big.Int))) {
```
krehermann, line 50: what is special about big.Int in the existing system? why aren't there other branches in this `if`?

bolekk (approved):

pkg/values/map.go
```diff
@@ -185,7 +185,9 @@ func unwrapsValues(f reflect.Type, t reflect.Type, data any) (any, error) {
 		n := reflect.New(t).Interface()
 		err := dv.UnwrapTo(n)
 		if err != nil {
-			return nil, err
+			// Do not return the error to allow mapstructure to retry with different types
```
krehermann, line 188: what happens when those retries fail? is the ultimate error swallowed or is it returned and intelligible?

nolag: Then mapstructure will try to map it with the type that it has and would fail anyways.

pkg/values/value.go
```diff
@@ -95,11 +102,17 @@ func Wrap(v any) (Value, error) {
...
 		if reflect.Indirect(reflect.ValueOf(v)).Kind() == reflect.Struct {
-			return createMapFromStruct(reflect.Indirect(reflect.ValueOf(v)).Interface())
+			return CreateMapFromStruct(reflect.Indirect(reflect.ValueOf(v)).Interface())
 		}
+	case reflect.String:
+		return Wrap(val.Convert(reflect.TypeOf("")).Interface())
+	case reflect.Bool:
+		return Wrap(val.Convert(reflect.TypeOf(true)).Interface())
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return Wrap(val.Convert(reflect.TypeOf(int64(0))).Interface())
```
krehermann, line 115: why does this work? in other places we have an explicit case for all the int type variations

pkg/values/value.go
```diff
@@ -251,7 +265,15 @@ func unwrapTo[T any](underlying T, to any) error {
 		}
 		*tb = underlying
 	default:
-		return fmt.Errorf("cannot unwrap to value of type: %T", to)
+		// Don't break for custom types that are the same underlying type
+		// eg: type FeedId string allows verification of FeedId's shape while unmarshalling
```
krehermann, line 269: nit add quotes to the embedded code example:
`type FeedId string`

pkg/values/value.go
nolag: Two reasons,

1 wrapping doesn't need to check boundaries (decoding breaks it down).  And in reflection you can do the equivalent of  (Int64)(int_value) for any int value in the same line of code, but without reflection you can't.

pkg/values/value.go
nolag: Can you suggest it, then I won't need re-approval from Bolek if I accept the change :D

pkg/values/int.go
```diff
@@ -90,5 +91,23 @@ func (i *Int64) UnwrapTo(to any) error {
 		return nil
 	}
 
+	rv := reflect.ValueOf(to)
+	if rv.Kind() == reflect.Ptr {
```
krehermann, line 95: i don't see any obvious new tests for this. is it covered?

pkg/values/big_int.go
nolag: This is in the class for *big.Int, if you can cast to a big int then you're a big int. If you can't then it's a problem.

pkg/values/int.go
nolag: yeah, the test for the equivalent type should cover this because it uses int.

pkg/values/big_int.go
krehermann: ah, thank, i missed that

krehermann (approved):

PR #720, opened 2024-08-21
Changes to support passing in gas limit to chain write on SubmitTransaction

ettec:


pkg/values/pb/values.pb.go
```diff
@@ -2,7 +2,7 @@
 // versions:
 // 	protoc-gen-go v1.31.0
 // 	protoc        v4.25.1
-// source: values.proto
```
cedric-cordenier, line 5: @ettec Can you remove changes to this file? It's not relevant to the change you're making here

cedric-cordenier (dismissed):

pkg/types/chain_writer.go
```diff
@@ -14,7 +14,8 @@ type ChainWriter interface {
 	//
 	// - `args` should be any object which maps a set of method param into the contract and method specific method params.
 	// - `transactionID` will be used by the underlying TXM as an idempotency key, and unique reference to track transaction attempts.
-	SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, meta *TxMeta, value *big.Int) error
+	SubmitTransaction(ctx context.Context, contractName, method string, args any, transactionID string, toAddress string, meta *TxMeta, value *big.Int,
+		gasLimit uint64) error
```
jmank88, line 18: Is `gas` a universal term? Is it always required?

jmank88:
This kind of breaking change will have cascading effects across many repos. Please open and link any corresponding PRs for review.

ilija42 (changes requested):
What is the reasoning behind this? Where will the gasLimit value come from, will we build a chain agnostic gas provider? Does passing this in even make sense in chain agnostic context, seems like a very EVM specific thing?
I understand if this is a blocking thing for keystone, but someone will have to do proper research on how this interface would work with all the different chains that Chain Writet will have to support

pkg/types/chain_writer.go
```diff
@@ -27,6 +27,7 @@ type ChainWriter interface {
 type TxMeta struct {
 	// Used for Keystone Workflows
 	WorkflowExecutionID *string
+	GasLimit            *big.Int
```
jmank88, line 30: Let's document this well. In particular, must every chain implement this feature? If not, what happens when it is not supported?

pkg/loop/internal/relayer/pluginprovider/chainwriter/converters.go
jmank88, line 1: Are there tests for these that should be updated to set this field?

pkg/types/chain_writer.go
ettec: sorted

pkg/loop/internal/relayer/pluginprovider/chainwriter/converters.go
ettec: sorted

pkg/types/chain_writer.go
```diff
@@ -27,6 +27,10 @@ type ChainWriter interface {
 type TxMeta struct {
 	// Used for Keystone Workflows
 	WorkflowExecutionID *string
+	// An optional maximum gas limit for the transaction.  It is at the discretion of the chainwriter implementation as
+	// to whether or not to respect this value.  If not set, the chainwriter implementation will be responsible for
+	// setting a gas limit for the transaction.
```
jmank88, line 32: Why is it OK/better to ignore a non-nil, as opposed to returning an error?

ettec: Good question, I suppose if the chainwriter expects this value it could error rather than default when not set, though I guess this would be a breaking change then.   @cedric-cordenier @DeividasK any opinion on this?

jmank88: I mean that in the scenario when a caller sets a `gasLimit`, but the implementation does not support it, then it must either:
1) ignore it
2) return an error

IMHO (2) is safer and more flexible, because it leaves it up to the caller to decide whether dropping the value is acceptable or not. They can choose to submit the request against with a nil value, or to propagate the returned error if having a custom `gasLimit` is required for that situation.

patrick-dowell: I think we'd like to ideally not break CCIP's integration with this change. Considering gasLimit wasn't specified in the TxMeta prevously and was done via the ChainWriterConfig, should we consider supporting that pattern - as in, if the gasLimit isn't specified in the TxMeta, we default to the config value?

ettec: I've updated the comment and added a new error type as per discussion.

So expected behaviour is that if `gasLimit` is specified and the ChainWriter implementation does not support per transaction gasLimit it will return a `ErrSettingTransactionGasLimitNotSupported` error, this allows the caller to gracefully handle this scenario as per Jordan's suggestion.  In the case where it is not set the ChainWriter implementation will do what it does right now, i.e. default it, so this change will not be breaking (for CCIP).

I'm going to update the evm chainwriter implementation to match the expected behaviour as part of the item fix.  Implementers of other chain writers will need notifying of this change, @patrick-dowell is that anyone other than you?

cedric-cordenier (approved):

pkg/loop/internal/relayer/pluginprovider/chainwriter/chain_writer.go
```diff
@@ -122,8 +122,9 @@ func WithServerEncoding(version chainreader.EncodingVersion) ServerOpt {
 
 func (s *Server) SubmitTransaction(ctx context.Context, req *pb.SubmitTransactionRequest) (*emptypb.Empty, error) {
 	err := s.impl.SubmitTransaction(ctx, req.ContractName, req.Method, req.Params, req.TransactionId, req.ToAddress, TxMetaFromProto(req.Meta), req.Value.Int())
+
 	if err != nil {
-		return nil, err
+		return nil, net.WrapRPCErr(err)
```
jmank88, line 127: This is only a client-side wrapper:
```suggestion
		return nil, err
```

pkg/types/chain_writer.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 )
 
+const (
+	ErrSettingTransactionGasLimitNotSupported = UnimplementedError("setting transaction gas limit is not supported")
```
jmank88, line 11: What are the differences between `Unimplemented` and `InvalidArgument`?
```suggestion
	ErrSettingTransactionGasLimitNotSupported = InvalidArgumentError("setting transaction gas limit is not supported")
```

pkg/loop/internal/relayer/pluginprovider/chainwriter/chain_writer.go
ettec: sorted

pkg/types/chain_writer.go
ettec: was 50/50 on this, seemed like either could apply, I've update to InvalidArgumentError

ilija42 (approved):

bolekk (approved):

jmank88 (approved):

PR #726, opened 2024-08-21
Fix SignedReport type in json schema

nolag:


pkg/capabilities/consensus/ocr3/ocr3_consensus-schema.json
```diff
@@ -5,25 +5,28 @@
     "SignedReport": {
       "type": "object",
       "properties": {
-        "WorkflowExecutionID": {
-          "type": "string"
+        "Report": {
+          "type": "string",
+          "contentEncoding": "base64"
```
nolag, line 10: We still need to figure out what to do longer term so that we can use this type for validation, otherwise it'll expect strings instead of the byte array.

cedric-cordenier (approved):

DeividasK (approved):

PR #733, opened 2024-08-26
feat(tooling-apis): adds job-distributor API

MStreet3:
## What
This PR adds protobuf definitions and a generated Go SDK for two main tooling applications: `job-distributor` and `orchestrator`.  

The `job-distributor` app is composed of four services defined as protobufs and is a generic distribution service for sending jobs to connected nodes.

The `orchestrator` app defines the contract between a node and a distributor/proposer.  This app is split into two protobuf services.

## Why
This is going into `chainlink-common` so that apps that must build clients on top of the `job-distributor` (or build an implementation of the service) have a single public reference point for the contract.  The `orchestrator` service definitions are centralized here as well because they are shared in at least three different repos.

## Details
* Adds the protobuf definitions for `job-distributor` and `orchestrator` services
* Adds generated Go SDKs for services
* Adds code generation tasks


AnieeG (approved):

krehermann (approved):

tooling-apis/job-distributor/job/job.proto
```diff
@@ -0,0 +1,206 @@
...
+package tooling.distributor.job;
+
+import "google/protobuf/timestamp.proto";
+import "job-distributor/shared/ptypes/label.proto";
+
+option go_package = "job-distributor/job";
+
+// JobService provides RPCs for managing jobs and job proposals within the
+// system.
+service JobService {
```
patrickhuie19, line 12: What are the lifecycles of Jobs and Job proposals? Should they have full CRUDL handler implementations?

tooling-apis/job-distributor/node/node.proto
```diff
@@ -0,0 +1,217 @@
...
+
+  // ListNodes returns a list of nodes, optionally filtered by the provided
+  // criteria.
+  rpc ListNodes(ListNodesRequest) returns (ListNodesResponse) {}
+
+  rpc ListNodeChainConfigs(ListNodeChainConfigsRequest)
+      returns (ListNodeChainConfigsResponse) {}
+
+  // RegisterNode registers a new node to the system.
+  rpc RegisterNode(RegisterNodeRequest) returns (RegisterNodeResponse) {}
```
patrickhuie19, line 30: Is there a case where "de-register" a node would be useful?

jmank88:
How many modules do we expect there to be?

Could some of them be combined? e.g. is it important for orchestrator/node and orchestrator/proposer to be completely independent modules? Could we even have one `sdk/go` module?

And for the .proto definitions. I personally like them alongside or even inside the package with the generated files. I think some version of that would simplify the source tree a bit.

jmank88 (changes requested):
I think we need to be very careful before creating this many modules

MStreet3:
Closed following discussion.

PR #735, opened 2024-08-27
Move interpolate key and keys from core to common so it can be reused by the upcoming test engine.

nolag:
See use in https://github.com/smartcontractkit/chainlink/pull/14248


pkg/workflows/execution/interpolation.go
```diff
@@ -0,0 +1,100 @@
+package execution
```
jmank88, line 1: Maybe consider `exec`?
```suggestion
package exec
```
https://go.dev/blog/package-names

pkg/workflows/execution/interpolation.go
```diff
@@ -0,0 +1,100 @@
...
+			inner, ok := v[r]
+			if !ok {
+				return "", fmt.Errorf("could not find ref part `%s` (ref: `%s`) in `%+v`", r, key, v)
+			}
+
+			val = inner
+		case []any:
+			i, err := strconv.Atoi(r)
+			if err != nil {
+				return "", fmt.Errorf("could not interpolate ref part `%s` (ref: `%s`) in `%+v`: `%s` is not convertible to an int", r, key, v, r)
```
jmank88, line 67: Should `err` be included?

pkg/workflows/execution/results.go
```diff
@@ -0,0 +1,13 @@
+package execution
+
+import "github.com/smartcontractkit/chainlink-common/pkg/values"
+
+type Results interface {
+	GetResultForStep(string) (*Result, bool)
```
jmank88, line 6: Does the `Get` prefix add value here?
```suggestion
	ResultForStep(string) (*Result, bool)
```
https://go.dev/doc/effective_go#Getters

pkg/workflows/execution/interpolation.go
nolag: was copied, but will add

jmank88 (approved):

pkg/workflows/exec/interpolation.go
```diff
@@ -0,0 +1,100 @@
...
+// it with its corresponding value in `state`.
+//
+// A key is valid if it contains at least two parts, with:
+//   - the first part being the workflow step's `ref` variable
+//   - the second part being one of `inputs` or `outputs`
+//
+// If a key has more than two parts, then we traverse the parts
+// to find the value we want to replace.
+// We support traversing both nested maps and lists and any combination of the two.
+func InterpolateKey(key string, state Results) (any, error) {
```
cedric-cordenier, line 22: Do we need to use an interface here? Under the hood this is just a map of State objects isn't it? Why not move that out too -- that feels like it simplifies the implementation

jmank88: If that is not possible, another in-between approach could be to accept a func here:
```suggestion
func InterpolateKey(key string, resultForStep func(string) (*Result, bool)) (any, error) {
```
Then caller just passes a reference to its method: `exec.InterpolateKey(key, results.ResultsForStep)

cedric-cordenier (approved):

PR #736, opened 2024-08-28
MERC-6190: Remove bid/ask fields from Mercury v4 schema

martin-cll:
We do not want them anymore in the v4 schema.


samsondav (approved):

pkg/loop/internal/pb/mercury/v4/datasource_v4.proto
```diff
@@ -28,8 +28,6 @@ message Block {
 // Observation is a gRPC adapter for [pkg/types/mercury/v1/Observation]
 message Observation {
     BigInt benchmarkPrice = 1;
-    BigInt bid = 2;
-    BigInt ask = 3;
```
jmank88, line 32: Should we mark them deprecated? IIRC there are two different ways to do it

martin-cll: We can mark them as deprecated, but since they've never been used in a deployed environment, I thought it would be easier to simply remove them for now.

jmank88: Removing is fine, but we should reserve them instead of leaving a gap without explanation:
https://protobuf.dev/programming-guides/proto3/#deleting

Otherwise deprecation is an 'option':
https://protobuf.dev/programming-guides/proto3/#options

martin-cll: I added the fields back with a deprecated tag.

jmank88 (approved):

samsondav (approved):

PR #743, opened 2024-08-29
Identical consensus builders

nolag:
Note, this does **not** implement identical consensus in OCR3 capability, that's https://smartcontract-it.atlassian.net/browse/CM-380

### DIFF
--- a/pkg/capabilities/consensus/ocr3/common_builders_generated.go
+++ b/pkg/capabilities/consensus/ocr3/common_builders_generated.go
@@ -8,47 +8,7 @@ import (
 
 type EncoderCap workflows.CapDefinition[Encoder]
 
-type EncoderConfigCap interface {
-	workflows.CapDefinition[EncoderConfig]
-	Abi() workflows.CapDefinition[string]
-	private()
-}
-
-// EncoderConfigCapFromStep should only be called from generated code to assure type safety
-func EncoderConfigCapFromStep(w *workflows.WorkflowSpecFactory, step workflows.Step[EncoderConfig]) EncoderConfigCap {
-	raw := step.AddTo(w)
-	return &encoderConfig{CapDefinition: raw}
-}
-
-type encoderConfig struct {
-	workflows.CapDefinition[EncoderConfig]
-}
-
-func (*encoderConfig) private() {}
-func (c *encoderConfig) Abi() workflows.CapDefinition[string] {
-	return workflows.AccessField[EncoderConfig, string](c.CapDefinition, "abi")
-}
-
-func NewEncoderConfigFromFields(
-	abi workflows.CapDefinition[string]) EncoderConfigCap {
-	return &simpleEncoderConfig{
-		CapDefinition: workflows.ComponentCapDefinition[EncoderConfig]{
-			"abi": abi.Ref(),
-		},
-		abi: abi,
-	}
-}
-
-type simpleEncoderConfig struct {
-	workflows.CapDefinition[EncoderConfig]
-	abi workflows.CapDefinition[string]
-}
-
-func (c *simpleEncoderConfig) Abi() workflows.CapDefinition[string] {
-	return c.abi
-}
-
-func (c *simpleEncoderConfig) private() {}
+type EncoderConfigCap workflows.CapDefinition[EncoderConfig]
 
 type SignedReportCap interface {
 	workflows.CapDefinition[SignedReport]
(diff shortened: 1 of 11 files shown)


pkg/capabilities/consensus/ocr3/common_builders_generated.go
```diff
@@ -6,6 +6,50 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/workflows"
 )
 
+type EncoderCap workflows.CapDefinition[Encoder]
+
+type EncoderConfigCap interface {
```
cedric-cordenier, line 11: This is probably specific to EVM, no?

nolag: Yeah, good call, I'll make it take a map[string]any.

pkg/capabilities/consensus/ocr3/ocr3_common-schema.json
```diff
@@ -27,6 +27,14 @@
...
       "additionalProperties": false,
       "required": ["Report", "Context", "Signature", "ID"]
+    },
+    "encoder": {
+      "type": "string",
+      "enum": ["EVM"]
+    },
+    "encoder_config": {
+      "type": "object",
+      "additionalProperties": true
```
DeividasK, line 37: Is this meant to be `any`? Why switch from https://github.com/smartcontractkit/chainlink-common/pull/743/files#diff-7b48ddffa1322b95afbcda50d5bd195e0d59e5069ed0a1ef651e3c31dd122b81L71-L84 ?

nolag: @cedric-cordenier pointed out that's only for EVM.  It should work for all encoders.

pkg/capabilities/consensus/ocr3/identical_consensus.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows"
+)
+
+// Note this isn't generated because generics isn't supported in json schema
+
+type IdenticalConsensusConfig[T any] struct {
+	Encoder       Encoder
+	EncoderConfig EncoderConfig
+}
```
DeividasK, line 13: I don't think I understand. Why does an identical consensus configuration have an encoder in it? I thought this would be an aggregation method and config.

cedric-cordenier: @nolag I think I agree with Deividas here: the aggregator + the encoder are two different parts of the consensus API.

pkg/workflows/testdata/fixtures/workflows/notstreamssepolia.yaml
```diff
@@ -11,7 +11,7 @@ ocr:
   report_id: '0001'
   encoder: EVM
   encoder_config:
-    abi: "(bytes32 FeedID, uint224 Price, uint32 Timestamp)[] Reports"
+    Abi: "(bytes32 FeedID, uint224 Price, uint32 Timestamp)[] Reports"
```
cedric-cordenier, line 14: @nolag Is this change correct? I think this breaks existing workflows if we're expecting the Abi to be capitalized

pkg/capabilities/consensus/ocr3/identical_consensus.go
nolag: I don't follow the "and config" this is the config, config is in the name.

pkg/workflows/testdata/fixtures/workflows/notstreamssepolia.yaml
nolag: I have a test in builder_test that verifies against the staging workflows.  This change is to a fake capability that I used in a demo.  Before the casing didn't matter because it wasn't a map but now it'll leave what I put.

cedric-cordenier (dismissed):

pkg/capabilities/consensus/ocr3/identical_consensus.go
nolag: After sync with @cedric-cordenier, the best would be to have an embedded struct with these two and report id.  Right now, I'm not sure how to get the json schema generator to embed a struct, so I'm going to leave them at the top level.

cedric-cordenier (approved):

bolekk (approved):

PR #748, opened 2024-09-03
Spike to assess impact and desirability of Capability API change

ettec:
To assess the impact of changing CallbackCapability Execute method from:

`Execute(ctx context.Context, r capabilities.CapabilityRequest) (<-chan capabilities.CapabilityResponse, error) `

to 

`Execute(ctx context.Context, r capabilities.CapabilityRequest) (capabilities.CapabilityResponse, error) `

Doing so will  protect against future bugs where the capability implementer sends more than the expected number of responses (e.g. this would break current remote target implementation) or forgets to close the channel.  As its used right now, in all places, callers of the API expect one result and typically block on the channel to wait for that result, i.e.they are effectively sync calls so the second API would be the most appropriate in all current use cases (that is the assumption that this PR is validating).   This does not preclude adding an additional Capability type in future that can stream results over a channel.


pkg/capabilities/consensus/ocr3/capability_test.go
```diff
@@ -165,11 +163,22 @@ func TestOCR3Capability_Eviction(t *testing.T) {
 		Inputs: inputs,
 	}
 
-	callback, err := cp.Execute(ctx, executeReq)
-	require.NoError(t, err)
-
-	fc.Advance(1 * time.Hour)
-	resp := <-callback
+	go func() {
```
jmank88, line 166: Let's track this and block so it doesn't leak. Something like this would work:
```suggestion
	done := make(chan struct{})
	t.Cleanup(func() { <-done })
	go func() {
		defer close(done)
```

ettec: looks like the test context is cancelled after t.Cleanup is called, so creating a child context from the test context and explicitly cancelling that in the test.

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -67,13 +67,7 @@ func CapabilityRequestToProto(req capabilities.CapabilityRequest) *CapabilityReq
 }
 
 func CapabilityResponseToProto(resp capabilities.CapabilityResponse) *CapabilityResponse {
-	errStr := ""
```
cedric-cordenier, line 70: @ettec This doesn't look right, did you intend to remove it?

pkg/loop/internal/core/services/capability/capabilities.go
```diff
@@ -331,73 +331,78 @@ func (c *callbackExecutableServer) UnregisterFromWorkflow(ctx context.Context, r
...
 	if err != nil {
-		return nil, fmt.Errorf("error executing capability request: %w", err)
+		return capabilities.CapabilityResponse{}, fmt.Errorf("error executing capability request: %w", err)
 	}
 
-	resp, err := responseStream.Recv()
+	message, err := responseStream.Recv()
 	if err != nil {
-		return nil, fmt.Errorf("error waiting for ack: %w", err)
+		return capabilities.CapabilityResponse{}, fmt.Errorf("error waiting for ack: %w", err)
```
cedric-cordenier, line 385: We're not waiting for an ack anymore, I think?

cedric-cordenier: @ettec Can we also remove the Ack type from the ResponseMessage proto type -- there's actually probably some additional simplification we can do there.

ettec: ah right, its waiting for the response, but its not an ack, I'll update the message

pkg/capabilities/consensus/ocr3/requests/handler.go
```diff
@@ -152,12 +151,9 @@ func (h *Handler) expirePendingRequests(ctx context.Context) {
 
 	for _, req := range h.pendingRequests {
 		if now.After(req.ExpiresAt) {
```
cedric-cordenier, line 153: Can this also be simplified now? i.e. can we just have this logic inline in the Execute call since we're no longer yielding back to the caller?

ettec: I'd prefer to do that simplification as a second PR as this one is quite beastly already

pkg/values/pb/values.pb.go
```diff
@@ -2,7 +2,7 @@
 // versions:
 // 	protoc-gen-go v1.31.0
 // 	protoc        v4.25.1
-// source: values.proto
```
cedric-cordenier, line 5: Could you uncommit these changes if you haven't actually changed anything related to values in this PR

ettec: ....I keep doing that, will sort

ettec: sorted

pkg/loop/internal/core/services/capability/capabilities.go
ettec: removed the ResponseMessage and related code, simplifies things a bit as you say

pkg/capabilities/consensus/ocr3/requests/handler.go
ettec: raised a JIRA against me for this-> Can we simplify request handler in OCR now the call in execute call is sync?

pkg/loop/internal/core/services/capability/capabilities.go
ettec: sorted the message

jmank88 (approved):

cedric-cordenier (approved):

PR #749, opened 2024-09-04
Expose ContractReader and ChainWriter of relayer in relayerSet

kidambisrinivas:
## Updates

- Implement ContractReader and ChainWriter in relayerset
  - Relayerset does not start gRPC service for relayer and mount it on-demand
  - So we had to expose ContractReader and ChainWriter of relayer from relayerset client itself

## Testing

```
$ go test -count=1 -v ./pkg/loop/internal/relayerset
=== RUN   Test_RelayerSet
2024-09-04T16:02:42.491+0100 [DEBUG] accepting initial connection: addr=/var/folders/1b/byq7w64n1x7gmp5cp9__vp_r0000gn/T/plugin2829362114
2024-09-04T16:02:42.492+0100 [DEBUG] stdio EOF, exiting copy loop
2024-09-04T16:02:42.492+0100 [DEBUG] stdio EOF, exiting copy loop
2024-09-04T16:02:42.493+0100 [DEBUG] making new client mux initial connection: addr=/var/folders/1b/byq7w64n1x7gmp5cp9__vp_r0000gn/T/plugin2829362114
2024-09-04T16:02:42.493+0100 [DEBUG] initial server connection accepted: addr=/var/folders/1b/byq7w64n1x7gmp5cp9__vp_r0000gn/T/plugin2829362114
2024-09-04T16:02:42.494+0100 [DEBUG] client muxer connected: addr=/var/folders/1b/byq7w64n1x7gmp5cp9__vp_r0000gn/T/plugin2829362114
2024-09-04T16:02:42.494+0100 [DEBUG] sending conn to default listener
2024-09-04T16:02:42.497+0100 [DEBUG] sending conn to default listener
2024-09-04T16:02:42.497+0100 [DEBUG] stdio: received EOF, stopping recv loop: err="rpc error: code = Unavailable desc = error reading from server: EOF"
--- PASS: Test_RelayerSet (0.01s)
PASS
ok      github.com/smartcontractkit/chainlink-common/pkg/loop/internal/relayerset       0.408s
```


jmank88:
Is this PR part of a set? If so, please link them together from the descriptions.

pkg/loop/internal/relayerset/relayer.go
```diff
@@ -12,13 +12,19 @@ import (
 )
 
 type relayerClient struct {
+	*relayer.RelayerClient
 	log              logger.Logger
 	relayerSetClient *Client
```
jmank88, line 17: What is the difference between the three different things named `client`?

kidambisrinivas: This approach was abandoned because relayer server is not started and mounted by relayerset. So the relayer client initialised here would not be able to call NewContractReader, as there would be no corresponding server.

cc @cedric-cordenier

patrickhuie19:
I don't follow - relayerset is supposed to serve relayers, which then hold a Chain Reader and Chain Writer. What does defining a Chain Reader and Chain Writer on a relayerset mean?

pkg/loop/internal/relayerset/relayer.go
jmank88: Sorry I am not following. Can you explain in english what each of these three different things that all share the same generic name of "client" are actually doing? I think it will lead to better names, or at least something we can document in the code.

pkg/loop/internal/relayer/relayer.go
```diff
@@ -182,7 +182,7 @@ func (k *keystoreServer) Sign(ctx context.Context, request *pb.SignRequest) (*pb
 
 var _ looptypes.Relayer = (*relayerClient)(nil)
 
-// relayerClient adapts a GRPC [pb.RelayerClient] to implement [Relayer].
+// relayerClient adapts a GRPC [pb.relayerClient] to implement [Relayer].
```
cedric-cordenier, line 185: The previous comment was correct @kidambisrinivas : see line 190, the `relayer` attribute is a `[pb.RelayerClient]`

cedric-cordenier (dismissed):

cedric-cordenier:
>
I don't follow - relayerset is supposed to serve relayers, which then hold a Chain Reader and Chain Writer. What does defining a Chain Reader and Chain Writer on a relayerset mean?

@patrickhuie19 This is just a way to save us from instantiating an extra server for the relayer.

The calls we would make normally are:
RelayerSet.Get -> Relayer
and then
Relayer.NewChainReader -> ChainReader

We could translate this to the GRPC world by having each call to RelayerSet.Get wrap the returned relayer in a server and register that to the GRPC server. However this is actually pretty inefficient since a relayer object on its own is useless; users will always want to use the relayer to instantiate eg. a chainreader or chainwriter.

So we can avoid the intermediate server for the relayer by just storing a reference to the relayerSet client and the relayer we want to fetch.

I.e. the calls described above instead would become:

RelayerSet.Get -> RelayerClient (i.e. a RelayerSetClient + relayerID) -- effectively this call just acts as a check that the relayer exists
RelayerClient.NewChainReader -> this is a call to RelayerSet.NewChainReader with (relayerID, []chainReaderConfig); the implementation will then fetch the relayer and call NewChainReader on it

pkg/loop/internal/relayerset/relayer.go
cedric-cordenier: There are just two clients here: a relayerClient and a relayerSetClient

A relayerClient is just a wrapper over the relayerSetClient + a relayID; this saves us having to register a relayer server every time we call RelayerSet.Get or RelayerSet.List since we defer fetching the relayer to the next call users make.

jmank88: There are three types in scope of this comment called "client":
- `relayerClient` or `relayerset.relayerClient`
- `relayer.RelayerClient`
- `Client` or `relayerset.Client`

jmank88: All of these type need documentation, or more intuitive names.

pkg/types/core/relayerset.go
```diff
@@ -27,7 +27,10 @@ type RelayArgs struct {
 	MercuryCredentials *types.MercuryCredentials
 }
 
+//go:generate mockery --quiet --name Relayer --output ./mocks/ --case=underscore
```
jmank88, line 30: Please use the `.mockery.yaml` instead of `go:generate`

jmank88 (changes requested):

pkg/loop/internal/relayerset/relayer.go
jmank88: Oh, was this removed?

kidambisrinivas: Yes, the original approach was removed because of the reason I mentioned above.

pkg/types/core/relayerset.go
kidambisrinivas: This interface is already in .mockery.yaml

patrickhuie19:
> > I don't follow - relayerset is supposed to serve relayers, which then hold a Chain Reader and Chain Writer. What does defining a Chain Reader and Chain Writer on a relayerset mean?
> 
> @patrickhuie19 This is just a way to save us from instantiating an extra server for the relayer.
> 
> The calls we would make normally are: RelayerSet.Get -> Relayer and then Relayer.NewChainReader -> ChainReader
> 
> We could translate this to the GRPC world by having each call to RelayerSet.Get wrap the returned relayer in a server and register that to the GRPC server. However this is actually pretty inefficient since a relayer object on its own is useless; users will always want to use the relayer to instantiate eg. a chainreader or chainwriter.
> 
> So we can avoid the intermediate server for the relayer by just storing a reference to the relayerSet client and the relayer we want to fetch.
> 
> I.e. the calls described above instead would become:
> 
> RelayerSet.Get -> RelayerClient (i.e. a RelayerSetClient + relayerID) -- effectively this call just acts as a check that the relayer exists RelayerClient.NewChainReader -> this is a call to RelayerSet.NewChainReader with (relayerID, []chainReaderConfig); the implementation will then fetch the relayer and call NewChainReader on it

That makes sense to me. Can we document that @kidambisrinivas ?

kidambisrinivas:
@patrickhuie19 Documented this

cedric-cordenier (approved):

jmank88 (approved):

PR #751, opened 2024-09-05
Add unit test runner for workflows

nolag:


pkg/workflows/testutils/runner.go
```diff
@@ -65,10 +126,118 @@ func (r *Runner) GetRegisteredMock(name string, step string) CapabilityMock {
...
+	}
+
+	val, err := exec.FindAndInterpolateAllKeys(input, r.results)
+	if err != nil {
+		return nil, err
+	}
+	return values.NewMap(val.(map[string]any))
+}
+
+func (r *Runner) iReady(ref string) bool {
```
cedric-cordenier, line 218: iReady? What does this stand for?

pkg/workflows/testutils/runner.go
```diff
@@ -34,15 +49,61 @@ func (r *Runner) Run(factory *workflows.WorkflowSpecFactory) error {
...
+}
+
+func (r *Runner) setupSteps(factory *workflows.WorkflowSpecFactory, spec workflows.WorkflowSpec) {
+	for _, step := range spec.Steps() {
+		if step.Ref == "" {
+			step.Ref = step.ID
+		}
+
+		r.idToStep[step.Ref] = step
+		if run := factory.GetFn(step.Ref); run != nil {
```
cedric-cordenier, line 91: This is really unobvious IMO -- can you add a comment here to explain what it's doing?

Eg. it seems that GetFn() returns nil if the step isn't custom, is that right?

pkg/workflows/testutils/compute_capability.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows"
+)
+
+type computeCapability struct {
+	sdk      workflows.SDK
+	callback func(sdk workflows.SDK, request capabilities.CapabilityRequest) capabilities.CapabilityResponse
+}
+
+func (c *computeCapability) Run(request capabilities.CapabilityRequest) capabilities.CapabilityResponse {
```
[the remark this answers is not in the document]
cedric-cordenier: I wouldn't bother with the extra layer of wrapping here: this is intended to be effectively a capability implementation so I would just make it match to the real API + forego the need to mock it.

pkg/workflows/testutils/runner.go
nolag: typo, isReady, will fix ty.

pkg/workflows/testutils/compute_capability.go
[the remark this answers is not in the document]
nolag: I'm not sure how we'll insert the CapabilityInfo part if you want to wrap running a capability, but if we want a real one to run during tests we can think about it more.  For now, I'm only using ID anyways.

cedric-cordenier (approved):

cedric-cordenier (approved):

pkg/workflows/testutils/runner.go
```diff
@@ -34,16 +56,69 @@ func (r *Runner) Run(factory *workflows.WorkflowSpecFactory) error {
...
-	// TODO https://smartcontract-it.atlassian.net/browse/KS-442, implement this function
-	_ = spec
-	return errors.New("TODO https://smartcontract-it.atlassian.net/browse/KS-442")
+	if err = r.ensureGraph(spec); err != nil {
+		return err
+	}
+
+	r.setupSteps(factory, spec)
+
+	return r.walk(spec, workflows.KeywordTrigger)
```
vreff, line 65: so a workflow always has a single trigger? is that set to change eventually?

nolag: Yes, and most likely.

pkg/workflows/testutils/runner_test.go
```diff
@@ -16,82 +17,140 @@ import (
...
 	})
 
-	t.Run("Captures errors", func(t *testing.T) {
-		t.Skip("TODO https://smartcontract-it.atlassian.net/browse/KS-442")
-		assert.Fail(t, "Not implemented")
+	t.Run("Run allows hard-coded values", func(t *testing.T) {
+		workflow := workflows.NewWorkflowSpecFactory(workflows.NewWorkflowParams{Name: "tester", Owner: "ryan"})
+		trigger := basictrigger.TriggerConfig{Name: "trigger", Number: 100}.New(workflow)
+		hardCodedInput := basicaction.NewActionOutputsFromFields(workflows.ConstantDefinition("hard-coded"))
+		tTransform := workflows.Compute2[basictrigger.TriggerOutputs, basicaction.ActionOutputs, bool](
```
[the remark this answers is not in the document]
nolag: ComputeN is custom code.  You have access to your config when you're building your workflow.  Others aren't your code and therefore need to dictate the structure of the config they want.

pkg/workflows/testutils/runner_test.go
```diff
@@ -16,82 +17,140 @@ import (
...
+			workflow,
+			"transform",
+			workflows.Compute2Inputs[basictrigger.TriggerOutputs, basicaction.ActionOutputs]{Arg0: trigger, Arg1: hardCodedInput},
+			func(SDK workflows.SDK, tr basictrigger.TriggerOutputs, hc basicaction.ActionOutputs) (bool, error) {
+				assert.Equal(t, "hard-coded", hc.AdaptedThing)
+				assert.NotNil(t, SDK)
+				assert.Equal(t, "cool", tr.CoolOutput)
```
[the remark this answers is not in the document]
nolag: [This folder](https://github.com/smartcontractkit/chainlink-common/tree/main/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities) has the ones that I refer to.  I used fun names because I didn't want to think to hard about real ones, since they are just for testing the engine and generator.

In particular, CoolOutput is defined [here](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/basictrigger/basic_trigger-schema.json#L23), which generates the Go type, the capability's builder and the mock.

pkg/workflows/testutils/runner.go
cedric-cordenier: Note a workflow can have multiple triggers, but it will only have a single trigger event (and therefore only a single "trigger" is the root for a given execution), which is why we can make this leap.

Lmk if that distinction doesn't make sense

bolekk (approved):

PR #752, opened 2024-09-06
[CAPPL-37] Implement Compute runner

cedric-cordenier:


pkg/workflows/sdk/runner_wasip1.go
```diff
@@ -0,0 +1,25 @@
...
+	"unsafe"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows"
+	sdkpb "github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk/pb"
+)
+
+//go:wasmimport env sendResponse
+func sendResponse(reqptr unsafe.Pointer, reqptrlen int32)
+
+func Run(factory workflows.WorkflowSpecFactory) {
```
nolag, line 14: I think we should use sdk.NewRunner() so that users can switch testing a workflow from native to compiling WASM easier.

Namely, I can make a function 

```
func RunWorkflow(runner: sdk.Runner) {
   // build workflow
   // run workflow
}
```

main can look like

```
func main() {
   pkg.RunWorkflow(sdk.NewRunner())
}
```

tests would use

`pkg.RunWorkflow(testutils.NewRunner())`

Thoughts?

nolag (approved):

agparadiso (approved):

PR #755, opened 2024-09-09
fix install protoc

jmank88:
install-protoc.sh needs to check the `protoc` version that will exceute, rather than an explicit path to a particular binary.

### DIFF
--- a/script/install-protoc.sh
+++ b/script/install-protoc.sh
@@ -13,7 +13,7 @@ os=$(uname)
 arch=$(uname -m)
 
 install_dir=$HOME/.local
-$install_dir/bin/protoc --version | grep $VERSION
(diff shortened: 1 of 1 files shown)


script/install-protoc.sh
```diff
@@ -13,7 +13,7 @@ os=$(uname)
 arch=$(uname -m)
 
 install_dir=$HOME/.local
-$install_dir/bin/protoc --version | grep $VERSION
+protoc --version | grep $VERSION
```
jmank88, line 16: With the qualified path, it can look like the correct version is installed even when that program will not be found on the PATH.

Atrax1 (approved):

george-dorin (approved):

PR #758, opened 2024-09-10
pkg/types/ccipocr3: add rmn sigs to report

makramkd:
Additionally, mark NewCommitPluginReport as deprecated.

### DIFF
--- a/pkg/types/ccipocr3/plugin_commit_types.go
+++ b/pkg/types/ccipocr3/plugin_commit_types.go
@@ -1,10 +1,27 @@
 package ccipocr3
 
+// CommitPluginReport contains the necessary information to commit CCIP
+// messages from potentially many source chains, to a single destination chain.
+//
+// It must consist of either:
+//
+//  1. a non-empty MerkleRoots array, or
+//  2. a non-empty PriceUpdates array
+//
+// If neither of the above is provided the report is considered empty and should
+// not be transmitted on-chain.
+//
+// In the event the MerkleRoots array is non-empty, it may also contain
+// RMNSignatures, if RMN is configured for some lanes involved in the commitment.
+// A report with RMN signatures but without merkle roots is invalid.
 type CommitPluginReport struct {
-	MerkleRoots  []MerkleRootChain `json:"merkleRoots"`
-	PriceUpdates PriceUpdates      `json:"priceUpdates"`
+	MerkleRoots   []MerkleRootChain   `json:"merkleRoots"`
+	PriceUpdates  PriceUpdates        `json:"priceUpdates"`
+	RMNSignatures []RMNECDSASignature `json:"rmnSignatures"`
 }
 
+// Deprecated: don't use this constructor, just create a CommitPluginReport struct directly.
+// Will be removed in a future version once all uses have been replaced.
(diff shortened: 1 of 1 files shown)


pkg/types/ccipocr3/plugin_commit_types.go
```diff
@@ -1,10 +1,27 @@
...
-	MerkleRoots  []MerkleRootChain `json:"merkleRoots"`
-	PriceUpdates PriceUpdates      `json:"priceUpdates"`
+	MerkleRoots   []MerkleRootChain   `json:"merkleRoots"`
+	PriceUpdates  PriceUpdates        `json:"priceUpdates"`
+	RMNSignatures []RMNECDSASignature `json:"rmnSignatures"`
 }
 
+// Deprecated: don't use this constructor, just create a CommitPluginReport struct directly.
+// Will be removed in a future version once all uses have been replaced.
 func NewCommitPluginReport(merkleRoots []MerkleRootChain, tokenPriceUpdates []TokenPrice, gasPriceUpdate []GasPriceChain) CommitPluginReport {
```
makramkd, line 25: This function is being deprecated to avoid breaking changes, adding a param would mean a breaking change and a desperate bump across chainlink-ccip and chainlink

pkg/types/ccipocr3/plugin_commit_types.go
```diff
@@ -37,6 +55,24 @@ func NewMerkleRootChain(
 	}
 }
 
+// RMNECDSASignature is the ECDSA signature from a single RMN node
+// on the RMN "Report" structure that consists of:
```
asoliman92, line 59: Is this the best place for this documentation? What happens if the report structure changes? Will be better if we link to the report struct directly or something.

dimkouv (commented):
How do we use FixedDestLaneUpdate fields of the query?

https://github.com/smartcontractkit/chainlink-ccip/blob/12c4fa8dbbadbdc7eca048e6b4fe5ce9d1ce97be/rmn_offchain.proto#L100

Just for offchain/plugin verifications?

pkg/types/ccipocr3/plugin_commit_types.go
makramkd: It's good to have documentation where possible.

makramkd:
> How do we use FixedDestLaneUpdate fields of the query?

The `FixedDestLaneUpdate` array will have to be equal to the `MerkleRoots` field in the plugin report.

pkg/types/ccipocr3/plugin_commit_types.go
asoliman92: It's good when it's not something that will change regularly or there's no central place to link to (the original contract for example). I won't block the PR because of this but it can be more harmful if it diverges from the current documentation.

makramkd: Yeah, linking to the contract is fine, unfortunately its not finalized. But will put a link here anyway, and boyscout principle means that someone should update it to the latest relevant link at some point 😉

dimkouv (approved):

pkg/types/ccipocr3/plugin_commit_types.go
makramkd: Added link

PR #761, opened 2024-09-11
feat(observability-lib): deploy logic in grafana module

Atrax1:
- Add in grafana module the functions :
  - DeployToGrafana
  - DeleteDashboard
  - GenerateJSON
- Simplify creating alert by removing from builder params Group and FolderUID
- Added tests to check json output generated from builded dashboards


observability-lib/cmd/builder.go
```diff
@@ -35,60 +34,75 @@ const (
...
-		SlackWebhookURL:   options.SlackWebhookURL,
-		SlackToken:        options.SlackToken,
-		SlackChannel:      options.SlackChannel,
-	}
-
+func BuildDashboardWithType(options *BuildOptions) (*grafana.Dashboard, error) {
 	switch options.TypeDashboard {
 	case TypeDashboardCoreNode:
-		db, err = corenode.NewDashboard(dashboardOptions)
+		return corenode.NewDashboard(&corenode.Props{
```
patrickhuie19, line 49: exploring: the `Props` (corenode, corenodecomponents, k8sresources) seem to share a fair amount of fields. Could the Props type be consolidated with validation added in NewDashboard to ensure the right fields are passed in? 

I do like that the right fields are asserted at build time though, so its a tradeoff.

observability-lib/grafana/dashboard.go
```diff
@@ -12,22 +16,150 @@ const (
...
+
+		// delete alert rules for the dashboard
+		for _, rule := range alertsRule {
+			_, _, errDeleteAlertRule := grafanaClient.DeleteAlertRule(*rule.Uid)
+			if errDeleteAlertRule != nil {
+				return errDeleteAlertRule
+			}
+		}
+
+		// Create alert rules for the dashboard
```
patrickhuie19, line 79: how do the alerts defined here play with the alerts we define in the alerts repo?

Atrax1: Those alerts are Grafana managed alert rules
image
The alerts in alerts repo are data source-managed alert rule

observability-lib/cmd/builder.go
Atrax1: The exported function NewDashboard from each modules needs to be easily usable from anywhere without using cmd. I'm open for idea on the params needed for the NewDashboard function exported from each module. For now I don't enforce through the library the common params when building dashboards and alerts.

patrickhuie19 (approved):

lukaszcl (approved):

PR #762, opened 2024-09-11
[CAPPL-42] SDK logger

agparadiso:
### Description

This pr introduces a logger to the SDK that can be used from the wasm guest.
It basically json encodes, sends to the host, json decodes in the host, and logs using the our custom logger.Logger.

In order to do this I've introduced a new constructor `NewWithZapLogger` that returns a new Logger based on a zap.Logger. With this we can build our own core and pass it down to the `logger.Logger`

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -1,6 +1,7 @@
 package logger
 
 import (
+	"io"
 	"reflect"
 	"testing"
 
@@ -63,6 +64,12 @@ func NewWith(cfgFn func(*zap.Config)) (Logger, error) {
 		return nil, err
 	}
 	return &logger{core.Sugar()}, nil
+}
+
+// NewWithSync returns a new Logger with a given SyncWriter.
+func NewWithSync(w io.Writer) Logger {
+	core := zapcore.NewCore(zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()), zapcore.AddSync(w), zapcore.InfoLevel)
+	return &logger{zap.New(core).Sugar()}
 }
 
 // Test returns a new test Logger for tb.
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -3,6 +3,7 @@ package host
 import (
 	"bytes"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -179,6 +180,56 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
 			}
 
 			return ErrnoSuccess
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	err = linker.FuncWrap(
+		"env",
+		"log",
+		func(caller *wasmtime.Caller, ptr int32, ptrlen int32) {
+			b, innerErr := safeMem(caller, ptr, ptrlen)
+			if innerErr != nil {
+				logger.Errorf("error calling log: %s", err)
+				return
+			}
+
+			var raw map[string]interface{}
+			innerErr = json.Unmarshal(b, &raw)
+			if innerErr != nil {
+				return
+			}
+
+			level := raw["level"]
+			delete(raw, "level")
+
+			msg := raw["msg"].(string)
+			delete(raw, "msg")
+			delete(raw, "ts")
+
+			var args []interface{}
+			for k, v := range raw {
+				args = append(args, k, v)
+			}
+
+			switch level {
+			case "debug":
+				logger.Debugw(msg, args...)
+			case "info":
+				logger.Infow(msg, args...)
+			case "warn":
+				logger.Warnw(msg, args...)
+			case "error":
+				logger.Errorw(msg, args...)
+			case "panic":
+				logger.Panicw(msg, args...)
+			case "fatal":
+				logger.Fatalw(msg, args...)
+			default:
+				logger.Infow(msg, args...)
+			}
 		},
 	)
 	if err != nil {
--- a/pkg/workflows/wasm/runner_wasip1.go
+++ b/pkg/workflows/wasm/runner_wasip1.go
@@ -6,11 +6,15 @@ import (
 
 	"google.golang.org/protobuf/proto"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	wasmpb "github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/pb"
 )
 
 //go:wasmimport env sendResponse
 func sendResponse(respptr unsafe.Pointer, respptrlen int32) (errno int32)
+
+//go:wasmimport env log
+func log(respptr unsafe.Pointer, respptrlen int32)
 
 func bufferToPointerLen(buf []byte) (unsafe.Pointer, int32) {
 	return unsafe.Pointer(&buf[0]), int32(len(buf))
@@ -46,6 +50,18 @@ func NewRunner() *Runner {
 
 			os.Exit(code)
 		},
+		SDK: Runtime{
+			Logger: logger.NewWithSync(&wasmWriteSyncer{}),
+		},
 		args: os.Args,
 	}
+}
+
+type wasmWriteSyncer struct{}
+
+// Write is used to proxy log requests from the WASM binary back to the host
+func (wws *wasmWriteSyncer) Write(p []byte) (n int, err error) {
+	ptr, ptrlen := bufferToPointerLen(p)
+	log(ptr, ptrlen)
+	return int(ptrlen), nil
 }
--- a/pkg/workflows/wasm/sdk.go
+++ b/pkg/workflows/wasm/sdk.go
@@ -1,7 +1,12 @@
 package wasm
 
-import "github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+)
 
-type Runtime struct{}
+type Runtime struct {
+	Logger logger.Logger
+}
 
 var _ sdk.Runtime = (*Runtime)(nil)
--- a/pkg/workflows/wasm/host/test/log/cmd/main.go
+++ b/pkg/workflows/wasm/host/test/log/cmd/main.go
@@ -1,0 +1,35 @@
+//go:build wasip1
+
+package main
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/basictrigger"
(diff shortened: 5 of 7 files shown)


pkg/workflows/wasm/wasm.go
```diff
@@ -72,13 +79,13 @@ func newModule(binary []byte) (*module, error) {
...
 			b, innerErr := safeMem(caller, unsafe.Pointer(uintptr(ptr)), ptrlen)
-			if err != nil {
+			if innerErr != nil {
 				outerErr = innerErr
 				return
 			}
 
 			innerErr = proto.Unmarshal(b, &resp)
-			if err != nil {
+			if innerErr != nil {
```
agparadiso, line 88: @cedric-cordenier modifying these two, I think this is the right error, but please check it

cedric-cordenier: Good catch :) Thank you!

pkg/workflows/wasm/test/cmd/main.go
```diff
@@ -24,6 +24,8 @@ func BuildWorkflow(config []byte) *workflows.WorkflowSpecFactory {
 
 func main() {
 	runner := sdk.NewRunner()
+	runner.SDK.LogInfo("building workflow...")
 	workflow := BuildWorkflow(runner.Config())
+	runner.SDK.LogInfo("running workflow...")
 	runner.Run(workflow)
```
cedric-cordenier, line 30: Could we also add some debug logging to the runner itself as part of this? I think it would help speed up debugging in the future and would generally be very useful.

pkg/workflows/wasm/sdk/runner_wasip1.go
```diff
@@ -18,11 +21,25 @@ var (
...
 func NewRunner() *Runner {
+
+	wws := &WasmWriteSyncer{
+		logFn: func(msg string) {
+			ptr, ptrlen := bufferToPointerLen(bytes.NewBufferString(msg).Bytes())
+			log(ptr, ptrlen)
+			return
+		},
+	}
+	core := zapcore.NewCore(zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()), zapcore.AddSync(wws), zapcore.InfoLevel)
```
cedric-cordenier, line 40: @agparadiso Note, calling AddSync means you don't need to implement Sync on line 76, so I would just implement the Write method down below

pkg/workflows/wasm/sdk/runner_wasip1.go
```diff
@@ -18,11 +21,25 @@ var (
...
+func log(respptr unsafe.Pointer, respptrlen int32)
+
 func bufferToPointerLen(buf []byte) (unsafe.Pointer, int32) {
 	return unsafe.Pointer(&buf[0]), int32(len(buf))
 }
 
 func NewRunner() *Runner {
+
+	wws := &WasmWriteSyncer{
+		logFn: func(msg string) {
```
cedric-cordenier, line 34: @agparadiso You're dividing up the logic between here and the writer implementation below -- this would probably be simpler if you moved all of this logic down to the writer, and just passed in the log function to the constructor here

pkg/workflows/wasm/sdk.go
```diff
@@ -1,7 +1,12 @@
...
 
-import "github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+	"go.uber.org/zap"
+)
 
-type Runtime struct{}
+type Runtime struct {
+	Logger *zap.Logger
```
cedric-cordenier, line 9: @agparadiso Can we make this a logger.Logger too if that's what we use as standard?

pkg/workflows/wasm/host/module.go
```diff
@@ -11,9 +11,10 @@ import (
 	"unsafe"
 
 	"github.com/bytecodealliance/wasmtime-go/v23"
+	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	ilogger "github.com/smartcontractkit/chainlink-common/pkg/logger"
```
cedric-cordenier, line 17: Can you remove the alias if you don't need it?

cedric-cordenier (approved):

ettec (approved):

jmank88:
The core logger package is very unusual, and only still exists to support historical quirks that don't appear to be used here. Can the common logger package be used instead?

jmank88 (changes requested):
~This constructor does not return a logger like the others in this package. Let's use the common logger instead.~

pkg/workflows/wasm/log_codec.go
jmank88, line 1: Why do we need to write our own custom json encoder?

pkg/workflows/wasm/host/module.go
```diff
@@ -185,6 +186,44 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
...
+		"env",
+		"log",
+		func(caller *wasmtime.Caller, ptr int32, ptrlen int32) {
+			b, innerErr := safeMem(caller, ptr, ptrlen)
+			if innerErr != nil {
+				logger.Errorf("error calling log: %s", err)
+				return
+			}
+
+			entry, fields, innerErr := wasm.NewLogCodec().Decode(b)
```
jmank88, line 199: It would be ideal if we could use the zap json encoder instead of something custom, but if this is truly necessary then we can make it more efficient. There is no need construct a codec each time since decoding is a pure func.

pkg/workflows/wasm/log_codec.go
```diff
@@ -0,0 +1,159 @@
...
+
+	buf := buffer.NewPool().Get()
+	buf.AppendBytes(logBytes)
+
+	return buf, nil
+}
+
+func (c *logCodec) Decode(b []byte) (*zapcore.Entry, []interface{}, error) {
+	c.Lock()
+	defer c.Unlock()
```
jmank88, line 59: Why lock? Isn't this a pure implementation that doesn't use `c`?

pkg/workflows/wasm/host/module.go
cedric-cordenier: I don't see an alternative to this: we tried to use the existing zap json encoder but this turned out to be surprisingly hard to do: the basic problem is that while you can encode the entries just fine, you won't be able to decode it on the other end to pass it to the underlying logger.

pkg/workflows/wasm/log_codec.go
cedric-cordenier: I missed this in my previous review :/ Yeah we don't need to lock here -- the implementation is pure + WASM is single-threaded.

cedric-cordenier:
@jmank88 Re. your comment about a new constructor, it's because we want to control the core that we're using. The existing constructors don't allow us to do that so easily so this seemed the simplest option.

pkg/workflows/wasm/host/module.go
jmank88: I'm not understanding the issue. Zap outputs standard JSON with well known fields. What is different about what you are doing?

Or more importantly, do we even need to decode it at all? Could we just pipe it through serialized instead?

cedric-cordenier: >Could we just pipe it through serialized instead

This was the initial approach that we tried, but as far as I can see from the API that's not possible.

The lowest level API that I could find for zap was the zapcore.Core interface: https://pkg.go.dev/go.uber.org/zap/zapcore#Core

This requires us to pass in an entry + some fields.

I guess we could pack the serialized bytes into an entry if that's what you mean?

pkg/workflows/wasm/log_codec.go
agparadiso: thanks for pointing this, its a left over. initially I had the buffer as part of the state. will remove it thanks!

pkg/workflows/wasm/host/module.go
agparadiso: I think I might be able to use the zapcore jsonEncoder and just adjust the decoder to its format. Will try this

cedric-cordenier (dismissed):

pkg/logger/logger.go
```diff
@@ -65,6 +66,16 @@ func NewWith(cfgFn func(*zap.Config)) (Logger, error) {
 	return &logger{core.Sugar()}, nil
 }
 
+// NewWasm returns a new Logger to be used in a Wasm binary that will proxy the logs to the host.
+func NewWasm(log func(respptr unsafe.Pointer, respptrlen int32)) Logger {
```
jmank88, line 70: Is it possible to move the `log` func definition as well? Since we don't need to swap it out, and it looks pretty foreign.
```go
//go:wasmimport env log
func log(respptr unsafe.Pointer, respptrlen int32)
```

cedric-cordenier: @jmank88 Do you mean move the goimport into this package? I'm not 100% keen on this since it could potentially make it harder to track what imports are being used. Atm they are all listed in runner_wasip1 so our imports (not the WASI ones) are at least clearly defined.

agparadiso: I think this needs to live in the runner_wasip1.go in order for the wasmimport to work. I even get a static check error when moving it to the logger. Can you confirm @cedric-cordenier?

jmank88: I guess I'm not following how these pieces fit together. The file that contains that func does not contain `//go:build wasip1` like many other files, so what is the difference with putting inside of this package?

If we are stuck for some reason, then maybe a better alternative is something like:
```suggestion
func NewWithSync(io.Writer) Logger {
```
With the caller passing in `wasm.WriteSyncer` :shrug:

agparadiso: I think it has to do with the name of the file, its called: `runner_wasip1.go` if you modify the name you get a `missing function body` error. thats why @cedric-cordenier  mentioned about creating `logger_wasip1.go`. I can def modify the Constructor to receive a io.Writer

pkg/workflows/wasm/runner_wasip1.go
```diff
@@ -46,6 +50,22 @@ func NewRunner() *Runner {
...
 }
+
+type wasmWriteSyncer struct {
+	logFn func(respptr unsafe.Pointer, respptrlen int32)
+}
+
+// Write is used to proxy log requests from the WASM binary back to the host
+func (wws *wasmWriteSyncer) Write(p []byte) (n int, err error) {
+	ptr, ptrlen := bufferToPointerLen(p)
+	wws.logFn(ptr, ptrlen)
```
jmank88, line 69: nit/ do we actually swap this out? Or could it call `log` directly?
```suggestion
	log(ptr, ptrlen)
```

jmank88 (approved):

pkg/workflows/wasm/runner_wasip1.go
agparadiso: we are not changing it, Im updating it. thanks!

jmank88 (dismissed):

jmank88 (approved):

cedric-cordenier (approved):

PR #765, opened 2024-09-11
Seperate sdk from workflow to as part of an effort to shrink the WASM binary size

nolag:
Note that the separation was also required for ocr3 because Prometheus is used and it doesn't support WASM compilations.


jmank88 (dismissed):

.github/workflows/pkg.yml
```diff
@@ -76,4 +76,6 @@ jobs:
           make rm-mocked
           make rm-builders
           make generate
-          git diff --stat --exit-code
+          # The wasm build includes the hash of the latest git commit.
+          # Builds are consistent, until you commit and that hash changes.
+          git diff --stat --exit-code -- . ':!pkg/workflows/wasm/host/test/cmd/testmodule.wasm'
```
cedric-cordenier, line 81: This makes me wonder whether we shouldn't just .gitignore *.wasm and have a command to build the wasm binaries; otherwise we'll have to keep adding to this list constantly

nolag: I would want to know what it's for.  Eg, if we have a WASM that isn't dependant on common itself then I don't think it's an issue.

I would be hapy for it to change when we understand it'll be a pattern.

cedric-cordenier: Happy to wait until this becomes a regular problem

jmank88: IMHO we need to solve this problem without resorting to a hack like this. What is fundamentally in the way of making this reproducible?

nolag: The WASM build depends on the Git repository it's in. Go uses the hash of the commit, along with the content of all the .go files it depends on, to generate metadata.

If you update any dependency (even just a comment), it would be fine to say that regeneration is needed. However, the binary won't reproduce once you commit, because Go sees that the hash of the package it's importing is different.

To reproduce, pull this locally and build. You'll see a +0-0 diff with a different shasum. If you run git reset --soft HEAD~1 and rebuild, it reproduces. You may need to do this a couple of times, as I'm not sure if I regenerated with the last commit.

I'm happy to have a quick sync if that would be easier.

cedric-cordenier (approved):

.github/workflows/pkg.yml
nolag: Per discussion, building the go in the test itself instead.

cedric-cordenier (approved):

PR #769, opened 2024-09-16
Allow remote refs in the generator

nolag:


go.mod
```diff
@@ -109,4 +109,4 @@ require (
 )
 
 // until merged upstream: https://github.com/omissis/go-jsonschema/pull/264
-replace github.com/atombender/go-jsonschema => github.com/nolag/go-jsonschema v0.16.0-smallSizedInts
+replace github.com/atombender/go-jsonschema => github.com/nolag/go-jsonschema v0.16.2-rtinianov-merge-request
```
nolag, line 112: This is my last official merge before the owner is reviewing.  I got small PR feedback two days ago and addressed it today.  Hopefully I'll remove this soon.

The upgrade allows go-jsonschmea to use remote refs.  It looked like someone accidentally disabled it because it had it before and the structs to do it still existed but weren't used.

pkg/capabilities/cli/cmd/generate-types/main.go
```diff
@@ -18,6 +19,7 @@ var goWorkflowTestTemplate string
 
 var dir = flag.String("dir", "", fmt.Sprintf("Directory to search for %s files, if a file is provided, the directory it is in will be used", cmd.CapabilitySchemaFilePattern.String()))
 var localPrefix = flag.String("local_prefix", "github.com/smartcontractkit", "The local prefix to use when formatting go files")
+var extraUrls = flag.String("extra_urls", "", "Comma separated list of extra URLs to fetch schemas from")
```
nolag, line 22: ’m not thrilled about requiring the user to specify which remote files they’re referring to, but I needed to do so in order to construct the namespace correctly for go-jsonschema to work as intended with my generator.

I think it’s acceptable because it forces users to be more conscious of their inclusions, especially when nested. Users probably should know what they depend on for a capability...

bolekk (dismissed):

cedric-cordenier (approved):

bolekk (approved):

PR #772, opened 2024-09-17
Remove mustEmbed from ContractReader

EasterTheBunny:
Temporarily removing mustEmbed from `ContractReader` interface due to complications involving mockery. To use the mustEmbed, a new interface mocking style must be implemented that doesn't directly reference `ContractReader` in common. This would negatively impact the timeline to CCIP at the current time.


jmank88 (approved):

winder (approved):
Thank you :pray:

PR #778, opened 2024-09-18
[CAPPL-58] Correctly stub out clock_time_get and poll_oneoff

cedric-cordenier:
* correctly stub out sleeping: the implementation is based on wazero's implementation of the syscalls, but with less support (eg. for file descriptor notifications, which we don't support anyway).
* this includes stubbing out `clock_time_get` using a clock which ticks every time it's called and which is advanced by any time.Sleep calls.
* make fuel optional
* make sendResponse return an errno so that we can bubble this back to callers.


agparadiso (dismissed):

pkg/workflows/wasm/host/errno.go
```diff
@@ -0,0 +1,165 @@
...
+	ErrnoNotsock
+	// ErrnoNotsup Not supported, or operation not supported on socket.
+	ErrnoNotsup
+	// ErrnoNotty Inappropriate I/O control operation.
+	ErrnoNotty
+	// ErrnoNxio No such device or address.
+	ErrnoNxio
+	// ErrnoOverflow Value too large to be stored in data type.
+	ErrnoOverflow
+	// ErrnoOwnerdead Previous owner died.
```
patrickhuie19, line 134: 👻

pkg/workflows/wasm/host/module.go
```diff
@@ -116,65 +125,60 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
...
 	cfg.SetEpochInterruption(true)
-	cfg.SetConsumeFuel(true)
+	if modCfg.InitialFuel > 0 {
+		cfg.SetConsumeFuel(true)
+	}
 
 	engine := wasmtime.NewEngineWithConfig(cfg)
 
+	if !modCfg.IsUncompressed {
+		rdr := brotli.NewReader(bytes.NewBuffer(binary))
```
patrickhuie19, line 135: how did you choose brotli for the compression/decompression?

cedric-cordenier: We did some benchmarking between gzip + brotli (and zopfli) and brotli offered the best mix of speed + compression performance.

ettec (approved):

agparadiso (approved):

PR #780, opened 2024-09-18
[CAPPL-60] Dynamic encoder selection in OCR consensus aggregator

bolekk:
The aggregation phase can now dynamically override the default encoder setting from the workflow spec.


pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -409,10 +410,26 @@ func (r *reportingPlugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]oc
...
-			enc, err := r.r.getEncoder(id.WorkflowId)
-			if err != nil {
-				lggr.Errorw("could not retrieve encoder for workflow", "error", err)
-				continue
+			var encoder pbtypes.Encoder
+			if newOutcome.EncoderName != "" {
+				lggr.Debugw("using encoder from outcome", "encoderName", newOutcome.EncoderName, "executionID", report.Id.WorkflowExecutionId)
+				encoderConfig, err2 := values.FromMapValueProto(newOutcome.EncoderConfig)
+				if err2 != nil {
+					lggr.Errorw("could not convert desired encoder config to values.Map", "error", err2, "executionID", report.Id.WorkflowExecutionId)
```
nolag, line 418: I'm not convinced it's better to assume a fallback than to fail.

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -409,10 +410,26 @@ func (r *reportingPlugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]oc
...
+				} else {
+					encoder, err2 = r.r.getEncoderByName(newOutcome.EncoderName, encoderConfig)
+					if err2 != nil {
```
nolag, line 421: same here

nolag (approved):
Approving to unblock, the decision around fallbacks can be decided later.

PR #781, opened 2024-09-19
ccip - Add Limit method to seq num range

dimkouv:
Add `Limit` method that returns a copy of the provided `SeqNumRange` limited up to `n` elements by truncating the end.

### DIFF
--- a/pkg/types/ccipocr3/generic_types.go
+++ b/pkg/types/ccipocr3/generic_types.go
@@ -62,6 +62,27 @@ func (s *SeqNumRange) SetEnd(v SeqNum) {
 	s[1] = v
 }
 
+// Limit returns a range limited up to n elements by truncating the end if necessary.
+// Example: [1 -> 10].Limit(5) => [1 -> 5]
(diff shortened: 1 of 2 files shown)


pkg/types/ccipocr3/generic_types.go
```diff
@@ -62,6 +62,27 @@ func (s *SeqNumRange) SetEnd(v SeqNum) {
 	s[1] = v
 }
 
+// Limit returns a range limited up to n elements by truncating the end if necessary.
+// Example: [1 -> 10].Limit(5) => [1 -> 5]
+func (s *SeqNumRange) Limit(n uint64) SeqNumRange {
+	limitedRange := NewSeqNumRange(s.Start(), s.End())
+
+	numElems := s.End() - s.Start() + 1
```
0xnogo, line 70: nit: I wonder start > end. It might add an extra safety to return an empty range. But not sure if this case is possible

0xnogo (approved):

makramkd (approved):

PR #784, opened 2024-09-19
[chore] Handle aliases in slices

cedric-cordenier:
## Updates

- Support unwrapping a list of bytes in values.Value to a list of alias types

```
// For example
type aliasUInt8 uint8

unwrapTo: []bytes to []aliasUInt8
```


pkg/values/value.go
```diff
@@ -278,10 +278,36 @@ func unwrapTo[T any](underlying T, to any) error {
...
+			for i := 0; i < rUnderlying.Len(); i++ {
+				el := rUnderlying.Index(i)
+				toEl := reflect.New(rToVal.Type().Elem())
+				err := unwrapTo(el.Interface(), toEl.Interface())
+				if err != nil {
+					return err
+				}
+
+				if rToVal.Len() > i {
+					rToVal.Index(i).Set(reflect.Indirect(toEl))
```
kidambisrinivas, line 301: Does this require 'to' to be initialised with the right list size?

pkg/values/value.go
```diff
@@ -278,10 +278,36 @@ func unwrapTo[T any](underlying T, to any) error {
...
+			reflect.Indirect(rTo.Convert(underlyingPtr)).Set(rUnderlying)
+			return nil
+		}
+
+		rToVal := reflect.Indirect(rTo)
+		if rToVal.Kind() == reflect.Slice && rUnderlying.Kind() == reflect.Slice {
+			for i := 0; i < rUnderlying.Len(); i++ {
+				el := rUnderlying.Index(i)
+				toEl := reflect.New(rToVal.Type().Elem())
+				err := unwrapTo(el.Interface(), toEl.Interface())
```
kidambisrinivas, line 295: Aha! Nice trick to recursively call unwrapTo and handle. In the base case where unwrapTo is called for each element, how is the following call to Set not throwing the same error:

```
reflect.Indirect(rTo.Convert(underlyingPtr)).Set(rUnderlying)
```

Error: 
```
panic: reflect: reflect.Value.Set using unaddressable value [recovered]
panic: reflect: reflect.Value.Set using unaddressable value
```

pkg/values/value_test.go
```diff
@@ -382,3 +382,71 @@ func Test_Copy(t *testing.T) {
...
+				}
+
+				return to
+			},
+		},
+		{
+			name:    "aliasMap -> map[string]any",
+			val:     func() any { return map[string]any{} },
+			alias:   func() any { return aliasMap{} },
+			convert: func(a any) any { return map[string]any(a.(aliasMap)) },
```
kidambisrinivas, line 431: Nice, so the unwrap method works for aliases of complex types like map[string]any as well

kidambisrinivas (approved):
LGTM

pkg/values/value.go
cedric-cordenier: Hey! Yeah, we avoid that problem by creating the pointer on line 294

kidambisrinivas: Ah ok. So thats what we were missing yesterday! Thanks!

kidambisrinivas (dismissed):

kidambisrinivas (approved):

agparadiso (approved):

PR #790, opened 2024-09-21
pkg/workflows/sdk: add WorkflowSpec.FormatChart for mermaid flowcharts

jmank88:
This PR adds `WorkflowSpec.FormatChart` for visualizing workflows as mermaid flowcharts.


pkg/workflows/sdk/testdata/fixtures/charts/notstreamssepolia.md
jmank88, line 1: Rendering each input as a separate edge is not very display friendly. I will consolidate to a single link.

pkg/workflows/sdk/testutils/utils.go
jmank88, line 1: The diff was comparing each as one single enormous line, so just pretty-printing for readability.

pkg/workflows/sdk/builder.go
```diff
@@ -119,6 +138,16 @@ func NewWorkflowSpecFactory(
...
 func (step *Step[O]) AddTo(w *WorkflowSpecFactory) CapDefinition[O] {
 	stepDefinition := step.Definition
+
+	if w.serialMode {
+		// ensure we depend on each previous step
+		for _, prevRef := range w.prevRefs {
+			if !stepDefinition.Inputs.HasRef(prevRef) {
+				stepDefinition.Condition = fmt.Sprintf("$(%s.success)", prevRef)
+			}
+		}
```
cedric-cordenier, line 148: Since it's serial you could just depend on the step right before it

jmank88: That is true if serial mode is left from the start, but if they just flipped it on, then we have to wait for any async nodes that were added since flipping off serial.

jmank88:
I split out the builder changes to #821, so this PR is now just flowcharts for the existing types.

pkg/workflows/sdk/workflow_spec.go
```diff
@@ -1,12 +1,129 @@
 package sdk
 
-import "github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+import (
+	"cmp"
+	"iter"
+	"slices"
+	"strings"
+	"text/template"
```
nolag, line 8: I'm a little worried about pulling in text/template to our WASM binaries.

Could we make helpers who live outside of the SDK to do this work?

pkg/workflows/sdk/workflow_spec.go
```diff
@@ -35,3 +152,57 @@ func (w *WorkflowSpec) Steps() []StepDefinition {
 	s = append(s, w.Targets...)
 	return s
 }
+
+func (w *WorkflowSpec) FormatChart() (string, error) {
```
cedric-cordenier, line 156: Are we duplicating logic here? Why don't we just parse the graph and execute the template from that? A lot of the logic we've added here will reimplement the parsing logic.

pkg/workflows/sdk/workflow_spec.go
```diff
@@ -1,12 +1,129 @@
...
 type StepInputs struct {
 	OutputRef string
 	Mapping   map[string]any
 }
 
+type Output struct {
+	Ref, Name string
+}
+
+func (i *StepInputs) HasRef(ref string) bool {
```
cedric-cordenier, line 22: Are we calling this from anywhere?

pkg/workflows/sdk/workflow_spec.go
```diff
@@ -35,3 +152,57 @@ func (w *WorkflowSpec) Steps() []StepDefinition {
...
+	{{ $ref }}[/"{{$name}}"\]
+	{{ else -}}
+	{{ $ref }}["{{$name}}"]
+	{{ end -}}
+	{{ if .Inputs.OutputRef -}}
+	{{ .Inputs.OutputRef }} --> {{ $step.Ref }}
+	{{ else -}}
+		{{ range $out := .Inputs.Outputs -}}
+			{{ if $out.Name -}}
+	{{ $out.Ref }} -- {{ $out.Name }} --> {{ $ref}}
```
cedric-cordenier, line 201: I think this is where we indicate that we're doing outputs.Value, right? If so, can we also represent the fact that we also shaping the incoming inputs?

Eg. for consensus we actually do:
{"observations": []any{"outputs.Value"}}

pkg/workflows/sdk/workflow_spec.go
jmank88: Ah, not anymore. Will move to #821

pkg/workflows/sdk/workflow_spec.go
jmank88: I was worried about including everything, and it's not obvious to me where to draw the line :shrug: 
We also have other options for presenting the data, like triggering actions on click or tool-tips on hover.
Or something way different, like subgraphs for each step, with multiple internal nodes:
```mermaid
flowchart
	subgraph event-trigger
		trigger-->event-outputs[outputs]
	end
	subgraph compute-a-step[compute a]
		compute-a-inputs[TriggerOutputs] --> compute-a[Execute] --> compute-a-outputs[Value]
	end
	subgraph compute-b-step[compute b]
		compute-b-inputs[TriggerOutputs] --> compute-b[Execute] --> compute-b-outputs[Value]
	end
	subgraph report-step[report]
		report[Build Report]
		report-inputs-compute-a[Value] --> report
		report-inputs-compute-b[Value] --> report
		report --> report-outputs[Report]
	end
	subgraph write-step[write]
		write-inputs[Report] --> write
	end
	event-outputs --> compute-a-inputs
	event-outputs --> compute-b-inputs
	compute-a-outputs --> report-inputs-compute-a
	compute-b-outputs --> report-inputs-compute-b
	report-outputs --> write-inputs
```
This level of granularity would allow us to express more visually, and utilize tool tips for more information, since they work on nodes but not edges.

pkg/workflows/sdk/workflow_spec.go
jmank88: Probably :shrug: I wasn't looking outside of this package. Which type is best suited for this?

pkg/workflows/sdk/workflow_spec.go
jmank88: Sure, as long as everything we need is exported, then it doesn't matter where it lives.

jmank88: Moved to workflows package as a method on `DependencyGraph`

pkg/workflows/sdk/workflow_spec.go
jmank88: Moved to a method on `DependencyGraph` but needs more refactoring to use the graph instead of custom code.

jmank88:
I swapped over to using the `DependencyGraph.Graph`, but it appears to be incomplete so I need to figure out how to augment it with the trigger dependencies. Back to draft for now.

jmank88:
> Hey @jmank88 , I am trying to figure out how we provide the visual to users via CLI. From the dev platform CLI perspective,
> 
> * We get the paths of local files from users and get the wasm binary and config as byte[]
> * We can get an instance of WorkflowSpec via [GetWorkflowSpec](https://github.com/smartcontractkit/chainlink-common/blob/b7b7f6310ac2815acf7a4deae0d3c5e33ac349d9/pkg/workflows/wasm/host/wasm.go#L15)
> * We can call something like GetChart() to get the visual? Is the visual a SVG or image to the local?
> 
> Want to hear your thoughts on the flow, and how to integrate. Thanks! This comment is a copy-pasta from [link](https://github.com/smartcontractkit/chainlink-common/pull/821#discussion_r1803613611)

I refactored the method to be on `DependencyGraph` instead, but yes you can just call `FormatChart()`. However,  it returns text to be interpreted by Mermaid for visualization, so we'd have to call in to that library for rendering to SVG or anything else.

shileiwill:
Returning text and we (client) render with Mermaid sounds great to me. This is gonna be a very useful feature from Dev Platform side. 

> But we might not be pursuing this direction any longer
May I know the new direction?

PR #797, opened 2024-09-24
[BCF-3392]  - ContractReaderByIDs Wrapper

ilija42:
### Adding a wrapper for this feature instead of making interface changes since not all use cases might want this

- Main concern is double versioning since ContractReader changes might break this interface so we would have to version both CR and this wrapper. Right now there are no further planned changes to the interface, so at least we should be safe until new chain family impl. starts.
- Had to change the fakeChainReader tester to support multiple addresses for same contract


pkg/chainreader/contract_reader.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ContractReaderByIDs wraps types.ContractReader to allow the caller to set custom contractIDs with Bind.
+type ContractReaderByIDs interface {
+	// Bind accepts a map of bound contracts where map keys are custom contractIDs to be used for calling the interface.
+	Bind(ctx context.Context, bindings map[string]types.BoundContract) error
+	Unbind(ctx context.Context, bindings map[string]types.BoundContract) error
+	GetLatestValue(ctx context.Context, contractID, readName string, confidenceLevel primitives.ConfidenceLevel, params, returnVal any) error
+	BatchGetLatestValues(ctx context.Context, request BatchGetLatestValuesRequestByCustomID) (BatchGetLatestValuesResultByCustomID, error)
+	QueryKey(ctx context.Context, contractID string, filter query.KeyFilter, limitAndSort query.LimitAndSort, sequenceDataType any) ([]types.Sequence, error)
+}
```
winder, line 21: Rather than create an interface, I think you can make `contractReaderByIDs` public

pkg/chainreader/contract_reader.go
```diff
@@ -0,0 +1,111 @@
...
+func WrapContractReaderByIDs(contractReader types.ContractReader) ContractReaderByIDs {
+	return &contractReaderByIDs{
+		cr: contractReader,
+	}
+}
+
+type contractReaderByIDs struct {
+	bindings sync.Map
+	cr       types.ContractReader
+}
```
winder, line 33: Does this need to embed `UnimplementedContractReader`?

pkg/chainreader/contract_reader.go
```diff
@@ -0,0 +1,111 @@
...
+	return crByIds.cr.Bind(ctx, toBind)
+}
+
+func (crByIds *contractReaderByIDs) Unbind(ctx context.Context, bindings map[string]types.BoundContract) error {
+	var toUnbind []types.BoundContract
+	for customContractID, boundContract := range bindings {
+		crByIds.bindings.Delete(customContractID)
+		toUnbind = append(toUnbind, boundContract)
+	}
+	return crByIds.cr.Unbind(ctx, toUnbind)
```
winder, line 53: If there's an error, could `crByIds` be in a bad state?

pkg/chainreader/contract_reader.go
```diff
@@ -0,0 +1,111 @@
...
+		wrappedRes[bcToID[bc.String()]] = batchResp
+	}
+
+	return wrappedRes, nil
+}
+
+func (crByIds *contractReaderByIDs) getBoundContract(contractID string) (types.BoundContract, error) {
+	binding, ok := crByIds.bindings.Load(contractID)
+	if !ok {
+		return types.BoundContract{}, fmt.Errorf("binding not found for contractID %s", contractID)
```
winder, line 103: We've found an error type for this to be very helpful: `ErrContractReaderNotFound = errors.New("contract reader not found")`

pkg/chainreader/contract_reader.go
ilija42: Not really, `UnimplementedContractReader` is mainly for relayer impl. to have a stub so that the build doesn't break. If there is an interface change to Contract Reader, chainlink-common won't be mergeable before `contractReaderByIDs` calls to ContractReader are updated. There is a separate concern here though, if the interface changes are breaking then the wrapper methods will have to be modified too, which means that we would have to version both the ContractReader interface and this wrapper.

pkg/chainreader/contract_reader.go
ilija42: I think with the current EVM impl. even in the worst scenario (LP having a dangling filter tracking a contract for no reason) this would be fine. I changed it though to future proof a bit more, we should make Unbind completely transactional so that the behaviour is easier to predict and handle when we get to multiple chain impl.

EasterTheBunny (approved):
LGTM

pkg/chainreader/contract_reader_test.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	require.NoError(t, crByIDs.Bind(ctx, map[string]types.BoundContract{bc1CustomID: bc1, bc2CustomID: bc2}))
+
+	// Mock Unbind function for error case
+	mockReader.unbindFunc = func(ctx context.Context, contracts []types.BoundContract) error {
+		return fmt.Errorf("some error")
+	}
+
+	// Test Unbind with error shouldn't remove bindings
+	require.Error(t, crByIDs.Unbind(ctx, map[string]types.BoundContract{bc1CustomID: bc1}))
```
patrickhuie19, line 44: /nit is this slightly clearer for the reader if the map is empty?
```suggestion
	require.Error(t, crByIDs.Unbind(ctx, map[string]types.BoundContract{}))
```

patrickhuie19: Ah, no, I think it's clearer as is given the Test successful Unbind case

patrickhuie19 (approved):

patrickhuie19:
I don't know if I fully grok the reason for the feature request, but the change looks good 👍

Farber98 (approved):

winder (approved):

PR #807, opened 2024-09-27
Extract expirable cache abstraction for reuse

ettec:
There's a requirement in the ReadContract capability (sits in the capability repo, has no dependencies on core) to cache with the same behaviour as the cache in this PR -> https://github.com/smartcontractkit/chainlink/pull/14496.    

This PR extracts the cache into a generic cache for use across both use cases.

fyi @cedric-cordenier

### DIFF
--- a/pkg/utils/cache/expirable_cache.go
+++ b/pkg/utils/cache/expirable_cache.go
@@ -1,0 +1,139 @@
+package cache
+
+import (
+	"sync"
+	"time"
+
+	"github.com/jonboulle/clockwork"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
+
+type StatsCollector interface {
+	OnCacheHit()
+	OnCacheMiss()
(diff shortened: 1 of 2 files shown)


george-dorin (approved):

kidambisrinivas (dismissed):

ettec:
> @ettec Can we chat about this? As I've expressed before, "a little duplication is better than a little dependency", and this very much adds a dependency between two otherwise unrelated capabilities.
> 
> The caching logic may be the same today, but there's no guarantee that it will be in the future, and this has also added a repo hop to the capability, with all of the associated pain that that brings (in fact, we'll have 3 repo hops if we change the interface itself). In short, IMO we are better off just copying the code at this stage.

sure, it seems fairly complex code to me and I would argue its not a little duplication, lets catch up on Monday, I won't merge it

kidambisrinivas (approved):

cedric-cordenier (approved):

PR #811, opened 2024-09-27
created integration test that asserts cursor functionality

EasterTheBunny:


silaslenihan (dismissed):

vyzaldysanchez (dismissed):

silaslenihan (approved):

pkg/types/interfacetests/chain_components_interface_tests.go
```diff
@@ -794,6 +794,57 @@ func runQueryKeyInterfaceTests[T TestingT[T]](t T, tester ChainComponentsInterfa
...
+						}
+
+						limit.Limit = query.CursorLimit(sequences[len(sequences)-1].Cursor, query.CursorFollowing, 2)
+						allSequences = append(allSequences, sequences...)
+					}
+
+					return len(allSequences) == len(testStructs) &&
+						reflect.DeepEqual(&testStructs[0], allSequences[0].Data) &&
+						reflect.DeepEqual(&testStructs[len(testStructs)-1], allSequences[len(testStructs)-1].Data)
+				}, tester.MaxWaitTimeForEvents(), 50*time.Millisecond)
```
jmank88, line 845: There's no need to check this often, especially with such complex code:
```suggestion
				}, tester.MaxWaitTimeForEvents(), 500*time.Millisecond)
```

pkg/types/interfacetests/chain_components_interface_tests.go
```diff
@@ -794,6 +794,57 @@ func runQueryKeyInterfaceTests[T TestingT[T]](t T, tester ChainComponentsInterfa
 				}, tester.MaxWaitTimeForEvents(), time.Millisecond*10)
```
jmank88, line 794: This is out of scope, but 10ms seems way too fast:
```suggestion
				}, tester.MaxWaitTimeForEvents(), time.Millisecond*500)
```
There is typically little payoff to doing these checks so fast. You risk keeping the CPU so busy that it slows down or cannot even make progress on the test, and the best case payoff is only to finish earlier by the max difference in the durations (assuming parallel tests).

jmank88 (dismissed):

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
```diff
@@ -621,21 +622,39 @@ func (f *fakeContractReader) QueryKey(_ context.Context, bc types.BoundContract,
...
+			}
+
+			if (len(filter.Expressions) == 0 || doAppend) && !skipAppend {
 				if isValueType {
 					value, err := values.Wrap(trigger.testStruct)
 					if err != nil {
 						return nil, err
 					}
-					sequences = append(sequences, types.Sequence{Data: &value})
+					sequences = append(sequences, types.Sequence{Cursor: fmt.Sprintf("%d", idx), Data: &value})
```
jmank88, line 645: Could shorten these a bit?
```suggestion
					sequences = append(sequences, types.Sequence{Cursor: strconv.Itoa(idx), Data: &value})
```

silaslenihan (approved):

jmank88 (approved):

Farber98 (approved):

ilija42 (commented):
Looks good, but there are failing tests

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
```diff
@@ -337,7 +337,7 @@ func (it *fakeContractReaderInterfaceTester) GenerateBlocksTillConfidenceLevel(t
 }
 
 func (it *fakeContractReaderInterfaceTester) MaxWaitTimeForEvents() time.Duration {
-	return time.Millisecond * 100
+	return time.Second * 10
```
ilija42, line 340: why, this is a 100x increase?

ilija42: Approved, but we should look into optimising these. Which test was this?

ilija42 (approved):

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go
silaslenihan: Any test that uses `MaxWaitTimeForEvents()` was failing in this PR before I made this change

silaslenihan (approved):

PR #813, opened 2024-09-30
CCIP-3555 Attestation encoder interfaces

mateusz-sekara:
Please see https://github.com/smartcontractkit/chainlink-ccip/pull/186 for details


pkg/types/ccipocr3/interfaces.go
```diff
@@ -31,3 +31,10 @@ type RMNCrypto interface {
 		signerAddresses []Bytes,
 	) error
 }
+
+// TokenDataEncoder is a generic interface for encoding offchain token data for different chain families.
+// Right now it supports only USDC/CCTP, but every new token that requires offchain data processsing
+// should be added to that interface and implemented in the downstream repositories (e.g. chainlink-ccip, chainlink).
+type TokenDataEncoder interface {
```
mateusz-sekara, line 38: It must be proto-compatible, so this interface must implement every supported token (we can't make it use generics). Considering that USDC is the only token supported now we shouldn't expect this interface to grow indefinitely

asoliman92 (approved):

pkg/types/ccipocr3/interfaces.go
makramkd: It should be alright, the amount of _unique_ attestation setups is probably pretty small, we're already planning on using a USDC-esque setup for another wrapped BTC token.

makramkd (approved):

PR #814, opened 2024-09-30
[CAPPL-41] SDK Fetch import

agparadiso:
### Description

This pr introduces the Fetch import, this will allow the wasm binary to perform http requests. The scope of this task is building the SDK side, in order to have the full functionality this still needs to be connected to the gateway. this will be tackled on a separate task


pkg/workflows/wasm/host/module.go
```diff
@@ -88,6 +89,7 @@ type ModuleConfig struct {
 	InitialFuel    uint64
 	Logger         logger.Logger
 	IsUncompressed bool
+	Fetch          func(wasm.TargetRequestPayload) (wasm.TargetResponsePayload, error)
```
cedric-cordenier, line 92: Nit can we name this FetchRequest and FetchResponse: Target is a type of capability and is not relevant in this context

pkg/workflows/wasm/host/module.go
```diff
@@ -236,6 +242,55 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
...
 	}
 
+	err = linker.FuncWrap(
+		"env",
+		"fetch",
+		func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
+			b, innerErr := safeMem(caller, reqptr, reqptrlen)
+			if innerErr != nil {
+				logger.Errorf("error calling fetch: %s", innerErr)
+				return 1 //errnumber
```
cedric-cordenier, line 252: Can we define an err to indicate a failure with a Fetch call, see: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/workflows/wasm/runner.go#L21 for where we do this kind of thing

pkg/workflows/wasm/host/module.go
```diff
@@ -236,6 +242,55 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
...
+			}
+
+			req := wasm.TargetRequestPayload{}
+			innerErr = json.Unmarshal(b, &req)
```
cedric-cordenier, line 256: @agparadiso Could we use proto here? We use this elsewhere for our the guest <> host SDK so we should be consistent.

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -75,6 +79,9 @@ func Test_GetWorkflowSpec(t *testing.T) {
 	spec, err := GetWorkflowSpec(
 		&ModuleConfig{
 			Logger: logger.Test(t),
+			Fetch: func(req wasm.TargetRequestPayload) (wasm.TargetResponsePayload, error) {
```
cedric-cordenier, line 82: @agparadiso GetWorkflowSpec shouldn't need to provide a Fetch implementation: can we modify the passed in config with a noop'd implementation of Fetch instead? The Fetch API shouldn't be available when making a GetWorkflowSpec call

pkg/workflows/wasm/runner_wasip1.go
```diff
@@ -51,7 +58,37 @@ func NewRunner() *Runner {
...
+			Logger: l,
+			Fetch: func(req TargetRequestPayload) TargetResponsePayload {
+
+				b, err := json.Marshal(req)
+				if err != nil {
+					os.Exit(CodeRunnerErr)
+				}
+				reqptr, reqptrlen := bufferToPointerLen(b)
+
+				respBuffer := make([]byte, 1024*5)
```
cedric-cordenier, line 70: We should try to make this configurable from the host somehow, probably by including it in the Compute request args. As it stands this will be hardcoded in any binaries which could be painful if a binary exceeds this limit. By passing it from the host it'll at least be configurable even if we just hardcode it there 🙈

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -168,6 +183,158 @@ func Test_GetWorkflowSpec_Logs(t *testing.T) {
 	}
 }
 
+func Test_Compute_Fetch(t *testing.T) {
+	binary := createTestBinary(fetchBinaryCmd, fetchBinaryLocation, true, t)
+
+	t.Run("OK_default_runtime_cfg", func(t *testing.T) {
+		logger, logs := logger.TestObserved(t, zapcore.InfoLevel)
+		expected := "Valid fetch response"
```
cedric-cordenier, line 191: @agparadiso This approach works for sure, but you run the risk of having brittle tests if you're checking logs to verify that something is happening or not.

As an alternative, you could transmit the response back to the host, and check against that in the host.

So in other words, your workflow would:

- Compute: make the request, return the response
- Then you get that response back on the host side from the call to Run
- if that response == expected you know the test has worked.

cedric-cordenier (dismissed):

pkg/workflows/wasm/host/module.go
```diff
@@ -236,6 +244,55 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
...
+			size = copyBuffer(caller, resplenBytes, resplenptr, uint32Size)
+			if size == -1 {
+				return ErrnoFault
+			}
+
+			return ErrnoSuccess
+		},
+	)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
agparadiso: this error would only happen if the link.FuncWrap() fails, the errors inside the actual `fetch` would be received in the guest side and would also be logged. 
I can add something like: `failed to wrap Fetch function` and probably add it for `log` and `sendResponse` for consistency. Would that make sense?

pkg/workflows/sdk/testutils/runtime.go
```diff
@@ -1,9 +1,19 @@
 package testutils
 
 import (
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
 )
 
-type runtime struct{}
+type NopRuntime struct{}
```
MStreet3, line 8: _nit_ `NoOp` instead of `Nop` as this was initially confusing with `Node Operator`

pkg/workflows/wasm/host/module.go
```diff
@@ -236,6 +244,55 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
...
+				logger.Errorf("error calling fetch: %s", innerErr)
+				return ErrnoFault
+			}
+
+			req := &wasmpb.FetchRequest{}
+			innerErr = proto.Unmarshal(b, req)
+			if innerErr != nil {
+				logger.Errorf("error calling fetch: %s", innerErr)
+				return ErrnoFault
+			}
```
MStreet3, line 262: _nit_ move wrapped function out of the constructor and in to it's own function

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -168,6 +185,152 @@ func Test_GetWorkflowSpec_Logs(t *testing.T) {
...
+		require.NoError(t, err)
+		err = r.Value.Underlying["Value"].UnwrapTo(&actual)
+		require.NoError(t, err)
+
+		assert.Equal(t, expected, actual)
+	})
+
+	t.Run("NOK_fetch_error_returned", func(t *testing.T) {
+		logger, logs := logger.TestObserved(t, zapcore.InfoLevel)
+		expectedErr := fmt.Errorf("test-error")
```
MStreet3, line 292: _nit_ `assert.AnError` from the `assert` package

pkg/workflows/wasm/host/module.go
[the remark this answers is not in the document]
agparadiso: added 👍🏼

cedric-cordenier (approved):

cedric-cordenier (approved):

ettec (approved):

PR #815, opened 2024-09-30
Allow the creation of maps from string to capability outputs.

nolag:
The current use case could be Beholder, if they want an explicit capability step.  This could be useful as when Beholder is called from within compute, there's no consensus yet.


ilija42 (approved):
Lack the context, but the implementation looks good

PR #816, opened 2024-10-01
(feat): Add Median Aggregator to OCR3 capability

justinkaseman:
### Description
Add Median Aggregator that takes in the output of an Action
```
type CapabilityResponse struct {
    Value *values.Map
}
```

and returns a single signed report with

```
Reports: 
[
  {
  FeedID: []byte
  RawReport:  nil,
  Price:      big.Int,
  Timestamp:  time.Time.Unix(),
  RemappedID: string
  },
...
]
```

### Caveats
- Input should have observations data as bytes, but this is not enforced. Open to ideas here.
- Currently supports only a single value, but should be straight forward to extend by providing multiple ValueKeys in config. and then return multiple reports
- This will need to be merged for a proper integration test within the Chainlink repo


pkg/capabilities/consensus/ocr3/median/median_aggregator.go
```diff
@@ -0,0 +1,256 @@
...
+	// Aggregator outputs reports in the following format:
+	//   []Reports{FeedID []byte, RawReport []byte, Price *big.Int, Timestamp int64}
+	// Example of a compatible EVM encoder ABI config:
+	//   (bytes32 FeedID, bytes RawReport, uint256 Price, uint64 Timestamp)[] Reports
+	TopLevelListOutputFieldName = "Reports"
+	FeedIDOutputFieldName       = "FeedID"
+	RawReportOutputFieldName    = "RawReport"
+	PriceOutputFieldName        = "Price"
+	TimestampOutputFieldName    = "Timestamp"
+	RemappedIDOutputFieldName   = "RemappedID"
```
cedric-cordenier, line 36: @justinkaseman The actual implementation of this shouldn't have any data feeds specific concepts in it, and shouldn't make any assumptions about the shape of the observation.

pkg/capabilities/consensus/ocr3/median/median_aggregator.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+type aggregatorConfig struct {
+	ValueKey        string
+	Deviation       decimal.Decimal    `mapstructure:"-"`
+	DeviationString string             `mapstructure:"deviation"`
+	FeedID          datastreams.FeedID `mapstructure:"feedId"`
+	Heartbeat       int
+	RemappedID      []byte `mapstructure:"-"`
+	RemappedIDHex   string `mapstructure:"remappedId"`
```
cedric-cordenier, line 46: FeedID, RemappedID and RemappedID hex are config options that are specific to data feeds, we shouldn't need them here.

pkg/capabilities/consensus/ocr3/median/median_aggregator.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+type medianAggregator struct {
+	clock       clockwork.Clock
+	config      aggregatorConfig
+	reportCodec datastreams.ReportCodec
```
cedric-cordenier, line 52: Same comment about the reportCodec

pkg/capabilities/consensus/ocr3/median/median_aggregator.go
```diff
@@ -0,0 +1,256 @@
...
+		// we only expect a single observation per node
+		if len(nodeObservations) == 0 || nodeObservations[0] == nil {
+			a.lggr.Warnf("node %d contributed with empty observations", nodeID)
+			continue
+		}
+		if len(nodeObservations) > 1 {
+			a.lggr.Warnf("node %d contributed with more than one observation", nodeID)
+			continue
+		}
+		payload := &capabilities.CapabilityResponse{}
```
cedric-cordenier, line 164: This isn't correct, the observation would just be a values.Value but we wouldn't know exactly what it is, so we'll need to do some iterating over the types (or alternatively we can assume it's a map if users provide an aggregationKey).

pkg/capabilities/consensus/ocr3/median/median_aggregator.go
```diff
@@ -0,0 +1,256 @@
...
+	oldFl, _ := oldPrice.Float64()
+	return diffFl / oldFl
+}
+
+func median(items [][]byte) []byte {
+	sort.Slice(items, func(i, j int) bool {
+		if len(items[i]) != len(items[j]) {
+			// NOTE: this doesn't account for extra leading zeros
+			return len(items[i]) < len(items[j])
+		}
```
MStreet3, line 204: why are you comparing the lengths here, is `bytes.Compare` not sufficient?

PR #818, opened 2024-10-01
[BCFR-147][common] - Add codec chain agnostic modifier for converting byte array address to string

Farber98:
### **Description**  
This PR introduces a chain-agnostic solution for transforming address byte arrays into string representations. To do so, it uses a modifier that supports chain-specific logic.  The modifier is injected during relayer contractReader runtime. 

### **Key Changes**
- **AddressModifier Interface:** Provides an abstraction for encoding/decoding addresses across different chains. Will be implemented in each relayer with corresponding chain-specific logic.

```
type AddressModifier interface {
    EncodeAddress([]byte) (string, error)
    DecodeAddress(string) ([]byte, error)
    Length() int
}
```

- **AddressBytesToStringModifierConfig**: This config struct skips the modifier in JSON serialization to avoid serialization issues and it will be injected in runtime.
```
type AddressBytesToStringModifierConfig struct {
    Fields   []string
    Modifier AddressModifier `json:"-"` 
}
```

### Resources
- [BCFR-147](https://smartcontract-it.atlassian.net/browse/BCFR-147)

### Related PRs
- https://github.com/smartcontractkit/chainlink-solana/pull/875
- https://github.com/smartcontractkit/chainlink/pull/14620

[BCFR-147]: https://smartcontract-it.atlassian.net/browse/BCFR-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

core ref: bcfr-147-add-address-to-string-modifier
solana ref: bcfr-147-add-address-to-string-modifier


go.mod
```diff
@@ -84,7 +84,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.4 // indirect
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
-	github.com/mr-tron/base58 v1.2.0 // indirect
+	github.com/mr-tron/base58 v1.2.0
```
jmank88, line 87: Please move up to the direct block. (I'm surprised that tidy didn't do this)

pkg/codec/byte_string_modifier.go
```diff
@@ -0,0 +1,466 @@
...
+	Base58Encoding EncodingType  = "base58"
+	// General
+	NoneChecksum ChecksumType = "none"
+)
+
+type AddressChecksum func([]byte) []byte
+
+// EIP55AddressChecksum Applies EIP55 checksum logic to the address
+// assuming input 'a' is already in hex form without the "0x" prefix
+func EIP55AddressChecksum(a []byte) []byte {
```
jmank88, line 36: How is this chain-agnostic? The gold standard is that we are able to integration new chains, only by authoring a LOOP Plugin, and without having to update any other modules, like chainlink/ or this one. This code looks _universal_ to me, meaning that it knows about every chain. That doesn't meet our requirements.

jmank88 (changes requested):
This repo must remain chain-agnostic, meaning that it must not contain any chain specifics at all, in order to meet our requirements. What problem are we trying to solve here?

pkg/codec/byte_string_modifier.go
Farber98: Thought the requirement was not importing libraries like `go-ethereum` directly.

Removing the `Checksum` part would make this development comply with the standards? Length and encoding seem something general that could be reused by different blockchains

jmank88: I think we need to zoom out. What problem are we trying to solve? This repository is imported by core, all products, and all chains. Core and all products are chain-agnostic, and so cannot use this. Where would it be called from?

ilija42: We want to write a codec modifier that parse addresses into strings, the issue is that all of the previous modifiers aren't doing anything chain specific and live in common. The modifiers get created based on chain agnostic codec configs [here](https://github.com/smartcontractkit/chainlink-common/blob/e1435d915916aef78e985938f9cc98fe501507f2/pkg/codec/config.go#L41), which directly initialise a chain agnostic modifier struct [here](https://github.com/smartcontractkit/chainlink-common/blob/e1435d915916aef78e985938f9cc98fe501507f2/pkg/codec/config.go#L184). 

One solution that I think could work(not completely sure), would be to inject a chain specific implementation of a modifier during Contract Reader init, just leave some kind of an empty interface in the modifier in common and set it when Contract Reader is initialised in the relayer.

Anothe solution is to remove change this modifier to support all of these formats(I just put together a random list):
- EVM is        20 bytes Base16 (Hexadecimal)
- Solana is     32 bytes Base58
- Starknet is  32 bytes Base16 (Hexadecimal)
-  Cosmos is  20 bytes Bech32
- Aptos is       32 bytes Base16 (Hexadecimal)
- Polkadot is  32 bytes SSBase58
- Cardano is   32 bytes Bech32
- Tezos is       20 bytes Base58Check
- Algorand is  32 bytes Base32
- Avalanche is 20 bytes Bech32
- Tron is           20 bytes Base58Check
- Near is          32 bytes Base58

I don't know cryptography well enough to determine if this approach makes sense and if chains have a reason to introduce new encodings

jmank88: > One solution that I think could work(not completely sure), would be to have that config call into a modifier interface and then to inject a chain specific implementation during Contract Reader init, just leave some kind of an empty interface in the modifier in common and set it when Contract Reader is initialised in the relayer.

This sounds right. I don't see any need to parameterize anything here. The relayers know what format to expect. I think this kind of extensible API was inevitable, because the codec runs on the server side.

pkg/codec/byte_string_modifier.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// addressToStringHookForOffChain converts byte arrays to their string representation for off-chain use.
+func addressToStringHookForOffChain(modifier AddressModifier) func(from reflect.Type, to reflect.Type, data any) (any, error) {
+	return func(from reflect.Type, to reflect.Type, data any) (any, error) {
+		byteArrTyp := reflect.ArrayOf(modifier.Length(), reflect.TypeOf(byte(0)))
+		strTyp := reflect.TypeOf("")
+
+		// if 'from' is a pointer to the byte array (e.g., *[20]byte), dereference it.
+		if from.Kind() == reflect.Ptr && from.Elem() == byteArrTyp {
```
ilija42, line 284: I don't 100% remember how this works, but I think we can return an error if fields aren't what we are expecting. This would mean that someone defined an unnecessary modifier right?

pkg/codec/byte_string_modifier.go
```diff
@@ -0,0 +1,307 @@
...
+func (t *bytesToStringModifier) RetypeToOffChain(onChainType reflect.Type, itemType string) (tpe reflect.Type, err error) {
+	defer func() {
+		// StructOf can panic if the fields are not valid
+		if r := recover(); r != nil {
+			tpe = nil
+			err = fmt.Errorf("%w: %v", types.ErrInvalidType, r)
+		}
+	}()
+
+	if len(t.fields) == 0 {
```
ilija42, line 73: what does this do?

pkg/codec/byte_string_modifier.go
```diff
@@ -0,0 +1,307 @@
...
+
+		t.onToOffChainType[onChainType] = offChainType
+		t.offToOnChainType[offChainType] = onChainType
+
+		return offChainType, nil
+	case reflect.Struct:
+		return t.getStructType(onChainType)
+	default:
+		return nil, fmt.Errorf("%w: cannot retype the kind %v", types.ErrInvalidType, onChainType.Kind())
+	}
```
ilija42, line 129: wdyt about this

```suggestion

	var offChainType reflect.Type
	switch onChainType.Kind() {
	case reflect.Pointer:
		// Recursively call RetypeToOffChain on the element type
		elm, err := t.RetypeToOffChain(onChainType.Elem(), "")
		if err != nil {
			return nil, err
		}

		offChainType = reflect.PointerTo(elm)
	case reflect.Slice:
		elm, err := t.RetypeToOffChain(onChainType.Elem(), "")
		if err != nil {
			return nil, err
		}

		offChainType = reflect.SliceOf(elm)
	case reflect.Array:
		addrType := reflect.ArrayOf(t.modifier.Length(), reflect.TypeOf(byte(0)))
		// Check for nested byte arrays (e.g., [20]byte)
		if onChainType.Elem() == addrType.Elem() {
			offChainType = reflect.ArrayOf(onChainType.Len(), reflect.TypeOf(""))
		} else {
			elm, err := t.RetypeToOffChain(onChainType.Elem(), "")
			if err != nil {
				return nil, err
			}

			offChainType = reflect.ArrayOf(onChainType.Len(), elm)
		}
	case reflect.Struct:
		return t.getStructType(onChainType)
	default:
		return nil, fmt.Errorf("%w: cannot retype the kind %v", types.ErrInvalidType, onChainType.Kind())
	}
	
	t.onToOffChainType[onChainType] = offChainType
	t.offToOnChainType[offChainType] = onChainType
	return offChainType, nil
```

ilija42 (commented):
There is a lot of comments that just explain what the code is doing, I would only leave these around the most confusing reflection code, everything else should be self explanatory from code, comment the why, not the what

pkg/codec/byte_string_modifier.go
Farber98: return the original type without performing any transformation, because no fields require modification

pkg/codec/byte_string_modifier.go
Farber98: It was in case we received a pointer to the array. I think we can just grab the underlying value and proceed

pkg/codec/byte_string_modifier.go
ilija42: makes sense, would it make sense to return an error here?

pkg/codec/byte_string_modifier.go
ilija42: I'm saying that if `from` is not of any type that we are expecting should we return an error?

pkg/codec/byte_string_modifier.go
```diff
@@ -114,97 +105,92 @@ func (t *bytesToStringModifier) RetypeToOffChain(onChainType reflect.Type, itemT
...
-			// Use reflection to extract the underlying value
 			rVal := reflect.ValueOf(val)
-			if rVal.Kind() == reflect.Ptr {
-				// Dereference the pointer if necessary
+
+			if !rVal.IsValid() {
+				return fmt.Errorf("invalid value for field %s", fieldName)
+			}
+
+			if rVal.Kind() == reflect.Ptr && !rVal.IsNil() {
```
ilija42, line 147: is nil a valid value here, if not the check above should check for both is nil and isValid and return an error

pkg/codec/byte_string_modifier.go
Farber98: It makes sense to me, I think it will help avoiding confusions. However, the [modifierBase RetypeToOffChain](https://github.com/smartcontractkit/chainlink-common/blob/8166e6555b2330ba0abfc747a6a78d0685394d89/pkg/codec/modifier_base.go#L30-L34) is not returning an error under this condition. Do you want me to change it there too?

Farber98: I think I found a way to reuse the modifier base functionality. Let me try something and I'll push some changes

pkg/codec/byte_string_modifier.go
Farber98: Realized this is not needed so I'll just remove it

pkg/codec/byte_string_modifier.go
Farber98: may panic if we put it in the above line. Makes sense to first check if  `rVal.Kind() == reflect.Ptr` before calling `IsNil`

pkg/codec/byte_string_modifier.go
Farber98: Removed the check from the new modifier as we will rely on the `modiferBase RetypeToOffChain` now. It seems better to leave that condition as it is in the `modifierBase` for this development as other modifiers start to fail. 

I can create a new ticket to check this later and see if it makes sense for all the modifiers to have an error there

ilija42 (approved):

silaslenihan (approved):
LGTM, Nicely done Juan!

EasterTheBunny (approved):

jmank88 (approved):

PR #823, opened 2024-10-02
Fix map and ToListDefinition, add tests for them in the builder, add a way to create a list of any from inputs

nolag:


pkg/workflows/sdk/builder.go
```diff
@@ -79,7 +90,15 @@ type singleCapList[O any] struct {
 }
 
 func (s singleCapList[O]) Index(i int) CapDefinition[O] {
-	return &capDefinitionImpl[O]{ref: s.CapDefinition.Ref().(string) + "." + strconv.FormatInt(int64(i), 10)}
+	listRef, ok := s.CapDefinition.Ref().(string)
+
+	// the value was from a hard-coded list, seems odd they are indexing into this
```
cedric-cordenier, line 95: I'm not following this comment: what makes you say that?

cedric-cordenier: Ah the tests make it clear -- I do feel like adding more detail to the comments would be helpful here:

You're basically saying that there are two cases to indexing here:
- It's a ref, in which case we just want to index the ref, i.e. ref -> ref.i
- It's a hardcoded list, in which case we literally want to index into it (i.e. [ref].0 -> ref)

For the latter, it doesn't feel like there's any bound protection, or is that handled elsewhere?

jmank88 (approved):

cedric-cordenier (approved):

PR #824, opened 2024-10-02
workflows: Validate workflowName is exactly 10 bytes

archseer:
Current validation allowed shorter names which would then later error in consensus phase

### DIFF
--- a/pkg/workflows/models_yaml.go
+++ b/pkg/workflows/models_yaml.go
@@ -63,7 +63,7 @@ func ParseWorkflowSpecYaml(data string) (sdk.WorkflowSpec, error) {
 type WorkflowSpecYaml struct {
 	// NOTE: Name and Owner are constrained the onchain representation in [github.com/smartcontractkit/chainlink-common/blob/main/pkg/capabilities/consensus/ocr3/types/Metadata]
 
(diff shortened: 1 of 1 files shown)


cedric-cordenier:
@archseer Hmmm... we should be padding the workflow name in consensus to get to 10. What was the error you experienced?

archseer:
Setting a shorter name will accept the workflow, then fail during workflow execution:

```
"caller":"ocr3/reporting_plugin.go:426","error":"incorrect length for id WorkflowName"
```

krehermann (approved):

pkg/workflows/models_yaml.go
```diff
@@ -63,7 +63,7 @@ func ParseWorkflowSpecYaml(data string) (sdk.WorkflowSpec, error) {
 type WorkflowSpecYaml struct {
 	// NOTE: Name and Owner are constrained the onchain representation in [github.com/smartcontractkit/chainlink-common/blob/main/pkg/capabilities/consensus/ocr3/types/Metadata]
 
-	Name string `json:"name,omitempty" jsonschema:"pattern=^[0-9A-Za-z_\\-]+$,maxLength=10"` // plain text string exactly 10 characters long, or  empty name allowed for anonymous workflows
+	Name string `json:"name,omitempty" jsonschema:"pattern=^[0-9A-Za-z_\\-]+$,minLength=10,maxLength=10"` // plain text string exactly 10 characters long, or  empty name allowed for anonymous workflows
```
krebernisak, line 66: What about "or empty name allowed for anonymous workflows"?

krebernisak (approved):

PR #828, opened 2024-10-03
keystone custom message proto

patrickhuie19:
Supports 

https://github.com/smartcontractkit/chainlink/pull/14510


pkg/beholder/pb/keystone/keystone_custom.proto
```diff
@@ -0,0 +1,21 @@
...
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/beholder/pb/keystone";
+
+package pb;
+
+message KeystoneCustomMessage {
+  string msg=1;
+  // https://protobuf.dev/programming-guides/proto3/#maps
+  // In go: if Value is empty for a key, nothing will be serialized
+  map<string, Value> labels=2;
```
cedric-cordenier, line 11: @patrickhuie19 Can we import the values library here instead? We shouldn't need to duplicate this type

pkg/beholder/pb/keystone/keystone_custom.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/beholder/pb/keystone";
+
+package pb;
+
+message KeystoneCustomMessage {
```
cedric-cordenier, line 7: For the required Keystone labels (workflowOwner, id, etc), are these expected to be added to the labels map?

pkg/beholder/pb/keystone/keystone_custom.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
cedric-cordenier, line 1: Reading the thread you linked me to yesterday, does it make sense to put this here? As opposed to in pkg/capabilities/pb?

pkg/beholder/pb/keystone/keystone_custom.proto
patrickhuie19: Yes, thats the expectation. I prefer the flexibility and scalability of this approach (a generic labels map) over an explicit definition of each field like I had before.

pkg/beholder/pb/keystone/keystone_custom.proto
patrickhuie19: Is capabilities the right dir for it? A msg with some labels could be used pretty much everywhere. Mb that's more appropriate in a beholder general dir? Then other consumers can perhaps consume it and add custom fields

WDYT (@cll-gg)?

pkg/values/pb/values.pb.go
```diff
@@ -494,7 +494,7 @@ func file_values_pb_values_proto_rawDescGZIP() []byte {
 }
 
 var file_values_pb_values_proto_msgTypes = make([]protoimpl.MessageInfo, 6)
-var file_values_pb_values_proto_goTypes = []interface{}{
+var file_values_pb_values_proto_goTypes = []any{
```
patrickhuie19, line 497: the `interface` to `any` updates unintentional, but afaik more idiomatic.

pkg/beholder/pb/keystone/keystone_custom.proto
patrickhuie19: 345595f38e0ec17b8a5f1f488aed12aca0414d86

pkg/beholder/pb/keystone/keystone_custom.proto
patrickhuie19: refactored to sit in pkg/beholder/pb

345595f38e0ec17b8a5f1f488aed12aca0414d86

krehermann (approved):

pkcll (approved):

MStreet3 (approved):

PR #829, opened 2024-10-03
BCFR-967 - Basic support for method writing and reading - Add logic to enable/disable test cases for chain components common test suite

pablolagreca:
As part of adding basic support for generating EVM bindings for CR/CW for solidity contracts we add functionality to be able to disable test cases not yet supported by a partial implementation so we can commit incremental changes without having full support from day one.

solana ref: BCFR-966


Farber98 (dismissed):
see this really useful!

ilija42 (commented):
the approach looks fine, can you modfiy the EVM or Solana tests to use this?

ilija42 (approved):

ilija42 (changes requested):
Shouldn't enabledTests be exported so that the relayers can have a list of enabled tests?

pkg/types/interfacetests/utils.go
```diff
@@ -20,11 +20,30 @@ type BasicTester[T any] interface {
...
 	GetAccountBytes(i int) []byte
+	IsDisabled(testID string) bool
+	DisableTests(testIDs []string)
 }
 
-type testcase[T any] struct {
-	name string
-	test func(t T)
+type TestSelectionSupport struct {
+	enabledTests map[string]bool
```
ilija42, line 28: maybe its more intuitive if we call this disabledTests, since the default is enabled?

ilija42 (commented):
Your latest commits aren't verified

ilija42 (changes requested):
PR looks good, but the commit is still not verified

ilija42 (dismissed):

Farber98 (approved):

pkg/types/interfacetests/chain_components_interface_tests.go
```diff
@@ -16,6 +16,40 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
+// GetLatestValue method
+const GetLatestValueAsValuesDotValue = "Gets the latest value as a values.Value"
```
jmank88, line 20: Since the package is not limited to contract reader, should these be qualified?
```suggestion
const ContractReaderGetLatestValueAsValuesDotValue = "Gets the latest value as a values.Value"
```
I think that might actually help readability somewhat, since it is recognizable as the test idiom.

pkg/types/interfacetests/chain_components_interface_tests.go
```diff
@@ -16,6 +16,40 @@ import (
...
+
+// BatchGet
+const BatchGetLatestValue = "BatchGetLatestValues works"
+const BatchGetLatestValueNoArgumentsPrimitiveReturn = "BatchGetLatestValues works without arguments and with primitive return"
+const BatchGetLatestValueMultipleContractNamesSameFunction = "BatchGetLatestValues allows multiple contract names to have the same function Name"
+const BatchGetLatestValueNoArgumentsWithSliceReturn = "BatchGetLatestValue without arguments and with slice return"
+const BatchGetLatestValueWithModifiersOwnMapstructureOverride = "BatchGetLatestValues wraps config with modifiers using its own mapstructure overrides"
+const BatchGetLatestValueDifferentParamsResultsRetainOrder = "BatchGetLatestValues supports same read with different params and results retain order from request"
+const BatchGetLatestValueDifferentParamsResultsRetainOrderMultipleContracts = "BatchGetLatestValues supports same read with different params and results retain order from request even with multiple contracts"
+const BatchGetLatestValueSetsErrorsProperly = "BatchGetLatestValues sets errors properly"
```
jmank88, line 44: Grouping would reduce some noise and provide something to anchor these comments:
```suggestion
// BatchGet
const (
	BatchGetLatestValue = "BatchGetLatestValues works"
	BatchGetLatestValueNoArgumentsPrimitiveReturn = "BatchGetLatestValues works without arguments and with primitive return"
	BatchGetLatestValueMultipleContractNamesSameFunction = "BatchGetLatestValues allows multiple contract names to have the same function Name"
	BatchGetLatestValueNoArgumentsWithSliceReturn = "BatchGetLatestValue without arguments and with slice return"
	BatchGetLatestValueWithModifiersOwnMapstructureOverride = "BatchGetLatestValues wraps config with modifiers using its own mapstructure overrides"
	BatchGetLatestValueDifferentParamsResultsRetainOrder = "BatchGetLatestValues supports same read with different params and results retain order from request"
	BatchGetLatestValueDifferentParamsResultsRetainOrderMultipleContracts = "BatchGetLatestValues supports same read with different params and results retain order from request even with multiple contracts"
	BatchGetLatestValueSetsErrorsProperly = "BatchGetLatestValues sets errors properly"
)
```

jmank88 (approved):

pkg/types/interfacetests/chain_components_interface_tests.go
pablolagreca: Done

pkg/types/interfacetests/chain_components_interface_tests.go
pablolagreca: Done

Farber98 (approved):

ilija42 (approved):

PR #830, opened 2024-10-03
TT-1303 INFOPLAT-1372 Add support for OTLP/HTTP exporters for beholder sdk

pkcll:
Add support for OTEL client base of OTLP/HTTP exporters

Companion PR
- https://github.com/smartcontractkit/atlas/pull/7204


pkg/beholder/config.go
```diff
@@ -11,6 +11,7 @@ type Config struct {
 	InsecureConnection       bool
 	CACertFile               string
 	OtelExporterGRPCEndpoint string
+	OtelExporterHTTPEndpoint string
```
jmank88, line 14: It seem redundant to have separate constructors and separate config fields. What about either:
- One constructor that checks for which field is set (and ensures both are not set)
- Two constructors that share one field: `OtelExporterEndpoint`

pkcll: Makes sense

pkcll: @jmank88 Addressed here https://github.com/smartcontractkit/chainlink-common/pull/830/commits/ec6684930d600f2d6c8517bd5f955101a4897f6e

anukin (approved):

vyzaldysanchez (approved):

pkg/beholder/httpclient.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil, fmt.Errorf("credentials: failed to append certificates")
+	}
+	return cp, nil
+}
+
+func newHTTPClient(cfg Config, otlploghttpNew otlploghttpFactory) (*Client, error) {
+	noop := NewNoopClient()
+	baseResource, err := newOtelResource(cfg)
+	if err != nil {
+		return noop, err
```
jmank88, line 41: Why return noop instead of nil? Typically when error is not nil, the rest of the return values are undefined. Can we leave it up to the caller to use a `NewNoopClient()` if they want to?

pkg/beholder/httpclient.go
```diff
@@ -0,0 +1,203 @@
...
+		),
+	}
+	if config.TraceSpanExporter != nil {
+		opts = append(opts, sdktrace.WithBatcher(config.TraceSpanExporter))
+	}
+	return sdktrace.NewTracerProvider(opts...), nil
+}
+
+func newHTTPMeterProvider(config Config, resource *sdkresource.Resource, tlsConfig *tls.Config) (*sdkmetric.MeterProvider, error) {
+	ctx := context.Background()
```
jmank88, line 180: Are these "unused"?

pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/830/commits/90367b568661d9822e17544193b2f1d4ac88a2a4

pkcll: In otlpmetrichttp.New ctx is not used, checking others

jmank88: OK, maybe worth noting in a comment :+1:

jmank88 (dismissed):

pkg/beholder/httpclient.go
pkcll: otlptracehttp.New does use context

pkcll: otlploghttp.New does not use context

pkcll: > OK, maybe worth noting in a comment 👍

added comments

4of9 (approved):

PR #831, opened 2024-10-04
feat(wasm): override random_get

MStreet3:
this PR allows overrides of `random_get` of `wasi` such that a cryptographically insecure, pre-seeded source is used for all generated randomness.  forcing determinism in a wasm module can ensure that binaries execute deterministically.  `WithDeterminism` is only used when building a workflow spec.

### DIFF
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -8,7 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"math/rand"
+
 	"strings"
 	"sync"
 	"time"
@@ -179,7 +179,7 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 		return nil, fmt.Errorf("error creating wasmtime module: %w", err)
 	}
 
-	linker, err := newWasiLinker(engine)
+	linker, err := newWasiLinker(modCfg, engine)
 	if err != nil {
 		return nil, fmt.Errorf("error creating wasi linker: %w", err)
 	}
@@ -268,17 +268,6 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 		return nil, err
 	}
 
-	if modCfg.Determinism != nil {
-		err = linker.FuncWrap(
-			"wasi_snapshot_preview1",
-			"random_get",
-			createRandomGet(modCfg),
-		)
-		if err != nil {
-			return nil, err
-		}
-	}
-
 	m := &Module{
 		engine: engine,
 		module: mod,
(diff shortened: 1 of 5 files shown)


pkg/workflows/wasm/host/wasip1.go
```diff
@@ -41,9 +44,34 @@ func newWasiLinker(engine *wasmtime.Engine) (*wasmtime.Linker, error) {
...
 	return linker, nil
 }
 
+func randomGet(caller *wasmtime.Caller, buf, bufLen int32) int32 {
+	// Get slice of memory for the buffer
+	bufPtr := &buf
+	bufFromPtr := (*[4]byte)(unsafe.Pointer(bufPtr))[:]
+
+	// Fix the random source with a hardcoded seed
+	randSource := rand.New(rand.NewSource(42)) //nolint:gosec
```
cedric-cordenier, line 65: @MStreet3 There should be some functions in this package that will allow you to interact with the buffers/pointers in a less slow level way (slightly less anyway).

Check out:
- safeMem to copy a buffer from memory
- copyBuffer to take a buffer and copy it into memory

In this case you probably want to create a new buffer with bufLen and then copy it into the WASM memory using copyBuffer

pkg/workflows/wasm/host/wasip1.go
```diff
@@ -41,9 +43,41 @@ func newWasiLinker(engine *wasmtime.Engine) (*wasmtime.Linker, error) {
...
 	return linker, nil
 }
 
+// randomGet overrides the random_get function in the WASI API and fixes the random source with
+// a hardcoded seed via insecure randomness.  This is to ensure that workflows are deterministic.
+// Randomness should not be used in workflows.
+func randomGet(caller *wasmtime.Caller, buf, bufLen int32) int32 {
+	var (
+		// Fix the random source with a hardcoded seed
+		randSource = rand.New(rand.NewSource(42)) //nolint:gosec
```
nolag, line 64: Shouldn't the randSource be tied to the execution and not the call?  Otherwise, all random numbers will always be the same.

Also, it may be beneficial to use the execution ID at that point so that different runs can have different values.

nolag (changes requested):

pkg/workflows/wasm/host/wasip1.go
cedric-cordenier: We won't necessarily have an execution ID @nolag 

IMO we should just freeze the randomness and make that clear to users -- do you see any legitimate uses of randomness in workflows?

nolag: We can make execution id 0, or use the workflow ID during DAG creation.  Is there any other point that we wouldn't have one?

NFT projects often use offline random for packs.  They give a hash of the cards you receive as the pack's ID so that you know it was created before it's purchased and they can't just give you any cards afterwards. 

I haven't seen any use cases here, but I'm worried that just because I haven't seen it doesn't mean there's no other legitimate use.

pkg/workflows/wasm/host/module.go
```diff
@@ -103,7 +116,26 @@ type Module struct {
 	stopCh chan struct{}
 }
 
-func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
+// WithDeterminism sets the Determinism field to a deterministic seed from a known time.
+//
+// "The Times 03/Jan/2009 Chancellor on brink of second bailout for banks"
```
cedric-cordenier, line 121: 🔥

pkg/workflows/wasm/host/module.go
```diff
@@ -236,6 +268,17 @@ func NewModule(modCfg *ModuleConfig, binary []byte) (*Module, error) {
 		return nil, err
 	}
 
+	if modCfg.Determinism != nil {
```
cedric-cordenier, line 271: @MStreet3 Can we put all wasi overrides in newWasiLinker? Your code as written has some coupling in it, in that it relies on AllowShadowing to be set in the newWasiLinker function

nolag (approved):
Ublocking for now because we don't have a need for randomness.  That said, I think a long-term solution should look at allowing the instance of the module execution to get deterministic randoms that aren't always the same value.  

To achieve this, we could use the id of the module execution to map to a random source that's initialized by step id + workflow execution id and use that when the user makes the callback.

cedric-cordenier (approved):

nolag (approved):

PR #832, opened 2024-10-07
ccip: Add unknown address types

winder:
Use new address types in chainlink, chainlink-common and chainlink-ccip.

Related:
* https://github.com/smartcontractkit/chainlink-ccip/pull/212
* https://github.com/smartcontractkit/chainlink/pull/14673

### DIFF
--- a/pkg/types/ccipocr3/common_types.go
+++ b/pkg/types/ccipocr3/common_types.go
@@ -24,6 +24,14 @@ func (a UnknownAddress) String() string {
 	return Bytes(a).String()
 }
 
+func (a UnknownAddress) MarshalJSON() ([]byte, error) {
+	return Bytes(a).MarshalJSON()
+}
+
+func (a *UnknownAddress) UnmarshalJSON(data []byte) error {
+	return (*Bytes)(a).UnmarshalJSON(data)
+}
+
 // UnknownEncodedAddress represents an encoded address with an unknown encoding.
 type UnknownEncodedAddress string
 
--- a/pkg/types/ccipocr3/common_types.go
+++ b/pkg/types/ccipocr3/common_types.go
@@ -7,6 +7,26 @@ import (
 	"strings"
 )
 
+// UnknownAddress represents a raw address with an unknown encoding.
+type UnknownAddress []byte
+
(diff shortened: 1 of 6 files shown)


pkg/types/ccipocr3/common_types.go
```diff
@@ -7,6 +7,26 @@ import (
 	"strings"
 )
 
+// UnknownAddress represents a raw address with an unknown encoding.
+type UnknownAddress []byte
```
mateusz-sekara, line 11: Thinking out loud, but do we need some kind of equals/compare method then? How do we want to use it in the plugin code?

```go
func (u *UnknownAddress) Compare(other UnknownAddress, selector ChainSelector) int
```

winder: What if the unknown addresses are from two different families, would it ever make sense to compare them? I'm imagining this PR as the first step towards a final solution which has `Address` and `EncodedAddress` types which know how to convert between each other.

Maybe at that point they could be compared.

Some experiments from a couple weeks ago: https://github.com/smartcontractkit/chainlink-ccip/pull/158

mateusz-sekara: Yeah, good question, I think it would be great to start with the example usage. If `Message` stores all addresses as `UnknownAddress`, it might be very cumbersome for the customer of that interface to do the casting back and forth everytime they need to compare addresses to each other.

That being said, maybe Compare/Equals should return an `error` when trying to compare incompatible address types? On the other hand, Message has `SourceChainSelector` / `DestChainSelector` so we should be able to automatically convert these addresses to the proper types, no?

winder: I think it's already the case that working with `Bytes` is a little cumbersome. For comparison you would call the `String()` function.

This first change improves type safety and readability slightly by adding a specific type that tells you if it's encoded or not.

I'm imagining the CCIPReader would convert these to a the real address type (with validation) at which point you no longer have to worry about the chain selector or errors during encode/decode.

mateusz-sekara (approved):

winder:
Closed in favor of #836

PR #834, opened 2024-10-07
[INFOPLAT-1071] Add CSA auth mechanism to Beholder

4of9:
## What 
- [Ticket here](https://smartcontract-it.atlassian.net/jira/software/c/projects/INFOPLAT/boards/260?selectedIssue=INFOPLAT-1071) 
- Adds mechanism to authenticate via CSA key signature in Beholder 
- Uses `grpc.WithPerRPCCredentials` despite the static header value in case this is dynamic in the future 
- Header format is `<version>:<csa_public_key_hex>:<signature_hex>` where the byte value of `<csa_public_key_hex>` is what's being signed
- Adds `csa_pub_key` attribute to all OTel message types

### DIFF
--- a/pkg/beholder/config.go
+++ b/pkg/beholder/config.go
@@ -1,6 +1,7 @@
 package beholder
 
 import (
+	"crypto/ed25519"
 	"time"
 
 	otelattr "go.opentelemetry.io/otel/attribute"
@@ -30,7 +31,13 @@ type Config struct {
 	// Batch processing is enabled by default
 	// Disable it only for testing
 	LogBatchProcessor bool
+
+	CSAAuthEnabled bool
+	CSAPublicKey   ed25519.PublicKey
+	CSASigner      CSASigner
 }
+
+type CSASigner func([]byte) []byte
 
 const (
 	defaultPackageName = "beholder"
@@ -58,6 +65,7 @@ func DefaultConfig() Config {
 		// Log
 		LogExportTimeout:  1 * time.Second,
 		LogBatchProcessor: true,
+		CSAAuthEnabled:    false,
 	}
 }
 
--- a/pkg/beholder/node_auth.go
+++ b/pkg/beholder/node_auth.go
@@ -1,0 +1,83 @@
+package beholder
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+)
+
+// NodeAuthenticator implements PerRPCCredentials interface from google.golang.org/grpc/credentials
+type NodeAuthenticator struct {
+	staticAuthHeaders        map[string]string
+	requireTransportSecurity bool
+}
+
+func NewNodeAuthenticator(config Config) (*NodeAuthenticator, error) {
+	if err := validateNodeAuthConfig(config); err != nil {
+		return nil, err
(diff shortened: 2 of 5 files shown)


pkg/beholder/config.go
```diff
@@ -30,8 +31,14 @@ type Config struct {
 	// Batch processing is enabled by default
 	// Disable it only for testing
 	LogBatchProcessor bool
+
+	CSAAuthEnabled bool
```
jmank88, line 35: Is an explicit bool necessary? Or could we infer this from the presence of the other fields?

pkg/beholder/node_auth.go
```diff
@@ -0,0 +1,83 @@
...
+// The current format for the header is:
+//
+// <version>:<csa_public_key_hex>:<signature_hex>
+//
+// where the byte value of <csa_public_key_hex> is what's being signed
+func deriveAuthHeaderValue(config Config) string {
+	messageBytes := config.CSAPublicKey
+	messageHex := hex.EncodeToString(messageBytes)
+
+	signature := config.CSASigner(messageBytes)
```
jmank88, line 79: What is proven by signing the same static message every time?

pkg/beholder/node_auth.go
```diff
@@ -0,0 +1,83 @@
...
+		requireTransportSecurity: !config.InsecureConnection,
+	}, nil
+}
+
+func (na *NodeAuthenticator) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+	authHeaders := make(map[string]string)
+	for k, v := range na.staticAuthHeaders {
+		authHeaders[k] = v
+	}
+	return authHeaders, nil
```
jmank88, line 31:
```suggestion
	return maps.Clone(na.staticAuthHeaders), nil
```

pkg/beholder/node_auth.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+)
+
+// NodeAuthenticator implements PerRPCCredentials interface from google.golang.org/grpc/credentials
+type NodeAuthenticator struct {
+	staticAuthHeaders        map[string]string
+	requireTransportSecurity bool
+}
+
+func NewNodeAuthenticator(config Config) (*NodeAuthenticator, error) {
```
jmank88, line 15: Does this need to be exported?
```suggestion
// nodeAuthenticator implements PerRPCCredentials interface from google.golang.org/grpc/credentials
type nodeAuthenticator struct {
	staticAuthHeaders        map[string]string
	requireTransportSecurity bool
}

func newNodeAuthenticator(config Config) (*NodeAuthenticator, error) {
```

pkg/beholder/node_auth.go
4of9: the signature is created just once on init and the header is stored

the `PerRPCCredentials` config is used here to provide the scaffolding to move to a dynamic signature in the future, though in our discussion it seems likely we'll stick with the static header in the near term for ease of LOOP plugin integration

4of9:
closing this in favor of a different implementation, PR to follow

PR #835, opened 2024-10-08
Implement RetirementReport format

samsondav:
core ref: 5edbb5a9d33514017e496044e7b3e4962eb55dfa

### DIFF
--- a/pkg/types/provider_llo.go
+++ b/pkg/types/provider_llo.go
@@ -1,11 +1,21 @@
 package types
 
 import (
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+
 	"github.com/smartcontractkit/chainlink-common/pkg/types/llo"
 )
 
+type LLOConfigProvider interface {
+	OffchainConfigDigester() ocrtypes.OffchainConfigDigester
+	// One instance will be run per config tracker
+	ContractConfigTrackers() []ocrtypes.ContractConfigTracker
+}
+
 type LLOProvider interface {
-	ConfigProvider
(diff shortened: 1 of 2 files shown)


brunotm (approved):

cedric-cordenier (dismissed):

pkg/types/provider_llo.go
```diff
@@ -1,11 +1,21 @@
...
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+
 	"github.com/smartcontractkit/chainlink-common/pkg/types/llo"
 )
 
+type LLOConfigProvider interface {
+	OffchainConfigDigester() ocrtypes.OffchainConfigDigester
+	// One instance will be run per config tracker
+	ContractConfigTrackers() []ocrtypes.ContractConfigTracker
+}
```
jmank88, line 13: Don't we need GRPC/protos for this?

jmank88: Oh, I didn't realize these are only superficially connected to the Relayer interfaces :thinking:

samsondav: Do we? I don't know how that would work. We always returned a contract config tracker here

jmank88: If we had support for the rest, then this wouldn't compile until you filled in the new ones, but we just have it stubbed out to error I guess :shrug:

jmank88 (approved):

martin-cll (approved):

brunotm (approved):

PR #836, opened 2024-10-08
ccip: Add unknown address types

winder:
Introduce new address types in a backwards compatible way.


mateusz-sekara (approved):

pkg/types/ccipocr3/v2/common_types.go
```diff
@@ -0,0 +1,176 @@
...
+		return Bytes32{}, fmt.Errorf("Bytes32 must start with '0x' prefix: %s", s)
+	}
+
+	b, err := hex.DecodeString(s[2:])
+	if err != nil {
+		return Bytes32{}, fmt.Errorf("failed to decode hex: %w", err)
+	}
+
+	var res Bytes32
+	copy(res[:], b)
```
0xAustinWang, line 105: Would we want to put the bytes at the end of the array? 
ie. should the result of the conversion be `"0x1234" -> "0x000...01234"` rather than `"0x1234" -> "0x12340000..."`?

pkg/types/ccipocr3/v2/generic_types.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		limitedRange.SetEnd(newEnd)
+	}
+
+	return limitedRange
+}
+
+// Overlaps returns true if the two ranges overlap.
+func (s SeqNumRange) Overlaps(other SeqNumRange) bool {
+	return s.Start() <= other.End() && other.Start() <= s.End()
```
0xAustinWang, line 86: Do we consider overlapping in the reverse direction? 
ie: 
other.Start() <= s.End() && s.Start() <= other.End()

If we want to add it, we can check which one of `other` or `s` is earlier, then do the comparison

pkg/types/ccipocr3/v2/generic_types.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// Contains returns true if the range contains the given sequence number.
+func (s SeqNumRange) Contains(seq SeqNum) bool {
+	return s.Start() <= seq && seq <= s.End()
+}
+
+func (s SeqNumRange) String() string {
+	return fmt.Sprintf("[%d -> %d]", s[0], s[1])
```
0xAustinWang, line 95: Seems like the SeqNumRange is inclusive of the endpoints, let's include that in a comment somewhere.

Any reason we use this specific arrow notation instead of a comma?

pkg/types/ccipocr3/v2/generic_types.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (c ChainSelector) String() string {
+	return fmt.Sprintf("ChainSelector(%d)", c)
+}
+
+// Message is the generic Any2Any message type for CCIP messages.
+// It represents, in particular, a message emitted by a CCIP onramp.
+// The message is expected to be consumed by a CCIP offramp after
+// translating it into the appropriate format for the destination chain.
+type Message struct {
```
0xAustinWang, line 108: I'm surprised this doesn't already exist somewhere

makramkd: This does exist, this is just the `v2` version of the struct

pkg/types/ccipocr3/v2/generic_types.go
makramkd: I believe its the same - this is basically checking for a set intersection which is commutative

pkg/types/ccipocr3/v2/common_types.go
```diff
@@ -0,0 +1,176 @@
...
+	b, err := NewBytesFromString(s)
+	if err != nil {
+		return nil, err
+	}
+	return UnknownAddress(b), nil
+}
+
+// String returns the hex representation of the unknown address.
+func (a UnknownAddress) String() string {
+	return Bytes(a).String()
```
makramkd, line 24: Should we prefix this by something to make it clear that its unknown?

```suggestion
	return fmt.Sprintf("unkown_%s", Bytes(a).String())
```

makramkd: Should unknown addresses string representations ever be passed in somewhere that will make use of that representation (e.g like the contract reader)?

pkg/types/ccipocr3/v2/common_types.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (a UnknownAddress) MarshalJSON() ([]byte, error) {
+	return Bytes(a).MarshalJSON()
```
makramkd, line 28: Ditto here re: prefix

pkg/types/ccipocr3/v2/common_types.go
makramkd: The former seems to be more of what I would expect, at least if the hex bytes represented a number

pkg/types/ccipocr3/v2/common_types.go
winder: Yeah, that would be great.

winder: I would rather not attach semantics to the string representation. It's mainly kept here for convenience and backwards compatibility.

pkg/types/ccipocr3/v2/common_types.go
winder: This code isn't changed from v1. There aren't many places using `Bytes32` after this change, we would have to review those to see if it makes sense to modify the behavior.

I guess my expectation would be `s == NewBytes32FromString(s).String()` which definitely isn't the case with the current implementation which also silently truncates the input.

makramkd (approved):

jmank88 (changes requested):
This looks like it includes CCIP specific code. Can we move it to chainlink-ccip?

pkg/types/ccipocr3/v2/generic_types.go
jmank88, line 1: Why do these types use json struct tags? Are we mixing layers of the API?

PR #838, opened 2024-10-10
Add a non-wasip1 version of NewRunner

cedric-cordenier:
This is so we can remove the wasi build tag from workflows and import the BuildWorkflow functions into tests.


pkg/workflows/wasm/runner_notwasip1.go
```diff
@@ -0,0 +1,7 @@
+//go:build !wasip1
```
nolag, line 1: Since the goal of this is to allow us to catch errors in the test WASM builds faster.  Right now, we need to run the tests, and capture error output from a failure message.  

What if we make our own tag like commonRunnerTesting instead so that users can't accidentally import this?  Then in the test directory, include a README.md that tells users if they are working on the WASMs, but want the syntax highlighting without changing their whole build they an add our flag.  In Goland, it's easy to add a new tag, and I'm assuming it's easy in VS Code. 

That way, IDEs can play nicer and display errors for the WASM binaries without requiring the whole build to be set to WASM (some other packages would show their own errors if you set your IDE to WASM), while this still can't be imported by mistake either.

pkg/workflows/wasm/host/test/runnerapi/cmd/main.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/cli/cmd/testdata/fixtures/capabilities/basictrigger"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+)
+
+func BuildWorkflow(config []byte) *sdk.WorkflowSpecFactory {
```
shileiwill, line 10: Thanks @cedric-cordenier for the change. The UX is a lot neater:
1. BuildWorkflow() can stay in the main.go, and doesnt need to be in a separate file.
2. The main.go doesnt need to have the wasip1 build constraint, actually no constraint needed at all.

agparadiso (approved):

pkg/workflows/wasm/runner_notwasip1.go
cedric-cordenier: The goal of this is really to allow the whole workflow to be in one main.go file, and for that file to be importable in tests. Atm the main.go file can only be compiled to WASM since no non-wasm NewRunner method exists, and if the file is imported it will fail to compile.

cedric-cordenier: I think this is pretty hard to misuse IMO: sure you could compile the binary incorrectly, but it would immediately fail validation in the CLI tool, since the compiled binary wouldn't be valid WASM.

cedric-cordenier: I think this is pretty hard to misuse IMO: sure you could compile the binary incorrectly, but it would immediately fail validation in the CLI tool, since the compiled binary wouldn't be valid WASM.

ettec (approved):

pkg/workflows/wasm/runner_notwasip1.go
nolag: > The goal of this is really to allow the whole workflow to be in one main.go file, and for that file to be importable in tests

I'm lost, why would main.go be importable, main should be the literal package main and all it needs to do is call a function with the right Runner.

func main() {
    pkg.RunWorkflow(wasm.NewRunner())
}

Then you can call pkg.BuildWorkflow in there.  You can test both BuildWorkflow and RunWorkflow with the mock runner.

If you build with wasm.NewRunner() then you can't test the runner's use at all because it'll panic.

cedric-cordenier: The problem is that the API isn't symmetric, so if a user writes a main.go with the workflow in it AND with the runner boilerplate, they can't write a test for just the BuildWorkflow portion of it -- since the go compiler will complain that NewRunner doesn't exist for non-wasip1 architectures.

By adding this, we just unblock that path -- this version of NewRunner should never be called; we should call the testutils.NewRunner version instead. This would be enforced by the CLI tool dev services is building; it'll automatically set the right compiler args to ensure we build with GOOS=wasip1.

I considered making the non-wasip1 version testutils.NewRunner (there's something nice about that, if magical), but that would force NewRunner to return an interface and therefore would force the user to cast the runner to the test runner if they want to use its special methods. For that reason I chose not to do that.

nolag: I'm still feeling a little lost on what they can't do in test today and what is bad to force them to do.

The only thing they can't test is the creation of the WASM runner, and this doesn't allow them to do that testing either.  You can test by making a function that takes an sdk.Runner.

PR #842, opened 2024-10-11
(feat): New OCR3 Consensus Capability Aggregator: Reduce

justinkaseman:
### Description
Take multiple observations and reduce down to a single report (Values.Map).
Intended to be flexible enough to cover 95% of single report use cases by allowing workflow authors to define how they would like their data points aggregated, what format it outputs as, along with different deviation strategies.

More complex aggregation, such as spawning multiple reports, would require a different feature for accepting a user supplied aggregator.


pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
```diff
@@ -0,0 +1,392 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+	"github.com/smartcontractkit/chainlink-common/pkg/values/pb"
+)
+
+type AggregatorConfig struct {
+	// Static data that is added into every report
+	AdditionalData map[string]any `mapstructure:"additionalData"`
```
cedric-cordenier, line 25: Is this going to be eg. the FeedID ?

cedric-cordenier: I think if other fields are provided we should still try and aggregate them.

I see the config being slighltly different:
- a user provides a list of fields to aggregate along with its aggregation method.
  eg. FeedID - Mode, Price - Median, Timestamp - Median
- they also provide a field that we want to check the deviation logic against; in our case Price

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
```diff
@@ -0,0 +1,392 @@
...
+			if oldValue == nil || currDeviation > field.Deviation.InexactFloat64() {
+				shouldReport = true
+			}
+		}
+	}
+
+	// Add static additonal data to the report
+	for key, value := range a.config.AdditionalData {
+		report[key] = value
+	}
```
cedric-cordenier, line 101: See my comment about additional data; we should still be aggregating this IMO

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
```diff
@@ -0,0 +1,392 @@
...
+
+type AggregatorConfig struct {
+	// Static data that is added into every report
+	AdditionalData map[string]any `mapstructure:"additionalData"`
+	// Configuration on how to aggregate one or more data points
+	Fields []AggregationField `mapstructure:"fields"`
+	// The top level field name that report data is put into
+	OutputFieldName string `mapstructure:"outputFieldName" json:"outputFieldName" default:"Reports"` // TODO: is this actually needed?
+	// The structure surrounding the report data that is put on to "OutputFieldName"
+	ReportFormat string `mapstructure:"reportFormat" json:"reportFormat" default:"map" jsonschema:"enum=map,enum=array"`
```
cedric-cordenier, line 31: Why is this needed? Feels like extra complexity, but I may well be missing the point :/

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+	a.lggr.Debugw("current state initialized", "state", currentState, "previousOutcome", previousOutcome)
+	return &currentState, nil
+}
+
+func (a *reduceAggregator) extractValues(observations map[ocrcommon.OracleID][]values.Value, aggregationKey string) []any {
+	events := map[string][]any{}
+	for nodeID, nodeObservations := range observations {
+		// TODO: check the following comment:
+		// we only expect a single observation per node
```
cedric-cordenier, line 172: I think this is fine for the MVP 👍

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
```diff
@@ -0,0 +1,392 @@
...
+		vmap, err := nodeObservations[0].Unwrap()
+		if err != nil {
+			a.lggr.Warnf("could not unwrap observation value from node %d: %v", nodeID, err)
+			continue
+		}
+
+		value := vmap.(map[string]any)[aggregationKey]
+
+		switch v := value.(type) {
+		case string:
```
cedric-cordenier, line 190: @justinkaseman I wonder if you'd be better off dealing with the type complexity as low down as possible, rather than unwrapping the types here and dealing with it again further down.

Concretely, you just pass the values.Value to either mode or median and let those functions figure out what types they support + how they aggregate.

Eg. for mode, you can just do what the identical aggregator does and marshal the proto to some bytes and sha it. For median, you can just convert each type to a decimal (I think that's what we agreed on anyway in our last chat) and return an error if any one of the values can't be handled.

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
justinkaseman: It _is_ extra complexity that I don't love having in -- but the purpose is:

The Data Feeds consumer expects that EncodableOutcome looks like:
```
{
"Reports": [
      {
         "FeedID": []byte,
         "Price": *big.Int,
         "Timestamp": int64
      },
...
}
```

Notice that it's an array of reports, so that it can handle multiple mercury reports coming through.

The EVM Encoder gets the config:
image

So this Aggregator config is to tell it to wrap in an array, even if there's only one report.

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
justinkaseman: That does seem cleaner. I'll try that.

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
justinkaseman: Yes exactly, you can see the CCIP Data Feeds use case recreated in [the uncommented test](https://github.com/smartcontractkit/chainlink-common/pull/842/files#diff-e3c476aac2b96c73893bd0d49244f6bc32b63e97c98c5876e6368000712209b8R210).

Yeahhh, that's fair - just run it all through aggregation. I had thought about that too. I'll make that change.

pkg/capabilities/consensus/ocr3/reduce/reduce_aggregator.go
cedric-cordenier: Ah I see, sorry I missed that initially. Makes sense 👍

pkg/capabilities/consensus/ocr3/ocr3cap/ocr3cap_data_feeds_consensus-schema.json
```diff
@@ -35,7 +35,8 @@
         "aggregation_method": {
           "type": "string",
           "enum": [
-            "data_feeds"
+            "data_feeds",
+            "reduce"
```
cedric-cordenier, line 39: @justinkaseman I don't think this is correct -- or at least it's not what we intended for the consensus capability: we expect each aggregation method to have its own config + struct, rather than extending the existing data feeds consensus one.

For identical consensus, I created this struct here: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/capabilities/consensus/ocr3/ocr3cap/identical_consensus.go, you'll need to create an equivalent for Reduce consensus.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+		if len(vals) < 2*f+1 {
+			return nil, fmt.Errorf("not enough observations provided %s, have %d want %d", field.InputKey, len(vals), 2*f+1)
+		}
+
+		singleValue, err := reduce(field.Method, vals)
+		if err != nil {
+			return nil, fmt.Errorf("unable to reduce on method %s, err: %s", field.Method, err.Error())
+		}
+
+		if field.DeviationType != "none" {
```
cedric-cordenier, line 85: Can we put this in a constant, along with the other deviation types :)

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+	DeviationType string `mapstructure:"deviationType" json:"deviationType,omitempty" jsonschema:"enum=percent,enum=absolute,enum=none"`
+}
+
+type reduceAggregator struct {
+	config ReduceAggConfig
+}
+
+var _ types.Aggregator = (*reduceAggregator)(nil)
+
+func (a *reduceAggregator) Aggregate(lggr logger.Logger, previousOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value, f int) (*types.AggregationOutcome, error) {
```
cedric-cordenier, line 59: This can be a follow up, but can we add some debug logging throughout to record what's happening during the aggregation step? Eg. current/target deviation and heartbeat values, along with the final shouldReport decision and the executionID.

From experience not having this makes it really hard to debug what's going on, and it just means we end up adding it later.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+	// The key that the aggregated data is put under
+	// If omitted, the InputKey will be used
+	OutputKey string `mapstructure:"outputKey" json:"outputKey"`
+	// How the data set should be aggregated
+	Method string `mapstructure:"method" json:"method" jsonschema:"enum=median,enum=mode" required:"true"`
+	// An optional check to only report when the difference from the previous report exceeds a certain threshold.
+	// Can only be used when the field is of a numeric type: string, decimal, int64, big.Int, time.Time, float64
+	// If no deviation is provided on any field, there will always be a report once minimum observations are reached.
+	DeviationString string          `mapstructure:"deviation"  json:"deviation,omitempty"`
+	Deviation       decimal.Decimal `mapstructure:"-"  json:",omitempty"`
```
cedric-cordenier, line 46: Are we missing heartbeat here?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+		if err != nil {
+			return nil, err
+		}
+		err = mv.UnwrapTo(currentState)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	zeroValue, _ := values.Wrap(decimal.NewFromInt(0))
```
cedric-cordenier, line 162: Hmmmm... I don't think the result is the same here if the function errors vs if it doesn't, so I would still check the error, or better yet, manually construct the value:

```
var val values.Value
val = values.NewDecimal(decimal.NewFromInt(0))
```

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+			continue
+		}
+
+		val, err := nodeObservations[0].Unwrap()
+		if err != nil {
+			lggr.Warnf("node %d contributed a Value that could not be unwrapped", nodeID)
+			continue
+		}
+
+		// extract the value from complex types and re-wrap them
```
cedric-cordenier, line 192: A better comment here would explain why we need to do this: i.e. what feature of the value/aggregation is making it necessary to do this :)

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+			} else {
+				lggr.Warnf("aggregation key %s provided, but value is not an indexable type", aggregationKey)
+			}
+		}
+	}
+
+	return vals
+}
+
+func reduce(method string, items []values.Value) (values.Value, error) {
```
cedric-cordenier, line 231: Much cleaner :) Well done!

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,436 @@
...
+	// validations & fill defaults
+	if len(parsedConfig.Fields) == 0 {
+		return ReduceAggConfig{}, errors.New("reduce aggregator must contain config for Fields to aggregate")
+	}
+	for i, field := range parsedConfig.Fields {
+		if len(field.DeviationType) == 0 {
+			field.DeviationType = "none"
+			parsedConfig.Fields[i].DeviationType = "none"
+		}
+		if field.DeviationType != "absolute" && field.DeviationType != "percent" && field.DeviationType != "none" {
```
cedric-cordenier, line 405: IMO this (and deviationType below) could use a helper:

eg.

```
isOneOf(field.DeviationType, []string{"percent", "absolute", "none"})
```

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: Yes!

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: This can be used as heartbeat by using an absolute deviation on the timestamp. I'll get some tests up to show that off.

pkg/capabilities/consensus/ocr3/ocr3cap/ocr3cap_data_feeds_consensus-schema.json
justinkaseman: Oops, yeah this shouldn't be there. Good catch. I'll make the new schema.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: Expanding this comment makes me realize that I could just pass `any` type around. Keeping it as a Value doesn't really give any benefit. Where it's used for mode to get the protobuf I could just wrap it there.

pkg/capabilities/consensus/ocr3/ocr3cap/ocr3cap_data_feeds_consensus-schema.json
justinkaseman: Like so? https://github.com/smartcontractkit/chainlink-common/pull/842/commits/2076c0f23f5e9107043d705ad5a5512f4aab82c2

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: May not be a problem to move it around as a value though :grin:

cedric-cordenier: Yeah that makes sense :)

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: Yes exactly right on the config for PoR.

Agreed on needing more documentation. I'll create a design doc for this.

pkg/capabilities/consensus/ocr3/ocr3cap/ocr3cap_data_feeds_consensus-schema.json
justinkaseman: Ahh, I see what you mean. Added.

cedric-cordenier (dismissed):

pkg/capabilities/consensus/ocr3/models.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 type config struct {
-	AggregationMethod string      `mapstructure:"aggregation_method" json:"aggregation_method" jsonschema:"enum=data_feeds"`
+	AggregationMethod string      `mapstructure:"aggregation_method" json:"aggregation_method" jsonschema:"enum=data_feeds,enum=reduce"`
```
bolekk, line 8: Is this syntax correct? Not "enum=a,b,c"?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,504 @@
...
+	ocrcommon "github.com/smartcontractkit/libocr/commontypes"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+	"github.com/smartcontractkit/chainlink-common/pkg/values/pb"
+)
+
+const (
+	AGGREGATION_METHOD_MEDIAN = "median"
```
bolekk, line 25: Maybe add a comment that it is not a true median (no averaging of two middle values)

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,504 @@
...
+		}
+		err = mv.UnwrapTo(currentState)
+		if err != nil {
+			return nil, fmt.Errorf("initializeCurrentState FromMapValueProto error: %s", err.Error())
+		}
+	}
+
+	zeroValue := values.NewDecimal(decimal.Zero)
+	for _, field := range a.config.Fields {
+		if _, ok := currentState[field.InputKey]; !ok {
```
bolekk, line 204: what if InputKey is an empty string?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,504 @@
...
+			return nil, fmt.Errorf("not enough observations provided %s, have %d want %d", field.InputKey, len(vals), 2*f+1)
+		}
+
+		singleValue, err := reduce(field.Method, vals)
+		if err != nil {
+			return nil, fmt.Errorf("unable to reduce on method %s, err: %s", field.Method, err.Error())
+		}
+
+		if field.DeviationType != DEVIATION_TYPE_NONE {
+			oldValue := (*currentState)[field.InputKey]
```
bolekk, line 107: What if InputKey is an empty string?

pkg/capabilities/consensus/ocr3/models.go
justinkaseman: It's the way that I see it in the example from json schema parser that we use
https://github.com/invopop/jsonschema?tab=readme-ov-file#example

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: It gets set to
```
{
  "": value
}
```

But that's a good edge case in that if there are two Fields with no input key then they will clash.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,504 @@
...
+	SubMapField bool `mapstructure:"subMapField"  json:"subMapField,omitempty"`
+}
+
+type reduceAggregator struct {
+	config ReduceAggConfig
+}
+
+var _ types.Aggregator = (*reduceAggregator)(nil)
+
+func (a *reduceAggregator) Aggregate(lggr logger.Logger, previousOutcome *types.AggregationOutcome, observations map[ocrcommon.OracleID][]values.Value, f int) (*types.AggregationOutcome, error) {
```
MStreet3, line 80: docstring on Aggregate

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
MStreet3, line 107: _nit_ personally this syntax is a bit rare/strange, might be nicer to just define a `Get(string) (any, error)` method on the current state's type

pkg/capabilities/consensus/ocr3/aggregators/identical.go
```diff
@@ -14,10 +14,10 @@ import (
 )
 
 type identicalAggregator struct {
```
MStreet3, line 16: _nit_ touched, would be nice to add docstring of what is an `identicalAggregator`

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
```diff
@@ -0,0 +1,928 @@
...
+					mockValue, err := values.Wrap(int64(100))
+					require.NoError(t, err)
+					return map[commontypes.OracleID][]values.Value{1: {mockValue}, 2: {mockValue}, 3: {mockValue}}
+				},
+				shouldReport: true,
+				expectedOutcome: map[string]any{
+					"Reports": []any{
+						map[string]any{
+							"": int64(100),
+						},
```
MStreet3, line 59: why is this a valid report, we don't want to error instead?

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
```diff
@@ -0,0 +1,928 @@
...
+							"Price": true,
+						},
+					},
+				},
+				expectedState: map[string]any{
+					"BenchmarkPrice": true,
+				},
+			},
+			{
+				name: "aggregate on non-indexable type",
```
MStreet3, line 337: what is being tested here exactly and why is this not a failure mode?  a report should reduce even if the observations are not a map?

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
```diff
@@ -0,0 +1,928 @@
...
+							"Price": big.NewInt(100),
+						},
+					},
+				},
+				expectedState: map[string]any{
+					"BenchmarkPrice": big.NewInt(100),
+				},
+			},
+			{
+				name: "buffer mode",
```
MStreet3, line 238: what is buffer mode and how would this work if the method was `median`? is that valid?

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
```diff
@@ -0,0 +1,928 @@
...
+				},
+				shouldReport: true,
+				expectedOutcome: map[string]any{
+					"Reports": []any{
+						map[string]any{
+							"Price": int64(1),
+						},
+					},
+				},
+				expectedState: map[string]any{"": int64(1)},
```
MStreet3, line 359: what is `expectedState` vs `expectedOutcome`, as in why doesn't this have `Price` as a key?

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
justinkaseman: Finding a mode on a data set of buffer / []byte types. Taking `median` of buffers isn't valid, because you can't sort them using a numeric comparison (`>`). We could support them by sorting them in another way though.

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
justinkaseman: `expectedState` is how the data is stored within the `metadata`, which is what gets passed in on every new aggregation round as what the "previous round" was.

`expectedOutcome` is what gets put as the workflow step's `result`, which currently is what ends up going to the chain-writer, and then on-chain.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,504 @@
...
+
+		(*currentState)[field.InputKey] = singleValue
+		if len(field.OutputKey) > 0 {
+			report[field.OutputKey] = singleValue
+		} else {
+			report[field.InputKey] = singleValue
+		}
+	}
+
+	// if SubMapKey is provided, move fields in a nested map
```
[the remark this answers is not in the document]
justinkaseman: It was a feature request for: data can nested data can be in the report, which enables the sub-piece to stay encoded as bytes in the contract, allowing storing it in the registry without knowing how it is decoded. The consumer from the registry will decode it all the way.

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
justinkaseman: Exactly, since the input to a capability is a `values.Value`, it could be either a List, Map, or just a non-complex type. We could either reject anything that isn't a Map, or map/list -- but to keep this flexible to future use cases, the code currently allows all of the above to come in.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: Good call out. Added words to AggregationField struct.

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
justinkaseman: Yes, I think you are right about this being an invalid report. It shouldn't be a map with an empty key, rather just a non-complex value.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: Since `{"": value}` shouldn't be a valid report, I added a new `reportFormat` for `value` which covers outputting a single value, and I changed it to now hold state as the output structure.

pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
justinkaseman: Renamed to bytes to make it more clear

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: See above, same change should cover this

pkg/capabilities/consensus/ocr3/aggregators/identical.go
justinkaseman: @bolekk how would you say `identicalAggregator` is different? One difference that I see is that it can enforce observations length. But maybe we can roll these together?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,523 @@
...
+	// If no deviation is provided on any field, there will always be a report once minimum observations are reached.
+	Deviation       decimal.Decimal `mapstructure:"-"  json:"-"`
+	DeviationString string          `mapstructure:"deviation"  json:"deviation,omitempty"`
+	// The format of the deviation being provided
+	// * percent - a percentage deviation
+	// * absolute - an unsigned numeric difference
+	DeviationType string `mapstructure:"deviationType" json:"deviationType,omitempty" jsonschema:"enum=percent,enum=absolute,enum=none"`
+	// The key to find a data point within the input data
+	// If omitted, the entire input will be used
+	InputKey string `mapstructure:"inputKey" json:"inputKey"`
```
ettec, line 62: is this a key or a path? ie for nested structures you will get map of maps, maybe Path would be a better name?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -0,0 +1,523 @@
...
+	case DEVIATION_TYPE_ABSOLUTE:
+		return diff, nil
+	case DEVIATION_TYPE_PERCENT:
+		if prevDeci.Cmp(decimal.NewFromInt(0)) == 0 {
+			if diff.Cmp(decimal.NewFromInt(0)) == 0 {
+				return decimal.NewFromInt(0), nil
+			}
+			return decimal.NewFromInt(math.MaxInt), nil
+		}
+		return diff.Div(prevDeci), nil
```
ettec, line 414: should there be a zero check here or is it not possible for prevDeci to be 0?

justinkaseman: The zero check is right above there :grin: 

```
		if prevDeci.Cmp(decimal.NewFromInt(0)) == 0 {
			if diff.Cmp(decimal.NewFromInt(0)) == 0 {
				return decimal.NewFromInt(0), nil
			}
			return decimal.NewFromInt(math.MaxInt), nil
		}
```

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: Leaving for now, but I'll make a note in the ticket for Matt's suggestion of Field paths to get this in

cedric-cordenier (approved):

bolekk (approved):

PR #844, opened 2024-10-11
Fix Latest timestamp metric to return timestamp in seconds

alecgard:
- Made a rookie mistake when first adding this metric and extracted the Seconds component of the time, rather than the time in seconds 🤦.


alecgard:
`Build External Repositories > Build Chainlink Image` workflow is failing, but this should be unrelated and appears to have been failing for a while: https://github.com/smartcontractkit/chainlink-common/actions/workflows/build_external.yml

aalu1418 (approved):

george-dorin (approved):

PR #845, opened 2024-10-11
[cappl-86] feat(workflows/wasm): emit msgs to beholder

MStreet3:
This PR provides a new method `Emitter() MessageEmitter` in the wasm SDK, which returns an emitter that sends custom messages to an external collector.  The `Emitter()` method is exposed on the guest `Runtime` for custom compute.

To implement `MessageEmitter`, the `wasip1` build now links and exposes a new function called `emit`.  The `wasip1` host implementation of `emit` depends on a separate `MessageEmitter` which is passed via the module config.   The passed `MessageEmitter` is typically an instance of an `events.Labeler`.

All custom compute requests that emit are required to include the `workflow_id`, `workflow_name` and `workflow_owner_address` or else module requests for compute will fail.

## Out of Scope
This PR does not address the need for the host's emit behavior to be context aware.  Follow up work will handle context within the WASM.


pkg/workflows/sdk/runtime.go
```diff
@@ -7,9 +7,29 @@ import (
...
+	Emit(msg string, labels map[string]any) error
+}
+
+type EmitterFunc func(msg string, labels map[string]any) error
+
+func (f EmitterFunc) Emit(msg string, labels map[string]any) error {
+	return f(msg, labels)
+}
+
+type ContextEmitter interface {
```
cedric-cordenier, line 21: No point in supporting context on the guest-side, the wasmtime API isn't itself context-aware so it would be a pointless arg.

We should however support it on the host side by using the context passed into the execute call.

pkg/workflows/wasm/host/adapters/beholder/emitter.go
```diff
@@ -0,0 +1,17 @@
...
+	b "github.com/smartcontractkit/chainlink-common/pkg/beholder"
+)
+
+type emitter struct{}
+
+func NewEmitter() *emitter {
+	return &emitter{}
+}
+
+func (e *emitter) Emit(msg string, labels map[string]any) error {
```
cedric-cordenier, line 15: Nit/style: this is small enough that it could live in a separate file in host, but that may be a preference more than anything (I like to keep things clustered together as much as possible)

pkg/workflows/wasm/host/module.go
```diff
@@ -285,6 +300,15 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 		return nil, fmt.Errorf("error wrapping fetch func: %w", err)
 	}
 
+	err = linker.FuncWrap(
+		"env",
+		"emit",
+		createEmitFn(logger, modCfg.Emitter),
```
cedric-cordenier, line 306: This should be context-aware, by leveraging the requestId + the store to add state to the ongoing request. Also we need to add a set of predefined tags (executionID, owner, workflowID, name -- see the ticket for the list) to each emission.

pkg/workflows/wasm/host/module.go
```diff
@@ -445,3 +469,46 @@ func fetchFn(logger logger.Logger, modCfg *ModuleConfig) func(caller *wasmtime.C
...
+) func(caller *wasmtime.Caller, msgptr, msglen int32) int32 {
+	logErr := func(err error) {
+		l.Errorf("error emitting message: %s", err)
+	}
+
+	return func(caller *wasmtime.Caller, msgptr, msglen int32) int32 {
+		b, err := safeMem(caller, msgptr, msglen)
+		if err != nil {
+			logErr(err)
+			return ErrnoFault
```
cedric-cordenier, line 485: The API can return an error, so can we return it in a response too? That would imply that we should have a resp buffer.

pkg/workflows/wasm/pb/wasm.proto
```diff
@@ -74,7 +70,13 @@ message FetchRequest {
...
   bool success = 1;
   string errorMessage = 2;
-  uint32 statusCode = 3; // NOTE: this is actually a uint8, but proto doesn't support this.
+  uint32 statusCode =
+      3; // NOTE: this is actually a uint8, but proto doesn't support this.
   values.Map headers = 4;
   bytes body = 5;
 }
+
+message CustomEmitMessage {
```
cedric-cordenier, line 79: Consider just calling this `EmitMessage` -- CustomEmitMessage doesn't say much more than that.

pkg/workflows/wasm/runner_wasip1.go
```diff
@@ -23,107 +23,138 @@ func log(respptr unsafe.Pointer, respptrlen int32)
...
+
+	b, err := proto.Marshal(&wasmpb.CustomEmitMessage{
+		Message: msg,
+		Labels:  values.ProtoMap(vm),
+	})
+
+	ptr, ptrlen := bufferToPointerLen(b)
+	errno := emit(ptr, ptrlen)
+	if errno != 0 {
+		os.Exit(CodeRunnerErr)
```
cedric-cordenier, line 152: @MStreet3 You can return an error here rather than exiting: let the caller handle this and choose what to do; if they want to exit, they can bubble up the error.

pkg/workflows/wasm/pb/wasm.proto
```diff
@@ -74,7 +70,13 @@ message FetchRequest {
 message FetchResponse {
   bool success = 1;
   string errorMessage = 2;
-  uint32 statusCode = 3; // NOTE: this is actually a uint8, but proto doesn't support this.
+  uint32 statusCode =
+      3; // NOTE: this is actually a uint8, but proto doesn't support this.
```
agparadiso, line 74: 💭  is this change intentionally?

pkg/workflows/wasm/host/module.go
```diff
@@ -19,31 +19,70 @@ import (
...
+// UnsafeReaderFunc abstractly defines the behavior of reading from WASM memory.
+type UnsafeReaderFunc func(c *wasmtime.Caller, ptr, len int32) ([]byte, error)
+
+// wasmMemoryAccessor is the default implementation for unsafely accessing the memory of the WASM module.
+func wasmMemoryAccessor(caller *wasmtime.Caller) []byte {
+	return caller.GetExport("memory").Memory().UnsafeData(caller)
+}
+
+// wasmRead returns a copy of the wasm module memory at the given pointer and size.
+func wasmRead(caller *wasmtime.Caller, ptr int32, size int32) ([]byte, error) {
```
MStreet3, line 42: renamed `safeMem`

pkg/workflows/wasm/host/module.go
```diff
@@ -19,31 +19,70 @@ import (
...
-	if ptr+size > int32(len(data)) {
+type UnsafeWriterFunc func(c *wasmtime.Caller, src []byte, ptr, len int32) int64
+
+type UnsafeFixedLengthWriterFunc func(c *wasmtime.Caller, ptr int32, val uint32) int64
+
+// UnsafeReaderFunc abstractly defines the behavior of reading from WASM memory.
+type UnsafeReaderFunc func(c *wasmtime.Caller, ptr, len int32) ([]byte, error)
+
+// wasmMemoryAccessor is the default implementation for unsafely accessing the memory of the WASM module.
+func wasmMemoryAccessor(caller *wasmtime.Caller) []byte {
```
MStreet3, line 37: pulled this out so we could unit test the actions on the slice

pkg/workflows/wasm/host/module.go
MStreet3: this is a failure to read the binary's memory, added a response buffer for the error on calling Emit

pkg/workflows/sdk/runtime.go
```diff
@@ -7,9 +7,26 @@ import (
 
 var BreakErr = capabilities.ErrStopExecution
 
+type Emitter interface {
+	// Emit sends a message with the given message and labels to the configured collector.
+	//
+	// TODO(mstreet3): Emit and custmsg.Labeler should be context aware.  Update signature once
+	// WASM can support context.
+	Emit(msg string, labels map[string]any) error
```
cedric-cordenier, line 15: Ergonomic nit: if you allow a struct to be passed in here users can omit labels if they don't need it

pkg/workflows/wasm/host/module.go
```diff
@@ -429,19 +424,177 @@ func fetchFn(logger logger.Logger, modCfg *ModuleConfig) func(caller *wasmtime.C
...
+		}
+		validated[k] = str
+	}
+
+	return custmsg.NewLabeler().WithMapLabels(validated).SendLogAsCustomMessage(msg)
+}
+
+// UnsafeWriterFunc defines behavior for writing directly to wasm memory.  A source slice of bytes
+// is written to the location defined by the ptr.
+type UnsafeWriterFunc func(c *wasmtime.Caller, src []byte, ptr, len int32) int64
```
cedric-cordenier, line 533: @MStreet3 Can we make these types private? They shouldn't leak outside the package.

pkg/workflows/wasm/runner.go
```diff
@@ -163,7 +163,7 @@ func (r *Runner) handleComputeRequest(factory *sdk.WorkflowSpecFactory, id strin
 		}
 	}
 
-	sdk := r.sdkFactory(drc)
+	sdk := r.sdkFactory(drc, WithRequestID(id), WithRequestMetaData(&creq.Metadata))
```
cedric-cordenier, line 166: id + requestMetadata aren't really options -- if they get omitted, then it'll be a misuse of the API. Wdyt about not having these as options and as having them as simple args?

pkg/workflows/wasm/sdk.go
```diff
@@ -33,3 +69,114 @@ func (r *Runtime) Fetch(req sdk.FetchRequest) (sdk.FetchResponse, error) {
...
+// createEmitFn builds the runtime's emit function implementation, which is a function
+// that handles marshalling and unmarshalling messages for the WASM to act on.
+func createEmitFn(
+	sdkConfig *RuntimeConfig,
+	l logger.Logger,
+	emit func(respptr unsafe.Pointer, resplenptr unsafe.Pointer, reqptr unsafe.Pointer, reqptrlen int32) int32,
+) func(string, map[string]any) error {
+	emitFn := func(msg string, labels map[string]any) error {
+		// Prepare the labels to be emitted
+		if sdkConfig.MetaData == nil {
```
cedric-cordenier, line 86: nit MetaData -> Metadata

pkg/workflows/wasm/host/module.go
```diff
@@ -100,6 +77,10 @@ type ModuleConfig struct {
 	IsUncompressed bool
 	Fetch          func(*wasmpb.FetchRequest) (*wasmpb.FetchResponse, error)
 
+	// Emitter is the function that will be called when the emit function is called in the WASM module.
+	// Implementation should emit to external consumer.
+	Emitter sdk.Emitter
```
cedric-cordenier, line 82: I don't think we should use the same interface for the guest and for the host. When we make the host method context aware, we'll need to start using different interfaces here, since the host emitter will accept a context, while the guest won't.

pkg/workflows/sdk/runtime.go
```diff
@@ -7,9 +7,26 @@ import (
...
+}
+
+type EmitterFunc func(msg string, labels map[string]any) error
+
+func (f EmitterFunc) Emit(msg string, labels map[string]any) error {
+	return f(msg, labels)
+}
```
agparadiso, line 22: I might not be seeing this correctly, but I think with just having the function defined inside the Runtime interface we already get the benefits of decoupling it. 

Instead of having the `Emitter` interface implementing the `Emit` method that has the same signature as the `EmitterFunc` type isn't simpler to just have the 

```
type Runtime interface {
	Logger() logger.Logger
	Fetch(req FetchRequest) (FetchResponse, error)
         Emit(msg string, labels map[string]any) error
}
```

or is it because we expect the Emitter interface to grow?

pkg/workflows/sdk/runtime.go
patrickhuie19: Hmm, the responsibilities of the initial `Labeler` struct and this `Emitter` interface seem to have overlap. With the addition of this interface it seems clearer to me that `SendLogAsCustomMessage` should be updated to be called `Emit`. Usages in core will have to be updated as well. 

@cedric-cordenier In core this is one reason why I chose not to have an emitting API with labels as a map. The API can always grow to include labels, but that there are risks there because the Labeler interface can be bypassed if Emit is implemented to only send the labels passed as attributes vs what is stored in the Labeler impl. 

Can we stick to `Emit(msg string)`?

pkg/workflows/sdk/runtime.go
MStreet3: `EmitterFunc` is a convenience type/adapter so that a function can satisfy the `Emitter` interface directly.  Like `[HttpHandlerFunc](https://pkg.go.dev/net/http#HandlerFunc)` from `net/http`.  Based on @cedric-cordenier comment above though, this can be moved into the module.  It's only used so that the `beholderEmitter` function can satisfy the interface.

pkg/workflows/wasm/host/module.go
```diff
@@ -429,19 +424,177 @@ func fetchFn(logger logger.Logger, modCfg *ModuleConfig) func(caller *wasmtime.C
...
+			return writeErr(err)
+		}
+
+		return ErrnoSuccess
+	}
+}
+
+// beholderEmitter sends a message to the Beholder service via the custmsg package's labeler.  Each label is
+// expected to be a string.  Fails if any label is not a string.
+func beholderEmitter(msg string, labels map[string]any) error {
```
patrickhuie19, line 518: IMO this consumer shows the impact of the current `Emit` interface on Labeler - the intent was for Labeler to be consumed at the highest level so that consumers can add labels at multiple depths in a call stack and pass it on.

MStreet3: hm, not sure I follow, what's the suggested change/issue?

patrickhuie19: It looks like you would have a consumer call `beholderEmitter`, which creates a new labeler with labels at that point of time, instead of creating a Labeler, adding labels through `With`, and then calling `Emit`. I could propose a specific change if you have an example of why `beholderEmitter` is needed.

pkg/workflows/wasm/runner_wasip1.go
```diff
@@ -24,103 +25,121 @@ func log(respptr unsafe.Pointer, respptrlen int32)
...
-			if err != nil {
-				// We somehow couldn't marshal the response, so let's
-				// exit with a special error code letting the host know
-				// what happened.
-				os.Exit(CodeInvalidResponse)
-			}
+		sendResponse: sendResponseFn,
+		sdkFactory: func(sdkConfig *RuntimeConfig, id string, md *capabilities.RequestMetadata, opts ...func(*RuntimeConfig)) *Runtime {
+			sdkConfig.RequestID = &id
+			sdkConfig.Metadata = md
```
cedric-cordenier, line 38: :/ This is worse IMO since you're coupling id + RequestMetadata with RuntimeConfig through a function. They are all being passed in from the same place, why not move these lines so that they happen before passing it in?

pkg/workflows/wasm/host/module.go
```diff
@@ -100,6 +90,10 @@ type ModuleConfig struct {
 	IsUncompressed bool
```
patrickhuie19, line 90: @MStreet3 the intent of a Labeler was to be wired alongside a logger. re: my previous comments on misuse of Labeler, I would consider a couple options:

* Remove `Emitter` and drill a `Labeler` here (Requires the updates to the Labeler API that I commented on earlier) 

Or 

* Satisfy `Emitter` through a Labeler, not through the `beholderEmitter` func. (Requires the updates to the Labeler API that I commented on earlier)

cedric-cordenier (dismissed):

pkg/workflows/wasm/host/module.go
cedric-cordenier: @patrickhuie19 IIUC correctly, you're saying we should preserve any existing labelers that are instantiated further up, is that correct?

This beholderEmitter would only be called if an emitter isn't passed in when calling NewModule. Where this is used I'd expect us to instantiate an emitter based off of an existing labeler.

patrickhuie19: >IIUC correctly, you're saying we should preserve any existing labelers that are instantiated further up, is that correct?

Yes, and you wouldn't need a new func (`beholderEmitter`) if you had an instantiated Labeler and instead used its Emit func

pkg/custmsg/custom_message.go
```diff
@@ -11,16 +11,41 @@ import (
...
+	for k, v := range labels {
+		newCustomMessageLabeler.labels[k] = v
+	}
+
+	return newCustomMessageLabeler
 }
 
 // With adds multiple key-value pairs to the CustomMessageLabeler for transmission With SendLogAsCustomMessage
-func (c Labeler) With(keyValues ...string) Labeler {
+func (c *labeler) With(keyValues ...string) *labeler {
```
patrickhuie19, line 48: /nit all the methods on Labeler are intended to be read-only, like with Logger, so I do not think pointer receivers are necessary

pkg/workflows/wasm/host/module.go
```diff
@@ -100,6 +94,14 @@ type ModuleConfig struct {
...
 	Fetch          func(*wasmpb.FetchRequest) (*wasmpb.FetchResponse, error)
 
+	// emitter is the function that will be called when the emit function is called in the WASM module.
+	// Implementation should emit to external consumer.  Private because a real emitter should be adapted
+	// from the Labeler.
+	emitter Emitter
```
patrickhuie19, line 100: can the WASM module consume `EmitLabeler` directly? If so, is this redundant?

pkg/workflows/wasm/host/module.go
```diff
@@ -91,6 +66,14 @@ type DeterminismConfig struct {
 	Seed int64
 }
 
+type EmitLabeler interface {
+	// Emit sends a message to the labeler's destination.
+	Emit(string) error
+
+	// WithMapLabels sets the labels for the message to be emitted.  Labels are cumulative.
+	WithMapLabels(map[string]string) EmitLabeler
```
MStreet3, line 74: the host works with maps from unmarshalled protos, so we have a different interface defined locally

agparadiso (dismissed):

cedric-cordenier (dismissed):

pkg/workflows/wasm/host/module.go
```diff
@@ -91,6 +66,14 @@ type DeterminismConfig struct {
 	Seed int64
 }
 
+type EmitLabeler interface {
```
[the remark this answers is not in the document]
cedric-cordenier: (Feel free to follow up or take as a nit)

pkg/workflows/wasm/sdk.go
```diff
@@ -1,26 +1,53 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/custmsg"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
 	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+	wasmpb "github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/pb"
+)
+
+const (
```
patrickhuie19, line 19: these are also defined in [pkg/capabilities/events/events.go](https://github.com/smartcontractkit/chainlink-common/blob/ceeb47375a5679524c088fe04a66105efc1f7a71/pkg/capabilities/events/events.go#L18). Is it possible to avoid duplication here?

patrickhuie19 (dismissed):

patrickhuie19 (dismissed):

pkg/values/map.go
```diff
@@ -20,7 +20,7 @@ func EmptyMap() *Map {
 	}
 }
 
-func NewMap(m map[string]any) (*Map, error) {
+func NewMap[T any](m map[string]T) (*Map, error) {
```
MStreet3, line 23: required all maps to be `map[string]any` now can also create values map from typed map (e.g., `map[string]string`)

agparadiso (approved):

cedric-cordenier (approved):

PR #846, opened 2024-10-11
Beholder: Add domain and entity to metadata

4of9:
- Adds required attributes `beholder_domain` and `beholder_entity` to custom messages
- [Companion Atlas PR](https://github.com/smartcontractkit/atlas/pull/7316)

### DIFF
--- a/pkg/beholder/client_test.go
+++ b/pkg/beholder/client_test.go
@@ -50,6 +50,8 @@ func TestClient(t *testing.T) {
 			"byte_key_1":           []byte("byte_val_1"),
 			"str_slice_key_1":      []string{"str_val_1", "str_val_2"},
 			"nil_key_1":            nil,
+			"beholder_domain":      "TestDomain",        // Required field
+			"beholder_entity":      "TestEntity",        // Required field
 			"beholder_data_schema": "/schemas/ids/1001", // Required field, URI
 		}
 	}
@@ -202,15 +204,39 @@ func TestEmitterMessageValidation(t *testing.T) {
 		{
 			name: "Invalid URI",
 			attrs: Attributes{
+				"beholder_domain":      "TestDomain",
+				"beholder_entity":      "TestEntity",
 				"beholder_data_schema": "example-schema",
 			},
 			exporterCalledTimes: 0,
 			expectedError:       "'Metadata.BeholderDataSchema' Error:Field validation for 'BeholderDataSchema' failed on the 'uri' tag",
 		},
 		{
-			name:                "Valid URI",
+			name: "Invalid BeholderDomain",
+			attrs: Attributes{
+				"beholder_data_schema": "/example-schema/versions/1",
+				"beholder_entity":      "TestEntity",
+				"beholder_domain":      "Test__Domain",
+			},
+			exporterCalledTimes: 0,
+			expectedError:       "'Metadata.BeholderDomain' Error:Field validation for 'BeholderDomain' failed on the 'no_double_underscore' tag",
+		},
+		{
+			name: "Invalid BeholderEntity",
+			attrs: Attributes{
+				"beholder_data_schema": "/example-schema/versions/1",
+				"beholder_entity":      "Test__Entity",
+				"beholder_domain":      "TestDomain",
+			},
+			exporterCalledTimes: 0,
+			expectedError:       "'Metadata.BeholderEntity' Error:Field validation for 'BeholderEntity' failed on the 'no_double_underscore' tag",
+		},
+		{
+			name:                "Valid Attributes",
 			exporterCalledTimes: 1,
 			attrs: Attributes{
+				"beholder_domain":      "TestDomain",
+				"beholder_entity":      "TestEntity",
 				"beholder_data_schema": "/example-schema/versions/1",
 			},
 			expectedError: "",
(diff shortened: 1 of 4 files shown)


pkg/beholder/client_test.go
```diff
@@ -49,6 +49,8 @@ func TestClient(t *testing.T) {
 			"byte_key_1":           []byte("byte_val_1"),
 			"str_slice_key_1":      []string{"str_val_1", "str_val_2"},
 			"nil_key_1":            nil,
+			"beholder_domain":      "TestDomain",        // Required field
+			"beholder_entity":      "TestEntity",        // Required field
 			"beholder_data_schema": "/schemas/ids/1001", // Required field, URI
```
pkcll, line 54: @4of9,  should we remove URI validation rule for `beholder_data_schema`? 
It could be just subject name in the future or something that is not URI format

pkcll: @4of9, we need to be careful with adding required fields. Before merging this all places where `Emit` is its used needs to be patched.

4of9: @pkcll I think the intention is to still include a URI to the schema file in `beholder_data_schema`, though we're not using it to derive the subject downstream

4of9: > Before merging this all places where Emit is its used needs to be patched.

@pkcll by this you mean all the places / external repos where the Beholder library is used?

pkcll: >  by this you mean all the places / external repos where the Beholder library is used?

yes

nanchano: we should probably make a note to look into versioning en the future

nanchano (approved):
Looks good

anukin (approved):

pkcll (approved):

pkg/beholder/message.go
```diff
@@ -16,6 +18,8 @@ type Message struct {
 type Metadata struct {
 	//	REQUIRED FIELDS
 	// Schema Registry URI to fetch schema
+	BeholderDomain     string `validate:"required,domain_entity"`
+	BeholderEntity     string `validate:"required,domain_entity"`
```
jmank88, line 22: What is the significance of these being `required`? What happens when they are omitted?

pkg/beholder/message.go
```diff
@@ -222,8 +232,32 @@ func NewMetadata(attrs Attributes) *Metadata {
...
+		if strings.Contains(str, "__") {
+			return false
+		}
+		if !validDomainAndEntityRegex.MatchString(str) {
+			return false
+		}
+		return true
+	})
+	if err != nil {
+		panic(err)
```
jmank88, line 254: Why panic? The caller has the ability to return error.

anukin (dismissed):

pkg/beholder/message.go
4of9: good point, will update

pkg/beholder/message.go
4of9: `required` is a built-in validator that ensures a field is non-empty. if it was omitted here then empty strings would be valid

jmank88: I mean that if these fields are not set, what happens at run time? (because I don't see them being set in this PR - have you tried importing this in to core yet?)

pkg/beholder/message.go
4of9: addressed in 78b4fa890248046a0c074b712889bd44551608d3

pkg/beholder/message.go
jmank88: Oh, I guess we are not actually using the `Message` type yet? :shrug: 
I'm still curious what happens here

4of9: I have not tried importing this into core yet

at runtime, if the fields are not set, custom messages will fail to send when Emit is called ([see here](https://github.com/smartcontractkit/chainlink-common/blob/4b45ad16ad7fa0b865decac91f60aa77b91bf147/pkg/beholder/client.go#L242))

this is following the existing pattern we have for our other required field `BeholderDataSchema`, the next line in this file

pkg/beholder/message.go
```diff
@@ -222,8 +232,35 @@ func NewMetadata(attrs Attributes) *Metadata {
 	return m
 }
 
-func (m *Metadata) Validate() error {
+// validDomainAndEntityRegex allows for alphanumeric characters and ._-
+var validDomainAndEntityRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
```
jmank88, line 236: I wonder if this is worth optimizing since it will be such high volume?

jmank88 (approved):

pkg/beholder/message.go
4of9: I think this regex would just be compiled once and then reused as it's at the package level, which should help with high volume

or is there a different optimization you had in mind?

pkcll (approved):

anukin (approved):

pkg/beholder/message.go
jmank88: This won't run in the playground, but if you run it locally it performs a rough benchmark: https://go.dev/play/p/oZngGcqQXqC

Regex is relatively slow, but it may not be significant:
```
Regex: 3.690749ms
ASCII: 64.001µs
Unicode: 327.251µs
```

PR #848, opened 2024-10-14
[CAPPL-87] Support breaking a workflow from inside custom compute

cedric-cordenier:


pkg/capabilities/capabilities.go
```diff
@@ -26,7 +26,7 @@ func (e errStopExecution) Error() string {
 }
 
 func (e errStopExecution) Is(err error) bool {
-	return err.Error() == errStopExecutionMsg
+	return strings.Contains(err.Error(), errStopExecutionMsg)
```
cedric-cordenier, line 29: This is needed because returning BreakErr in a custom compute step will get dumped as a string, thus breaking the error chain.

The behaviour associated with this PR is tested in the follow up PR in core.

agparadiso (approved):

ettec (approved):

PR #854, opened 2024-10-16
[chore] Add Beholder custom event for Workflows and Capabilities

cedric-cordenier:


pkg/capabilities/events/pb/events.proto
```diff
@@ -0,0 +1,13 @@
...
+
+package events;
+
+import "values/pb/values.proto";
+
+message OperationalEvent {
+  string message = 1;
+  values.Map labels = 2;
+}
+
```
MStreet3, line 13: looks good to me, just missing a proto message that can handle the response with the possible error

cedric-cordenier: I added the message :)

MStreet3 (approved):

ettec (approved):

PR #855, opened 2024-10-16
contract reader api change for block meta data

ettec:
Part of the work for -> https://smartcontract-it.atlassian.net/browse/CAPPL-108

ContractReader API change discussed  here-> https://docs.google.com/document/d/1jyJ2eXIR0EP8eO30Sj4eVj-jJFW34OsnwIzV5mYwKKY/edit?tab=t.0#heading=h.3hiih1q4rib

### DIFF
--- a/pkg/types/contract_reader.go
+++ b/pkg/types/contract_reader.go
@@ -17,10 +17,6 @@ const (
 	ErrNotFound                    = NotFoundError("not found")
 )
 
-type BlockMetaData struct {
-	Height uint64
-}
-
 // ContractReader defines essential read operations a chain should implement for reading contract values and events.
 type ContractReader interface {
 	services.Service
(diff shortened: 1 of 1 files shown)


pkg/types/contract_reader.go
```diff
@@ -46,6 +50,9 @@ type ContractReader interface {
 	// Passing in a *values.Value as the returnVal will encode the return value as an appropriate value.Value instance.
 	GetLatestValue(ctx context.Context, readIdentifier string, confidenceLevel primitives.ConfidenceLevel, params, returnVal any) error
 
+	// GetLatestValueWithBlockMetaData should be used in the same was as GetLatestValue, but also returns the block metadata.
+	GetLatestValueWithBlockMetaData(ctx context.Context, readIdentifier string, confidenceLevel primitives.ConfidenceLevel, params, returnVal any) (BlockMetaData, error)
```
krehermann, line 54: need to add to unimplmented contract reader

pkg/types/contract_reader.go
```diff
@@ -46,6 +50,9 @@ type ContractReader interface {
 	// Passing in a *values.Value as the returnVal will encode the return value as an appropriate value.Value instance.
 	GetLatestValue(ctx context.Context, readIdentifier string, confidenceLevel primitives.ConfidenceLevel, params, returnVal any) error
 
+	// GetLatestValueWithBlockMetaData should be used in the same was as GetLatestValue, but also returns the block metadata.
```
ilija42, line 53:
```suggestion
	// GetLatestValueWithBlockMetaData should be used in the same way as GetLatestValue, but also returns the block metadata.
```

pkg/types/contract_reader.go
```diff
@@ -17,6 +17,10 @@ const (
 	ErrNotFound                    = NotFoundError("not found")
 )
 
+type BlockMetaData struct {
+	Height uint64
+}
```
ettec, line 22: we can add to this as needed, don't think we need to finalise on what data to expose at this point?

pkg/types/contract_reader.go
ettec: ooh, thanks, doing it

pkg/types/contract_reader.go
```diff
@@ -17,6 +17,10 @@ const (
 	ErrNotFound                    = NotFoundError("not found")
 )
 
+type BlockMetaData struct {
```
ilija42, line 20: you can reuse Head type that is defined in this same file

pkg/types/contract_reader.go
ettec: done, ta

pkg/types/contract_reader.go
ettec: sorted

ilija42 (approved):

cedric-cordenier (approved):

PR #865, opened 2024-10-17
Validation for required attributes

nanchano:
Rules:
1. CSA key can't be empty
2. CSA Signature can't be empty
3. Data Schema can't be empty
4. Domain can't be empty
5. Set up Timestamp automatically

### DIFF
--- a/pkg/beholder/client_test.go
+++ b/pkg/beholder/client_test.go
@@ -50,10 +50,10 @@ func TestClient(t *testing.T) {
 			"byte_key_1":           []byte("byte_val_1"),
 			"str_slice_key_1":      []string{"str_val_1", "str_val_2"},
 			"nil_key_1":            nil,
-			"beholder_domain":      "TestDomain",        // Required field
-			"beholder_entity":      "TestEntity",        // Required field
-			"beholder_data_schema": "/schemas/ids/1001", // Required field, URI
-      "node_csa_key":         "node_csa_val",           // Required
+			"beholder_domain":      "TestDomain",             // Required field
+			"beholder_entity":      "TestEntity",             // Required field
+			"beholder_data_schema": "/schemas/ids/1001",      // Required field, URI
+			"node_csa_key":         "node_csa_val",           // Required
 			"node_csa_signature":   "mode_csa_signature_val", // Required
 		}
 	}
--- a/pkg/beholder/message.go
+++ b/pkg/beholder/message.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-  "time"
+	"time"
 
 	"github.com/go-playground/validator/v10"
 	"go.opentelemetry.io/otel/attribute"
--- a/pkg/beholder/example_test.go
+++ b/pkg/beholder/example_test.go
@@ -47,7 +47,7 @@ func ExampleNewClient() {
 			"beholder_data_schema", "/custom-message/versions/1", // required
 			"beholder_domain", "ExampleDomain", // required
 			"beholder_entity", "ExampleEntity", // required
-     	"node_csa_key", "node_csa_val", // required
+			"node_csa_key", "node_csa_val", // required
 			"node_csa_signature", "mode_csa_signature_val", // required
 			"beholder_data_type", "custom_message",
 			"foo", "bar",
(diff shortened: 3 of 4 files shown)


pkg/beholder/message.go
```diff
@@ -211,13 +216,15 @@ func (m *Metadata) FromAttributes(attrs Attributes) *Metadata {
...
+		case "timestamp":
+			m.Timestamp = v.(time.Time)
 		}
 	}
 	return m
 }
 
 func NewMetadata(attrs Attributes) *Metadata {
-	m := &Metadata{}
+	m := &Metadata{Timestamp: time.Now().UTC()}
```
pkcll, line 227: @nanchano, We should try not to use attributes with unbounded cardinality. 
It can cause various problems see this thread for details 
https://chainlink-core.slack.com/archives/C06TEMFRPQD/p1729082258236499

pkg/beholder/message.go
```diff
@@ -22,9 +23,9 @@ type Metadata struct {
 	// The version of the CL node.
 	NodeVersion string
 	// mTLS public key for the node operator. This is used as an identity key but with the added benefit of being able to provide signatures.
-	NodeCsaKey string
+	NodeCsaKey string `validate:"required"`
```
pkcll, line 26: @4of9, are these going to be added as resource attributes to beholder config or we want to pass them as regular attributes?

pkg/beholder/message.go
cll-gg: I could be wrong, but the unbounded cardinality issue is a problem for metrics, not for custom events, right?

nanchano: How would we wanna represent a timestamp then? As a string and let clients parse it?

pkg/beholder/message.go
4of9: I was thinkin `NodeCsaKey` would be included as a resource attribute so users don't have to add it to every message

pkg/beholder/client_test.go
```diff
@@ -49,7 +49,9 @@ func TestClient(t *testing.T) {
 			"byte_key_1":           []byte("byte_val_1"),
 			"str_slice_key_1":      []string{"str_val_1", "str_val_2"},
 			"nil_key_1":            nil,
-			"beholder_data_schema": "/schemas/ids/1001", // Required field, URI
+			"beholder_data_schema": "/schemas/ids/1001",      // Required field, URI
+			"node_csa_key":         "node_csa_val",           // Required
+			"node_csa_signature":   "mode_csa_signature_val", // Required
```
4of9, line 54: I don't think this would be an attribute, the signature will be coming through the authentication headers

pkg/beholder/message.go
pkcll: Same applies to `NodeCsaSignature`

pkcll: We might want to add validation for resource attributes as well. 
E.g check if `NodeCsaKey` is set then `NodeCsaSignature` should be set as well

pkg/beholder/message.go
pkcll: It probably should be inside the custom message payload. 
FYI there is `Timestamp` field exist in otel log record  
https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition

> I could be wrong, but the unbounded cardinality issue is a problem for metrics, not for custom events, right?
yes, but attributes can be reused between different telemetry types e.g span attributes are copied to metric attributes.

> not for custom events, right?
I depends how we are going to store it. 

We can put it in the attribute and get back to it later. 
Whats the use case for having it in attributes?

pkcll: Loki labels suffer from hight cardinality as well:
See also https://utcc.utoronto.ca/~cks/space/blog/sysadmin/GrafanaLokiCardinalityProblem
https://grafana.com/docs/loki/latest/get-started/labels/#cardinality

```
This is high cardinality, and it can lead to significant performance degredation.

When we talk about cardinality we are referring to the combination of labels and values and the number of streams they create. High cardinality is using labels with a large range of possible values, such as ip, or combining many labels, even if they have a small and finite set of values, such as using status_code and action.

High cardinality causes Loki to build a huge index and to flush thousands of tiny chunks to the object store. Loki currently performs very poorly in this configuration. If not accounted for, high cardinality will significantly reduce the operability and cost-effectiveness of Loki.
```

cll-gg: Right, but custom events are not sent to Loki, so the cardinality issue is at play here, I think

nanchano:
closing.
csa validations herE: https://github.com/smartcontractkit/chainlink-common/pull/877
timestamp implementation (and account id potentially) will come in a different PR

PR #867, opened 2024-10-18
INFOPLAT-1376 Be able to configure retries for beholder otel exporters

pkcll:
Be able to configure retries for otel exporters.
NOTE: the retries are enabled by default in OTel SDK


pkg/beholder/config.go
```diff
@@ -18,20 +18,40 @@ type Config struct {
...
 	LogExportTimeout time.Duration
 	// Batch processing is enabled by default
 	// Disable it only for testing
 	LogBatchProcessor bool
 }
 
+type RetryConfig struct {
+	// Enabled indicates whether to not retry sending batches in case of
+	// export failure.
+	Enabled bool
```
jmank88, line 41: It looks like we could omit this explicit enabled field, if we recognized an empty/zero-value struct as disabled. If that is too implicit, another option would be `*RetryConfig` pointer fields :shrug:

pkcll: Addressed here https://github.com/smartcontractkit/chainlink-common/pull/867/commits/4cb30150f84112a6d4d228ebc812ae0600d000ba

pkg/beholder/client.go
```diff
@@ -90,16 +90,20 @@ func newGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
 			return nil, err
 		}
 	}
-	sharedLogExporter, err := otlploggrpcNew(
+	opts := []otlploggrpc.Option{
 		otlploggrpc.WithTLSCredentials(creds),
 		otlploggrpc.WithEndpoint(cfg.OtelExporterGRPCEndpoint),
-		otlploggrpc.WithRetry(otlploggrpc.RetryConfig{
+	}
+	if cfg.EmitterExporterRetryConfig != nil {
```
pkcll, line 97: By default, the retry is enabled in the OTel SDK. This check is needed to avoid disabling default retry unintentionally when EmitterExporterRetryConfig is not set.

pkg/beholder/config.go
```diff
@@ -118,3 +114,38 @@ func TestDefaultConfigHTTPClient() Config {
...
+	copy := *c
+	return &copy
+}
+
+// Calculate if retry is enabled
+func (c *RetryConfig) Enabled() bool {
+	if c == nil {
+		return false
+	}
+	return c.InitialInterval > 0 && c.MaxInterval > 0 && c.MaxElapsedTime > 0
```
pkcll, line 128: Calculated field

pkg/beholder/config.go
```diff
@@ -19,20 +19,45 @@ type Config struct {
 	EmitterExportTimeout time.Duration
 	// Batch processing is enabled by default
 	// Disable it only for testing
-	EmitterBatchProcessor bool
+	EmitterBatchProcessor      bool
+	EmitterExporterRetryConfig *RetryConfig
```
jmank88, line 23: Is this truly specific to the Emitter? If yes, should it be? i.e. why not `LogExporterRetryConfig`? :thinking:

pkcll: I plan to setup separate Exporter for Logger once we add support for it. Currently Emitter and Logger share otel sdk logger exporter

jmank88: Why is it a problem to share one logger exporter? It would be much simpler to have a single one registered globally.

pkcll: There could be noisy neighbor problem in case of verbose Logger. That can result in filling up internal exporter queue and dropped logs&custom-messages records. This is in theory, I have not verified that, and the benefit of having separate log exporter is also hypothetical as of now.

jmank88: But then this seems like an inconsistent starting point because we already have the global logger. Can we start with a global logger retry config and then introduce this field later if/when we separate them?

pkcll: Addressed here https://github.com/smartcontractkit/chainlink-common/pull/867/commits/da1ff9e8eedb326795ceb8493b3d5e2391e9f42a

pkg/beholder/config.go
```diff
@@ -20,19 +20,45 @@ type Config struct {
 	// Batch processing is enabled by default
 	// Disable it only for testing
 	EmitterBatchProcessor bool
+	// Retry config for shared log exporter, used by Emitter and Logger
+	LogRetryConfig *RetryConfig
```
jmank88, line 24: nit/ Could move down below now

jmank88 (dismissed):

pkg/beholder/config.go
pkcll: Updated https://github.com/smartcontractkit/chainlink-common/pull/867/commits/728e195dbf7b70b63d5058a7f8bc603c579c39ee

jmank88 (approved):

Atrax1 (approved):

PR #875, opened 2024-10-22
Move labelers to new pkg

vyzaldysanchez:


patrickhuie19 (commented):
could we keep `custmsg` and `metrics` as top level packages? i.e. remove `labelers` from the path

patrickhuie19 (approved):

krehermann (approved):

PR #876, opened 2024-10-22
removing values.Map from BaseMessage proto as Beholder only supports root level protos

patrickhuie19:
Temporary workaround while we are blocked on INFOPLAT-1386

Originally tried to embed values.Map into the proto, but it was very messy. This assumes that all labels will be strings, which is true for the short term.


vyzaldysanchez (changes requested):
LGTM, but we need CI working.

vyzaldysanchez (approved):

krehermann (approved):

krehermann (approved):

PR #877, opened 2024-10-22
Beholder CSA Authentication

4of9:
## What

- Adds CSA authentication support to Beholder, including LOOPPs
- Adds the CSA public key as an Otel resource attribute 
- A signer is used to craft a static header that is included on all Otel messages. LOOPPs don't have access to the CSA keystore, so the auth header is passed through env config

## Notes
- [Ticket here](https://smartcontract-it.atlassian.net/jira/software/c/projects/INFOPLAT/boards/260?selectedIssue=INFOPLAT-1071)
- [Companion Atlas PR](https://github.com/smartcontractkit/atlas/pull/7397)
- [Companion Chainlink PR](https://github.com/smartcontractkit/chainlink/pull/15040)
- [Contains fixes from this PR](https://github.com/smartcontractkit/chainlink-common/pull/902)

### Challenges

Two main challenges were in mind when arriving at an authentication mechanism for Beholder:
- There are many node operators with many nodes, so leaning on an existing system would be preferred. Rolling something out that required new credentials to be exchanged would have a lot of operational overhead
- The Chainlink node is first sending telemetry to a sidecar collector and then to the gateway

### Why no mTLS

mTLS can secure the connection between the sidecar collector and the gateway, while also providing identity and authentication

This was decided against as we wouldn't be able to lean on an existing system. While CSA keys do power mTLS today in WSRPC, they're being [used in a non-standard way](https://github.com/smartcontractkit/wsrpc/blob/main/credentials/tls.go#L71-L90) with the certificates being generated in-memory. To use these certificates in the sidecar collector would require writing them to disk (a new security consideration). 

Furthermore, mTLS is one of the reasons WSRPC is being deprecated. Since it needs to be what terminates TLS, we're not able to use L7 load balancers with more advanced security features. Using mTLS in the Gateway would put us in a similar position. It's feasible this could be made to work with a load balancer that supports mTLS, but this wasn't greatly explored 

### Why auth header

Headers can be passed through from the Chainlink node, to the sidecar collector, and up to the Gateway. We can also leverage the existing CSA keys as a means of identifying nodes. Thus, using a header for authentication seemed like the path of least resistance.


pkg/beholder/authenticator.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"maps"
+)
+
+type Authenticator struct {
+	headers   map[string]string
+	publicKey []byte
+}
+
+func NewAuthenticator(config Config) (*Authenticator, error) {
```
jmank88, line 14: Does this need to be exported?

pavel-raykov:
> ## What
> * Adds CSA authentication support to Beholder, including LOOPPs
> * Adds the CSA public key as an Otel resource attribute
> * Main flow is the `Authenticator` is given a signer and a public key. It uses the signer to craft a static header that is included on all Otel messages. The public key is what's being signed
> * LOOPPs don't have access to the CSA keystore, so here the `Authenticator` is instead passed the auth header as env config
> 
> ## Notes
> * [Ticket here](https://smartcontract-it.atlassian.net/jira/software/c/projects/INFOPLAT/boards/260?selectedIssue=INFOPLAT-1071)
> * [Companion Atlas PR](https://github.com/smartcontractkit/atlas/pull/7397)
> * [Companion Chainlink PR](https://github.com/smartcontractkit/chainlink/pull/15040)
> * [Contains fixes from this PR](https://github.com/smartcontractkit/chainlink-common/pull/902)
> 
> ## Issues
> * I've seen this work locally for `keystone` LOOPPs, but not for Aptos. It seems Aptos is started before Beholder is initialized so it receives the `noop` client. Beholder requires they Keystore to be unlocked with these changes and [I've already moved Beholder init to right after unlock](https://github.com/smartcontractkit/chainlink/pull/15040/files#diff-cf030e38d7afbc9435885ff2e8eb2070170ad23f4237522829cdce9cc719150bR403), so I'm curious if we can delay the Aptos init

Could you please explain:

1. What data is authenticated.
2. Which entity authenticates it.
3. Which entity verifies it.

4of9:
Hey @pavel-raykov 👋 

1. The goal is to create an authentication header that will be passed along with every Otel message. This header is signed and cryptographically verifies a NOP. Currently the data that's being signed is a NOP's public CSA key
2. The Otel Collector Gateway will extract the header and send it to the CSA Auth Service, which will authenticate (please see [the design document](https://docs.google.com/document/d/1ZmgmbBXFkOVh_wKVeedPVXjfa0eb3Pgk3v3EYujcRoY/edit?tab=t.0#heading=h.4e7ng0tekbkc) for more info on the auth service)
3. The CSA Auth Service will also verify
4. The keys being used are the CSA keys that are already present in the Chainlink node. They're currently used to send telemetry, amongst other things. The private keys are generated by the node and the public keys are communicated as a part of the process a NOP goes through when onboarding a new node

pavel-raykov:
> Hey @pavel-raykov 👋
> 
> 1. The goal is to create an authentication header that will be passed along with every Otel message. This header is signed and cryptographically verifies a NOP. Currently the data that's being signed is a NOP's public CSA key
> 2. The Otel Collector Gateway will extract the header and send it to the CSA Auth Service, which will authenticate (please see [the design document](https://docs.google.com/document/d/1ZmgmbBXFkOVh_wKVeedPVXjfa0eb3Pgk3v3EYujcRoY/edit?tab=t.0#heading=h.4e7ng0tekbkc) for more info on the auth service)
> 3. The CSA Auth Service will also verify
> 4. The keys being used are the CSA keys that are already present in the Chainlink node. They're currently used to send telemetry, amongst other things. The private keys are generated by the node and the public keys are communicated as a part of the process a NOP goes through when onboarding a new node

Sorry, I am not sure I fully understand: it seems that you want to sign a part of the header (the CSA public key so far) just to ensure that only NOP can send these messages. But I am not sure I understand why you need this:
 - isn't node operator and otel gateway already connected via a mTLS connection?
 - assuming this is needed, the signed part of the header is always the same (only this CSA public key); so what prevents someone who sees a single authenticated header once to "forge" arbitrary messages by just copying this signature?

4of9:
Great questions!

- We're not using mTLS on the hop between the NOP's Otel collector and the Otel Gateway. I think @pkcll can add more context here
- This approach is not sufficient to guard against replay attacks. Having a static header was agreed upon as a good first pass with a dynamic header to follow. I would like to add a time-based component to the data that's being signed so signatures can be considered invalid after some duration. I've created a [story to capture that here](https://smartcontract-it.atlassian.net/jira/software/c/projects/INFOPLAT/boards/260?selectedIssue=INFOPLAT-1443), please feel free to review

pavel-raykov:
> Great questions!
> 
> * We're not using mTLS on the hop between the NOP's Otel collector and the Otel Gateway. I think @pkcll can add more context here
> * This approach is not sufficient to guard against replay attacks. Having a static header was agreed upon as a good first pass with a dynamic header to follow. I would like to add a time-based component to the data that's being signed so signatures can be considered invalid after some duration. I've created a [story to capture that here](https://smartcontract-it.atlassian.net/jira/software/c/projects/INFOPLAT/boards/260?selectedIssue=INFOPLAT-1443), please feel free to review

Thanks for the replies. I can of course approve this PR if it is time-sensitive, but first, if possible, I would like to understand the complexity of enabling mTLS between nodes and the gateway. I am asking for this because 1. not only it is a standard approach for securing the connection, but 2. it does not require us to write any adhoc authentication code (which as we agree is already vulnerable [at least so far]).

pkg/beholder/authenticator.go
4of9: Addressed in 210eb67fecd000768569ca6f2c43ea18859269ce

pkg/beholder/global.go
```diff
@@ -43,6 +43,10 @@ func GetEmitter() Emitter {
 	return GetClient().Emitter
 }
 
+func GetAuthenticator() *Authenticator {
+	return GetClient().Authenticator
+}
```
jmank88, line 48: Can we drop this now too? (and unexport `Authenticator`?)

pkcll:
> We're not using mTLS on the hop between the NOP's Otel collector and the Otel Gateway. I think @pkcll can add more context here

There is TLS enabled (one way, not mTLS) fro traffic from from sidecar to gateway collector.
Currently there is public static Cloudflare HTTPS endpoint. It receives traffic from sidecar collector, terminates TLS and sends unencrypted traffic to otel-collector gateway.

pkcll:
> so what prevents someone who sees a single authenticated header once to "forge" arbitrary messages by just copying this signature?
The HTTP request headers are inside TLS envelope for traffic between sidecar collector and gateway collector

pkcll:
@pavel-raykov, thank you for the comments!

> it does not require us to write any adhoc authentication code 

Agree, it would be great to avoid that indeed. We considered CSA key based authentication as temporary solution.

> I would like to understand the complexity of enabling mTLS between nodes and the gateway. 
- CSA key (private, public) are stored in memory and not available to be reused for collector mTLS. 
- It would require NOP's to generate additional node private key, node cert, sign node cert by Certificate Authority (CA).
  generate gateway key,cert; sign gateway cert by CA.

Open question: 
- Is it feasible to ask NOP's to generate additional key/certificate for sidecar collector? 

If yes

- Should we rely of publicly available CA or setup private CA ?
- How to ensure private key is stored securely on the node?
- Should it be one key/certificate per sidecar node or one per NOP?

pkcll:
> This approach is not sufficient to guard against replay attacks. 

We use HTTPS between sidecar and gateway and attacker can't copy CSA key/signature headers but only able to intersect encrypted traffic and replay it. 
Cloudflare should have solutions for replay attack (see [here](https://developers.cloudflare.com/magic-wan/reference/anti-replay-protection/)). We can explore that further.

4of9:
> This approach is not sufficient to guard against replay attacks.

I think "replay attacks" was likely the wrong term to use. What I mean to illustrate is the current implementation essentially creates a session token that doesn't expire. While transport is secured via TLS, an attacker which got hold of this header could use it to send bad telemetry payloads indefinitely

patrickhuie19:
> I think "replay attacks" was likely the wrong term to use. What I mean to illustrate is the current implementation essentially creates a session token that doesn't expire. While transport is secured via TLS, an attacker which got hold of this header could use it to send bad telemetry payloads indefinitely

Including an ISO8601 timestamp with ms precision should prevent against this. You can store a hash of the request and the timestamp in an in-memory key-value cache on the server. If there are timestamp collisions in a validity window (lets say 2 minutes) of the timestamp you can route them to a handler that deals with this replay edge case.

pkg/beholder/global.go
4of9: I think we still need this to be exported? Both are being used to retrieve the headers to pass to the LOOPP e.g. `beholder.GetAuthenticator().GetHeaders()`

jmank88: I think we should avoid that: https://github.com/smartcontractkit/chainlink/pull/15040#discussion_r1826309910

4of9:
@jmank88  with the suggestion to remove globals, the latest commit greatly simplifies the PR. I've updated the companion Chainlink PR as well

@pavel-raykov as discussed, I've added some of the context to the PR description as to why this approach was chosen

Also note, this PR is branched from the latest `chainlink-common` commit on the `aptos-init` branch in Chainlink to ease local testing. Once all feedback is addressed I'll merge in `main` and fix the conflicts

pkg/loop/server.go
```diff
@@ -90,12 +90,15 @@ func (s *Server) start() error {
...
 			attributes = tracingConfig.Attributes()
 		}
+
 		beholderCfg := beholder.Config{
 			InsecureConnection:       envCfg.TelemetryInsecureConnection,
 			CACertFile:               envCfg.TelemetryCACertFile,
 			OtelExporterGRPCEndpoint: envCfg.TelemetryEndpoint,
 			ResourceAttributes:       append(attributes, envCfg.TelemetryAttributes.AsStringAttributes()...),
 			TraceSampleRatio:         envCfg.TelemetryTraceSampleRatio,
+			AuthHeaders:              envCfg.TelemetryAuthHeaders,
```
pavel-raykov, line 100: Sorry, is it clear how AuthHeaders are created? I.e., is one supposed to call the BuildAuthHeaders function with via some command line tool prior starting the node?

pavel-raykov: Ok, it is done here https://github.com/smartcontractkit/chainlink/pull/15040/.

pavel-raykov (approved):

pkg/beholder/auth.go
```diff
@@ -0,0 +1,23 @@
...
+
+// BuildAuthHeaders creates the auth header value to be included on requests.
+// The current format for the header is:
+//
+// <version>:<public_key_hex>:<signature_hex>
+//
+// where the byte value of <public_key_hex> is what's being signed
+func BuildAuthHeaders(signer func([]byte) []byte, pubKey []byte) map[string]string {
+	messageBytes := pubKey
+	signature := signer(messageBytes)
```
patrickhuie19, line 19: /nit but do we need to support multiple signing functions? Can we lock the functionality down in this module to save consumers from possibly duplicating/using invalid signing methods?

4of9: good point, I think realistically we'll just be supporting ed25519 unless something takes the place of CSA keys

4of9: addressed here 88df7ab02bd80b886d7840ef0161a23b2031e6d8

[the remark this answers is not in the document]
4of9: moved to accepting an ed25519 key instead of a signer [with this feedback](https://github.com/smartcontractkit/chainlink-common/pull/877#discussion_r1833222924)

pkg/beholder/auth.go
```diff
@@ -0,0 +1,26 @@
...
+var authHeaderVersion = "1"
+
+// BuildAuthHeaders creates the auth header value to be included on requests.
+// The current format for the header is:
+//
+// <version>:<public_key_hex>:<signature_hex>
+//
+// where the byte value of <public_key_hex> is what's being signed
+func BuildAuthHeaders(privKey ed25519.PrivateKey, pubKey ed25519.PublicKey) map[string]string {
+	messageBytes := pubKey
```
jmank88, line 21: If we are passing the private key, can we just ask it for the public?
```suggestion
	messageBytes := privKey.Public()
```
I guess it adds a type cast...

4of9: addressed here b969b66e6141c80565d51ecd64a8bb39d4c09811

jmank88 (approved):

pkcll (approved):

pkcll (approved):

pavel-raykov (approved):

PR #878, opened 2024-10-22
pkg/services: add (*Engine).Tracer()

jmank88:
This PR adds an `(*Engine).Tracer() trace.Tracer` method to return the named tracer derived from the `Logger.Name()`.


pkg/services/service.go
```diff
@@ -79,6 +86,16 @@ func (e *Engine) GoTick(ticker *timeutil.Ticker, fn func(context.Context)) {
 	})
 }
 
+// TraceAs wraps fn in a tracing span with service attributes and the given name.
+// Consider wrapping funcs passed to Go or GoTick.
+func (e *Engine) TraceAs(name string, fn func(context.Context)) func(context.Context) {
```
patrickhuie19, line 91: would it be helpful to add a variadic strings param for key value pairs to associate with the trace?

jmank88: I think I would rather just expose the Tracer than make this complex. Any kind of variadic after the func can be awkward since these are often large closures.

jmank88: Update to `Tracer() trace.Tracer` and added an example

vyzaldysanchez (approved):

patrickhuie19 (approved):

pkg/services/service.go
```diff
@@ -162,12 +184,15 @@ const scopeName = "github.com/smartcontractkit/chainlink-common/pkg/services"
...
 	s := &service{
-		tracer: otel.GetTracerProvider().Tracer(scopeName),
-		cfg:    c,
+		cfg: c,
 		eng: Engine{
 			StopChan:      make(StopChan),
 			SugaredLogger: lggr,
-			conds:         make(map[string]error),
+			tracer: otel.GetTracerProvider().Tracer(scopeName, trace.WithInstrumentationAttributes(
+				attribute.String("service.name", c.Name),
```
pkcll, line 192: Strictly speaking `service.name` belongs to ResourceAttributes and its one of the reserved OTel attributes 
https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_service_name
https://opentelemetry.io/docs/languages/js/resources/

This one here is InstrumentationScope attribute 
https://opentelemetry.io/docs/concepts/instrumentation-scope/

Alternatively: 
-  `service.name` could be set via beholder config.ResourceAttributes from where it will propagate to resource attributes of metric,tracer,logger providers
- use another attribute here `chainlink.service` or something 
  
But if its better fits chainlink model its fine to set it here

pkcll (dismissed):

pkg/services/service.go
jmank88: `Name` is a non-unique, short, internal service name - often just the Go type.
`lggr.Name()` is the fully-qualified service name, including the chain of ancestor service names from the hierarchy, making it unique within the node and tied to the logs.

What would you suggest that we name them?

pkcll: > What would you suggest that we name them?

Idk, but something different from `service.name` since its reserved.
Examples: `service.full_name`, `service.path`, `service.scope`



  - _The [Instrumentation scope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope) 
  represents a logical unit within the application code with which the emitted telemetry can be associated._

jmank88: It sounds like `service` is the wrong scope though? I think these are more like: `<thing>.name` & `<thing>.type`/`<thing>.kind`

pkcll: Right, that <thing> could be something that represents logic code component
Could be: `component.name`, 'unit.name', 'block.name', 'package.name'

patrickhuie19: It seems like `service.name` is intended by OTEL to be at the application level, whereas we use service within the context of an application. to differentiate, we could call it `chainlink_service.name`?

jmank88: I would think there is some standard since named loggers are common. `component` seem appealing :shrug:

patrickhuie19: 🤷 https://github.com/search?q=trace.WithInstrumentationAttributes&type=code&p=1 seems fairly ad hoc. I'm ok with component though my instinct is that a component is composed of multiple lower level services

jmank88: The [zapcore.Entry](https://pkg.go.dev/go.uber.org/zap/zapcore#Entry) passed to [otlpCore.Write](https://github.com/agoda-com/opentelemetry-go/blob/main/otelzap/core.go#L73) has a `LoggerName` field but it is not used :shrug:

jmank88: [log.LoggerProvider.Logger](https://pkg.go.dev/go.opentelemetry.io/otel/log#LoggerProvider) takes a name as the first argument, as we can see being used here: https://github.com/uptrace/opentelemetry-go-extra/blob/otelzap/v0.3.2/otelzap/otelzap.go#L68
However, it is documented as the _package_ name, and we do not use package scoped loggers. It may be OK since it is just a recommendation though.

jmank88: Ah, so the tracer takes a name just like logger, so those should be aligned. That just leaves the `Name` which is actually redundant since it will be the suffix of the fully qualified name. Maybe should just drop it?

jmank88: Updated to simplify so this isn't an issue.

EasterTheBunny (approved):

EasterTheBunny (approved):

pkcll (approved):

patrickhuie19 (approved):

PR #883, opened 2024-10-23
[CAPPL-121] context propagation

agparadiso:
### Description

This pr implements context propagation. 
The way this is implemented on the wasm host-guest communication is by modifying the store (the shared buffer between host-guest) including the ctx.


pkg/workflows/wasm/host/module.go
```diff
@@ -24,31 +25,42 @@ import (
...
 	}
 
-	r.m[id] = resp
+	// we only add the response as the context has already been added
+	if found && req.response == nil {
+		r.m[id].response = storedReq.response
+		return nil
+	}
+
+	r.m[id] = req
```
cedric-cordenier, line 53: @agparadiso I think you're right: we can just let this be the same as before (i.e. revert the diff to this method), and assume we'll only add it once

If we want to mutate the entry, we can get the entry and modify it 👍

pkg/workflows/wasm/host/module.go
```diff
@@ -326,7 +338,18 @@ func (m *Module) Close() {
...
+	if request == nil {
+		return nil, fmt.Errorf("invalid request: can't be nil")
+	}
+
+	if request.Id == "" {
+		return nil, fmt.Errorf("invalid request: can't be empty")
+	}
+
+	// we add the request context to the store to make it available to the Fetch fn
+	m.requestStore.add(request.Id, &RequestData{ctx: ctx})
```
cedric-cordenier, line 351: I'm surprised linting hasn't complained here -- you need to check that we didn't error here

pkg/workflows/wasm/host/module.go
```diff
@@ -401,31 +429,42 @@ func containsCode(err error, code int) bool {
...
 
-func fetchFn(logger logger.Logger, modCfg *ModuleConfig) func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
-	const fetchErrSfx = "error calling fetch"
+func fetchFn(logger logger.Logger, modCfg *ModuleConfig, store *store) func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
+	logFetchErr := func(err error) { logger.Errorf("error calling fetch: %s", err.Error()) }
 	return func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
 		b, innerErr := wasmRead(caller, reqptr, reqptrlen)
 		if innerErr != nil {
-			logger.Errorf("%s: %s", fetchErrSfx, innerErr)
+			logFetchErr(innerErr)
```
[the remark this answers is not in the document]
agparadiso: 🤔 you might be right. I had some conflicts when merging main branch and might have brought back this by mistake. checking...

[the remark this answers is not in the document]
agparadiso: info for reviewers: this hasn't been addressed before, it will be done on this new task [CAPPL-195](https://smartcontract-it.atlassian.net/browse/CAPPL-195)

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -459,28 +476,92 @@ func Test_Compute_Fetch(t *testing.T) {
 			assert.Equal(t, expectedEntries[i].Log.Message, logs.AllUntimed()[i].Entry.Message)
 		}
 	})
+
+	t.Run("OK_context_propagation", func(t *testing.T) {
```
MStreet3, line 480: should test that the context also cancels a function call by making a fetch that waits on ctx.Done()

agparadiso: 👍🏼  thanks! added!

pkg/workflows/wasm/host/module.go
```diff
@@ -366,22 +386,27 @@ func (m *Module) Run(request *wasmpb.Request) (*wasmpb.Response, error) {
 	_, err = start.Call(store)
 	switch {
 	case containsCode(err, wasm.CodeSuccess):
-		resp, innerErr := m.r.get(request.Id)
+		storedRequest, innerErr := m.requestStore.get(request.Id)
```
cedric-cordenier, line 389: We should actually pop the item here, not just get it. This is so that we ensure we're removing items once we've completed the request.

pkg/workflows/wasm/host/module.go
```diff
@@ -318,7 +324,21 @@ func (m *Module) Close() {
...
+	if request == nil {
+		return nil, fmt.Errorf("invalid request: can't be nil")
+	}
+
+	if request.Id == "" {
+		return nil, fmt.Errorf("invalid request: can't be empty")
+	}
+
+	// we add the request context to the store to make it available to the Fetch fn
+	err := m.requestStore.add(request.Id, &RequestData{ctx: ctx})
```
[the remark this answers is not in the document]
agparadiso: thanks @jmank88, I've modified it to store a function that returns a context on [this commit](https://github.com/smartcontractkit/chainlink-common/pull/883/commits/cce6e43ace30f3aad9f1a38d6c55fd3228afa7fc)

now the struct looks like this:
```
type RequestData struct {
	response *wasmpb.Response
	ctxFn    func() context.Context
}
```
wdyt about it?

pkg/workflows/wasm/host/module.go
```diff
@@ -366,22 +400,27 @@ func (m *Module) Run(request *wasmpb.Request) (*wasmpb.Response, error) {
 	_, err = start.Call(store)
 	switch {
 	case containsCode(err, wasm.CodeSuccess):
-		resp, innerErr := m.r.get(request.Id)
+		storedRequest, innerErr := m.requestStore.pop(request.Id)
```
cedric-cordenier, line 403: @agparadiso I don't think this is enough to remove the job in all cases: if we fall through any of the other cases the entry we added will still remain.

Probably a better way to handle this is to move the pop so that it's global, i.e. move to just after line 400.

Then handle the error, and where you need to use a response, you can

pkg/workflows/wasm/host/module.go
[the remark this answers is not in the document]
jmank88: I think we need to prefer solutions that follow the package recommendations of passing context as the first function argument:
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

pkg/workflows/wasm/host/module.go
```diff
@@ -393,39 +432,56 @@ func containsCode(err error, code int) bool {
...
+			return ErrnoFault
+		}
+
+		if storedRequest.ctxFn() == nil {
+			logFetchErr(errors.New("context is nil"))
 			return ErrnoFault
 		}
 
-		fetchResp, innerErr := modCfg.Fetch(req)
+		fetchResp, innerErr := modCfg.Fetch(storedRequest.ctxFn(), req)
```
jmank88, line 468: If this is too messy:
```suggestion
		var fetchResp Whatever
		var innerErr error 
		storedRequest.callWithCtx(func(ctx context.Context) {
			fetchResp, innerErr = modCfg.Fetch(ctx, req)
		})
```
Then the signature can be expanded to match the return args:
```suggestion
		fetchResp, innerErr := storedRequest.callWithCtx(func(ctx context.Context) (Whatever, error) {
			return modCfg.Fetch(ctx, req)
		})
```

pkg/workflows/wasm/host/module.go
agparadiso: good catch, I switched the implementation to have a `delete` method instead of `pop` and `defer delete()` after adding it to the store

pkg/workflows/wasm/host/module.go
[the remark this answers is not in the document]
agparadiso: thanks for the example you provided below, I think now I was able to capture what you meant in the [last commit](https://github.com/smartcontractkit/chainlink-common/pull/883/commits/ca4c9ab5086ac93e27507b31d64d1a87de6f9b1c)

pkg/workflows/wasm/host/module.go
```diff
@@ -393,39 +429,57 @@ func containsCode(err error, code int) bool {
...
-func fetchFn(logger logger.Logger, modCfg *ModuleConfig) func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
-	const fetchErrSfx = "error calling fetch"
+func createFetchFn(
+	logger logger.Logger,
+	reader unsafeReaderFunc,
+	writer unsafeWriterFunc,
+	sizeWriter unsafeFixedLengthWriterFunc,
+	modCfg *ModuleConfig, store *store,
+) func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
+	logFetchErr := func(err error) { logger.Errorf("error calling fetch: %s", err.Error()) }
```
jmank88, line 439: Is this helper just for convenience? Because this approach ends up using the wrong `caller` file/line-number. There are work arounds, but it is generally best to inline these.

agparadiso: we used to have everything inline but it was getting a bit messy with `Fetch` & `Emit` functions so we started extracting it. you mean the stacktrace/debugging would have the wrong file/line-number?  was not aware of this, Im only modifying the name here, but will def bring this up

jmank88 (approved):

cedric-cordenier (approved):

PR #885, opened 2024-10-23
refactor(custmsg+wasm): adjust message emitter iface

MStreet3:
passing the `custmsg.Labeler` to `host.ModuleConfig.Labeler` is blocked because the types do not match.  this PR fixes the type mismatch:

* removes the `MessageEmitter` interface from `wasm/host` as it is self-referential and therefore cannot be satisfied by an implementation that does not import `wasm/host`.

* `MessageEmitter` is instead defined in `custmsg` and imported into `wasm/host`.  Now `custmsg.Labeler` implements `custmsg.MessageEmitter` and can be passed directly to `host.ModuleConfig`.

### DIFF
--- a/pkg/custmsg/custom_message.go
+++ b/pkg/custmsg/custom_message.go
@@ -11,6 +11,20 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
+type MessageEmitter interface {
+	// Emit sends a message to the labeler's destination.
+	Emit(string) error
+
+	// WithMapLabels sets the labels for the message to be emitted.  Labels are cumulative.
+	WithMapLabels(map[string]string) MessageEmitter
+
+	// With adds multiple key-value pairs to the emission.
+	With(keyValues ...string) MessageEmitter
+
+	// Labels returns a view of the current labels.
+	Labels() map[string]string
+}
+
 type Labeler struct {
 	labels map[string]string
 }
@@ -21,7 +35,7 @@ func NewLabeler() Labeler {
 
 // WithMapLabels adds multiple key-value pairs to the CustomMessageLabeler for transmission
 // With SendLogAsCustomMessage
-func (l Labeler) WithMapLabels(labels map[string]string) Labeler {
+func (l Labeler) WithMapLabels(labels map[string]string) MessageEmitter {
 	newCustomMessageLabeler := NewLabeler()
 
 	// Copy existing labels from the current agent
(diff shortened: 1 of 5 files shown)


cedric-cordenier (approved):

pkg/custmsg/custom_message.go
```diff
@@ -11,17 +11,31 @@ import (
...
-func NewLabeler() Labeler {
-	return Labeler{labels: make(map[string]string)}
+func NewLabeler() labeler {
+	return labeler{labels: make(map[string]string)}
 }
 
 // WithMapLabels adds multiple key-value pairs to the CustomMessageLabeler for transmission
 // With SendLogAsCustomMessage
-func (l Labeler) WithMapLabels(labels map[string]string) Labeler {
+func (l labeler) WithMapLabels(labels map[string]string) MessageEmitter {
```
patrickhuie19, line 38: idiomatic go to accept interfaces as params and return structs. could we do that here?

pkg/custmsg/custom_message.go
```diff
@@ -38,7 +52,7 @@ func (l Labeler) WithMapLabels(labels map[string]string) Labeler {
 }
 
 // With adds multiple key-value pairs to the CustomMessageLabeler for transmission With SendLogAsCustomMessage
-func (l Labeler) With(keyValues ...string) Labeler {
+func (l labeler) With(keyValues ...string) MessageEmitter {
```
patrickhuie19, line 55: same comment as above

pkg/custmsg/custom_message.go
MStreet3: not a hard axiom.  In this case there is an exception for two reasons:
1. `labeler` is the only implementation of `MessageEmitter` (see Effective Go on [generality](https://go.dev/doc/effective_go#generality) of the interface rule)
2. the interface is self-referential, if the methods don't return `MessageEmitter` then `labeler` won't implement `MessageEmitter`

MStreet3: `Labeler` was originally public, but all of its fields are private, which makes it redundant to the interface so I made it private.  That's a third reason to not return `labeler` directly.

justinkaseman (dismissed):

pkg/custmsg/custom_message.go
patrickhuie19: Let's keep `Labeler` public for continued direct consumption by the engine and registry syncer in core.

patrickhuie19 (approved):

justinkaseman (approved):

PR #888, opened 2024-10-24
[CAPPL-132] Handle secrets in runner SDK

cedric-cordenier:
* Add a variant of Compute which accepts user-provided config.
* Add a SecretValue and interpolate secrets in the test runner.


pkg/workflows/sdk/builder.go
```diff
@@ -238,3 +238,35 @@ func AnyMap[M ~map[string]any](inputs CapMap) CapDefinition[M] {
...
+		keys: append(s.keys, key),
+	}
+}
+
+func (s Secrets) Ref() string {
+	if len(s.keys) == 0 {
+		return "$(ENV.secrets)"
+	}
+
+	str := strings.Join(s.keys, ".")
```
nolag, line 270: This is interesting, do we expect this use case?

pkg/workflows/sdk/builder.go
```diff
@@ -238,3 +238,35 @@ func AnyMap[M ~map[string]any](inputs CapMap) CapDefinition[M] {
...
 }
+
+// Alt 1:
+func XSecrets() string {
+	return "$(ENV.secrets)"
+}
+
+func XSecretNamed(name string) string {
+	return fmt.Sprintf("$(ENV.secrets.%s)", name)
+}
```
nolag, line 249: I like this one, but I'm opened to either approach.  

Maybe just secret, no need for Named.  It's common pattern to have foos() and foo(name string).  I wonder if they should return a special type

type Seceret string

then you can implement CapDefinition[String] with so it can be passed as input to a capability, and it can also easily be cast to a string.  Thoughts?

pkg/workflows/sdk/gen/compute.go.tmpl
```diff
@@ -6,6 +6,39 @@ import (
...
+        return nil, errors.New("`config` is a reserved keyword inside Compute config")
+    }
+    cm["config"] = "$(ENV.config)"
+
+    if _, ok := cm["binary"]; ok {
+        return nil, errors.New("`binary` is a reserved keyword inside Compute config")
+    }
+    cm["binary"] = "$(ENV.binary)"
+
+    return cm, nil
```
nolag, line 39: do we need a check for outputs too?

pkg/workflows/sdk/gen/compute.go.tmpl
```diff
@@ -6,6 +6,39 @@ import (
     "github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
+type ComputeConfig[C any] struct {
```
nolag, line 9: This struct can go in compute.go, since we don't need more than one of them, there's no need to generate it :).

pkg/workflows/sdk/builder.go
cedric-cordenier: We don't, secrets is just a map[string]string, so I removed this

pkg/workflows/sdk/builder.go
cedric-cordenier: Please take another look and let me know what you think :) Note: atm secrets (and env var interpolation in general) only works on Config; but we could easily change this if necessary.

pkg/workflows/sdk/gen/compute.go.tmpl
cedric-cordenier: I'm not sure :/ Can you say more about why you think we need it?

pkg/workflows/sdk/builder.go
```diff
@@ -238,3 +249,23 @@ func AnyMap[M ~map[string]any](inputs CapMap) CapDefinition[M] {
...
+
+func (s Secret) self() CapDefinition[string] {
+	return s
+}
+
+func Secrets() Secret {
+	return "$(ENV.secrets)"
+}
+
+func SecretNamed(named string) Secret {
```
nolag, line 269: nit: I think Secret(name sting) is more natural.  The arg implies what you're doing, and it's clear that one gets all secrets and the other gets on secret.

pkg/workflows/sdk/compute.go
```diff
@@ -20,3 +26,40 @@ func (c *computeOutputCap[T]) Value() CapDefinition[T] {
...
+
+	uc, err := wc.Unwrap()
+	if err != nil {
+		return nil, err
+	}
+
+	cm, ok := uc.(map[string]any)
+	if !ok {
+		return nil, errors.New("could not convert config into map")
+	}
```
nolag, line 48: Do you think it's worth a quick check up front to see if C is already a map[string]any?

var cmp map[string, any]
var ok bool

if cm, ok = uc.(map[string]any); !ok {
   c.toMap() // this function is just the lines below.
}

nolag (approved):

pkg/workflows/sdk/builder.go
cedric-cordenier: This is what I had initially but then I don't know what to call the Secret type; maybe SecretValue?

cedric-cordenier: I've called the Secret type SecretValue and renamed this to Secret

agparadiso (approved):

ettec (approved):

PR #890, opened 2024-10-24
Add Timestamp to Beholder attributes

nanchano:


pkg/beholder/message.go
```diff
@@ -166,6 +169,8 @@ func OtelAttr(key string, value any) otellog.KeyValue {
 		return otellog.Bool(key, v)
 	case []byte:
 		return otellog.Bytes(key, v)
+	case time.Time:
+		return otellog.Int64(key, v.Unix())
```
cll-gg, line 173: In `beholder/client.go` we're setting `timestamp` to `time.Now().UTC()` which is a string representing a date time. Here however we're using the epoch timestamp. Or are these unrelated?

nanchano: UTC() returns a `time.Time` which can be parsed into a string or any kind of unix timestamp. I decided to use a unix timestamp here but we can render it as a string too, don't really have a preference.

pkg/beholder/client.go
```diff
@@ -239,6 +240,11 @@ func newOtelResource(cfg Config) (resource *sdkresource.Resource, err error) {
 // Open question: what are pros/cons for using use map[]any vs use otellog.KeyValue
 func (e messageEmitter) Emit(ctx context.Context, body []byte, attrKVs ...any) error {
 	message := NewMessage(body, attrKVs...)
+
+	if _, ok := message.Attrs["timestamp"]; !ok {
+		message.Attrs["timestamp"] = time.Now().UTC()
+	}
```
jmank88, line 246: It looks like `log.Record`s already have a https://pkg.go.dev/go.opentelemetry.io/otel/log#Record.Timestamp and an https://pkg.go.dev/go.opentelemetry.io/otel/log#Record.ObservedTimestamp. Could we lever those built-ins instead?

pkcll: yep, I think I mentioned that somewhere in other PR previously

nanchano: Done

pkg/beholder/client.go
```diff
@@ -247,10 +247,17 @@ func newOtelResource(cfg Config) (resource *sdkresource.Resource, err error) {
 // Open question: what are pros/cons for using use map[]any vs use otellog.KeyValue
 func (e messageEmitter) Emit(ctx context.Context, body []byte, attrKVs ...any) error {
 	message := NewMessage(body, attrKVs...)
+	record := message.OtelRecord()
+
+	if _, ok := message.Attrs["timestamp"]; !ok {
```
patrickhuie19, line 252: This feels like a separate concern than Emit - can we handle attribute hydration higher up the stack?

pkg/beholder/client.go
jmank88: I expected this to be a much more simplifying change. Why do we still need to pass this around as metadata/attribute if we are using the built-in field?

nanchano:
Closing as it won't be done

PR #892, opened 2024-10-25
(fix): Enforce Consensus Capability config field key_id

justinkaseman:


pkg/capabilities/consensus/ocr3/models.go
```diff
@@ -12,7 +12,7 @@ type config struct {
 	ReportID          string      `mapstructure:"report_id" json:"report_id" jsonschema:"required,pattern=^[a-f0-9]{4}$"`
 	RequestTimeoutMS  int64       `mapstructure:"request_timeout_ms" json:"request_timeout_ms"`
 
-	KeyID string `mapstructure:"key_id" json:"key_id"`
+	KeyID string `mapstructure:"key_id" json:"key_id,omitempty" jsonschema:"required"`
```
bolekk, line 15: Apart from this, we should also check if the desired key ID is available, probably somewhere very early in the consensus step. I'm not sure if that's easy without some refactoring...

justinkaseman: When you say if the desired key ID is available, we would use this key ID from the workflow spec, where would we be checking against?

bolekk: Against all possible bundle IDs available in OCR3OnchainKeyringMultiChainAdapter (essentially in Keystore). That's why it might be a bit trickier because I don't think we have easy access to that data where we need to run a check... but maybe I'm wrong

bolekk (approved):

agparadiso (dismissed):

bolekk (approved):

PR #893, opened 2024-10-25
feat(observability-lib): improve alerting provisioning

Atrax1:
Alert rules are now updated when already existing instead of deleted all alerts attached to dashboard and then re-create them
State is not resetted when re-deploying alerts
Notification policy fix, no more duplicates 
PrettifyingJSON option for log panel
Opacity for timeseries panel fix default to 0
Add log panel to Node general dashboard
Disable provenance for contact points
Add pagerduty notification template


jmank88 (approved):

Atrax1:
> > Disable provenance for contact points
> 
> > Opacity for timeseries panel fix default to 0
> 
> @Atrax1 missing some context here - could you fill me in? What do these changes accomplish?

Yes sure,

Provenance is the ability to modify created contact points through API from UI
Opacity is related to the graph opacity on the panels

patrickhuie19 (approved):

PR #894, opened 2024-10-25
[CAPPL-128] limit the amount of fetch calls per request

agparadiso:
### Description

This pr introduces a hard limit on the amount of Fetch calls you can make per request. This can be configure using the config, if not present a default max of 5 (open to suggestions on what the number should be) is set.

[CAPPL-128](https://smartcontract-it.atlassian.net/browse/CAPPL-128)

### Supports

[CAPPL-128]: https://smartcontract-it.atlassian.net/browse/CAPPL-128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/workflows/wasm/host/module.go
```diff
@@ -139,6 +140,10 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 		}
 	}
 
+	if modCfg.MaxFetchRequests == 0 {
+		modCfg.MaxFetchRequests = 5
```
cedric-cordenier, line 144: @agparadiso Can we declare this as a const like we do on line 75

pkg/workflows/wasm/host/module.go
```diff
@@ -434,10 +439,20 @@ func createFetchFn(
 	reader unsafeReaderFunc,
 	writer unsafeWriterFunc,
 	sizeWriter unsafeFixedLengthWriterFunc,
-	modCfg *ModuleConfig, store *store,
+	modCfg *ModuleConfig,
+	store *store,
 ) func(caller *wasmtime.Caller, respptr int32, resplenptr int32, reqptr int32, reqptrlen int32) int32 {
-	const errFetchSfx = "error calling fetch"
+	fetchCounter := int64(0)
```
cedric-cordenier, line 445: @agparadiso This won't work :/ We need to store the counter in the store, since it should be a request-level counter.

pkg/workflows/wasm/host/module.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 type RequestData struct {
+	counter     int64
```
cedric-cordenier, line 30: Can we rename this to fetchRequestsCounter :)

pkg/workflows/wasm/host/module.go
```diff
@@ -69,24 +70,26 @@ func (r *store) delete(id string) {
...
 var (
-	defaultTickInterval = 100 * time.Millisecond
-	defaultTimeout      = 2 * time.Second
-	defaultMaxMemoryMBs = 256
-	DefaultInitialFuel  = uint64(100_000_000)
+	defaultTickInterval     = 100 * time.Millisecond
+	defaultTimeout          = 2 * time.Second
+	defaultMaxMemoryMBs     = 256
+	DefaultInitialFuel      = uint64(100_000_000)
+	defaultMaxFetchRequests = int64(5)
```
cedric-cordenier, line 77: @agparadiso This can just be an int I think: it's only used by our code, not by wasmtime-go (which would require an int64, and is why we use that for MaxMemoryMBs for example).

pkg/workflows/wasm/host/module.go
```diff
@@ -451,12 +460,19 @@ func createFetchFn(
...
 			logger.Errorf("%s: %s", errFetchSfx, innerErr)
 			return ErrnoFault
 		}
 
+		// limit the number of fetch calls we can make per request
+		if storedRequest.counter >= modCfg.MaxFetchRequests {
+			logger.Errorf("%s: max number of fetch request %d exceeded", errFetchSfx, modCfg.MaxFetchRequests)
+			return ErrnoFault
+		}
+		defer func() { storedRequest.counter++ }()
```
cedric-cordenier, line 474: You don't really need to defer this; just increment it after you've checked it :)

cedric-cordenier (approved):

PR #897, opened 2024-10-25
Fix a bug where only one file is used for user type generation

nolag:


pkg/capabilities/cli/cmd/generate_user_types.go
```diff
@@ -13,16 +13,17 @@ func GenerateUserTypes(info UserGenerationInfo) error {
...
+	for i, file := range dir {
 		fileName := file.Name()
 		if file.IsDir() || !strings.HasSuffix(fileName, ".go") {
 			continue
 		}
 
 		rawContent, err2 := os.ReadFile(path.Join(info.Dir, fileName))
 		if err2 != nil {
-			return err2
+			errs = append(errs, err2)
```
krehermann, line 26: nit/ you can build the error inline rather than two steps
var error
if err2 != nil {
err= errors.Join(err, err2)
}
...
if err !=nil {
return err
}

nolag: Did it for now, lmk if I should change it back.

krehermann (approved):

jmank88 (approved):

PR #898, opened 2024-10-29
update head data to be optional

ettec:
The underlying head tracker returns a pointer to head data, this updates the GetLatestValueWithHeadData to do the same.  Also, implements the loop logic for the GetLatestValueWithHeadData API method.


ilija42 (dismissed):

ilija42 (changes requested):
Look at `TestGetLatestValue` in `pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader_test.go`, this method should have similar tests

ilija42 (approved):

ilija42 (changes requested):
there is a test failure

ettec:
Dead - GetLatestValueWithHeadData added as part of another PR.

PR #901, opened 2024-10-29
custmsg domain chainlink --> platform

patrickhuie19:
https://smartcontract-it.atlassian.net/browse/INFOPLAT-1403

### DIFF
--- a/pkg/custmsg/custom_message.go
+++ b/pkg/custmsg/custom_message.go
@@ -117,7 +117,7 @@ func sendLogAsCustomMessageW(msg string, labels map[string]string) error {
 
 	err = beholder.GetEmitter().Emit(context.Background(), payloadBytes,
 		"beholder_data_schema", "/beholder-base-message/versions/1", // required
-		"beholder_domain", "chainlink", // required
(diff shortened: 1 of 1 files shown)


pkg/custmsg/custom_message.go
```diff
@@ -117,7 +117,7 @@ func sendLogAsCustomMessageW(msg string, labels map[string]string) error {
 
 	err = beholder.GetEmitter().Emit(context.Background(), payloadBytes,
 		"beholder_data_schema", "/beholder-base-message/versions/1", // required
-		"beholder_domain", "chainlink", // required
+		"beholder_domain", "platform", // required
 		"beholder_entity", "BaseMessage", // required
```
jmank88, line 121: Why do we use underscores instead of dot separation for these?

jmank88 (approved):

justinkaseman (approved):

pkg/custmsg/custom_message.go
patrickhuie19: @cll-gg may know, the naming of these keys was decided by the Beholder team

PR #903, opened 2024-10-30
Extract chain reads from ChainService relayer into Chain Reader

ilija42:


pkg/types/relayer.go
```diff
@@ -78,33 +78,34 @@ type NodeStatus struct {
...
+	NewContractReader(ctx context.Context, contractReaderConfig []byte) (ContractReader, error)
 	// LatestHead returns the latest head for the underlying chain.
 	LatestHead(ctx context.Context) (Head, error)
 	// GetChainStatus returns the ChainStatus for this Relayer.
 	GetChainStatus(ctx context.Context) (ChainStatus, error)
 	// ListNodeStatuses returns the status of RPC nodes.
 	ListNodeStatuses(ctx context.Context, pageSize int32, pageToken string) (stats []NodeStatus, nextPageToken string, total int, err error)
+}
+
+// ChainService is a sub-interface that encapsulates the explicit interactions with a chain, rather than through a provider.
```
jmank88, line 93: This comment is no longer true

pkg/types/relayer.go
```diff
@@ -78,33 +78,34 @@ type NodeStatus struct {
...
+type ChainService interface {
+	Service
```
jmank88, line 95: Why is it OK to change this interface? Seems like it should include `ChainReader`:
```suggestion
type ChainService interface {
	Service
	ChainReader
```

pkg/types/relayer.go
```diff
@@ -78,33 +78,34 @@ type NodeStatus struct {
...
 	// If balanceCheck is true, the balance will be checked before submitting.
 	Transact(ctx context.Context, from, to string, amount *big.Int, balanceCheck bool) error
 }
 
-// Relayer extends ChainService with providers for each product.
+// Relayer extends ChainService and ChainReader with providers for each product.
 type Relayer interface {
 	ChainService
-
+	ChainReader
```
jmank88, line 104: Oh, maybe we want:
```go
type Relayer interface {
	ChainReader
	ChainWriter
	...
}

type ChainService interface {
	Service
	ChainReader
	ChainWriter
}
type ChainWriter interface {
	// Transact submits a transaction to transfer tokens.
	// If balanceCheck is true, the balance will be checked before submitting.
	Transact(ctx context.Context, from, to string, amount *big.Int, balanceCheck bool) error

	// NewChainWriter returns a new ChainWriter.
	// The format of config depends on the implementation.
	NewChainWriter(ctx context.Context, config []byte) (ChainWriter, error)
}
```

pkg/types/relayer.go
ilija42: Sry, should've marked this as draft, I haven't given it too much thought yet, but I did assume that we want the end state to be ChainWriter and ChainReader without ChainService.

pkg/types/relayer.go
ilija42: Yeah, but the ChainWriter has to be renamed to ContractWriter first

pkg/loop/internal/types/types.go
```diff
@@ -44,14 +44,6 @@ type OCR3CapabilityProvider interface {
...
 	types.ChainService
-
-	// NewChainWriter returns a new ChainWriter.
-	// The format of config depends on the implementation.
-	NewChainWriter(ctx context.Context, chainWriterConfig []byte) (types.ChainWriter, error)
-
-	// NewContractReader returns a new ContractReader.
-	// The format of contractReaderConfig depends on the implementation.
-	NewContractReader(ctx context.Context, contractReaderConfig []byte) (types.ContractReader, error)
 	NewConfigProvider(context.Context, types.RelayArgs) (types.ConfigProvider, error)
```
ilija42, line 47: Should this contain the chainService? Have to check how Relayer interface is used

pkg/types/chain_writer.go
```diff
@@ -11,6 +11,9 @@ const (
 	ErrSettingTransactionGasLimitNotSupported = InvalidArgumentError("setting transaction gas limit is not supported")
 )
 
+type ContractWriter = ChainWriter
+
+// Deprecated: Use ContractWriter instead.
 type ChainWriter interface {
```
jmank88, line 17: Swapping these makes the "correct" type more direct and sets up removal to be a little cleaner later :shrug: 
```suggestion
// Deprecated: Use ContractWriter instead.
type ChainWriter = ContractWriter

type ContractWriter interface {
```

pkg/types/relayer.go
```diff
@@ -78,33 +78,38 @@ type NodeStatus struct {
...
+	NewContractReader(ctx context.Context, contractReaderConfig []byte) (ContractReader, error)
 	// LatestHead returns the latest head for the underlying chain.
 	LatestHead(ctx context.Context) (Head, error)
 	// GetChainStatus returns the ChainStatus for this Relayer.
 	GetChainStatus(ctx context.Context) (ChainStatus, error)
 	// ListNodeStatuses returns the status of RPC nodes.
 	ListNodeStatuses(ctx context.Context, pageSize int32, pageToken string) (stats []NodeStatus, nextPageToken string, total int, err error)
+}
+
+type IChainWriter interface {
```
jmank88, line 93: This is a bummer... :thinking: We could rename it later of course - are you planning to do that? Or leave the deprecated type around for a while?

ilija42: Everyone that has context on this is slammed with work, I've been chipping away at the backlog when I have time. Maybe I'll do it, or I'll try to somehow prioritise the backlog

pkg/types/chain_writer.go
ilija42: ah, didn't realise that was legal

pkg/types/relayer.go
jmank88: Is this conflict resolved now?
```suggestion
type ChainWriter interface {
```

pkg/loop/internal/types/types.go
pablolagreca, line 47: Can you document NewConfigProvider function?.

pkg/types/relayer.go
```diff
@@ -78,33 +78,38 @@ type NodeStatus struct {
...
 	// If balanceCheck is true, the balance will be checked before submitting.
 	Transact(ctx context.Context, from, to string, amount *big.Int, balanceCheck bool) error
 }
 
+// ChainService encapsulates ChainReader and IChainWriter sub-interfaces that encapsulate explicit read/write interactions with a chain and smart contract read/write components ContractReader and ContractWriter.
+type ChainService interface {
+	Service
+
+	ChainReader
+	IChainWriter
```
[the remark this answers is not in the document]
ilija42: Because ChainWriter is yet to be renamed to ContractWriter

pkg/loop/internal/types/types.go
ilija42: Not sure what NewConfigProvider does, seems out of scope her

pkg/types/relayer.go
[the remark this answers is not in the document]
jmank88: Can we do that rename first in order to make this change smoother? Then we don't need to introduce additional types.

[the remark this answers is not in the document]
ilija42: Started the proccess here https://github.com/smartcontractkit/chainlink-common/pull/956 🙃

pkg/loop/internal/types/types.go
pablolagreca: Ahh, don't worry. I thought it was an addition in this PR.

pablolagreca (approved):

PR #905, opened 2024-10-30
Add codec wrapper modifier

ilija42:


mateusz-sekara (dismissed):
I don't have enough knowledge about chain reader and its concepts to review it from the functional perspective. High level looks good to me, how would the config for the ccip case (`BigInt`) look like?

pablolagreca (changes requested):

ilija42:
> I don't have enough knowledge about chain reader and its concepts to review it from the functional perspective. High level looks good to me, how would the config for the ccip case (`BigInt`) look like?

The config for MsgsBetweenSeqNums would have to add this
```
&codec.WrapperModifierConfig{
											Fields: map[string]string{
												"FeeToken":       "Int",
												"FeeTokenAmount": "Int",
											},
										},
```

pablolagreca (approved):

mateusz-sekara (approved):

PR #909, opened 2024-10-31
[CAPPL-182] Add context to message emitter

cedric-cordenier:
* Add context to message emitter WASM import
* Speed up tests by making them use uncompressed binaries by default.

### Supports

- https://github.com/smartcontractkit/chainlink/pull/15059


pkg/workflows/wasm/host/module_test.go
```diff
@@ -48,14 +49,25 @@ func newMockMessageEmitter(e func(string, map[string]string) error) custmsg.Mess
...
+		err := store.add(
+			reqId,
+			&RequestData{ctx: func() context.Context { return tests.Context(t) }})
+		require.NoError(t, err)
 		emitFn := createEmitFn(
 			logger.Test(t),
+			store,
 			newMockMessageEmitter(func(_ string, _ map[string]string) error {
 				return nil
 			}),
```
MStreet3, line 66: I would modify the tests now to enable you to make an assertion on the context, add a key to the context and verify it can be read out in the message emitter

pkg/workflows/wasm/host/module_test.go
```diff
@@ -22,7 +23,7 @@ type mockMessageEmitter struct {
 	labels map[string]string
 }
 
-func (m *mockMessageEmitter) Emit(msg string) error {
+func (m *mockMessageEmitter) Emit(ctx context.Context, msg string) error {
```
MStreet3, line 26: change type on `e` to accept a context then in the unit tests you can make assertions on the context

MStreet3 (commented):
one comment on testing

pkg/workflows/wasm/host/module_test.go
MStreet3:
```suggestion
			store,
			newMockMessageEmitter(func(ctx context.Context, _ string, _ map[string]string) error {
			         v := context.Value(ctx)
			         require.Equal(t, v, "test-value)
				return nil
			}),
```

pkg/workflows/wasm/host/module_test.go
```diff
@@ -48,14 +49,25 @@ func newMockMessageEmitter(e func(string, map[string]string) error) custmsg.Mess
...
 func Test_createEmitFn(t *testing.T) {
 	t.Run("success", func(t *testing.T) {
+		store := &store{
+			m:  make(map[string]*RequestData),
+			mu: sync.RWMutex{},
+		}
+		reqId := "random-id"
+		err := store.add(
+			reqId,
+			&RequestData{ctx: func() context.Context { return tests.Context(t) }})
```
MStreet3, line 59:
```suggestion
			&RequestData{ctx: func() context.Context { 
			ctx :=  tests.Context(t)
			
			return context.WithValue(ctx, "key", "test-value")
			}})
```

MStreet3 (approved):

PR #910, opened 2024-11-01
Add pkg/pg with dialects.go & txdb.go

reductionista:
[NONEVM-739](https://smartcontract-it.atlassian.net/browse/NONEVM-739)

### Supports
https://github.com/smartcontractkit/chainlink/pull/15064
https://github.com/smartcontractkit/chainlink-solana/pull/921

[NONEVM-739]: https://smartcontract-it.atlassian.net/browse/NONEVM-739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

# Description

This  mostly just moves  `chainlink/core/internal/testutils/pgtest/txdb.go` and `chainlink/core/store/dialects/dialects.go`  into a new common package `chainlink-common/pkg/pg`, updating imports accordingly.
The purpose of this is so that it can be imported and used by `chainlink-solana` for unit testing Solana ORM''s (for example of actual use, see the linked `chainlink-solana` PR this supports)

Since `txdb.go` was a bit of a mess, it has been cleaned up a bit.

- The global `init()` function invoked on package load has been replaced with `RegisterTxDb()` which accepts a dbUrl as a param instead of  automatically reading it from the `CL_DATABASE` env var. Since there are only two places in the codebase which depend on our custom `txdb` dialect of sql being registered, this can simply be called in both of those places (inside `NewSqlxDb()` and `NewConnection()`) instead of relying on the global `init()`.
- Prevents calling `sql.Register()` more than once if `RegisterTxDb()` is called more than once (calling `sql.Register()` more than once with the same driver name will result in an error).
- Replaces passing of `context.Background` to queries with passing of a new context which gets cancelled when the db is closed. (Note: db here refers to the txdb, which is just a single connection in the lower level pg db, corresponding to the dsn passed in which is a randomly generated UUID.)
- Corrected spelling errors

This also moves some unit tests of txdb itself from `chainlink` into `chainlink-common`. Since they need an actual database, but don''t need any fixtures this is an ideal use case for an in-memory test db, something in between the empty interface added earlier and spinning up a full docker postgresql container with the schemas and tables all set up. `go-duckdb` has been used for this, which is now easy to add to our dependencies since we''re already at go 1.23. If the CL_CHAINLINK env var is set locally, txdb can be tested with a full ordinary postgresql database backing it... otherwise it falls back to testing it with the in-memory db backing it.

core ref: ec9e488802243585e0ae5827723f6136d6e7739b
solana ref: 6fd1891d0fbc79b0f10be353f6bd97446588209a


pkg/pg/pg.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+)
+
+func NewSqlxDB(t testing.TB, dbURL string) *sqlx.DB {
+	tests.SkipShortDB(t)
+	err := RegisterTxDb(dbURL)
+	if err != nil {
+		t.Errorf("failed to register txdb dialect: %s", err.Error())
```
dhaidashenko, line 19: should not this be a fatal error?
Otherwise caller has to check if returned db is nil.

pkg/pg/txdb_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+)
+
+func TestTxDBDriver(t *testing.T) {
+	dbURL, ok := os.LookupEnv("CL_DATABASE_URL")
+	require.True(t, ok, "CL_DATABASE_URL must be set")
+	db := NewSqlxDB(t, dbURL)
```
dhaidashenko, line 22: I've noticed weird behavior in my PR. If two tests use db and run sequentially, second fails with `conn closed` error.

reductionista: Ah, I think I know why that's happening. I changed txdb to use the test context for queries instead of `context.Background`, thinking that was an improvement over what's currently in `chainlink` repo.  But apparently that won't work--we'll have to put it back to `context.Background`, I guess there's just no context that would survive long enough for this.

pkg/pg/txdb.go
```diff
@@ -0,0 +1,531 @@
...
+	// must lock here to avoid racing with Open
+	d.Lock()
+	defer d.Unlock()
+
+	if d.conns[c.dsn] != c {
+		return nil // already been replaced
+	}
+	delete(d.conns, c.dsn)
+	if len(d.conns) == 0 && d.db != nil {
+		close(d.abort)
```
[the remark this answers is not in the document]
reductionista: Good point, but the underlying problem was a bit more than that. Sharing only a single abort channel across all connections doesn't work since different tests close their connections at different times. I changed it to have only 1 abort channel per connection now--sort of.  The only exception is if two connections have the same dsn string passed to them, but it looks like we always generate them randomly when calling `Open()` so hopefully there can't be a conflict.

pkg/utils/tests/tests.go
```diff
@@ -68,3 +68,19 @@ func RequireSignal(t *testing.T, ch <-chan struct{}, failMsg string) {
...
+
+// SkipShort skips tb during -short runs, and notes why.
+func SkipShort(tb testing.TB, why string) {
+	if testing.Short() {
+		tb.Skipf("skipping: %s", why)
+	}
+}
+
+// SkipShortDB skips tb during -short runs, and notes the DB dependency.
+func SkipShortDB(tb testing.TB) {
```
jmank88, line 80: I'm not sure if we should introduce this pattern here. It was added because of the core repo's default dependency on an external DB to run tests, which we do not want to encourage by default.

reductionista: Once the evm extraction is done most of the database tests we currently have will be in a separate evm repo that can't import `chainlink`.   At a minimum, `chainlink`, `chainlink-evm`, and `chainlink-solana` will all need to import this function from somewhere... but every relay will need it eventually, assuming we need to run CCIP on it.  In fact, chainlnk-cosmos also needs it--presently all of the unit tests for tx mgr in the cosmos repo are manually commented out which is not a good situation. IMO we should fix that so they call a function like this.

reductionista: This is the first I've heard that we might want to get rid of external db dependencies in tests.  We can get by in most tests by mocking the ORM, but I don't see how we could test the ORM's without having an external db with all of the right tables set up. Has anyone proposed a plan to move away from that?

jmank88: > At a minimum, chainlink, chainlink-evm, and chainlink-solana will all need to import this function from somewhere

EVM seems exceptional since it has historical baggage coming from core, but this is trivial to copy in that case. I don't see why we would introduce this problem to solana or anything else though.

I am not proposing that we do not test against external DBs. The point is just that a developer should be able to run `go test` without any extra baggage. If there are dependencies required, then typically you opt-in to those tests by e.g. including a build tag, setting a flag/env-var, etc.

pkg/pg/txdb_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"time"
+
+	_ "github.com/marcboeker/go-duckdb"
+
+	"github.com/google/uuid"
+	"github.com/jmoiron/sqlx"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"context"
```
jmank88, line 17: Does the linter normally fail to flag unsorted std lib? :confused: That is surprising

pkg/pg/txdb_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+)
+
+func TestTxDBDriver(t *testing.T) {
+	dbURL, ok := os.LookupEnv("CL_DATABASE_URL")
+	if !ok {
+		fmt.Sprintf("CL_DATABASE_URL not set--falling back to testing txdb backed by an in-memory db")
+		dbURL = string(InMemoryPostgres)
+	}
+	db := NewSqlxDB(t, dbURL)
```
jmank88, line 28: Would it make sense to package this up as a test helper?

pkg/pg/txdb_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+)
+
+func TestTxDBDriver(t *testing.T) {
+	dbURL, ok := os.LookupEnv("CL_DATABASE_URL")
+	if !ok {
+		fmt.Sprintf("CL_DATABASE_URL not set--falling back to testing txdb backed by an in-memory db")
```
jmank88, line 25: Did you mean `t.Log`?
```suggestion
		t.Log("CL_DATABASE_URL not set--falling back to testing txdb backed by an in-memory db")
```

pkg/utils/tests/tests.go
jmank88: Given that we have the in-memory version, I don't think we need this short-DB association. However, since the in-memory version might not work for every test, we could use a similar `SkipExternalDB` sort of mechanism to skip those tests when the memory DB is used.

pkg/pg/txdb_test.go
reductionista: My guess would be that it makes sure they are sorted within the groups you choose, but won't combine different groups if they are separated with a blank line. Lately it seems GoLand keeps inserting imports in weird places, I wonder if there's an adjustment I could make to my settings to get it to choose more wisely.

jmank88: I thought `goimports` would put std lib first, but GoLand certainly has a setting at least

pkg/pg/txdb_test.go
reductionista: We could, although I can't think of a good use case for this other than testing txdb itself (ie, this test). So making it a test helper, especially if it's in a place where it can be imported, might confuse someone.

Tests that depend on the db migration scripts being run (ie, require the standard tables and indexes set up) should call `NewSqlxDB()` to connect to an actual db and just fail if they can't.  Tests that don't depend on those scripts should call `NewInMemoryDataSource()`. Usually if testing with `NewInMemoryDataSource` is something that can work, then you wouldn't have any reason to attempt testing it with an actual db. In this case, it makes sense because there is some code in the `Open()` method of the txdb driver which parses the db url. This gets skipped if you're testing with an in-memory data source but everything else is tested.  So the full db tests everything but using the in-memory db in its place tests like 95%

jmank88: > Tests that depend on the db migration scripts being run

This is part of the pattern that we are trying to get away from though. It should not be the default.

>  Usually if testing with NewInMemoryDataSource is something that can work, then you wouldn't have any reason to attempt testing it with an actual db

Why do you say this? If we have the ability to trivially integration test against real postgres, then we should use it.

reductionista: > Why do you say this? If we have the ability to trivially integration test against real postgres, then we should use it.

I was mostly saying that because it seems unlikely anything could pass in the in-memory db but fail in a real db. (Unless we're testing the connection to the db itself, or that the db is set up properly... in which case I wouldn't expect an in-memory db to be useful.)  I suppose there are probably some ways, but in general I would assume that the set of sql commands supported by an in-memory db would be a small subset of those supported by a fully fledged database.

> This is part of the pattern that we are trying to get away from though. It should not be the default.

I'm not sure what the motivation for wanting to change existing patterns is, but if the reason is because we want to minimize use of the external db to conserve resources then making this a helper function and encouraging its use would be diametrically opposed to that. If we always try to hit the external db first, and only use the in-memory one as a backup if that fails then we'd waste a whole lot of resources in CI since 100% of them will go to the db. I'm not sure what the purpose of an in-memory db would even be then. In the short term, it could help us with local testing before we get the db isolation work Keith planned out done. But once each relay has its own migration scripts, it shouldn't be a problem.

reductionista: To clarify:

In my last paragraph I was assuming this meant that you think we should always call this helper function instead of using an in-memory db by itself:
>  If we have the ability to trivially integration test against real postgres, then we should use it.

So that's why I say "100% of them will go to the db". If instead by that, you just mean that we ought to leave the option open to do that for some tests, rather than an all-or-nothing approach then my objection above wouldn't apply.

pkg/utils/tests/tests.go
reductionista: > I am not proposing that we do not test against external DBs. The point is just that a developer should be able to run go test without any extra baggage.

Ok, I thought you were saying something different.  I think this might explain some confusion I had in the other thread I responded to here--I hadn't refreshed my browser in a while so hadn't read this.

I just always keep the same db set up locally, so for the most part have never had any issues with it.  The only problem I've run into occasionally is forgetting to run `make testdb` when I upgrade to a new version of chainlink. Usually it becomes obviously fairly quickly tho when tests start failing.

I'm not opposed to the long term goal of trying to allow devs to run tests without having a db set up. But I have a feeling if we get there it would have to be pretty far in the future. I'm not sure an in-memory db would ever be able to substitute for all of the types of tests we need. It's possible we could do that for some of the relays in the future, but trying to do that for Solana would not be possible within the time frame of when this needs to be released. I suppose we could keep it as-is for now and update it after we've got the db isolation stuff done, but I don't think we have much of an idea of what duckdb's capabilities are--that would require some research to see if it's even feasible.

pkg/pg/txdb_test.go
reductionista: Update: in light of having read your other thread, I understand now that the motivation is not conserving CI resources but making it easier for devs to run tests locally. At least for some tests, it may still not be worth using the extra CI resources just to get that slight bit of extra certainty that the function you're testing behaves as expected. But given your motivations, it does seem like you're recommending an all-or-nothing approach. Since unless all of the tests were able to run in-memory you wouldn't get rid of the dependency... right?

pkg/utils/tests/tests.go
reductionista: My main point:  unless and until we get to a point where we know that 100% of the tests can be run in-memory, there will still be a need for `SkipShortDB()`

reductionista: Alright, making it as flexible as possible sounds good to me.

I think that criterion suggests that we should add both `SkipShortDB()` and the helper function you recommended to `chainlink-common`.  Let each modules decide which one if either they need.  Not exposing either of them is being opinionated.

pkg/pg/txdb.go
```diff
@@ -0,0 +1,548 @@
...
+	"sync"
+	"testing"
+
+	"github.com/jmoiron/sqlx"
+	"go.uber.org/multierr"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils"
+)
+
+// txdb is a simplified version of https://github.com/DATA-DOG/go-txdb
```
pavel-raykov, line 20: Since you are introducing our own custom txdb here (which is a copy from chainlink), have you considered just using the DATA-DOG one (which is at least regularly updated)?

pkg/pg/txdb_test.go
jmank88: No, it does not imply that we need to use short at all:
https://github.com/smartcontractkit/chainlink-common/pull/910#discussion_r1859423319

pkg/utils/tests/tests.go
jmank88: My point is that this library should not be misleading anyone in to thinking that `-short` has anything at all to do with databases. That is a per project decision, and is used as a workaround for bad UX that we have been stuck with. This repo should not promote it.

pkg/pg/txdb.go
jmank88: cc @samsondav

samsondav (dismissed):

pkg/pg/txdb.go
samsondav: Its not a fork its a rewrite and its relatively small. Just put it here and import wherever you need it then its only needed once.

If it ain't broken don't fix it, and this ain't broke.

reductionista: We have so many tests in chainlink repo currently dependent on txdb, that I'm pretty sure migrating them to anything else would be a huge project. From what I understand, our txdb is based on DATA-DOG's txdb but customized toward chainlink's specific needs--but I haven't done a comparison of the two myself. (Although it's possible DATA-DOG supports more now than they did when ours was built?)

The context for the present PR is that we were looking for the fastest way to get db support including unit testing for Solana up and running so we can meet some very aggressive milestones toward getting CCIP up and running on Solana.  So we've stayed as closely as possible for now to the existing paradigm in order to avoid introducing unknown delay factors.  I have cleaned up our txdb however, and made some minor improvements such as better context handling.

pkg/pg/txdb_test.go
reductionista: Ah, I guess this is what you had in mind...
> If there are dependencies required, then typically you opt-in to those tests by e.g. including a build tag, setting a flag/env-var, etc.

So you're just recommending we move to using one of these methods instead of --short... that makes sense

pkg/pg/txdb.go
reductionista: > and we will now do the same in the chainlink-common

Just to be clear, this is a move not a copy... the PR this is linked to in chainlink repo changes all of the imports of it to point here and removes it from that repo

dhaidashenko (approved):

pkg/sqlutil/pg/pg.go
```diff
@@ -0,0 +1,35 @@
...
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/jmoiron/sqlx"
+	"github.com/scylladb/go-reflectx"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func NewSqlxDB(t testing.TB, dbURL string) *sqlx.DB {
```
jmank88, line 14: Should the name say test? Maybe instead of sqlx?
```suggestion
func NewTestDB(t testing.TB, dbURL string) *sqlx.DB {
```

pkg/sqlutil/pg/pg.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	db, err := sqlx.Open(string(TransactionWrappedPostgres), uuid.New().String())
+	require.NoError(t, err)
+	t.Cleanup(func() { assert.NoError(t, db.Close()) })
+	db.MapperFunc(reflectx.CamelToSnakeASCII)
+
+	return db
+}
+
+func DbUrlOrInMemory(t testing.TB) string {
```
jmank88, line 28: I think the go naming convention here is unfortunately caps. But maybe the package provides enough context to drop DB. And if we say Test, then the in memory part can be documented instead 🤷
```suggestion
func TestURL(t testing.TB) string {
```

pkg/sqlutil/pg/pg.go
reductionista: We can change it to whatever we want here... but I'd prefer not to change the name of the wrapper function in the core repo that calls this. Since that's called in unit tests all throughout the codebase, the CODEOWNERS requirement for the search-and-replace would be a nightmare

dhaidashenko (approved):

jmank88 (approved):

PR #913, opened 2024-11-04
feat(observability-lib): multiple alert rules can be attached to panel

Atrax1:
Multiple alert rules can be attached to a single panel

### DIFF
--- a/observability-lib/dashboards/core-node/component.go
+++ b/observability-lib/dashboards/core-node/component.go
@@ -2,6 +2,7 @@ package corenode
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/grafana/grafana-foundation-sdk/go/alerting"
 	"github.com/grafana/grafana-foundation-sdk/go/cog"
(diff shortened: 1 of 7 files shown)


jmank88 (approved):

observability-lib/dashboards/core-node/component.go
```diff
@@ -373,47 +418,10 @@ func headlines(p *Props) []*grafana.Panel {
...
-							Params: []float64{90},
-							Type:   grafana.TypeThresholdTypeLt,
-						},
-					},
-				},
-			},
+		AlertsOptions: []grafana.AlertOptions{
+			healthAverageAlertRule(p, 90, map[string]string{"severity": "info"}),
+			healthAverageAlertRule(p, 70, map[string]string{"severity": "warning"}),
+			healthAverageAlertRule(p, 50, map[string]string{"severity": "critical"}),
```
Atrax1, line 424: So here is that we generate 3 alerts that got different severity depending on the threshold (second param of the function is the threshold)

skudasov (approved):

observability-lib/grafana/dashboard.go
```diff
@@ -103,9 +104,11 @@ func (db *Dashboard) DeployToGrafana(options *DeployOptions) error {
 			alert.FolderUID = folder.UID
 			alert.Annotations["__dashboardUid__"] = *newDashboard.UID
 
-			panelId := panelIDByTitle(db.Dashboard, alert.Title)
+			panelId := panelIDByTitle(db.Dashboard, alert.Annotations["panel_title"])
+			// we can clean it up as it was only used to get the panelId
+			delete(alert.Annotations, "panel_title")
 			if panelId != "" {
-				alert.Annotations["__panelId__"] = panelIDByTitle(db.Dashboard, alert.Title)
+				alert.Annotations["__panelId__"] = panelId
```
patrickhuie19, line 111: why this `__` prefix + suffix combo?

observability-lib/dashboards/core-node/component.go
patrickhuie19: Very necessary, nice

patrickhuie19 (approved):

observability-lib/grafana/dashboard.go
Atrax1: its internal grafana API that requires this

PR #915, opened 2024-11-04
feat(observability-lib): builder to create independently grafana resources

Atrax1:
Builder refactor to allow creating contact points and notification policies without creating a dashboard


observability-lib/grafana/builder.go
```diff
@@ -29,16 +29,23 @@ type BuilderOptions struct {
...
-	}
+	builder := &Builder{}
 
-	builder := &Builder{
-		dashboardBuilder: dashboard.NewDashboardBuilder(options.Name).
-			Tags(options.Tags).
-			Refresh(options.Refresh).
-			Time(options.TimeFrom, options.TimeTo).
-			Timezone(options.TimeZone),
+	if options.Name != "" {
```
Atrax1, line 34: Only creating a dashboard builder if we provide a name

observability-lib/grafana/dashboard.go
```diff
@@ -17,20 +17,20 @@ const (
 	TypePlatformDocker     TypePlatform = "docker"
 )
 
-type Dashboard struct {
+type Observability struct {
 	Dashboard            *dashboard.Dashboard
 	Alerts               []alerting.Rule
 	ContactPoints        []alerting.ContactPoint
 	NotificationPolicies []alerting.NotificationPolicy
 }
```
Atrax1, line 25: renaming as the struct is holding all type of grafana resources not only dashboard anymore

observability-lib/api/notification-policy.go
```diff
@@ -8,11 +8,32 @@ import (
...
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i := range a {
+		foundMatch := false
+		for j := range b {
+			if reflect.DeepEqual(a[i], b[j]) {
+				foundMatch = true
+				break
```
patrickhuie19, line 21: this doesn't look like an exact match - could you leave a comment on the method explaining when a and b should be equal to each other?

observability-lib/grafana/builder.go
```diff
@@ -104,33 +111,39 @@ func (b *Builder) AddNotificationPolicy(notificationPolicies ...*alerting.Notifi
...
 
-	var alerts []alerting.Rule
-	for _, alertBuilder := range b.alertsBuilder {
-		alert, errBuildAlert := alertBuilder.Build()
-		if errBuildAlert != nil {
-			return nil, errBuildAlert
+	if b.dashboardBuilder != nil {
+		db, errBuildDashboard := b.dashboardBuilder.Build()
+		if errBuildDashboard != nil {
+			return nil, errBuildDashboard
```
patrickhuie19, line 120: /nit, but when passing errors up the stack, it can be helpful for debugging to wrap the error with a `fmt.Errorf` with a description of what the consumer observed with that error, i.e. 

```
return nil, fmt.Errorf("failure in building dashboard: %w, errBuildDashboard")
```

observability-lib/grafana/builder.go
```diff
@@ -104,33 +111,39 @@ func (b *Builder) AddNotificationPolicy(notificationPolicies ...*alerting.Notifi
...
+			}
 
-			alertBuildWithTags := alertBuilder.Labels(tags)
-			alertWithTags, errBuildAlertWithTags := alertBuildWithTags.Build()
-			if errBuildAlertWithTags != nil {
-				return nil, errBuildAlertWithTags
+			// Add common tags to alerts
+			if b.alertsTags != nil && len(b.alertsTags) > 0 {
+				tags := maps.Clone(b.alertsTags)
+				maps.Copy(tags, alert.Labels)
```
patrickhuie19, line 134: what is the difference between the tags and alert.Labels at this point?

observability-lib/api/notification-policy.go
Atrax1: associated tests here : https://github.com/smartcontractkit/chainlink-common/pull/915/files#diff-5b4adecb60cf2e79cc22806a969468909ea9f8bfa52bd464a0add735ef682cbeR10-R46 it should explain what is expected

observability-lib/grafana/builder.go
Atrax1: right I'll take care of this in future iterations

observability-lib/grafana/builder.go
Atrax1: Its the same I use tags as a name but grafana calls them Labels I could rename vars

patrickhuie19: non-blocking, but if you could clarify in a future iteration please

observability-lib/api/notification-policy.go
Atrax1: agree Im adding the comment

anukin (approved):

PR #916, opened 2024-11-05
Misc changes for working capabilities and workflows

kidambisrinivas:
## Updates

- Register contract reader service as LOOPP gRPC server
- Bug fixes to WASM workflow targets
- Support pointer type for SortByTimestamp

## Testing

```
$ go test -count=1 -v -run="^TestRunner_Run_GetWorkflowSpec" ./pkg/workflows/wasm
=== RUN   TestRunner_Run_GetWorkflowSpec
--- PASS: TestRunner_Run_GetWorkflowSpec (0.00s)
PASS
ok      github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm 0.324s
```


pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader.go
```diff
@@ -624,6 +624,8 @@ func convertLimitAndSortToProto(limitAndSort query.LimitAndSort) (*pb.LimitAndSo
 			tp = pb.SortType_SortTimestamp
 		case query.SortBySequence:
 			tp = pb.SortType_SortSequence
+		case *query.SortBySequence:
+			tp = pb.SortType_SortSequence
```
cedric-cordenier, line 628: 🤔 Do we need this? We don't support pointer based sortTypes for the other options in this list.

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader.go
```diff
@@ -902,5 +904,7 @@ func convertSequencesFromProto(pbSequences []*pb.Sequence, sequenceDataType any)
 }
 
 func RegisterContractReaderService(s *grpc.Server, contractReader types.ContractReader) {
-	pb.RegisterServiceServer(s, &goplugin.ServiceServer{Srv: contractReader})
+	service := goplugin.ServiceServer{Srv: contractReader}
+	pb.RegisterServiceServer(s, &service)
+	pb.RegisterContractReaderServer(s, NewServer(contractReader))
```
cedric-cordenier, line 909: Thanks for pushing this 👍

pkg/values/map.go
```diff
@@ -14,6 +15,40 @@ type Map struct {
 	Underlying map[string]Value
 }
 
+// TODO: this is temporary until the gateway can correctly unmarshal Web API trigger requests.
+func (m Map) MarshalJSON() ([]byte, error) {
```
cedric-cordenier, line 19: @kidambisrinivas This doesn't seem right to me -- what's making this necessary upstream?

Note you can marshal a map to json by unwrapping it, and then marshaling the resulting map[string]any. Can you do that wherever you need this instead?

pkg/workflows/wasm/pb/wasm.go
```diff
@@ -75,7 +75,7 @@ func WorkflowSpecToProto(spec *sdk.WorkflowSpec) (*WorkflowSpec, error) {
 		if err != nil {
 			return nil, fmt.Errorf("error translating step definition to proto: %w", err)
 		}
-		ws.Targets = append(ws.Consensus, tt)
```
cedric-cordenier, line 78: Whoops :) Thanks!

cedric-cordenier: Can you also extend the corresponding tests so that this is regression tested?

pkg/workflows/wasm/host/module.go
```diff
@@ -161,10 +161,12 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 	// Take the max of the default and the configured max memory mbs.
 	// We do this because Go requires a minimum of 16 megabytes to run,
 	// and local testing has shown that with less than 64 mbs, some
-	// binaries may error sporadically.
+	// binaries may error sporadically. At 64 MB max memory, we have
+	// some OOM issues with compute steps while using debugger.
 	modCfg.MaxMemoryMBs = int64(math.Max(float64(defaultMaxMemoryMBs), float64(modCfg.MaxMemoryMBs)))
 
 	cfg := wasmtime.NewConfig()
+	defer cfg.Close()
```
cedric-cordenier, line 169: I think you need to add this to the module struct: the config is used in the engine which outlives this funtion.

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader.go
kidambisrinivas: Without this, we were getting the error "Unknown sort by type", when testing out full workflows

pkg/workflows/sdk/compute_generated.go
```diff
@@ -76,6 +76,58 @@ func Compute1[I0 any, O any](w *WorkflowSpecFactory, ref string, input Compute1I
 	return &computeOutputCap[O]{(&Step[ComputeOutput[O]]{Definition: def}).AddTo(w)}
 }
 
+func Compute1WithMetadata[I0 any, O any](w *WorkflowSpecFactory, ref string, input Compute1Inputs[I0], compute func(Runtime, I0, capabilities.RequestMetadata) (O, error)) ComputeOutputCap[O] {
```
jinhoonbang, line 79: was this auto generated?

pkg/values/map.go
kidambisrinivas: @jinhoonbang  confirmed that local testing works with those two methods commented out. seems like all gateway requests work. Removing this

pkg/workflows/sdk/compute_generated.go
kidambisrinivas: Seems like it. @RyanRHall  Could you advice on whether this change is needed and if so, how to re-generate this code after recent changes in this PR?

pkg/workflows/wasm/host/module.go
cedric-cordenier: Can we rename it? The Close needs to be tied to the module Close, so we need to store it on the module.

pkg/workflows/sdk/compute_generated.go
cedric-cordenier: @nolag (not Ryan Hall)

pkg/workflows/wasm/host/module.go
kidambisrinivas: Fixed this, as per our discussion

pkg/workflows/sdk/compute_generated.go
kidambisrinivas: Thanks for the right tag :)

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader.go
kidambisrinivas: Yes, this is for the logeventtrigger capability, when run as a LOOPP gRPC process

pkg/workflows/sdk/compute_generated.go
nolag: This is generated code.  The generator lives in the folder gen in the same directory.  My understanding is that we're only using this to hack consensus and that it's not something we want to check into develop.  Once we have read consensus, we won't use this.  I vote against merging this change.

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader.go
kidambisrinivas: This was the only type required for log event trigger and adding this works for our use case. We thought, other types could be added as needed.

pkg/workflows/wasm/pb/wasm.go
kidambisrinivas: Extended tests as suggested

pkg/loop/internal/relayer/pluginprovider/contractreader/contract_reader.go
cedric-cordenier: Can we spend some time understanding why we are making this change, and then fix it properly? (i.e. what, if anything, has changed upstream that requires this? What is the right way to use this API -- is it to use values, pointers, both? And if so, can we fix it for all supported values?)

Just adding the value you need right now is just asking us to change this later when someone starts to use other values.

cedric-cordenier: I did a bit of digging on this:

The answer is that query.LimitAndSort.SortBy is an interface type -- and although all existing implementations are values (SortByTimestamp, SortByBlock ...), the interfaces is also satisfiable by the equivalent pointer (since Go does implicit receiver conversions).

So the right fix here is to accept both values and pointers for all items in the enum.

nolag: You should also put the two cases on the same line.  Ie 

case query.SortBySequence, *query.SortBySequence:

pkg/workflows/sdk/compute_generated.go
kidambisrinivas: As discussed offline, I've added this function to generator and added a comment specifying that this function is temporary and not intended to be used outside testing. This function would be removed once more tooling is available for testing.

jinhoonbang (approved):

nolag (approved):

kidambisrinivas:
Not able to merge this PR because of unverified commits. Created a [new PR](https://github.com/smartcontractkit/chainlink-common/pull/921) due to this

PR #917, opened 2024-11-05
Extend `FetchRequest` fields for beholder support

vyzaldysanchez:


patrickhuie19 (dismissed):

pkg/workflows/wasm/sdk.go
```diff
@@ -171,12 +171,16 @@ func createFetchFn(
 		}
 
 		b, err := proto.Marshal(&wasmpb.FetchRequest{
-			Id:        *sdkConfig.RequestID,
-			Url:       req.URL,
-			Method:    req.Method,
-			Headers:   values.ProtoMap(headerspb),
-			Body:      req.Body,
-			TimeoutMs: req.TimeoutMs,
+			Id:                  *sdkConfig.RequestID,
```
patrickhuie19, line 174: for future cleanup, but linters in core assert `ID` instead of `Id`

patrickhuie19 (approved):

pkg/workflows/wasm/pb/wasm.proto
```diff
@@ -66,6 +66,10 @@ message FetchRequest {
   bytes body = 4;
   uint32 timeoutMs = 5;
   string id = 6;
+  string workflowId = 7;
```
cedric-cordenier, line 69: @vyzaldysanchez Could we add a comment here to explain why we're adding this here? (It's because it's needed for the ID of the request + for metrics

vyzaldysanchez: On it.

cedric-cordenier: Actually it's probably worth pulling this out into a metadata sub-message to not pollute the overall message

patrickhuie19 (approved):

pkg/workflows/wasm/pb/wasm.proto
```diff
@@ -59,13 +59,22 @@ message Response {
   }
 }
 
+// NOTE: This message was added because it is needed to be used as part of the request and for metrics.
```
bolekk, line 62: This comment is not very useful, remove it?

bolekk (approved):

PR #919, opened 2024-11-06
generate a mock DataSource for use in unit testing

reductionista:
This seems like a good thing to have in general, I just happened to find it useful for the two linked PR's below, where you can see an example of it being used. In this case, an empty interface would have been good enough, but having it backed by an actual in-memory db allows us to do a few more things with it and should make it behave a bit more like a realistic mock. What won't work for now is setting up the tables and fixtures chainlink components usually expect, so it would only be good for cases where not much is done with the data source. We'll still need txdb for most things which is moved into chainlink-common in another set of linked PR's.

### Supports
- https://github.com/smartcontractkit/chainlink/pull/15127
- https://github.com/smartcontractkit/chainlink-solana/pull/916

[NONEVM-876]: https://smartcontract-it.atlassian.net/browse/NONEVM-876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

solana ref: dce6483e2bd7bde0eb564c5e3ed703186c232f90
core ref: 3ad3e7514fc86a3a3935162e3cba67a410858dde


jmank88:
We have been trying to avoid sharing mocks between modules because it splinters the version of mockery being used in the main module.

jmank88:
On the other hand, if we wanted to provide some kind of "fake" in-memory implementation, then that could live here in a test package.

reductionista:
> We have been trying to avoid sharing mocks between modules because it splinters the version of mockery being used in the main module.

That seems like a big self-imposed limitation to have. Is there no way to keep the mockery versions in sync between these repos?

reductionista:
> On the other hand, if we wanted to provide some kind of "fake" in-memory implementation, then that could live here in a test package.

That's a good idea.  For now, the only thing I need is anything that satisfies the interface.  So I could just commit it with unimplemented methods that return nil. Although having an actual in-memory version of a data source would certainly be very useful for what's coming up... so I'll take a look and see if that's something I can easily add.

jmank88:
> > We have been trying to avoid sharing mocks between modules because it splinters the version of mockery being used in the main module.
> 
> That seems like a big self-imposed limitation to have. Is there no way to keep the mockery versions in sync between these repos?

There is no limitation. Anybody that wants to use a mock is free to generate one for themselves.

reductionista:
> There is no limitation. Anybody that wants to use a mock is free to generate one for themselves.

Not sure what you mean by this. When the structure you need to mock is in `chainlink-common`, you're not supposed to generate the mock in chainlink-common. So, are you saying it should be generated separately by each thing importing it? Would the generation be automatic, or just done manually and copied over?

jmank88:
> > There is no limitation. Anybody that wants to use a mock is free to generate one for themselves.
> 
> Not sure what you mean by this. When the structure you need to mock is in `chainlink-common`, you're not supposed to generate the mock in chainlink-common. So, are you saying it should be generated separately by each thing importing it? Would the generation be automatic, or just done manually and copied over?

https://github.com/smartcontractkit/chainlink/blob/develop/.mockery.yaml#L343-L350

.mockery.yaml
```diff
@@ -35,3 +35,7 @@ packages:
   github.com/smartcontractkit/chainlink-common/pkg/beholder/internal:
     interfaces:
       OTLPExporter:
+
+  github.com/smartcontractkit/chainlink-common/pkg/sqlutil:
+    interfaces:
+      DataSource:
```
jmank88, line 41: Can remove this now, right?
```suggestion
```

pkg/sqlutil/test_data_source.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/jmoiron/sqlx"
+	_ "github.com/proullon/ramsql/driver"
+	"github.com/stretchr/testify/require"
+)
+
+var _ DataSource = &TestDataSource{}
+
+// TestDataSource is a simple in-memory DataSource type which can be used for some types of unit testing.
+type TestDataSource struct {
+	db *sqlx.DB
```
jmank88, line 17: Is this wrapper type necessary? `*sqlx.DB` already implements `DataSource` directly.

pkg/sqlutil/test_data_source.go
```diff
@@ -0,0 +1,85 @@
...
+	return ds.db.Rebind(_a0)
+}
+
+// SelectContext provides a mock function with given fields: ctx, dest, query, args
+func (ds *TestDataSource) SelectContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error {
+	return ds.SelectContext(ctx, dest, query, args...)
+}
+
+// NewTestDataSource creates a new instance of DataSource.
+func NewTestDataSource(t *testing.T) *TestDataSource {
```
jmank88, line 80: I think `Test` is too ambiguous. How about something like:
```suggestion
func NewInMemoryDataSource(t *testing.T) DataSource {
```

pkg/sqlutil/test_data_source.go
```diff
@@ -0,0 +1,85 @@
...
+	db, err := sqlx.Open("ramsql", "testdb")
```
jmank88, line 81:
```suggestion
	db, err := sqlx.Open("ramsql", t.Name())
```

pkg/sqlutil/test_data_source.go
```diff
@@ -0,0 +1,85 @@
+package sqlutil
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/jmoiron/sqlx"
+	_ "github.com/proullon/ramsql/driver"
```
jmank88, line 9: @pavel-raykov did you consider this library?

pkg/sqlutil/test_data_source.go
reductionista: Not at present, although I was thinking we'll probably want to extend it to do other things in the future. For example, a test could add a hook that intercepts each request and gives a custom response... or modify the response in some way, or just assert some condition.

reductionista: I guess I could leave the constructor named the same but just return a `sqlx.DB` for now... that probably makes more sense, since we could still extend it later without modifying anything that's calling the constructor.

jmank88: I think we could leverage our existing sqlutil hook code later if we wanted to.

pkg/sqlutil/test_data_source.go
pavel-raykov: Ramsql does not claim to be compatible with postgres. For example it cannot create the following table:

"CREATE TABLE test (tstamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP)"

with the error 	"Syntax error near default CURRENT_TIMESTAMP" (this for example immediately renders it incompatible with goose that uses CURRENT_TIMESTAMP in the goose_versions table in the db to keep track of the migrations).

If we want a tiny lightweight in-memory db which can closely mock postgres I would propose to use duckdb. (One more advantage of it is that you can start it by typing duckdb in console as opposed to ramsql that does not have a standalone version.)

pkg/sqlutil/test_data_source.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+
+	"github.com/jmoiron/sqlx"
+	_ "github.com/proullon/ramsql/driver"
+	"github.com/stretchr/testify/require"
+)
+
+// NewInMemoryDataSource returns a new in-memory DataSource
+func NewInMemoryDataSource(t *testing.T) *sqlx.DB {
```
jmank88, line 12: Returning `DataSource` would allow us slip in a hook later without affecting the callers:
```suggestion
func NewInMemoryDataSource(t *testing.T) DataSource {
```

pkg/sqlutil/test_data_source.go
reductionista: ramsql does come with a standalone version actually:
```
[MB-TQH59VKW66:~/workspace/chainlink-common> ramsql
ramsql> CREATE TABLE foo (a INTEGER);
Query OK. 1 rows affected
ramsql>
```
I'll take a look at duckdb, if that's more compatible with postgres thanks

pavel-raykov: And also, both ramSQL and duckdb are incompatible with goose, so this is one more reason not use them. (Unfortunately, I don't know of any other in-memory db that can be used  [cockroach is better, but it also cannot execute all our migrations =( ])

go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/smartcontractkit/chainlink-common
 
-go 1.22.0
+go 1.23
 
-toolchain go1.22.7
```
reductionista, line 5: This and the other changes to go.mod are all from `go get github.com/marcboeker/go-duckdb`. It requires go 1.23

pkg/sqlutil/mocks/data_source.go
jmank88, line 1: Why don't chainlink and chainlink-solana generate their own mock?

jmank88 (changes requested):

pkg/sqlutil/mocks/data_source.go
reductionista: Isn't it better to have it in one place rather than two? Maybe I misunderstood what you were saying earlier about this.

jmank88: What are we trying to achieve? If they want to use a real generated mock from mockery, then they should generate it themselves via their local mockery config in order to align with the rest of their code. What is this file supposed to represent? It looks like a hybrid stub/mock.

reductionista: For now I just need something they can each pass to satisfy the interface. Generating it once in `chainlink-common` seemed like the simplest and cleanest approach, but you mentioned we are trying to avoid doing that.

If I understood correctly, the reason we want to avoid that pattern is because there might be slight incompatibilities between different versions of mockery/testify between different repos in some cases. Since most of the specific boilerplate it generates is unnecessary for the  current use case anyway, I figured replacing it with a simple stub that doesn't have any mockery version dependency might be better.

I think you're right though--I see a few good reasons why generating them in their own repos will make life easier:

- Only 2 linked PR's for now instead of 3
- No need to update chainlink-common later if we need to extend the functionality
- Less likely to need to extend it since it will support all of the standard mocking methods.

reductionista: The only thing I'm a bit unclear about is: where should mockery generate these mocks? Since the data structure doesn't live in either of those repos, there is no natural place to put it.

jmank88: I'm still not clear on what we are trying to achieve. Do we need a mock or not? If not, then how about:
```go
package sqltest

func NewNoOpDataSource() sqlutil.DataSource {
	return &noOpDataSource{}
}

type NoOpDataSource struct {}
```

reductionista: @jmank88 I didn't see your last comment till now, I guess we must have both left ours at the same time.

```
package sqltest

func NewNoOpDataSource() sqlutil.DataSource {
	return &noOpDataSource{}
}

type noOpDataSource struct {}
```
Yeah, that would work for the present use case--assuming we include all of the empty implementations of each of the methods on `noOpDataSource` after it.

It does seem likely we'd want to replace this with either actual mocks at some point or with the in-memory db, but I guess we can make that decision later when it's more clear what's needed.

jmank88: > It does seem likely we'd want to replace this with either actual mocks at some point or with the in-memory db, but I guess we can make that decision later when it's more clear what's needed.

We can add mocks or an in memory version later, but why would we replace anything? Which to use is a case by case decision for callers to make, and any test that can pass using a no-op implementation should prefer to use one instead of wasting overhead on something more complex.

pkg/sqltest/data_source.go
jmank88, line 1: How about `pkg/sqlutil/sqltest/data_source.go`? Like https://pkg.go.dev/net/http/httptest

pkg/sqlutil/sqltest/data_source.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/sqlutil"
+)
+
+// NewNoOpDataSource returns an empty DataSource type which will satisfy the interface
+func NewNoOpDataSource() sqlutil.DataSource {
+	return &noOpDataSource{}
+}
+
+type noOpDataSource struct{}
```
pavel-raykov, line 17: Sorry, is it clear why you cannot use dataSource from pkg/sqlutil/hook_test.go instead of noOpDataSource ?

pavel-raykov: From what I understand noOpDataSource is intended to be used in some tests. Would dataSource from pkg/sqlutil/hook_test.go just work as it is now? (Or alternatively, one could also call this HookableDataSource that can be initialized with hookBehaviour = false / true)

jmank88: > Would dataSource from pkg/sqlutil/hook_test.go just work as it is now?

I don't think that an exported "no op" type should be arbitrarily sleeping for any amount of time.

> (Or alternatively, one could also call this HookableDataSource that can be initialized with hookBehaviour = false / true [this is just another way to implement the desire behavior without embedding])

Can you elaborate?

pavel-raykov: > > Would dataSource from pkg/sqlutil/hook_test.go just work as it is now?
> 
> I don't think that an exported "no op" type should be arbitrarily sleeping for any amount of time.

I agree, but it would sleep only if GetContext/SelectContext is called - I am not sure how often this would happen.

> 
> > (Or alternatively, one could also call this HookableDataSource that can be initialized with hookBehaviour = false / true [this is just another way to implement the desire behavior without embedding])
> 
> Can you elaborate?

Instead of embedding the common functions and having two structs HookedDataSource and NoOpDataSource, one could have one HookableDataSource that would behave like HookedDataSource if initialized with true and like NoOpDataSource if initialized with false.

jmank88: > Instead of embedding the common functions and having two structs HookedDataSource and NoOpDataSource, one could have one HookableDataSource that would behave like HookedDataSource if initialized with true and like NoOpDataSource if initialized with false.

If there is a way to expose a _general_ type that can be _customized_ to do exactly what the current hook type is doing, then that would be great. But I don't think it would make sense to put this specialized behavior behind a flag.

pavel-raykov: Sorry, do you mean that something like

```
type dataSource struct {
	hookable bool
}

func (q *dataSource) SelectContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error {
	if q.hookable {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(selDur):
		}
	}
	return nil
}
```

is not good?

jmank88: I think that behavior is arbitrary in the context of a general exported type. It is just what one particular set of tests needed at the time, and there is no reason to assume that any other users would want the same behavior (in principle or with those particular const durations).

pavel-raykov: I agree, I am just trying to find a way to avoid code duplication. I would also be fine with the embed solution.

pavel-raykov (approved):

PR #925, opened 2024-11-08
[KS-507] Make Streams Trigger ID (name+version) configurable

bolekk:
Needed for upgrading trigger version

### DIFF
--- a/pkg/capabilities/triggers/mercury_trigger.go
+++ b/pkg/capabilities/triggers/mercury_trigger.go
@@ -16,18 +16,13 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
-const triggerID = "streams-trigger@1.0.0"
-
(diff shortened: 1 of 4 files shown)


pkg/capabilities/triggers/mercury_trigger.go
```diff
@@ -54,10 +49,22 @@ type subscriber struct {
...
 		tickerResolutionMs = defaultTickerResolutionMs
 	}
+	if capName == "" {
+		capName = defaultCapabilityName
+	}
+	if capVersion == "" {
+		capVersion = defaultCapabilityVersion
+	}
+
+	var capInfo = capabilities.MustNewCapabilityInfo(
```
cedric-cordenier, line 63: Using Must is fine as a global, but we should probably avoid it here, since this could panic.

cedric-cordenier (approved):

PR #927, opened 2024-11-09
Enable batching for beholder emitter in LOOP Server

pkcll:
Enable batching for beholder emitter in LOOP Server

Companion PR for https://github.com/smartcontractkit/chainlink/pull/15173

### DIFF
--- a/pkg/loop/config.go
+++ b/pkg/loop/config.go
@@ -46,16 +46,16 @@ type EnvConfig struct {
 	TracingTLSCertPath     string
 	TracingAttributes      map[string]string
 
-	TelemetryEnabled            bool
-	TelemetryEndpoint           string
-	TelemetryInsecureConnection bool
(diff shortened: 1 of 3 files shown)


pkg/loop/config.go
```diff
@@ -51,6 +54,8 @@ type EnvConfig struct {
 	TelemetryTraceSampleRatio   float64
 	TelemetryAuthHeaders        map[string]string
 	TelemetryAuthPubKeyHex      string
+	EmitterBatchProcessor       bool
+	EmitterExportTimeout        time.Duration
```
patrickhuie19, line 58: should these fields be consumed by the beholder client?

jmank88 (approved):

patrickhuie19 (approved):

pkg/loop/config.go
pkcll: yes

PR #934, opened 2024-11-12
[CAPPL-270/271] Fix Consensus bugs

cedric-cordenier:
- Fix "result is not a pointer error" in the reduce aggregator
- Continue rather than error if we encounter an aggregation error


pkg/capabilities/consensus/ocr3/aggregators/reduce_test.go
```diff
@@ -517,22 +582,6 @@ func TestReduceAggregator_Aggregate(t *testing.T) {
 					return map[commontypes.OracleID][]values.Value{}
 				},
 			},
-			{
```
cedric-cordenier, line 520: I removed this because this isn't actually a problem in the current implementation and doesn't error

ettec (approved):

agparadiso (approved):

bolekk (approved):

PR #939, opened 2024-11-13
[CAPPL-276] update current state only if should report is true

agparadiso:
### Description

This pr introduces a fix in which we want to update the state only if the shouldReport is true.
We have defined that it should be true when:
- the deviation type is `DEVIATION_TYPE_NONE` despite a change or not on the value.
- the current deviation is greater than the defined deviation.

This also introduces another fix in which we are not initialising the state with zeroValue, this was making the evaluation that defines if we shouldReport incorrect given it was checking for `nil` but it was never nil. 

[CAPPL-276](https://smartcontract-it.atlassian.net/browse/CAPPL-276)

### Supports

[CAPPL-276]: https://smartcontract-it.atlassian.net/browse/CAPPL-276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
@@ -117,7 +117,9 @@ func (a *reduceAggregator) Aggregate(lggr logger.Logger, previousOutcome *types.
 			lggr.Debugw("checked deviation", "key", field.OutputKey, "deviationType", field.DeviationType, "currentDeviation", currDeviation.String(), "targetDeviation", field.Deviation.String(), "shouldReport", shouldReport)
 		}
 
-		(*currentState)[field.OutputKey] = singleValue
+		if shouldReport {
+			(*currentState)[field.OutputKey] = singleValue
+		}
 		if len(field.OutputKey) > 0 {
 			report[field.OutputKey] = singleValue
 		} else {
--- a/pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
@@ -25,12 +25,17 @@ import (
 const (
 	AGGREGATION_METHOD_MEDIAN = "median"
 	AGGREGATION_METHOD_MODE   = "mode"
-	DEVIATION_TYPE_NONE       = "none"
-	DEVIATION_TYPE_PERCENT    = "percent"
-	DEVIATION_TYPE_ABSOLUTE   = "absolute"
-	REPORT_FORMAT_MAP         = "map"
-	REPORT_FORMAT_ARRAY       = "array"
-	REPORT_FORMAT_VALUE       = "value"
+	// DEVIATION_TYPE_NONE is no deviation check
+	DEVIATION_TYPE_NONE = "none"
+	// DEVIATION_TYPE_ANY is any difference from the previous value to the next value
+	DEVIATION_TYPE_ANY = "any"
+	// DEVIATION_TYPE_PERCENT is a numeric percentage difference
+	DEVIATION_TYPE_PERCENT = "percent"
+	// DEVIATION_TYPE_ABSOLUTE is a numeric absolute difference
+	DEVIATION_TYPE_ABSOLUTE = "absolute"
+	REPORT_FORMAT_MAP       = "map"
+	REPORT_FORMAT_ARRAY     = "array"
+	REPORT_FORMAT_VALUE     = "value"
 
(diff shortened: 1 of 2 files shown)


pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -185,6 +180,29 @@ func (a *reduceAggregator) Aggregate(lggr logger.Logger, previousOutcome *types.
 	}, nil
 }
 
+func (a *reduceAggregator) shouldReport(lggr logger.Logger, field AggregationField, singleValue values.Value, currentState *map[string]values.Value) (bool, error) {
+	if field.DeviationType == DEVIATION_TYPE_NONE {
+		return true, nil
+	}
```
cedric-cordenier, line 186: @agparadiso This isn't correct: this means that if we have a struct where a single field doesn't have a deviation defined on it, we'll always report.

That means we'll always report for our existing struct, which looks like:
- Price <- 5 % deviation
- Timestamp <- 300 s deviation
- FeedID <- no deviation

If the deviation type is none, we should instead look at the previous entry and only report if it's different.

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -185,6 +180,29 @@ func (a *reduceAggregator) Aggregate(lggr logger.Logger, previousOutcome *types.
...
+
+	oldValue := (*currentState)[field.OutputKey]
+	if oldValue == nil {
```
cedric-cordenier, line 189: Can we add a comment here to say that a nil value means it hasn't been initialized and this is the first pass?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -201,14 +219,6 @@ func (a *reduceAggregator) initializeCurrentState(lggr logger.Logger, previousOu
 		currentState = mv.Underlying
 	}
 
-	zeroValue := values.NewDecimal(decimal.Zero)
```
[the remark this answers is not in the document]
cedric-cordenier: Discussed with Justin offline: it looks like we do need to handle this somewhere else if we're not setting it to zero -- in which case, please make sure to cover this in a test before we merge.

The fact that you could change this without a test failing means we're missing some test coverage.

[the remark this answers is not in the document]
agparadiso: I did add a check [here](https://github.com/smartcontractkit/chainlink-common/pull/939/files#diff-88ee9bf84f3810ebff31bda3dc7af497a9dfd3bee5b904fb123c0834aafde3fdR189), if this is not validated there are tests complaining about a nil value

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
agparadiso: addressed on the last commit 👍🏼

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -185,6 +182,53 @@ func (a *reduceAggregator) Aggregate(lggr logger.Logger, previousOutcome *types.
...
+		switch v := unwrappedOldValue.(type) {
+		case []byte:
+			if !bytes.Equal(v, unwrappedSingleValue.([]byte)) {
+				return true, nil
+			}
+		default:
+			if unwrappedOldValue != unwrappedSingleValue {
+				return true, nil
+			}
+		}
```
agparadiso, line 214: 🤔 for `DEVIATION_TYPE_NONE` Im considering field types as `[]byte` or `bool` are there any other type I need to check?

pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
justinkaseman: > If the deviation type is none, we should instead look at the previous entry and only report if it's different.

@cedric-cordenier we added a new `deviationType` to cover this behavior, called "any". Basically if the value ever changes, send a report.

This is different from "none", which is never send a report if this value changes. (the default behavior)

cedric-cordenier: Sounds good!

cedric-cordenier (approved):

ettec (approved):

PR #943, opened 2024-11-14
fix generate diff check

jmank88:
Our `git diff` checks (used to verify tidiness and that generated files are up to date) fail in cases where there are completely new files. This is both due to the files being unstaged as well as the default behavior of `git diff` to be to ignore them. We can call `git add -all` first, and include the `--cached` flag to get the behavior we want.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -63,10 +63,12 @@ jobs:
       - name: Ensure "make gomodtidy" has been run
         run: |
           make gomodtidy
-          git diff --minimal --exit-code
+          git add --all
+          git diff --minimal --cached --exit-code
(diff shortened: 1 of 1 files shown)


jmank88:
This is meant to be failing initially in order to demonstrate the fix, but it is still passing :facepalm: :confused:

.github/workflows/pkg.yml
```diff
@@ -69,4 +69,5 @@ jobs:
           make rm-mocked
           make rm-builders
           make generate
+          git add --all
```
jmank88, line 72: This works locally, but not in CI :thinking:

george-dorin (approved):

ilija42 (approved):

PR #948, opened 2024-11-21
Pass AuthHeaders from beholder config to to loop/Tracing

pkcll:
Pass AuthHeaders from beholder config to to loop/Tracing

This is needed to enable CSA based authentication for loop/Tracing


patrickhuie19 (approved):

cll-gg:
@patrickhuie19 @jmank88: thanks for the approval. Could you merge it as well? We don't seem to have the permissions.

PR #949, opened 2024-11-21
(fix): Handle nested types as output of Compute capability

justinkaseman:
### Description
The Compute capability cannot return deeply nested types, such as `time.Time`.

This happens at a point where the output is put into a struct and then converted to a map in `CreateMapFromStruct`.
The issue is that this function uses the `mapstructure` package does not handle nested types.

Inputting `{2024-11-21 01:25:58.565404719 +0000 UTC}`
turns into `map[Value:map[]]`
instead of what we expect, which is `map[Value:2024-11-21 01:25:58.565404719 +0000 UTC]`.

### Solution
Use reflection to turn values into interfaces


pkg/values/value.go
```diff
@@ -303,12 +302,18 @@ func fromBigIntValueProto(biv *pb.BigInt) *BigInt {
...
-	err := mapstructure.Decode(v, &resultMap)
-	if err != nil {
-		return nil, err
+	resultMap := map[string]any{}
+
+	// use reflect to handle nested types as an interface
+	rv := reflect.ValueOf(v)
+	rt := reflect.TypeOf(v)
+	for i := 0; i < rv.NumField(); i++ {
+		if !rt.Field(i).IsExported() {
```
cedric-cordenier, line 311: @justinkaseman We can just skip unexported fields, i.e. treat them as private. This is pretty standard eg. across json.Marshal/Unmarshal, proto.Marshal/Unmarshal etc.

MStreet3 (approved):

cedric-cordenier (approved):

ettec (approved):

ettec (approved):

PR #953, opened 2024-11-25
[CAPPL-205] Remove OCR3 capability from registry when closed

justinkaseman:
### Caveats
Requires a companion `chainlink` repository change


pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -59,6 +62,11 @@ func (o *factory) Start(ctx context.Context) error {
 
 func (o *factory) Close() error {
 	return o.StopOnce("OCR3ReportingPlugin", func() error {
+		err := (*o.capabilityRegistry).Remove(o.capability)
```
bolekk, line 65: The factory object is not adding it to the Registry, is it? I think we should handle removal in the same object that Add()-ed it in the first place (where is that?)

pkg/types/core/capabilities_registry.go
```diff
@@ -17,4 +17,5 @@ type CapabilitiesRegistry interface {
 	GetTarget(ctx context.Context, ID string) (capabilities.TargetCapability, error)
 	List(ctx context.Context) ([]capabilities.BaseCapability, error)
 	Add(ctx context.Context, c capabilities.BaseCapability) error
+	Remove(c capabilities.BaseCapability) error
```
bolekk, line 20: I wonder how we make sure that malicious/buggy capabilities don't add/remove entries for others. Maybe we don't have to? Can we say it's node's responsibility to run legitimate capabilities...?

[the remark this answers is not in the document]
justinkaseman: The `Close` function of a service doesn't have any context. Should I have context here and then when removing from the registry just pass empty context?

pkg/capabilities/consensus/ocr3/factory.go
justinkaseman: No, it's [here](https://github.com/smartcontractkit/chainlink-common/blob/CAPPL-205/pkg/capabilities/consensus/ocr3/ocr3.go#L99). I thought about that too, agreed that it makes sense to move.

pkg/types/core/capabilities_registry.go
justinkaseman: > it's node's responsibility to run legitimate capabilities
I agree with this with regards to 3rd party capabilities. I imagine there would be:
- a code signature of the verified source code. Maybe this is something we add to the Capability Registry?
- more testing rails on what is allowed, e.g. ensure there are no conflicting Capability `Info`s and run LOOPP initialize function on a mock CapRegistry
- some official stamp of "CLL reviewed"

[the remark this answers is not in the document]
justinkaseman: Here is what that will look like in each capability https://github.com/smartcontractkit/chainlink-common/pull/953#discussion_r1859139581

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -264,6 +264,37 @@ func (cr *capabilitiesRegistryClient) Add(ctx context.Context, c capabilities.Ba
...
+	}
+
+	// Check the capability and the CapabilityType match here as the ServeNew method does not return an error
+	err = validateCapability(c, info.CapabilityType)
+	if err != nil {
+		return err
+	}
+
+	var cRes net.Resource
+	id, cRes, err := cr.ServeNew(info.ID, func(s *grpc.Server) {
```
cedric-cordenier, line 280: @justinkaseman Hmmmm... this is confusing -- why do we need to set up a server here?

I would have expected that calling info followed by remove would be enough.

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -264,6 +264,37 @@ func (cr *capabilitiesRegistryClient) Add(ctx context.Context, c capabilities.Ba
...
+		pbRegisterCapability(s, cr.BrokerExt, c, info.CapabilityType)
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = cr.grpc.Remove(ctx, &pb.RemoveRequest{
+		CapabilityID: id,
+		Type:         pb.ExecuteAPIType(getExecuteAPIType(info.CapabilityType)),
```
cedric-cordenier, line 289: I also don't think you need this: Remove should just accept that Base capability interface (the one that exposes info) since all it needs to Remove something is the id.

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -526,6 +557,29 @@ func (c *capabilitiesRegistryServer) Add(ctx context.Context, request *pb.AddReq
 	return &emptypb.Empty{}, nil
 }
 
+func (c *capabilitiesRegistryServer) Remove(ctx context.Context, request *pb.RemoveRequest) (*emptypb.Empty, error) {
+	conn, err := c.Dial(request.CapabilityID)
```
cedric-cordenier, line 561: Ah I understand now: I think this is a case where I wouldn't be afraid to just have Remove take an ID rather than the actual capability

I think that will simplify a lot, in the client you can just call Info() then Remove(info.ID) and propagate that through here -- no new clients, connections, etc. needed, since the call just operates on data.

cedric-cordenier (approved):

pkg/types/core/capabilities_registry.go
cedric-cordenier: We should be able to solve this for standard capabilities by moving the removal to the core node; for the OCR3 capability that's a bit more difficult though and we'll need to do it via the loopp boundary

PR #957, opened 2024-11-28
INFOPLAT-1539  Beholder Log Batch Processor More Settings

gheorghestrimtu:
we need to make the following settings configurable:

Batching Processor Parameters ([link](https://opentelemetry.io/docs/specs/otel/logs/sdk/#batching-processor)): Configurable parameters like maxQueueSize, scheduledDelayMillis, and maxExportBatchSize should be exposed to adjust the processing behavior based on load.

Configurable parameters:

exporter: the exporter where the LogRecords are pushed.

maxQueueSize: the maximum queue size. After the size is reached, logs are dropped. The default value is 2048.

scheduledDelayMillis: the delay interval in milliseconds between two consecutive exports. The default value is 1000.

exportTimeoutMillis: how long the export can run before it is cancelled. The default value is 30000.

maxExportBatchSize: the maximum batch size of every export. It must be smaller or equal to maxQueueSize. The default value is 512.

 

### Supports

[INFOPLAT-1539]: https://smartcontract-it.atlassian.net/browse/INFOPLAT-1539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cll-gg (dismissed):

hendoxc (dismissed):

pkcll (dismissed):

pkg/beholder/config.go
```diff
@@ -16,7 +16,10 @@ type Config struct {
...
 	ResourceAttributes []otelattr.KeyValue
 	// Message Emitter
-	EmitterExportTimeout time.Duration
+	EmitterExportTimeout      time.Duration
+	EmitterExportInterval     time.Duration
+	EmitterExportMaxBatchSize int
+	EmitterMaxQueueSize       int
 	// Batch processing is enabled by default
 	// Disable it only for testing
 	EmitterBatchProcessor bool
```
jmank88, line 25: This block of fields has grown quite large. I think we should either insert an empty line between each of Otel/Message/Trace/Metric/Log, or convert these field comments to be trailing so the sections are not disrupted:
```suggestion
	EmitterBatchProcessor bool // Enabled by default. Disable only for testing.
```

jmank88, line 25: This block of fields has grown quite large. I think we should either insert an empty line between each of Otel/Message/Trace/Metric/Log, or convert these field comments to be trailing so the sections are not disrupted:
```suggestion
	EmitterBatchProcessor bool // Enabled by default. Disable only for testing.
```

gheorghestrimtu:
There is no corresponding core side PR for this change. These changes are meant to be backwards compatible.

jmank88:
> There is no corresponding core side PR for this change. These changes are meant to be backwards compatible.

OK, but if we expect LOOPPs to use these settings, then the core host process needs to forward them along as environment variables.

gheorghestrimtu:
@pkcll @4of9 could you please help answer @jmank88 question from above:

> Do we expect all configuration from core to propagate to the LOOPPs?

pkcll (dismissed):

jmank88:
> > > There is no corresponding core side PR for this change. These changes are meant to be backwards compatible.
> > 
> > 
> > OK, but if we expect LOOPPs to use these settings, then the core host process needs to forward them along as environment variables (which this PR needs to support).
> 
> @jmank88 Is it ok if we merge this PR separately without companion PR for chainlink core? This change is needed for load testing.

Not if you intend for LOOPPs to have access to these fields. This PR would need additional updates in that case.

jmank88:
To clarify, circling back with additional changes might be fine, but this repository will need to be updated again in order to support passing the new fields as environment variables. It is usually much simpler to update both places with new fields so the module remains consistent.

pkcll:
> To clarify, circling back with additional changes might be fine, but this repository will need to be updated again in order to support passing the new fields as environment variables. It is usually much simpler to update both places with new fields so the module remains consistent.

Working on the commit for that

jmank88 (approved):

cll-gg (approved):

jmank88 (approved):

PR #962, opened 2024-12-04
[chore] Add function to generate workflowID

cedric-cordenier:
Loosely based on the implementation in dev-platform without the prefix + lengths since the result was being hashed anyway.

### DIFF
--- a/pkg/workflows/utils.go
+++ b/pkg/workflows/utils.go
@@ -3,7 +3,7 @@ package workflows
 import (
 	"crypto/sha256"
 	"encoding/hex"
-	"fmt"
+	"strings"
 )
 
 func EncodeExecutionID(workflowID, eventID string) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/workflows/utils.go
```diff
@@ -19,3 +20,42 @@ func EncodeExecutionID(workflowID, eventID string) (string, error) {
 
 	return hex.EncodeToString(s.Sum(nil)), nil
 }
+
+func GenerateWorkflowIDFromStrings(owner string, workflow []byte, config []byte, secretsURL string) (string, error) {
+	ownerb, err := hex.DecodeString(owner)
```
ibrajer, line 25: Suggestion: I find this function very useful because you don't have to write all ETH address validations yourself: https://pkg.go.dev/github.com/ethereum/go-ethereum/common#IsHexAddress

cedric-cordenier: Thanks -- we have a rule against importing geth inside chainlink-common since it's supposed to be chain agnostic so we can't use that unfortunately.

🤔 Come to think of it, I'm not sure we can actually assume the owner will be 20 bytes long either for the same reason. Let me check what we do elsewhere

ibrajer: Good point, address length may vary from chain to chain. Let's drop all validations aside from the hex characters check, and we'll run address validation prior to calling this function (at least the CLI tool will cause it will know which chain the user has set).

cedric-cordenier: Sounds good 👍

justinkaseman (approved):

PR #963, opened 2024-12-04
INFOPLAT-1559 Adds auth header token expiry functionality

hendoxc:
# What
- Adds `BuildAuthHeadersV2` for signing publickey + timestamp.

# Why
- First step towards token expiration using CSA Auth.

# Notes
Further implementation for implementing token refresh will in follow up tickets


pkg/beholder/auth.go
```diff
@@ -11,17 +13,81 @@ var authHeaderKey = "X-Beholder-Node-Auth-Token"
...
 //
 // where the byte value of <public_key_hex> is what's being signed
-func BuildAuthHeaders(privKey ed25519.PrivateKey) map[string]string {
+// and <signature_hex> is the signature of the public key.
+// The version `2` is:
+//
+// <version>:<public_key_hex>:<timestamp>:<signature_hex>
+//
+// where the byte value of <public_key_hex> and <timestamp> are what's being signed
+func BuildAuthHeaders(privKey ed25519.PrivateKey, opts ...BuildAuthHeadersOpt) map[string]string {
```
jmank88, line 37: Is there a compelling reason to use functional options? I am worried about how they pollute the package API namespace with verbose functions. Would something simple and direct suffice?
```suggestion
func (AuthHeadersConfig) Build(privKey ed25519.PrivateKey) map[string]string {
```

hendoxc: Allows existing backwards compatibility of `BuildAuthHeaders`. But I agree,  I think the most direct approach here would be to expose `BuildAuthHeadersV2`

jmank88: `BuildAuthHeadersV2` sounds good, but FWIW adding variadic args may not break callers, but it does change the function signature and is not compatible by definition. This article is a good read: https://go.dev/blog/module-compatibility#adding-to-a-function
> Sometimes the techniques of adding a new function and adding options can be combined by making the options struct a method receiver. Consider the evolution of the net package’s ability to listen at a network address.

https://pkg.go.dev/net@go1.11#ListenConfig

hendoxc: Thanks ^ this was helpful, I took a look at the `crypto/tls` `Dial` example

pkg/beholder/auth.go
```diff
@@ -2,26 +2,75 @@ package beholder
...
 	"fmt"
+	"time"
 )
 
 // authHeaderKey is the name of the header that the node authenticator will use to send the auth token
 var authHeaderKey = "X-Beholder-Node-Auth-Token"
 
 // authHeaderVersion is the version of the auth header format
 var authHeaderVersion = "1"
+var authHeaderVersionV2 = "2"
```
jmank88, line 15: nit/ VersionVersion? :shrug: 
```suggestion
var authHeaderVersion2 = "2"
```

jmank88 (approved):

pkg/beholder/auth.go
```diff
@@ -2,26 +2,75 @@ package beholder
...
+		config = defaultAuthHeaderConfig()
+	}
+	if config.version == "" {
+		config.version = authHeaderVersion2
+	}
+	// If timestamp is not set, use the current time
+	if config.timestamp == 0 {
+		config.timestamp = time.Now().UnixMilli()
+	}
+	// If timestamp is negative, set it to 0. negative values cause overflow on convertsion to uint64
```
4of9, line 54:
```suggestion
	// If timestamp is negative, set it to 0. negative values cause overflow on conversion to uint64
```

pkg/beholder/auth.go
```diff
@@ -2,26 +2,75 @@ package beholder
...
+	// 0 timestamps will be rejected by the server as being too old
+	if config.timestamp < 0 {
+		config.timestamp = 0
```
4of9, line 57: this sorta conflicts in spirit to the previous block where we set the timestamp to `now` if it's `0`

should this be moved up?

pkg/beholder/auth_test.go
```diff
@@ -2,21 +2,137 @@ package beholder
...
+	assert.Equal(t, csaPrivKey.Public().(ed25519.PublicKey), ed25519.PublicKey(pubKeyBytes))
+
+	// Parse the timestamp
+	timestampParsed, err := strconv.ParseInt(timestampStr, 10, 64)
+	require.NoError(t, err)
+
+	// Verify the timestamp is within the last 5 seconds
+	// This verifies that default configuration is to use the current time
+	now := time.Now().UnixMilli()
+	assert.InDelta(t, now, timestampParsed, 5000, "timestamp should be within the last 5 seconds")
```
4of9, line 96: nitpick: could tighten up the time range by recording `now` at the start of the test, sleeping a few milliseconds, and then comparing against that

pkg/beholder/auth.go
hendoxc: So you're saying if a negative value is passed in, we should auto correct to current time?

hendoxc: I think let's keep as is then

pkg/beholder/auth_test.go
```diff
@@ -2,21 +2,138 @@ package beholder
...
+	require.NoError(t, err)
+	assert.Equal(t, csaPrivKey.Public().(ed25519.PublicKey), ed25519.PublicKey(pubKeyBytes))
+
+	// Parse the timestamp
+	timestampParsed, err := strconv.ParseInt(timestampStr, 10, 64)
+	require.NoError(t, err)
+
+	// Verify the timestamp is within the last 50ms
+	// This verifies that default configuration is to use the current time
+	assert.InDelta(t, now, timestampParsed, 50, "timestamp should be within the last 50ms")
```
pkcll, line 97: nit: Should it be just after or equal `now` ?  or you can use mock clock github.com/jonboulle/clockwork

pkcll (approved):

pkg/beholder/auth.go
```diff
@@ -2,26 +2,75 @@ package beholder
...
 	pubKey := privKey.Public().(ed25519.PublicKey)
 	messageBytes := pubKey
 	signature := ed25519.Sign(privKey, messageBytes)
-	headerValue := fmt.Sprintf("%s:%x:%x", authHeaderVersion, messageBytes, signature)
 
-	return map[string]string{authHeaderKey: headerValue}
+	return map[string]string{authHeaderKey: fmt.Sprintf("%s:%x:%x", authHeaderVersion1, messageBytes, signature)}
+}
+
+func BuildAuthHeadersV2(privKey ed25519.PrivateKey, config *AuthHeaderConfig) map[string]string {
```
cll-gg, line 43:
```suggestion
// BuildAuthHeadersV2 creates the auth header value to be included on requests.
// See documentation on BuildAuthHeaders for more info. 
func BuildAuthHeadersV2(privKey ed25519.PrivateKey, config *AuthHeaderConfig) map[string]string {
```

pkg/beholder/auth.go
```diff
@@ -2,26 +2,75 @@ package beholder
...
+	if config == nil {
+		config = defaultAuthHeaderConfig()
+	}
+	if config.version == "" {
+		config.version = authHeaderVersion2
+	}
```
cll-gg, line 49: Not sure how important it is, but with the current code we theoretically allow a caller to build v2 headers while passing in a different `version`, which will get set in the header key.

pkg/beholder/auth.go
```diff
@@ -2,26 +2,75 @@ package beholder
...
+	messageBytes := append(pubKey, timestampUnixMsBytes...)
+	signature := ed25519.Sign(privKey, messageBytes)
+
+	return map[string]string{authHeaderKey: fmt.Sprintf("%s:%x:%d:%x", config.version, pubKey, config.timestamp, signature)}
+}
+
+func defaultAuthHeaderConfig() *AuthHeaderConfig {
+	return &AuthHeaderConfig{
+		version:   authHeaderVersion2,
+		timestamp: time.Now().UnixMilli(),
```
cll-gg, line 74: Is it intended that we don't use `config.timestamp` here? Seems counter-intuitive

hendoxc: not sure I understand

pkg/beholder/auth.go
hendoxc: Yes, by design, it means this function can support upgrading to a new version

pkg/beholder/auth.go
cll-gg: Ah I think I mixed things up, nvm!

cll-gg (approved):

4of9 (approved):

hendoxc:
Closing in favour of
https://github.com/smartcontractkit/chainlink-common/pull/974

PR #964, opened 2024-12-05
feat(observability-lib): remove consumers + refactor cmd

Atrax1:
- Remove consumers from the library
- Refactor cmd
- Improve managing notification policies + add associated tests

The reason of moving away consumers is to avoid deploying our grafana resources from chainlink-common CI


observability-lib/api/notification-policy_test.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func Pointer[T any](d T) *T {
```
patrickhuie19, line 10: we tend to replicate something like this across many of our testing packages. 
https://smartcontract-it.atlassian.net/browse/BCF-3448. No need to fix in this PR

patrickhuie19 (approved):

pavel-raykov (approved):

PR #965, opened 2024-12-06
Removed flakey testcases and optimized for parallel test runs

silaslenihan:
This PR removes some problematic checks from the ChainComponents interface tests. We test the ChainReader's ability to read a value at a specific confidence level. The tests work by submitting a transaction to the unconfirmed state, then trying to read the value at the Finalized level expecting an error. The issue here is that on certain chains, finality is instantaneous, and even when running on the Simulated Backend, sometimes there is a race condition where finality is achieved before those checks can go through. Since these tests are supposed to be chain agnostic (including for instant finality chains) we should remove these two checks to avoid unwanted behavior on instant finality chains and race conditions on EVM.  

Additionally, this PR enables parallel running of ChainComponents Tests in order to drastically reduce runtime. 

### Supports
- https://github.com/smartcontractkit/chainlink/pull/15537
- https://github.com/smartcontractkit/chainlink-solana/pull/986

### DIFF
--- a/pkg/loop/internal/goplugin/service.go
+++ b/pkg/loop/internal/goplugin/service.go
@@ -3,6 +3,7 @@ package goplugin
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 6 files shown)


george-dorin (dismissed):

pkg/loop/internal/relayer/pluginprovider/contractreader/__debug_bin3704910103
jmank88, line 1: What is this binary file? Accidental?

silaslenihan: Yep, accidentally left the debugger open. Removed.

ilija42 (approved):

jmank88 (approved):

PR #967, opened 2024-12-10
[INFOPLAT-1592] Address high CPU utilization when telemetry is enabled

pkcll:
**Context** 
 
High CPU usage for Telemetry reported by NOPs running Asset DON node.

Profiles examples taken from DON nodes:

image
image

It seems that NewBatchSpanProcessor is the offender

**Description**

- [loop/EnvConfig] parse sets TelemetryEmitterBatchProcessor, TelemetryEmitterExportTimeout
- [beholder/client] BatchProcessor ExportTimeout option is non-zero value

### DIFF
--- a/pkg/beholder/client.go
+++ b/pkg/beholder/client.go
@@ -14,7 +14,6 @@ import (
 	sdklog "go.opentelemetry.io/otel/sdk/log"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	sdkresource "go.opentelemetry.io/otel/sdk/resource"
-	"go.opentelemetry.io/otel/sdk/trace"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	oteltrace "go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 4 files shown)


pkg/loop/config_test.go
```diff
@@ -106,12 +137,66 @@ func TestEnvConfig_parse(t *testing.T) {
...
+	if len(a) != len(b) {
+		return false
+	}
+	for k, v := range a {
+		if b[k] != v {
+			return false
+		}
+	}
+	return true
+}
```
jmank88, line 198: Does testify have helpers for comparing map entries?

pkg/beholder/client.go
```diff
@@ -314,9 +321,12 @@ func newTracerProvider(config Config, resource *sdkresource.Resource, creds cred
 	if err != nil {
 		return nil, err
 	}
-
+	batcherOpts := []sdktrace.BatchSpanProcessorOption{}
+	if config.TraceBatchTimeout > 0 {
```
pkcll, line 325: The fix

pkg/loop/config_test.go
cll-gg: Why not use `reflect.DeepEqual()`?

pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/967/commits/22280e07c68a8770b8e10ef4f0969553d63fd862

patrickhuie19 (approved):

4of9 (approved):

PR #974, opened 2024-12-13
INFOPLAT-1562 dynamic expiring auth headers

hendoxc:
[INFOPLAT-1560](https://smartcontract-it.atlassian.net/browse/INFOPLAT-1560)

## What

Allows usage of dynamic auth headers  by implementing `grpc.PerRPCCredentials` & setting as a `DialOption` on the otel client

## Why

Makes tokens expire, in the case that one is leaked or intercepted.
- someone can indeed steal a token, update the timestamp portion of the token, then try send data to the gateway endpoint, however since pubkey bytes + timestamp bytes are what's being signed, the signature part of the token is invalid, and will be rejected by the gateway
- on the gateway, version `2` tokens have their timestamp part checked `time.Now > timestamp > time.Now - serverTTL`

## Notes

Current users of the client can still configure static headers using the `AuthHeaders` field of beholder client `Config`. To enable dynamic headers they instead should configure `AuthHeaderProvider`

[INFOPLAT-1560]: https://smartcontract-it.atlassian.net/browse/INFOPLAT-1560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/beholder/auth.go
```diff
@@ -1,27 +1,145 @@
...
+		a.refresh()
+	}
+	return a.headers
+}
 
-// authHeaderVersion is the version of the auth header format
-var authHeaderVersion = "1"
+// refresh creates a new signed auth header token and sets the lastUpdated time to now
+func (a *authHeaderPerRPCredentials) refresh() {
+	a.lastUpdated = time.Now()
```
cll-gg, line 84: Probably doesn't matter much, but conceptually I would put this line after actually updating the headers

pkg/beholder/auth.go
```diff
@@ -1,27 +1,145 @@
...
+	return map[string]string{authHeaderKey: fmt.Sprintf("%s:%x:%x", authHeaderVersion1, messageBytes, signature)}
+}
+
+// buildAuthHeadersV2 creates the auth headers to be included on requests.
+// Version `2` is:
+//
+// <version>:<public_key_hex>:<timestamp>:<signature_hex>
+//
+// where the concatenated byte value of <public_key_hex> & <timestamp> is what's being signed
+func buildAuthHeadersV2(privKey ed25519.PrivateKey, config *AuthHeaderConfig) map[string]string {
```
cll-gg, line 124: Can we get away now with making this function unexported? The exported v1 version is used in https://github.com/smartcontractkit/chainlink/blob/develop/core/services/keystore/beholder.go#L16

pkg/beholder/auth.go
```diff
@@ -1,27 +1,145 @@
...
+	switch a.version {
+		// refresh doesn't actually do anything for version 1 since we are only signing the public key
+		// this for backwards compatibility and smooth transition to version 2
+		case authHeaderVersion1:
+			a.headers = BuildAuthHeaders(a.privKey)
+		case authHeaderVersion2:
+			a.headers = buildAuthHeadersV2(a.privKey, &AuthHeaderConfig{timestamp: a.lastUpdated.UnixMilli()})
+		default:
+			a.headers = buildAuthHeadersV2(a.privKey, &AuthHeaderConfig{timestamp: a.lastUpdated.UnixMilli()})
+	}
```
cll-gg, line 94: This seems not thread-safe, maybe we should make `headers` atomic or add a mutex?

pkg/beholder/auth.go
```diff
@@ -1,27 +1,145 @@
...
+func (a *authHeaderPerRPCredentials) Credentials() credentials.PerRPCCredentials {
+	return a
+}
+
+func (a *authHeaderPerRPCredentials) GetRequestMetadata(_ context.Context, _ ...string) (map[string]string, error) {
+	return a.getHeaders(), nil
+}
+
+func (a *authHeaderPerRPCredentials) RequireTransportSecurity() bool {
+	return false
```
4of9, line 71: should maybe make this configurable, even though our main use case is over localhost

pkg/beholder/auth.go
```diff
@@ -1,27 +1,145 @@
...
+}
+
+// get headers returns the auth headers, refreshing them if they are expired
+func (a *authHeaderPerRPCredentials) getHeaders() map[string]string {
+	if time.Since(a.lastUpdated) > a.headerTTL {
+		a.refresh()
+	}
```
4of9, line 78: I'm thinking of the scenario where a header is returned at second 59, then it's expired by the time it hits our infra

what do you think about adding a configurable buffer here? or is there a better way to address that?

pkg/beholder/auth.go
hendoxc: yes, since it won't be *directly* called by anyone using/configuring a Beholder client. Users should configure a `AuthHeaderProvider` for their client,  which calls `buildAuthHeadersV2` under the hood when auth headers are set via `PerGRPCCredendtials`

pkg/beholder/auth.go
hendoxc: Have made a ticket to not lose this

pkg/beholder/auth.go
```diff
@@ -1,27 +1,158 @@
...
+}
+
+// AuthHeaderProviderConfig configures AuthHeaderProvider
+type AuthHeaderProviderConfig struct {
+	HeaderTTL                time.Duration
+	Version                  string
+	RequireTransportSecurity bool
+}
+
+func NewAuthHeaderProvider(privKey ed25519.PrivateKey, config *AuthHeaderProviderConfig) AuthHeaderProvider {
```
jmank88, line 46: Another option is to use a method for construction:
```suggestion
func (config AuthHeaderProviderConfig) New(privKey ed25519.PrivateKey) AuthHeaderProvider {
```
This simplifies the callers, which don't have to repeat the package or the type name:
```diff
- beholder.NewAuthHeaderProvider(privKey, &beholder.AuthHeaderProviderConfig{HeaderTTL: time.Minute})
+ beholder.AuthHeaderProviderConfig{HeaderTTL: time.Minute}.New(privKey)
```
And you could keep this simple func around too, for the cases passing `nil`:
```suggestion
func NewAuthHeaderProvider(privKey ed25519.PrivateKey) AuthHeaderProvider {
	return AuthHeaderProviderConfig{}.New(privKey)
```

4of9 (approved):

pkg/beholder/client.go
```diff
@@ -345,7 +354,11 @@ func newMeterProvider(config Config, resource *sdkresource.Resource, creds crede
 	opts := []otlpmetricgrpc.Option{
 		otlpmetricgrpc.WithTLSCredentials(creds),
 		otlpmetricgrpc.WithEndpoint(config.OtelExporterGRPCEndpoint),
-		otlpmetricgrpc.WithHeaders(config.AuthHeaders),
+	}
+	if config.AuthHeaderProvider != nil {
+		opts = append(opts, otlpmetricgrpc.WithDialOption(grpc.WithPerRPCCredentials(config.AuthHeaderProvider.Credentials())))
```
pkcll, line 359: It seems there should be a relationship between otlpmetricgrpc.WithTLSCredentials (line 355) and 
config.AuthHeaderProvider.Credentials().RequireTransportSecurity

If TLS is enabled AuthHeaderProvider.Credentials().RequireTransportSecurity()should always return true. 
Im wondering whats the good way to handle that?

pkg/beholder/client.go
```diff
@@ -92,7 +93,11 @@ func newGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
 	opts := []otlploggrpc.Option{
 		otlploggrpc.WithTLSCredentials(creds),
 		otlploggrpc.WithEndpoint(cfg.OtelExporterGRPCEndpoint),
-		otlploggrpc.WithHeaders(cfg.AuthHeaders),
+	}
+	if cfg.AuthHeaderProvider != nil {
+		opts = append(opts, otlploggrpc.WithDialOption(grpc.WithPerRPCCredentials(cfg.AuthHeaderProvider.Credentials())))
```
pkcll, line 98: Nice!

pkcll (approved):

pkg/beholder/client.go
pkcll: Lets test what happens when RequireTransportSecurity and InsecureConnection mismatch.

pkcll (approved):

4of9 (dismissed):

pkg/loop/config.go
```diff
@@ -57,6 +58,7 @@ type EnvConfig struct {
 	TelemetryTraceSampleRatio          float64
 	TelemetryAuthHeaders               map[string]string
 	TelemetryAuthPubKeyHex             string
```
cll-gg, line 60: Is it intended that these fields are kept even though they're not used anymore?

cll-gg (approved):

pkg/loop/config.go
```diff
@@ -57,6 +58,7 @@ type EnvConfig struct {
...
+	TelemetryAuthHeaderProvider        beholder.AuthHeaderProvider
```
pkcll, line 61: This EnvConfig options are passed to the LOOPP process via env variables (see TestEnvConfig_AsCmdEnv, TestEnvConfig_parse). You can pass static raw data like that from core to LOOPP but for passing TelemetryAuthHeaderProvider instance you need different approach. 
IIRC privateKey part of CSA key is NOT available in the LOOPP process (only in the Core). 
Signing key from the LOOPP process might require exposing endpoint on the Core side or some other approach. I think @4of9 should have more context on that.

4of9: that's a good callout @pkcll. IIRC a LOOPP can communicate back up to the core process through gRPC, though I haven't looked too much into that. the header would need to be fetched that way as indeed the CSA key stays with the main process

patrickhuie19: Just to clarify: dialing occurs now when
* the loop boots
* connection to the otel endpoint is lost and must be re-dialed

In the CRE architecture currently, its worth noting that loops are long lived. The second scenario occurring would result in long periods of downtime if a valid signature could not be generated.

patrickhuie19: > LOOPP process might require exposing endpoint on the Core side or some other approach
This is possible. hosts and clients in hashicorp go-plugin model can communicate bidirectionally. In this specific case, we would request the core node open up a server to serve fresh signatures to the client in the loop.

pkg/beholder/auth.go
```diff
@@ -1,27 +1,170 @@
...
-	return map[string]string{authHeaderKey: headerValue}
+	return map[string]string{authHeaderKey: fmt.Sprintf("%s:%x:%x", authHeaderVersion1, messageBytes, signature)}
+}
+
+// buildAuthHeadersV2 creates the auth headers to be included on requests.
+// Version `2` is:
+//
+// <version>:<public_key_hex>:<timestamp>:<signature_hex>
+//
+// where the concatenated byte value of <public_key_hex> & <timestamp> is what's being signed
```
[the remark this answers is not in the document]
hendoxc: If version of the token is not matching token scheme we expect, we consider the token `malformed` and reject the data

 e.g we get a token that is version `1`, we'd try to assert we can see a string that looks exactly like `<version>:<public_key_hex>:<signature_hex>`

pkg/beholder/auth.go
```diff
@@ -1,27 +1,170 @@
...
+func buildAuthHeadersV2(privKey ed25519.PrivateKey, config *AuthHeaderConfig) map[string]string {
+	if config == nil {
+		config = &AuthHeaderConfig{}
+	}
+	if config.version == "" {
+		config.version = authHeaderVersion2
+	}
+	// If timestamp is negative or 0, set it to current timestamp.
+	// negative values cause overflow on conversion to uint64
```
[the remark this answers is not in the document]
hendoxc: No production use-cases, any invalid values are corrected for
```
if config.timestamp <= 0 {
		config.timestamp = time.Now().UnixMilli()
	}
```

pkg/beholder/auth.go
```diff
@@ -1,27 +1,170 @@
...
+// AuthHeaderProviderConfig configures AuthHeaderProvider
+type AuthHeaderProviderConfig struct {
+	HeaderTTL                time.Duration
+	Version                  string
+	RequireTransportSecurity bool
+}
+
+// authHeaderPerRPCredentials is a PerRPCCredentials implementation that provides the auth headers
+type authHeaderPerRPCCredentials struct {
+	privKey                  ed25519.PrivateKey
```
patrickhuie19, line 40: Passing a private key around will make eventual isolation of the keystore difficult. If possible, I'd like to see us use the keystore directly here

PR #976, opened 2024-12-17
Generic heartbeat service

patrickhuie19:
### Supports
May eventually support https://github.com/smartcontractkit/chainlink/pull/15700/files

### DIFF
--- a/pkg/services/heartbeat.go
+++ b/pkg/services/heartbeat.go
@@ -1,0 +1,59 @@
+package services
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/timeutil"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/services/heartbeat.go
```diff
@@ -0,0 +1,59 @@
...
+// used in the core node and for loops. It accepts a named logger,
+// a beat, a setup func to initialize resources used on each beat,
+// a beat function to define the behavior on each beat, and a close func
+// for resource teardown
+type Heartbeat struct {
+	Service
+	eng *Engine
+
+	beat time.Duration
+	lggr logger.Logger
```
pkcll, line 22: Shall we pass beholder client as a dependency here as well. Its same type of concern as logger.

patrickhuie19: Hmm - my thinking with breaking this out was that this heartbeat service could be used in core, loops, and the beholder client as well

PR #977, opened 2024-12-19
Fix `padWorkflowName()`

vyzaldysanchez:

### DIFF
--- a/pkg/capabilities/consensus/ocr3/types/aggregator.go
+++ b/pkg/capabilities/consensus/ocr3/types/aggregator.go
@@ -1,7 +1,6 @@
 package types
 
 import (
-	"encoding/hex"
 	"strings"
 
 	ocrcommon "github.com/smartcontractkit/libocr/commontypes"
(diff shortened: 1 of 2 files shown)


pkg/capabilities/consensus/ocr3/types/aggregator.go
```diff
@@ -24,10 +25,16 @@ type Metadata struct {
...
 func (m *Metadata) padWorkflowName() {
-	if len(m.WorkflowName) < 10 {
+	b, err := hex.DecodeString(m.WorkflowName)
+	if err == nil && len(b) < 10 {
+		// Each byte is 2 characters, so we need to pad with 0s
+		neededBytes := append(b, make([]byte, 10-len(b))...)
+		m.WorkflowName = hex.EncodeToString(neededBytes)
+	} else if len(m.WorkflowName) < 10 {
+		// Pad with spaces
 		suffix := strings.Repeat(" ", 10-len(m.WorkflowName))
```
bolekk, line 38: This should never be the case. You also don't really have to hex-decode. Just pad with "0" chars up to length 20.

bolekk (changes requested):

pkg/capabilities/consensus/ocr3/types/aggregator.go
bolekk: cc @krehermann  - cap_encoder.go in core validates that and fails it it's not a valid hex string that decodes to 10 bytes. We should require it here too.

krehermann: ok. if it's got to be that then let's move the validation to this struct. can be a follow up change.

code comments aren't sufficient if there are restrictions on the values. and doesn't make sense to have validation far away in a different repo. @vyzaldysanchez can you do that as a follow up (write a Validate method for this struct)

krehermann (approved):

pkg/capabilities/consensus/ocr3/types/aggregator_test.go
```diff
@@ -16,25 +16,32 @@ func TestMetadata_padWorkflowName(t *testing.T) {
...
 			},
-			want: "          ",
+			want: "00000000000000000000",
+		},
+		{
+			name: "padWorkflowName non-hex string",
+			fields: fields{
+				WorkflowName: "not-hex",
+			},
+			want: "not-hex0000000000000",
```
bolekk, line 44: lol, technically correct  ;)

bolekk (approved):

pkg/capabilities/consensus/ocr3/types/aggregator.go
vyzaldysanchez: sure thing!

PR #980, opened 2024-12-20
feat(workflows): adds workflow name normalizer

MStreet3:
normalizes a string by taking the sha256 hash and truncating the result to the first 10 bytes.

### DIFF
--- a/pkg/workflows/utils.go
+++ b/pkg/workflows/utils.go
@@ -72,3 +72,15 @@ func GenerateWorkflowID(owner []byte, name string, workflow []byte, config []byt
 
 	return sha, nil
 }
+
+// HashTruncateName returns the SHA-256 hash of the workflow name truncated to the first 10 bytes.
+func HashTruncateName(name string) [10]byte {
+	// Compute SHA-256 hash of the input string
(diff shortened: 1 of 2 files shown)


cedric-cordenier (dismissed):

cedric-cordenier (approved):

justinkaseman (dismissed):
LGTM!

pkg/workflows/utils.go
```diff
@@ -72,3 +72,15 @@ func GenerateWorkflowID(owner []byte, name string, workflow []byte, config []byt
 
 	return sha, nil
 }
+
+// NormalizeWorkflowName returns the first 10 bytes of the SHA-256 hash of the input string.
+func NormalizeWorkflowName(name string) [10]byte {
```
jmank88, line 77: Can we name this a little more directly? And consider that we are already in the workflows package, so repeating workflow is noisy for the caller:
```suggestion
func NameHashPrefix(name string) [10]byte {
```
🤷

shileiwill (dismissed):
LGTM

justinkaseman (approved):

jmank88 (approved):

PR #982, opened 2025-01-07
feat(observability-lib): can specify tooltip on timeseries panels and enable by default

Atrax1:
- Can specify tooltip on timeseries panels and enable by default


patrickhuie19 (dismissed):

observability-lib/grafana/panels.go
```diff
@@ -88,6 +88,37 @@ func newTransform(options *TransformOptions) dashboard.DataTransformerConfig {
...
+
+	builder := common.NewVizTooltipOptionsBuilder().
+		Mode(options.Mode).
+		Sort(options.Sort)
+
+	if options.MaxWidth != nil {
+		builder.MaxWidth(*options.MaxWidth)
+	}
+
+	if options.MaxWidth != nil {
```
cll-gg, line 115:
```suggestion
	if options.MaxHeight != nil {
```

Atrax1: indeed thanks for catching it just fixed

cll-gg (approved):

anukin (approved):

PR #984, opened 2025-01-07
[CRE-39] (fix): Add more guards & nil checks to WASM compute

justinkaseman:


chudilka1 (changes requested):
@justinkaseman, would you be so kind as to add unit tests for the changes? ATM, the coverage for these files is 50% (the required is > 75%)

agparadiso (approved):

pkg/workflows/wasm/host/module.go
```diff
@@ -596,21 +600,24 @@ func createLogFn(logger logger.Logger) func(caller *wasmtime.Caller, ptr int32,
 			args = append(args, k, v)
 		}
 
+		reg, _ := regexp.Compile(`[^\w]`)
+		sanitizedMsg := reg.ReplaceAllString(msg, " ")
```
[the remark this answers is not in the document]
justinkaseman: Changed it to `*`

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -211,7 +211,7 @@ func Test_Compute_Logs(t *testing.T) {
 	require.Len(t, logs.AllUntimed(), 1)
 	expectedEntries := []Entry{
 		{
-			Log: zapcore.Entry{Level: zapcore.InfoLevel, Message: "building workflow..."},
+			Log: zapcore.Entry{Level: zapcore.InfoLevel, Message: "building workflow***"},
```
patrickhuie19, line 214: could you help me understand why the '*'s are more secure/useful for the custom compute logs?

cedric-cordenier (approved):

pkg/workflows/wasm/host/wasm_test.go
justinkaseman: Because these custom compute logs are potentially externally supplied, we are replacing anything that isn't alpha-numeric to avoid [log injection](https://www.aptive.co.uk/blog/log-injection-attack/)

justinkaseman: Risk of this is it has poor non-english language support
We could loosen it up to only take out new lines, brackets, and back/forward slashes.

patrickhuie19: Is there a third party lib we can use thats battle tested?

justinkaseman: Possibly, we can look into that. I didn't see one that we already have, so it would require approval for a new dependency.

I'm just going to split it out, because it's becoming a big can of worms

patrickhuie19 (approved):

agparadiso (approved):

cedric-cordenier:
^ To add to the above, there are lots of integration tests which verify the behaviour of our guest <> host protocol. Because this crosses the WASM boundary it won't be captured in coverage.

cedric-cordenier (approved):

PR #986, opened 2025-01-08
pkg/logger/logkey: new package for log key glossary

jmank88:


pavel-raykov:
> This PR introduces a new `package logkey` to serve as a glossary of logger keys in order to align common fields.

What is the intended use of the keys from logkey? Is this for labelling with ZAP's zap.String(key, value) construction, or is it for arbitrary string / int constants that can be used for logging in a centralized way?

jmank88:
> > This PR introduces a new `package logkey` to serve as a glossary of logger keys in order to align common fields.
> 
> What is the intended use of the keys from logkey? Is this for labelling with ZAP's zap.String(key, value) construction, or is it for arbitrary string / int constants that can be used for logging in a centralized way?

We don't typically use the zap field API like that, but that would be valid. More often we just call the `w` variants: `Logger.<level>w("<message>", logkey.ChainID, chainID)`

pavel-raykov:
> > > This PR introduces a new `package logkey` to serve as a glossary of logger keys in order to align common fields.
> > 
> > 
> > What is the intended use of the keys from logkey? Is this for labelling with ZAP's zap.String(key, value) construction, or is it for arbitrary string / int constants that can be used for logging in a centralized way?
> 
> We don't typically use the zap field API like that, but that would be valid. More often we just call the `w` variants: `Logger.<level>w("<message>", logkey.ChainID, chainID)`

I see, so are you planning to extract all the possible keys in some kind of automatic way? Or would you just keep adding them here continuously?

jmank88:
> > > > This PR introduces a new `package logkey` to serve as a glossary of logger keys in order to align common fields.
> > > 
> > > 
> > > What is the intended use of the keys from logkey? Is this for labelling with ZAP's zap.String(key, value) construction, or is it for arbitrary string / int constants that can be used for logging in a centralized way?
> > 
> > 
> > We don't typically use the zap field API like that, but that would be valid. More often we just call the `w` variants: `Logger.<level>w("<message>", logkey.ChainID, chainID)`
> 
> I see, so are you planning to extract all the possible keys in some kind of automatic way? Or would you just keep adding them here continuously?

We could try to start with some automation, but adding keys manually would be sufficient for most common inconsistencies.

I did consider providing zap field support as well, but I'm not sure how that should work. Even just `ChainID` is an interesting example. We probably would want to encourage (or even enforce) `string` based chain IDs, but we don't want to force callers to make a `string` to pass in, since that is inefficient in cases where the log message does not need to be serialized. We'd rather also accept `fmt.Stringer`. So `zapfield.ChainID("string")` isn't flexible enough, we'd want a version to accept `fmt.Stringer` also. It wasn't obvious to me how to sort that out in a clean way, so I left it out of scope for now.

pavel-raykov:
> > > > > This PR introduces a new `package logkey` to serve as a glossary of logger keys in order to align common fields.
> > > > 
> > > > 
> > > > What is the intended use of the keys from logkey? Is this for labelling with ZAP's zap.String(key, value) construction, or is it for arbitrary string / int constants that can be used for logging in a centralized way?
> > > 
> > > 
> > > We don't typically use the zap field API like that, but that would be valid. More often we just call the `w` variants: `Logger.<level>w("<message>", logkey.ChainID, chainID)`
> > 
> > 
> > I see, so are you planning to extract all the possible keys in some kind of automatic way? Or would you just keep adding them here continuously?
> 
> We could try to start with some automation, but adding keys manually would be sufficient for most common inconsistencies.
> 
> I did consider providing zap field support as well, but I'm not sure how that should work. Even just `ChainID` is an interesting example. We probably would want to encourage (or even enforce) `string` based chain IDs, but we don't want to force callers to make a `string` to pass in, since that is inefficient in cases where the log message does not need to be serialized. We'd rather also accept `fmt.Stringer`. So `zapfield.ChainID("string")` isn't flexible enough, we'd want a `fmt.Stringer` version, and probably `int64` too. It wasn't obvious to me how to sort that out in a clean way, so I left it out of scope for now.

Sound good. Thanks for explaining!

pavel-raykov (dismissed):

pavel-raykov (approved):

pkg/logger/logkey/logkey.go
```diff
@@ -0,0 +1,10 @@
+package logkey
+
+const (
+	ChainID       = "chainID"       // string
+	ContractID    = "contractID"    // string
+	FeedID        = "feedID"        // string - hex-encoded 32-byte value, prefixed with "0x", all lowercase
```
jmank88, line 6: Perhaps lowercase is needlessly strict? I copied this from a `type FeedID string`.

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
```diff
@@ -112,12 +113,12 @@ func (a *dataFeedsAggregator) Aggregate(lggr logger.Logger, previousOutcome *typ
 		previousReportInfo := currentState.FeedInfo[feedIDStr]
 		feedID, err2 := datastreams.NewFeedID(feedIDStr)
 		if err2 != nil {
-			lggr.Errorw("could not convert %s to feedID", "feedID", feedID)
+			lggr.Errorw("could not convert %s to feedID", logkey.FeedID, feedID)
```
jmank88, line 116: Is `logkey` the best package name? We might want to use these with Prometheus too, and so we could generalize and shorten to just `keys`:
```suggestion
			lggr.Errorw("could not convert %s to feedID", keys.FeedID, feedID)
```
But this would have name collisions in some places.

Something like `labels` is perhaps less collision prone, but more verbose:
```suggestion
			lggr.Errorw("could not convert %s to feedID", labels.FeedID, feedID)
```

Any other ideas?

pavel-raykov: logconst or logentry?

pkg/logger/logkey/logkey.go
pavel-raykov: I think it is fine either way. What exactly do you mean by strict? I.e., in which situation lower case is better/worse than the upper case?

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
jmank88: I think it would need to be fewer than 6 characters to be an improvement on `logkey`

pkg/logger/logkey/logkey.go
jmank88: I mean that it seems superficial to expect/enforce that we use only lowercase. It is trivial to convert to only lowercase at the call site if that is what you need, and it could be a big pain to chase around all the cases and align them as lower case. Perhaps filtering logs is easier with all lowercase? That might be worth it.

krehermann, line 6: this will become a catch all across products and teams. eg FeedID is already specific product detail.

is that the intention?

jmank88: Most of these were already used in this module. But I agree there is a question of how much should be at this level, which made we worry about how to manage importing multiple packages like this if necessary. However, most of the concerns with centralization are not present (no extra imports, values should be immutable over time, etc.), so maybe nbd to put stuff here :shrug: My gut is that main modules could have some unique things that don't need to be promoted here, but most other shared keys from library modules could be.

jmank88: If we also had a `chainlink-framework` glossary, and `chainlink` had one itself, then it would be awkward to import all three in the current form. If we had a means of extending the base glossary, then we could import just one thing with all the keys. I'm not sure how to do this cleanly though. Interface methods are extra characters and overhead. Struct fields are mutable, plus still some overhead and indirection. Maybe the mutation risk is worth it... :thinking:

krehermann (dismissed):
discussed offline.

to move forward, agree to keep this simple and audit the growth/ussage of the package with [intent to mv/refactor if it becomes too ungainly](https://smartcontract-it.atlassian.net/browse/CRE-150)

pkg/capabilities/consensus/ocr3/datafeeds/feeds_aggregator.go
jmank88: Would `package lk` be too opaque? Typically they should be meaningful, but in this case the usage should always make it clear :thinking: It would also make aliasing packages less obnoxious

jmank88: Update to `package lk` to try it out. I think I like it.

PR #989, opened 2025-01-09
[CRE-42] Fix partial or truncated writes

agparadiso:
### Description

This fix validates that the size specified matches the actual length of the data being written in order to avoid partial/truncated writes or unwanted data if the size is bigger than the actual data written. More info on the ticket
[CRE-42](https://smartcontract-it.atlassian.net/browse/CRE-42)

### DIFF
--- a/pkg/workflows/wasm/host/module_test.go
+++ b/pkg/workflows/wasm/host/module_test.go
@@ -553,7 +553,7 @@ func Test_read(t *testing.T) {
 }
 
 func Test_write(t *testing.T) {
-	t.Run("successfully write to slice", func(t *testing.T) {
+	t.Run("OK-successfully_write_to_slice", func(t *testing.T) {
 		giveSrc := []byte("hello, world")
 		memory := make([]byte, 12)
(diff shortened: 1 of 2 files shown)


pkg/workflows/wasm/host/module_test.go
```diff
@@ -553,29 +553,43 @@ func Test_read(t *testing.T) {
 }
 
 func Test_write(t *testing.T) {
-	t.Run("successfully write to slice", func(t *testing.T) {
+	t.Run("OK-successfully_write_to_slice", func(t *testing.T) {
```
krehermann, line 556: these OK/NOK are out of character with the rest of the code base; recommend removing them

and, there is no need to explicitly underscore the names; the go tooling does it under the hood

krehermann (approved):

pkg/workflows/wasm/host/module_test.go
agparadiso: removed 👍🏼

krehermann (approved):

PR #990, opened 2025-01-09
Nested Value Codec Access

EasterTheBunny:
This commit introduces the ability to use a codec to encode/decode individual fields of a struct, including traversal to nested structs.

[NONEVM-1063](https://smartcontract-it.atlassian.net/browse/NONEVM-1063)

[NONEVM-1063]: https://smartcontract-it.atlassian.net/browse/NONEVM-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/codec/encodings/type_codec.go
```diff
@@ -138,6 +149,29 @@ func decode(c map[string]TypeCodec, raw []byte, into any, itemType string, exact
 	return codec.Convert(reflect.ValueOf(val), reflect.ValueOf(into), nil)
 }
 
-func (c CodecFromTypeCodec) GetMaxDecodingSize(ctx context.Context, n int, itemType string) (int, error) {
-	return c.GetMaxEncodingSize(ctx, n, itemType)
+func getCodec(c map[string]TypeCodec, itemType string) (TypeCodec, error) {
+	// itemType could recurse into nested structs
+	path := extendedItemType(itemType)
+
+	// itemType could recurse into nested structs
```
reductionista, line 156: Did you mean to write the same comment twice here?

reductionista (dismissed):

reductionista:
Tested this with my branch, and seems to work! I notice there is a test failure (race condition?) but other than that LGTM!

pkg/codec/encodings/type_codec.go
EasterTheBunny: Ooops I didn't mean to do that.

pkg/codec/encodings/type_codec.go
```diff
@@ -121,11 +127,16 @@ func (c LenientCodecFromTypeCodec) Decode(ctx context.Context, raw []byte, into
...
 
+func (c CodecFromTypeCodec) GetMaxDecodingSize(ctx context.Context, n int, itemType string) (int, error) {
+	return c.GetMaxEncodingSize(ctx, n, itemType)
+}
+
 func decode(c map[string]TypeCodec, raw []byte, into any, itemType string, exactSize bool) error {
-	ntcwt, ok := c[itemType]
-	if !ok {
-		return fmt.Errorf("%w: cannot find type %s", types.ErrInvalidType, itemType)
+	ntcwt, err := getCodec(c, itemType)
```
reductionista, line 135: You may have already realized this, but unfortunately this doesn't work.  You can't use a codec for a subfield to decode `raw`, since raw is the entire encoded struct, not just one subfield.

I wish this had occurred to me on Friday, but I think there's a fundamental problem with this interface--of having `CreateType` return only the subfield type and then passing that to `Decode`.

Stepping back a moment:  The only reason we need LogPoller to decode the raw data at all is because, unlike with EVM-encoded data, there's no way to predict what offset into a Borsh-encoded struct a subfield will be at without Borsh-decoding the entire struct.

So whatever we do to solve this problem, we have to decode the whole struct--not just a subfield. Because `Decode` doesn't have type information for the whole struct, it can't do that. So returning only the subfield type from CreateType is not useful.

We *could* fix this by having `CreateType` return the type of the entire struct, and passing that to Decode. Then having Decode decode the whole struct, and then extract the requested element out of it based on the `itemType` path. But that's an unnecessarily convoluted way to solve a much simpler problem.

The first two steps of what we need to solve it in the most straightforward way are already working without this PR:
1. Pass eventType as itemType to CreateType, and have it return the type of the whole struct.
2. Pass that into Decode and have that return the decoded struct.

All we need that's new is the third step:
3. A function which takes a struct as input and returns a nested subfield of that struct based on a dot-separated path.

This is what I originally implemented using reflect, and could probably dig up again.  I'm leaning now towards that approach, which wouldn't require modifying the codec... since doing this part doesn't require any knowledge of registered types or codecs, and doesn't involve any translation from on chain to off chain or vice versa.  It still seems like something the codec  package *might* be able to do more easily than writing it from scratch, but if so I have no idea how.

EasterTheBunny: Yes, this only works if you want to encode/decode a single field. It was never expected to be able to _extract_ a single field from a fully encoded struct without encoding that single field first.

Example:

```
type TypeA struct {
    A int
    B string
}
```

You can decode the whole struct with `Decode(ctx, allData, &structInstance, "")` where `allData` is the encoded value of the _entire_ struct. However, you cannot do `Decode(allData, &fieldA, "A")` and get only the int data.

Alternatively you can do something like the following:

```
// this would be the entire encoded struct
var allData []byte

var structInstance TypeA
_ = theCodec.Decode(allData, &structInstance, "TypeA")

// this will encode the value of the B field
strBts, _ := theCodec.Encode(ctx, structInstance.B, "TypeA.B")
```

I would envision in `LogPoller` that you could have different encoding and decoding codecs where you want to decode a struct and encode a single field, but use different encodings for each.

amit-momin: Just wanted to mention that Silas wrote something that could be used for step 3 for CW [here](https://github.com/smartcontractkit/chainlink-solana/blob/74b7a72c2706857920cc1e01b85fd4722559c798/pkg/solana/chainwriter/helpers.go#L85). If you we go this route, it might be worth pulling it into a util file so we can use it for CR too.

reductionista: @amit-momin Very interesting! Yes, that seems pretty similar to the helper I had to implement... we should def look at combining them when we get a chance. This is the one I needed to get everything working:
https://github.com/smartcontractkit/chainlink-solana/pull/1002/files#diff-43599bcdcf8c088e3eac4f93fd79202871475fcdeb2dcc9494a8935b9f991a07R439-R491

reductionista: > // this would be the entire encoded struct
> var allData []byte
>
> var structInstance TypeA
> _ = theCodec.Decode(allData, &structInstance, "TypeA")
>
> // this will encode the value of the B field
> strBts, _ := theCodec.Encode(ctx, structInstance.B, "TypeA.B")

That's very different from anything LogPoller does. Even if we completely redesigned it to work this way, it wouldn't be able to work. The only reason we're decoding anything in the logs is because we need to know the offset of where the data is in order to search for it in queries... if you don't know the offset, then encoding it in-place is pointless--either way there is no way to search for it.  There would also be no way to build an index on something like that, since every subkey would be at a different offset. That's why it has to extract them and indexes them separately in their own column.

reductionista: > However, you cannot do Decode(allData, &fieldA, "A") and get only the int data.

FWIW, this part actually did work. The example I used is the same as what you have: it was able to decode A successfully, but only because it was the first element in the struct so the offset happened to be 0.  That's why I thought it was working at one point. It's passing "TypeA.B" as the itemType that failed. It couldn't decode the string because it thought the string started at the beginning of the data representing the integer A.

ilija42 (commented):
The general approach looks valid, I'll do a more thorough review after modifiers are done

pkg/codec/encodings/struct.go
```diff
@@ -113,3 +120,46 @@ func (s *structCodec) SizeAtTopLevel(numItems int) (int, error) {
...
+
+	structType, ok := codec.(StructTypeCodec)
+	if !ok {
+		return nil, fmt.Errorf("%w: extended path not traversable for type %s", types.ErrInvalidType, itemType)
+	}
+
+	return structType.FieldCodec(tail)
+}
+
+type extendedItemType string
```
ilija42, line 152: nit
```suggestion
// extendedItemType is used for parsing through nested item types
type extendedItemType string
```

pkg/codec/by_item_type_modifier.go
```diff
@@ -22,13 +22,17 @@ type byItemTypeModifier struct {
 	modByitemType map[string]Modifier
 }
 
+// RetypeToOffChain attempts to apply a modifier using the provided itemType. To allow access to nested fields, this
+// function applies no modifications if a modifier by the specified name is not found.
```
ilija42, line 26: >  To allow access to nested fields, this function applies no modifications if a modifier by the specified name is not found.
This comment makes it sound as if the func won't throw an error if a modifier is not found

pkg/codec/by_item_type_modifier.go
```diff
@@ -40,13 +44,17 @@ func (b *byItemTypeModifier) TransformToOffChain(onChainValue any, itemType stri
...
-	if !ok {
-		return nil, fmt.Errorf("%w: cannot find modifier for %s", types.ErrInvalidType, itemType)
+	val any,
+	itemType string,
+	transform func(Modifier, any, string) (any, error),
+) (any, error) {
+	head, tail := extendedItemType(itemType).next()
+
+	if mod, ok := b.modByitemType[head]; ok {
+		return transform(mod, val, tail)
```
ilija42, line 54: Why is it now okay for the modifier to not exist?

pkg/codec/encodings/struct.go
```diff
@@ -113,3 +120,46 @@ func (s *structCodec) SizeAtTopLevel(numItems int) (int, error) {
...
+	if fieldName == "" {
+		return nil, fmt.Errorf("%w: field name required", types.ErrInvalidType)
+	}
+
+	idx, ok := s.fieldLookup[fieldName]
+	if !ok {
+		return nil, fmt.Errorf("%w: cannot find type %s", types.ErrInvalidType, itemType)
+	}
+
+	codec := s.fields[idx]
```
ilija42, line 138: maybe add a length check to prevent panic, although these should always be the same size

pkg/codec/encodings/struct.go
```diff
@@ -113,3 +120,46 @@ func (s *structCodec) SizeAtTopLevel(numItems int) (int, error) {
...
+
```
ilija42, line 139:
```suggestion
	// if itemType wasn't referencing a nested field
```

pkg/codec/encodings/struct.go
```diff
@@ -113,3 +120,46 @@ func (s *structCodec) SizeAtTopLevel(numItems int) (int, error) {
 	}
 	return size, nil
 }
+
+func (s *structCodec) FieldCodec(itemType string) (TypeCodec, error) {
```
ilija42, line 124: wdyt about adding a test that checks if a proper FieldCodec has been retrieved

pkg/codec/modifier_base.go
```diff
@@ -17,58 +17,88 @@ type modifierBase[T any] struct {
...
 		m.offToOnChainType[onChainType] = onChainType
 		m.onToOffChainType[onChainType] = onChainType
-		return onChainType, nil
-	}
+		m.onChainStructType = onChainType
+		m.offChainStructType = onChainType
 
-	if cached, ok := m.onToOffChainType[onChainType]; ok {
-		return cached, nil
+		return typeForPath(onChainType, itemType)
```
ilija42, line 57: This whole block is a bit hard to understand

Does something like this make sense?
`	if len(m.fields) == 0 {
		m.offToOnChainType[onChainType] = onChainType
		m.onToOffChainType[onChainType] = onChainType
		m.onChainStructType = onChainType
		m.offChainStructType = onChainType

		return typeForPath(onChainType, itemType)
	}
	
	// itemType != "" means that we want to retype the whole type
	// else means that we are retyping sub-fields of the whole type
	var onChainStructType reflect.Type
	if itemType != "" {
		onChainStructType = m.onChainStructType
	} else {
		onChainStructType = onChainType
		// sub-fields aren't cached
		if cached, ok := m.onToOffChainType[onChainStructType]; ok {
			return typeForPath(cached, itemType)
		}
	}`

pkg/codec/renamer.go
```diff
@@ -30,26 +32,72 @@ type renamer struct {
...
+func (r *renamer) TransformToOffChain(onChainValue any, itemType string) (any, error) {
+	// itemType references the on-chain type
+	// remap to the off-chain field name
+	if itemType != "" {
+		var ref string
+
+		parts := strings.Split(itemType, ".")
+		if len(parts) > 0 {
+			ref = parts[len(parts)-1]
+		}
```
ilija42, line 44: What is the expected input for itemType here? If you have 
```
type testStruct2 struct {
		A string
		B int64
		C int64
	}
	
	type testStruct struct {
		A string
		B int64
		C int64
		D testStruct2
	}

	type nestedTestStruct struct {
		A string
		B testStruct
		C []testStruct
		D string
	}

```

and itemType is  B.D this will fail?

pkg/codec/renamer.go
```diff
@@ -30,26 +32,72 @@ type renamer struct {
...
+		for on, off := range r.fields {
+			if ref == on {
+				// B.A -> C == B.C
+				parts[len(parts)-1] = off
+				itemType = strings.Join(parts, ".")
+
+				break
+			}
+		}
+	}
```
ilija42, line 55: wdyt about making the comment a bit more specific, something like:
 "Rename field/subfield path in itemType to match the modifier renaming"

pkg/codec/encodings/struct.go
EasterTheBunny: I ended up making this an exported function named `ItemTyper`.

pkg/codec/by_item_type_modifier.go
EasterTheBunny: I updated this comment to have more clarity.

pkg/codec/by_item_type_modifier.go
EasterTheBunny: I changed this to the way it was to maintain backward compatibility.

pkg/codec/modifier_base.go
```diff
@@ -17,58 +17,88 @@ type modifierBase[T any] struct {
...
 			tpe = nil
 			err = fmt.Errorf("%w: %v", types.ErrInvalidType, r)
 		}
 	}()
+
+	// if itemType is empty, store the type mappings
+	// if itemType is not empty, assume a sub-field property is expected to be extracted
+	onChainStructType := onChainType
+	if itemType != "" {
+		onChainStructType = m.onChainStructType
```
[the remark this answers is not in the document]
EasterTheBunny: Yes and it gets set at the end. This is done to check the cache and is the same as the original behavior.

pkg/codec/modifier_base.go
EasterTheBunny: The `itemType` being empty is what indicates the entire struct is expected to be retyped. Else is for a field extraction.

pkg/codec/modifier_base.go
```diff
@@ -17,58 +17,88 @@ type modifierBase[T any] struct {
...
 
-	if cached, ok := m.onToOffChainType[onChainType]; ok {
-		return cached, nil
+		return typeForPath(onChainType, itemType)
 	}
 
 	var offChainType reflect.Type
-	switch onChainType.Kind() {
+
+	// the onChainStructType here should always reference the full on-chain struct type
```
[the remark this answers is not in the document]
EasterTheBunny: We always need a reference to the full struct because retyping a nested field will input the `onChainType` for the _field_ and not the struct. That switch is handled by this:

```
onChainStructType := onChainType
if itemType != "" {
	onChainStructType = m.onChainStructType
}
```

pkg/codec/encodings/struct.go
```diff
@@ -113,3 +120,34 @@ func (s *structCodec) SizeAtTopLevel(numItems int) (int, error) {
...
+	idx, ok := s.fieldLookup[fieldName]
+	if !ok {
+		return nil, fmt.Errorf("%w: cannot find type %s", types.ErrInvalidType, itemType)
+	}
+
+	if idx >= len(s.fields) {
+		return nil, fmt.Errorf("%w: invalid field index for type %s", types.ErrInvalidType, itemType)
+	}
+
+	codec := s.fields[idx]
```
ilija42, line 140: codec colliedes with package name, maybe fieldCodec makes sense?

pkg/codec/encodings/struct.go
```diff
@@ -113,3 +120,34 @@ func (s *structCodec) SizeAtTopLevel(numItems int) (int, error) {
...
+		return nil, fmt.Errorf("%w: field name required", types.ErrInvalidType)
+	}
+
+	idx, ok := s.fieldLookup[fieldName]
+	if !ok {
+		return nil, fmt.Errorf("%w: cannot find type %s", types.ErrInvalidType, itemType)
+	}
+
+	if idx >= len(s.fields) {
+		return nil, fmt.Errorf("%w: invalid field index for type %s", types.ErrInvalidType, itemType)
```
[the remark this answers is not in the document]
EasterTheBunny: Harsher? Are you thinking it should panic or be worded differently?

pkg/codec/byte_string_modifier.go
```diff
@@ -25,7 +25,18 @@ type AddressModifier interface {
...
+	fields []string,
+	modifier AddressModifier,
+) Modifier {
+	return NewPathTraverseAddressBytesToStringModifier(fields, modifier, false)
+}
+
+func NewPathTraverseAddressBytesToStringModifier(
+	fields []string,
+	modifier AddressModifier,
+	enablePathTraverse bool,
```
ilija42, line 38: `enablePathTraverse` is unused

pkg/codec/by_item_type_modifier.go
```diff
@@ -15,38 +15,74 @@ func NewByItemTypeModifier(modByItemType map[string]Modifier) (Modifier, error)
...
+// NewNestableByItemTypeModifier returns a Modifier that uses modByItemType to determine which Modifier to use for a
+// given itemType. If itemType is structured as a dot-separated string like 'A.B.C', the first part 'A' will be used to
+// match in the mod map and the remaining list will be provided to the found Modifier 'B.C'.
+func NewNestableByItemTypeModifier(modByItemType map[string]Modifier) (Modifier, error) {
+	if modByItemType == nil {
+		modByItemType = map[string]Modifier{}
+	}
+
+	return &byItemTypeModifier{
+		modByitemType: modByItemType,
```
ilija42, line 31:
```suggestion
		modByItemType: modByItemType,
```

pkg/codec/epoch_to_time.go
```diff
@@ -11,6 +11,10 @@ import (
 
 // NewEpochToTimeModifier converts all fields from time.Time off-chain to int64.
 func NewEpochToTimeModifier(fields []string) Modifier {
+	return NewPathTraverseEpochToTimeModifier(fields, false)
+}
+
+func NewPathTraverseEpochToTimeModifier(fields []string, enablePathTraverse bool) Modifier {
```
ilija42, line 17: `enablePathTraverse` unused

pkg/codec/precodec.go
```diff
@@ -11,12 +11,24 @@ import (
 
 // PreCodec creates a modifier that will run a preliminary encoding/decoding step.
 // This is useful when wanting to move nested data as generic bytes.
-func NewPreCodec(fields map[string]string, codecs map[string]types.RemoteCodec) (Modifier, error) {
+func NewPreCodec(
```
ilija42, line 14: Comment should start with NewPreCodec

pkg/codec/encodings/struct.go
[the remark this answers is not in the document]
ilija42: This is a fatal error from codec perspective, but I don't think we want a panic, so just worded differently

ilija42 (approved):

amit-momin (approved):

PR #993, opened 2025-01-13
Various updates to observabilty library

leeyikjiun:
### Changes
- Add text panel
- Make Title and Decimals in panel nullable
- Add LineWidth and DrawStyle to time series panel
- Add more configs to log panel
- Add Hide to custom variable
- Add timezone and graph tooltip to dashboard builder

### Notes
Making title and decimals nullable is a BREAKING CHANGE.
- Sometimes an empty title is needed to have a dummy panel to fill the available space.
- Leaving the decimal as `nil` which will become `auto` in Grafana is better than manually figuring out the real dp.

### DIFF
--- a/observability-lib/grafana/panels.go
+++ b/observability-lib/grafana/panels.go
@@ -10,6 +10,7 @@ import (
 	"github.com/grafana/grafana-foundation-sdk/go/prometheus"
 	"github.com/grafana/grafana-foundation-sdk/go/stat"
 	"github.com/grafana/grafana-foundation-sdk/go/table"
+	"github.com/grafana/grafana-foundation-sdk/go/text"
 	"github.com/grafana/grafana-foundation-sdk/go/timeseries"
 )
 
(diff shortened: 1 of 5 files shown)


observability-lib/grafana/panels.go
```diff
@@ -453,20 +483,45 @@ func NewTablePanel(options *TablePanelOptions) *Panel {
...
+	DedupStrategy    common.LogsDedupStrategy
+	SortOrder        common.LogsSortOrder
 }
 
 func NewLogPanel(options *LogPanelOptions) *Panel {
 	setDefaults(options.PanelOptions)
 
+	if options.EnableLogDetails == nil {
+		options.EnableLogDetails = Pointer[bool](true)
+	}
```
Atrax1, line 498: So we enable by default log details

Atrax1 (approved):

observability-lib/grafana/panels.go
leeyikjiun: Yes, most of the defaults that I have added are the same defaults if you were to use the UI.

krehermann (approved):

PR #1022, opened 2025-02-04
Various updates to Observability Library

leeyikjiun:
- Fix alert query type not instant when the query is instant
- Add disable resolve message and uid to contact point
- Add stacking mode to time series panel
- Add multi and include all to custom variable
- Use csv when the variable key and value is the same

Example of variable values:
Before:
`crit : crit , error : error , warn : warn , info : info , debug : debug `
After:
`error, warn, info, debug, crit`

### DIFF
--- a/observability-lib/grafana/alerts.go
+++ b/observability-lib/grafana/alerts.go
@@ -35,6 +35,7 @@ func newRuleQuery(query RuleQuery) *alerting.QueryBuilder {
 		RefId(query.RefID)
 
 	if query.Instant {
+		model.QueryType("instant")
 		model.Instant()
 	}
 
--- a/observability-lib/grafana/contact-point.go
+++ b/observability-lib/grafana/contact-point.go
@@ -3,14 +3,23 @@ package grafana
 import "github.com/grafana/grafana-foundation-sdk/go/alerting"
 
 type ContactPointOptions struct {
-	Name     string
-	Type     alerting.ContactPointType
-	Settings map[string]interface{}
+	Name                  string
+	Type                  alerting.ContactPointType
+	Settings              map[string]interface{}
+	DisableResolveMessage bool
+	Uid                   string
 }
 
 func NewContactPoint(options *ContactPointOptions) *alerting.ContactPointBuilder {
-	return alerting.NewContactPointBuilder().
+	builder := alerting.NewContactPointBuilder().
 		Name(options.Name).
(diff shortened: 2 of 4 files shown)


observability-lib/grafana/notification-template.go
```diff
@@ -3,21 +3,28 @@ package grafana
...
 	"github.com/grafana/grafana-foundation-sdk/go/alerting"
 	"gopkg.in/yaml.v3"
 )
 
-func NewNotificationTemplatesFromFile(filepath string) (map[string]alerting.NotificationTemplate, error) {
+func NewNotificationTemplatesFromFile(filePath string) (map[string]alerting.NotificationTemplate, error) {
+	fileName := strings.TrimSuffix(filepath.Base(filePath), filepath.Ext(filePath))
+	if fileName == "notification-templates" {
+		fileName = "chainlink"
+	}
```
Atrax1, line 17: why is this necessary ?

Atrax1 (approved):

Atrax1:
@leeyikjiun Some test to fix as well

observability-lib/grafana/notification-template.go
leeyikjiun: Because I'm using the filename as the template name. So a `dev-platform-por.yaml` will end up becoming `dev-platform-por-slack-notification-template`.

We can rename `notification-templates.yaml` to `chainlink.yaml` as well. Either way is fine.

Atrax1: yes remove that logic I think better to rename file

Atrax1 (approved):
Nice 👍

observability-lib/api/notification-template.go
```diff
@@ -45,7 +45,7 @@ func (c *Client) DeleteNotificationTemplate(name string) (DeleteNotificationTemp
 	}
 
 	statusCode := resp.StatusCode()
-	if statusCode != 200 {
+	if statusCode != 204 {
```
jmank88, line 48:
```suggestion
	if statusCode != http.StatusNoContent {
```

jmank88 (approved):

observability-lib/api/notification-template.go
leeyikjiun: Great suggestion! Let me fix all the status codes in one go in another PR. cc @Atrax1

PR #1023, opened 2025-02-06
Adds consensus cap config proto

vyzaldysanchez:
This addresses the following stories:

- [smartcontract-it.atlassian.net/browse/CRE-141](https://smartcontract-it.atlassian.net/browse/CRE-141)
- [smartcontract-it.atlassian.net/browse/CRE-142](https://smartcontract-it.atlassian.net/browse/CRE-142)
- https://smartcontract-it.atlassian.net/browse/CRE-143
- https://smartcontract-it.atlassian.net/browse/CRE-144

For this to make sense, https://github.com/smartcontractkit/chainlink/pull/16247 needs to be taken a look at, since it is what uses this.

Then all this will end up having a C-D PR to complete the cycle.

Related PRs:

- <https://github.com/smartcontractkit/chainlink/pull/16247>
- <https://github.com/smartcontractkit/chainlink-deployments/pull/710/>


pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -37,19 +39,24 @@ func newFactory(s *requests.Store, c *capability, batchSize int, outcomePruningT
...
 }
 
-func (o *factory) NewReportingPlugin(ctx context.Context, config ocr3types.ReportingPluginConfig) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
+func (o *factory) NewReportingPlugin(_ context.Context, config ocr3types.ReportingPluginConfig) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
+	var configProto types.ReportingPluginConfig
+	err := proto.Unmarshal(config.OffchainConfig, &configProto)
+	if err != nil {
+		return nil, ocr3types.ReportingPluginInfo{}, err
+	}
 	rp, err := newReportingPlugin(o.store, o.capability, o.batchSize, config, o.outcomePruningThreshold, o.lggr)
```
bolekk, line 48: Here o.batchSize should become configProto.maxRequestBatchSize

bolekk: You can do the same for outcomePruningThreshold - make it configurable via OCR onchain config.

pkg/capabilities/consensus/ocr3/types/ocr3_config_types.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+option go_package = "capabilities/consensus/ocr3/types";
+
+package ocr3_config_types;
+
+message ReportingPluginConfig {
+  uint32 maxQueryLengthBytes = 1;
+  uint32 maxObservationLengthBytes = 2;
+  uint32 maxReportLengthBytes = 3;
```
bolekk, line 10: Missing maxOutcomeLength and maxReportCount

pkg/capabilities/consensus/ocr3/types/ocr3_config_types.proto
```diff
@@ -0,0 +1,13 @@
...
+  uint32 maxRequestBatchSize = 4;
```
bolekk, line 11: Call is just maxBatchSize, not "request"

pkg/capabilities/consensus/ocr3/types/ocr3_config_types.proto
```diff
@@ -0,0 +1,16 @@
...
+
+message ReportingPluginConfig {
+  uint32 maxQueryLengthBytes = 1;
+  uint32 maxObservationLengthBytes = 2;
+  uint32 maxReportLengthBytes = 3;
+  uint32 maxOutcomeLengthBytes = 4;
+  uint32 maxReportCount = 5;
+  uint32 maxBatchSize = 6;
+  uint64 outcomePruningThreshold = 7;
+  bool uniqueReports = 8;
```
bolekk, line 15: You can remove uniqueReports. It's hardcoded to true in the contract so it also needs to stay =true in the node.

pkg/capabilities/consensus/ocr3/types/ocr3_config_types.proto
```diff
@@ -0,0 +1,16 @@
...
+
+option go_package = "capabilities/consensus/ocr3/types";
+
+package ocr3_config_types;
+
+message ReportingPluginConfig {
+  uint32 maxQueryLengthBytes = 1;
+  uint32 maxObservationLengthBytes = 2;
+  uint32 maxReportLengthBytes = 3;
+  uint32 maxOutcomeLengthBytes = 4;
```
bolekk, line 11: nit: swap the order of Outcome and Report to resemble libOCR struct.

pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -21,35 +23,32 @@ type factory struct {
...
 
-func (o *factory) NewReportingPlugin(ctx context.Context, config ocr3types.ReportingPluginConfig) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
-	rp, err := newReportingPlugin(o.store, o.capability, o.batchSize, config, o.outcomePruningThreshold, o.lggr)
-	info := ocr3types.ReportingPluginInfo{
+func (o *factory) NewReportingPlugin(_ context.Context, config ocr3types.ReportingPluginConfig) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
+	var configProto types.ReportingPluginConfig
+	err := proto.Unmarshal(config.OffchainConfig, &configProto)
+	if err != nil {
+		return nil, ocr3types.ReportingPluginInfo{}, err
+	}
```
bolekk, line 39: We need to apply defaults if any fields are set to 0. Otherwise we're risking a big failure during deployment.

pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -3,14 +3,21 @@ package ocr3
...
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/requests"
-
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/types"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 )
 
+const (
+	defaultMaxPhaseOutputBytes = 1000000 // 1 MB
+	defaultMaxReportCount      = 20
```
bolekk, line 18: Can you move defaultBatchSize and defaultOutcomePruningThreshold here, for consistency?

pkg/capabilities/consensus/ocr3/factory.go
```diff
@@ -21,35 +28,53 @@ type factory struct {
...
 }
 
-func (o *factory) NewReportingPlugin(ctx context.Context, config ocr3types.ReportingPluginConfig) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
-	rp, err := newReportingPlugin(o.store, o.capability, o.batchSize, config, o.outcomePruningThreshold, o.lggr)
-	info := ocr3types.ReportingPluginInfo{
+func (o *factory) NewReportingPlugin(_ context.Context, config ocr3types.ReportingPluginConfig) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
+	var configProto types.ReportingPluginConfig
+	err := proto.Unmarshal(config.OffchainConfig, &configProto)
+	if err != nil {
+		return nil, ocr3types.ReportingPluginInfo{}, err
```
bolekk, line 43: Can you add a comment saying that "an empty byte array will be unmarshalled into zero values without error" please? I just confirmed that it is true in a unit test. It's important because when we deploy it, the existing contract might not have any config set and we can't break because of that.

bolekk (approved):

bolekk (dismissed):

krehermann (commented):
i'm confused by this change. How is it going to be used? how does it relate to configuring the contract in a deployment? is there a relationship with this 

https://github.com/smartcontractkit/chainlink/blob/a9264ab9275052dd6a53b9d832753ea34b326a95/deployment/keystone/changeset/deploy_ocr3.go#L43

https://github.com/smartcontractkit/chainlink/blob/a9264ab9275052dd6a53b9d832753ea34b326a95/deployment/keystone/changeset/internal/ocr3config.go#L35

AFAICT, this change isn't used and doesn't do anything new. What am I missing?

@bolekk do you have something in your mind?

bolekk:
Yes, see https://github.com/smartcontractkit/chainlink/pull/16247

vyzaldysanchez:
For this to make sense, https://github.com/smartcontractkit/chainlink/pull/16247 needs to be taken a look at, since it is what uses this.

After this is merged, we are working on https://smartcontract-it.atlassian.net/browse/CRE-144 and https://smartcontract-it.atlassian.net/browse/CRE-143 which will also be required to have a core PR.

Then all this will end up having a C-D PR to complete the cycle.

pkg/capabilities/consensus/ocr3/ocr3.go
```diff
@@ -88,11 +75,16 @@ func NewOCR3(config Config) *Capability {
 	return cp
 }
 
+func (o *Capability) SetRequestTimeout(timeout time.Duration) {
+	o.config.RequestTimeout = &timeout
```
krehermann, line 79: this looks odd, why do we need to set in both?

pkg/capabilities/consensus/ocr3/types/ocr3_config_types.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+option go_package = "capabilities/consensus/ocr3/types";
+
+package ocr3_config_types;
+
+import "google/protobuf/duration.proto";
+
+message ReportingPluginConfig {
+  uint32 maxQueryLengthBytes = 1;
```
krehermann, line 10: can you add a comment that links to the libocr fields that these are shadowing?

pkg/capabilities/consensus/ocr3/ocr3.go
vyzaldysanchez: Initially, we wanted to pass `Capability` around. I just pushed it as is to demonstrate to @bolekk that we might not need to pass it around and just access `o.config.capability.setRequestTimeout()` instead.

So I'm fully expecting to just remove this setter altogether and the passing around of `Capability` from this PR.

bolekk: So you want to pass o.config.capability into newFactory()...? That's fine, you can just implement the final version that you think will work.

krehermann (approved):

cedric-cordenier (commented):
Has this been tested on staging and confirmed to work? We are getting close to cutting our release for PoR so a change like this is making me a bit hesitant.

vyzaldysanchez:
@cedric-cordenier we added tests in core with a view on the data, and also were able to generate the state file for staging [here](https://github.com/smartcontractkit/chainlink-deployments/pull/710/files#diff-67584d4732a5ba7fa9d5b6f8ec9654276a0811570ddf8f7487fc25192b983249) successfully.

cedric-cordenier:
Could we please deploy these changes to staging -- it looks like it actually changes the behaviour of the OCR3 cap.

cedric-cordenier (changes requested):
Please deploy to staging + verify that OCR3 is still working before merging.

vyzaldysanchez:
@cedric-cordenier I haven't tested on staging just yet, but I have added tests on CLD with all the dependencies updating validating the new fields added to the struct and how it gets loaded from the orc3config files: https://github.com/smartcontractkit/chainlink-deployments/pull/710/files#diff-7a610c88816e8596e843dc7d97f6713c5c81750a97a992e7a92d482d77bf8051R45

Please take a look.

cc: @krehermann

bolekk (approved):

cedric-cordenier (approved):

PR #1025, opened 2025-02-08
[CAPPL-308] Add non-data dependency to chain reader

cedric-cordenier:
Add a new field to the chainreader config, called NonDataDependency; this allows an SA to specify a step that should act as a dependency to the current chainreader step.

Without this, users have to resort to hacks like stuffing the dependent step ref in another field of the input


ettec:
> Add a new field to the chainreader config, called NonDataDependency; this allows an SA to specify a step that should act as a dependency to the current chainreader step.
> 
> Without this, users have to resort to hacks like stuffing the dependent step ref in another field of the input

As discussed, as we are moving to Nodag and the 'proper' fix to this in complex and involved, on balance it makes sense to do this less than ideal workaround for now.

ettec (approved):

george-dorin (approved):

PR #1028, opened 2025-02-12
pkg/services: add EngineRecv

jmank88:


pkg/services/service.go
```diff
@@ -107,6 +107,19 @@ func (e *Engine) GoTick(ticker *timeutil.Ticker, fn func(context.Context)) {
 	})
 }
 
+func EngineRecv[C any, CH ~<-chan C](e *Engine, ch CH, fn func(C)) {
```
jmank88, line 110: This is clunky to do without generic method parameters. In one or two more releases we'll be able to do:
```suggestion
func (e *Engine) GoRecv[C any, CH ~<-chan C](ch CH, fn func(C)) {
```

jmank88: Updated to 1.27rc2

PR #1029, opened 2025-02-13
[CAPPL-536] Workflow update not picked up by Syncer

agparadiso:


pkg/loop/internal/core/services/capability/capabilities.go
```diff
@@ -270,6 +292,17 @@ func (t *triggerExecutableClient) RegisterTrigger(ctx context.Context, req capab
 		return nil, fmt.Errorf("error registering trigger: %w", err)
 	}
 
+	// In order to ensure the registration is successful, we need to wait for the first message from the server.
+	// This will be an ack or error message. If the error is not nil, we return an error.
+	ackMsg, err := responseStream.Recv()
```
justinkaseman, line 297: Should we add a timeout on how long we will wait for this?

cedric-cordenier: We pass in the context when calling RegisterTrigger above; that will scope the stream as a whole

pkg/loop/internal/core/services/capability/capabilities.go
```diff
@@ -270,6 +292,17 @@ func (t *triggerExecutableClient) RegisterTrigger(ctx context.Context, req capab
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to receive registering trigger ack message: %w", err)
+	}
+
+	if ackMsg.GetResponse().GetError() != "" {
+		return nil, errors.New(fmt.Sprintf("failed registering trigger: %s", ackMsg.GetResponse().GetError()))
+	}
```
cedric-cordenier, line 304: @agparadiso This code is a bit ambiguous I think: what happens if the server sends a response without an ack? We should clearly error in that case as that would mean the protocol has been violated.

pkg/loop/internal/core/services/capability/capabilities.go
```diff
@@ -215,9 +215,31 @@ func (t *triggerExecutableServer) RegisterTrigger(request *capabilitiespb.Trigge
...
+			Message: &capabilitiespb.TriggerResponseMessage_Response{
+				Response: &capabilitiespb.TriggerResponse{
+					Error: err.Error(),
+				},
+			},
+		}
+		if err = server.Send(msg); err != nil {
+			return fmt.Errorf("failed sending error response for trigger registration %s: %w", request, err)
+		}
 		return fmt.Errorf("error registering trigger: %w", err)
```
cedric-cordenier, line 230: No need to return an error in this case, you already returned it back to the caller successfully, just return nil.

cedric-cordenier: Better yet, I think `return server.Send(msg)` is enough

pkg/loop/internal/core/services/capability/capabilities.go
agparadiso: changed to send: `TriggerResponseMessage_Ack` and fail `if ackMsg.GetAck() == nil` much cleaner, thanks🙇🏼

cedric-cordenier (dismissed):

pkg/loop/internal/core/services/capability/capabilities.go
agparadiso: true, done 👍🏼

justinkaseman (approved):

cedric-cordenier (approved):

PR #1030, opened 2025-02-13
Adding core.Keystore to LOOP Initialization

patrickhuie19:
Note that this PR originally introduced a CSASigner type. https://github.com/smartcontractkit/chainlink-common/pull/1050 ran into a similar need to hide a Ed25519 private key behind an interface to aid with the EVM extraction. We will pass a core.Keystore type around to LOOPs, but call it a CSASigner to make it explicit. 

### Supports
CsaAuth service in Beholder


pkg/loop/internal/pb/csa_signer.proto
jmank88, line 1: Do we need a new proto? This look the same as the existing Keystore passed in to relayers.

patrickhuie19: I debated that and chose this approach so:
 * it was clear to the reader that the aim of this is not for relayer keystore
 * we have the lightest weight grpc wrappers
 
 Do you see an angle for using the generated code for the relayers?

jmank88: Consider that the evm relayer will need to be passed a CSA keystore, alongside the evm `core.Keystore`. It would be unfortunate to have to duplicate all of the wrappers due to type differences that are really only names.

jmank88:
The CSA key used in the EVM relayer gets pass through to wsrpc for x509 cert generation, and the underlying sign methods uses the `crypto.SignerOpts` type. It is open ended but we may need to support a `uint` hash type, and `string` context for the signing on the other end to use.

pkg/loop/internal/pb/csa_signer.proto
patrickhuie19: They are identical now, but it seemed reasonable to me that they could be evolved independently. It's unlikely the the CSASigner interface changes, but that's because I know what LOOPs will need to authenticate Beholder. Is the keystore signer similar? Do we ever expect it to do more things than sign a message?

jmank88: Sorry I missed this.

I think we'll move towards this or similar in the future: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/loop/internal/keystore/types.go, so if we can get by with the existing interface for now, then we'll have fewer changes.

PR #1037, opened 2025-02-19
SRE-1635-Add-panel-overrides

leeyikjiun:
Example usage:
```go
Overrides: []*grafana.Override{
	{
		Matcher: grafana.NewByNameMatcher("error"),
		Properties: []*grafana.Property{
			grafana.NewColorProperty(&grafana.ColorPropertyOptions{
				Mode:       grafana.ColorModeFixed,
				FixedColor: "red",
			}),
		},
	}, {
		Matcher: grafana.NewByNameMatcher("info"),
		Properties: []*grafana.Property{
			grafana.NewColorProperty(&grafana.ColorPropertyOptions{
				Mode:       grafana.ColorModeFixed,
				FixedColor: "green",
			}),
		},
	},
},
```

### DIFF
--- a/observability-lib/grafana/panels.go
+++ b/observability-lib/grafana/panels.go
@@ -91,6 +91,25 @@ func newTransform(options *TransformOptions) dashboard.DataTransformerConfig {
 	}
 }
 
+func newOverride(override *Override) (matcher dashboard.MatcherConfig, properties []dashboard.DynamicConfigValue) {
+	matcher = dashboard.MatcherConfig{
+		Id:      override.Matcher.ID,
+		Options: override.Matcher.Options,
+	}
+
+	for _, property := range override.Properties {
+		properties = append(
+			properties,
+			dashboard.DynamicConfigValue{
+				Id:    property.ID,
+				Value: property.Value,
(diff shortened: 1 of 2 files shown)


observability-lib/grafana/panels.go
```diff
@@ -91,6 +91,25 @@ func newTransform(options *TransformOptions) dashboard.DataTransformerConfig {
...
+		properties = append(
+			properties,
+			dashboard.DynamicConfigValue{
+				Id:    property.ID,
+				Value: property.Value,
+			},
+		)
+	}
+
+	return
```
Atrax1, line 110: the function signature expects to return something here

Atrax1 (changes requested):

observability-lib/grafana/panels.go
leeyikjiun: It's a named return values. You don't have to explicitly write the return values here.

Atrax1: ah yes right naked return

chudilka1 (approved):

PR #1039, opened 2025-02-19
Add interface support for MedianPluginOption

samsondav:
### Supports

https://github.com/smartcontractkit/chainlink-feeds/pull/82
https://github.com/smartcontractkit/chainlink/pull/16438


patrickhuie19 (dismissed):

jmank88 (changes requested):
I think we should find a way to avoid this detail leaking through to this layer. It is meant to be 1-1 with the grpc services. Can the caller just do a type cast to look for the extended method instead?

patrickhuie19:
> I think we should find a way to avoid this detail leaking through to this layer. It is meant to be 1-1 with the grpc services. Can the caller just do a type cast to look for the extended method instead?

It seemed reasonable to me to support an opts like pattern to the MedianFactory. What would the pattern with the type cast look like?

pkg/types/core/median.go
```diff
@@ -2,15 +2,23 @@ package core
...
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
 )
 
+type DeviationFuncDefinition interface {
+	Func(lggr logger.Logger) median.DeviationFunc
+	json.Marshaler
+	json.Unmarshaler
+}
```
jmank88, line 18: I think this should be a pure message type. Different funcs can be called dynamically depending on what data is in the message, but merging the two as a single type creates complexity.

samsondav: `NewMedianFactory` already takes a bunch of interface args and stuff tho?

jmank88: Every interface requires a client/server pair. Is that really necessary in this case?

jmank88 (approved):

patrickhuie19 (approved):

PR #1050, opened 2025-02-26
pkg/types/core: add Ed25519Signer

jmank88:
[smartcontract-it.atlassian.net/browse/CRE-305](https://smartcontract-it.atlassian.net/browse/CRE-305)

Supports:
- https://github.com/smartcontractkit/chainlink/pull/16577

### DIFF
--- a/pkg/types/core/keystore.go
+++ b/pkg/types/core/keystore.go
@@ -1,10 +1,72 @@
 package core
 
-import "context"
+import (
+	"context"
+	"crypto"
+	"crypto/ed25519"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+)
 
 type Keystore interface {
 	Accounts(ctx context.Context) (accounts []string, err error)
 	// Sign returns data signed by account.
 	// nil data can be used as a no-op to check for account existence.
 	Sign(ctx context.Context, account string, data []byte) (signed []byte, err error)
+}
+
+var _ crypto.Signer = &Ed25519Signer{}
+
+type SignFn func(ctx context.Context, account string, data []byte) (signed []byte, err error)
+
+// Ed25519Signer implements crypto.Signer and services.StartClose.
+type Ed25519Signer struct {
+	stopCh  services.StopChan
+	account string
+	pubKey  crypto.PublicKey
+	signFn  func(ctx context.Context, account string, data []byte) (signed []byte, err error)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/types/core/keystore.go
```diff
@@ -1,10 +1,38 @@
...
+		ctx:     func() context.Context { return ctx },
+		account: account,
+		signFn:  signFn,
+	}
+}
+
+func (c *CryptoSigner) Public() crypto.PublicKey { return c.account }
+
+func (c *CryptoSigner) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
+	//TODO sanity check rand & opts?
```
krehermann, line 36: are you going to implement something more here?

pkg/types/core/keystore.go
```diff
@@ -1,10 +1,38 @@
...
 	Accounts(ctx context.Context) (accounts []string, err error)
 	// Sign returns data signed by account.
 	// nil data can be used as a no-op to check for account existence.
 	Sign(ctx context.Context, account string, data []byte) (signed []byte, err error)
 }
+
+var _ crypto.Signer = &CryptoSigner{}
+
+type CryptoSigner struct {
+	ctx     func() context.Context
```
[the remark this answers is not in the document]
jmank88: Because we are adapting a method with a context to a signature without one. We could plug it with `context.Background()`, but we want to yield for things like shutdown, rather than block on a retry loop in the grpc keystore client.

pkg/types/core/keystore.go
jmank88: It depends what is required to support our usage. I'd like to keep it as simple as possible, and hopefully not have to pass opts over the wire at all. But even in that case, we may want to be strict about validating which opts are passed in so there aren't any surprises.

pavel-raykov (approved):

pkg/types/core/keystore.go
jmank88: Converted to a `Ed25519Signer` since that is all we need for CSA, and made it stricter here, tolerating only `crypto/rand.Reader` and `crypto.Hash(0)` for `x509.PureEd25519`, and returning a `ed25519.PublicKey`.

jmank88: TLS x509 is the only usage currently, for Mercury/LLO.

pkg/types/core/keystore.go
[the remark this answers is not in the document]
jmank88: Modified to have a `StopChan` to make these on the fly instead, so the constructor is not confusing.

pavel-raykov (approved):

patrickhuie19 (approved):

PR #1057, opened 2025-03-07
[CAPPL-620] add MaxFetchResponseSizeBytes to module config

agparadiso:
### Description

This pr adds the MaxFetchResponseSizeBytes to the `ModuleConfig` 
[CAPPL-620](https://smartcontract-it.atlassian.net/browse/CAPPL-620)

### DIFF
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -78,6 +78,7 @@ var (
 	defaultMaxFetchRequests          = 5
 	defaultMaxCompressedBinarySize   = 20 * 1024 * 1024  // 20 MB
 	defaultMaxDecompressedBinarySize = 100 * 1024 * 1024 // 100 MB
+	defaultMaxFetchResponseSizeBytes = 5 * 1024 * 1024   // 5 MB
 )
 
 type DeterminismConfig struct {
@@ -96,6 +97,7 @@ type ModuleConfig struct {
 	MaxFetchRequests          int
 	MaxCompressedBinarySize   uint64
 	MaxDecompressedBinarySize uint64
+	MaxFetchResponseSizeBytes uint64
 
 	// Labeler is used to emit messages from the module.
 	Labeler custmsg.MessageEmitter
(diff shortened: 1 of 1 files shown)


cedric-cordenier (approved):

cedric-cordenier (changes requested):
Once it is on the moduleConfig, how do you pass it to the binary? The limit applies to the guest, not the host

agparadiso:
> Once it is on the moduleConfig, how do you pass it to the binary? The limit applies to the guest, not the host

this was already in place, what was missing was getting this value from the jobspec.
The parameter goes down to the sdk config and then is used:
- [createEmitFn](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/workflows/wasm/sdk.go#L125)
- [createFetchFn](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/workflows/wasm/sdk.go#L209)

pkg/workflows/wasm/sdk.go
```diff
@@ -149,6 +149,11 @@ func createEmitFn(
 
 		// Attempt to read and handle the response from the host memory
 		responseSize := binary.LittleEndian.Uint32(resplenBuffer)
+
+		if responseSize > uint32(sdkConfig.MaxFetchResponseSizeBytes) {
```
cedric-cordenier, line 153: @agparadiso Can we rename MaxFetchResponseSizeBytes to something more generic given that it's being used in createEmitFn as well as createFetchFn?

Have you also checked that we don't need to apply a similar limit to createLogFn? (I think not since it doesn't return anything)

agparadiso: >Have you also checked that we don't need to apply a similar limit to createLogFn? (I think not since it doesn't return anything)

I have the same reading, on fetch we where getting back the response payload, on emit we can get an error back, but on log its not returning anything

cedric-cordenier (approved):

justinkaseman (approved):

PR #1058, opened 2025-03-08
Improve wrapper and property extractor modifiers

ilija42:
- Change property extractor modifier to support extracting fields nested under slices by returning slice of slices of extracted type
- Change wrapper modifier to support wrapping variables outside a struct


pkg/codec/property_extractor.go
```diff
@@ -188,30 +189,65 @@ func (e *propertyExtractor) getPropTypeFromStruct(onChainType reflect.Type) (ref
...
+				prevIsSlice = true
+			}
+		} else if prevLocation.Type.Kind() == reflect.Slice {
+			prevIsSlice = true
+		}
+	}
+
 	curLocations.updateTypeFromSubkeyMods(fieldName)
-	field, ok := curLocations.fieldByName(fieldName)
+	field, ok = curLocations.fieldByName(fieldName)
```
amit-momin, line 222: nit: Couldn't we keep this the way it was? Doesn't seem like initializing at the top is being used for anything. Ignore if this is a linter issue.
```
var field *reflect.StructField
var ok bool
```

pkg/codec/property_extractor.go
```diff
@@ -344,7 +380,30 @@ func extractElement(src any, field string) (reflect.Value, error) {
 	}
 
 	if len(extractMaps) != 1 {
-		return reflect.Value{}, fmt.Errorf("%w: cannot find %s", types.ErrInvalidType, field)
+		var sliceValue reflect.Value
+		var sliceInitialized bool
+		for _, fields := range extractMaps {
+			val, ok := fields[name]
+			if !ok {
+				continue
```
amit-momin, line 388: I'm probably lacking context but just wanted to double check. It's ok if `sliceValue.Len() != len(extractMaps)`. If we're extracting from a slice of structs, wouldn't we want to populate `sliceValue` with nil for that index? If this doesn't affect anything upstream, you can ignore this.

pkg/codec/property_extractor.go
ilija42: fixed

pkg/codec/property_extractor.go
ilija42: SliceValue is being created from extract maps, I don't think that the length can end up being mismatched

amit-momin: Looks like we initialize it with length 0 below though `sliceValue = reflect.MakeSlice(sliceType, 0, 0)`. If we fail to find name in fields, couldn't we have fewer items in sliceValue?

ilija42: `sliceValue.Len() != len(extractMaps)` this doesn't mean much on its own, we only care about elements of the map that belong the slice that we are building out, extract maps can also contain fields with other name. For e.g.
```
typ E struct{
  F string
}

typ A struct{
     B []E
     C var
}
```

extract maps will have multiple entries of B and one C, we just want the slice elements of C

archseer (approved):

patrickhuie19 (approved):

PR #1062, opened 2025-03-13
Send Metering Report to Beholder

EasterTheBunny:
This commit adds support for sending a proto message beholder that defines its own schema detail apart from the default string `BaseMessage` type. For testing purposes, a new test utility is included to set the beholder global for testing and message delivery assertions.

To support metering, cost/unit details were added to existing messages and a new report message was created to send to beholder.

[CRE-283](https://smartcontract-it.atlassian.net/browse/CRE-283)

[CRE-283]: https://smartcontract-it.atlassian.net/browse/CRE-283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/beholder/client_test.go
```diff
@@ -1,4 +1,4 @@
-package beholder
+package beholder_test
```
Atrax1, line 1: 👍

pkg/custmsg/custom_message.go
```diff
@@ -12,7 +14,7 @@ import (
 
 type MessageEmitter interface {
 	// Emit sends a message to the labeler's destination.
-	Emit(context.Context, string) error
+	Emit(context.Context, any) error
```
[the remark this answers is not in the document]
EasterTheBunny: I think this goes against the high level goals of beholder. They should be able to read the proto and decode the message directly.

pkg/utils/tests/beholder.go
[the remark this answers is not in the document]
EasterTheBunny: I'm not sure I follow you. The goal I was after was a utility specifically for testing that accepts a `*testing.T` value and provides some useful assertions on the collected data. This way we can verify that specific messages have made their way through the beholder interface. Since the beholder team wants to do retries as well in the future, we will likely need to add error simulation as well.

pkg/utils/tests/beholder.go
```diff
@@ -0,0 +1,166 @@
...
+func (b BeholderTester) Len(t *testing.T, attrKVs ...any) int {
+	t.Helper()
+
+	found := b.msgsForKVs(t, attrKVs...)
+
+	return len(found)
+}
+
+// GreaterOrEqual asserts the total number of messages received that match the provided attribute key/value pairs.
+func (b BeholderTester) GreaterOrEqual(t *testing.T, length int, attrKVs ...any) bool {
```
patrickhuie19, line 40: I might consider removing this API - I think consumers can take this complexity on themselves

pkg/utils/tests/beholder.go
```diff
@@ -0,0 +1,166 @@
...
+	t.Helper()
+
+	found := b.msgsForKVs(t, attrKVs...)
+
+	return assert.GreaterOrEqual(t, len(found), length)
+}
+
+func (b BeholderTester) msgsForKVs(t *testing.T, attrKVs ...any) []beholder.Message {
```
patrickhuie19, line 48: /nit comment

Atrax1 (approved):

PR #1067, opened 2025-03-15
pkg/beholder: add NewAuthHeaders accepting crypto.Signer

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-339

Supports:
- https://github.com/smartcontractkit/chainlink/pull/16798

### DIFF
--- a/pkg/beholder/auth.go
+++ b/pkg/beholder/auth.go
@@ -1,7 +1,9 @@
 package beholder
 
 import (
+	"crypto"
 	"crypto/ed25519"
+	"crypto/rand"
 	"fmt"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/beholder/auth.go
```diff
@@ -17,11 +19,14 @@ var authHeaderVersion = "1"
 // <version>:<public_key_hex>:<signature_hex>
 //
 // where the byte value of <public_key_hex> is what's being signed
-func BuildAuthHeaders(privKey ed25519.PrivateKey) map[string]string {
-	pubKey := privKey.Public().(ed25519.PublicKey)
+func BuildAuthHeaders(ed25519Signer crypto.Signer) (map[string]string, error) {
```
jmank88, line 22: My original intent was to preserve existing calls that pass PKs, but there is an error to handle, so now the signature is broken. Should we introduce a new func instead? And just deprecate this one?

PR #1074, opened 2025-03-21
fix(observability-lib): notification policies provisioning

Atrax1:
- remove recursion on checking if policy already exists as children can be equals and only routes from parent matters

### DIFF
--- a/observability-lib/api/notification-policy.go
+++ b/observability-lib/api/notification-policy.go
@@ -56,9 +56,6 @@ func policyExist(parent alerting.NotificationPolicy, newNotificationPolicy alert
 		if matchersEqual && receiversEqual {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


observability-lib/api/notification-policy.go
jmank88, line 1: What is the effect of calling these if their return values are not used?

jmank88: Could do `_ = ...` to make it more explicit so a future reader doesn't question if it was an accident.

Atrax1: I removed it I actually only want to check parent routes, the children ones can be equals but still need to updated

jmank88 (approved):

EasterTheBunny (approved):

PR #1080, opened 2025-03-24
CRE-293 Add Metering Detail to Capabilities Response

EasterTheBunny:
Every node must independently report metering information while the engine expects an aggregated list of all nodes. Instead of having two different response types: one for a node and one as an aggregate, a `CapabilityResponse` can contain metadata with an array of values. A node is expected to send a single entity in the array while an aggregated response would have multiple.

This change gives the ability to surface metering data from 1 or many nodes without moving to more complex types.

[CRE-293](https://smartcontract-it.atlassian.net/browse/CRE-293)

### Supports

- https://github.com/smartcontractkit/chainlink/pull/16918

[CRE-293]: https://smartcontract-it.atlassian.net/browse/CRE-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/capabilities/pb/capabilities.proto
+++ b/pkg/capabilities/pb/capabilities.proto
@@ -79,10 +79,10 @@ service TriggerExecutable {
 message CapabilityResponse {
   values.Map value = 1;
   string error = 2;
-  CapabilityResponseMetadata metadata = 3;
+  ResponseMetadata metadata = 3;
 }
 
-message CapabilityResponseMetadata {
+message ResponseMetadata {
   repeated MeteringReportNodeDetail metering = 1;
 }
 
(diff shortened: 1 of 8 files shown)


pkg/capabilities/pb/capabilities.proto
```diff
@@ -79,6 +79,11 @@ service TriggerExecutable {
 message CapabilityResponse {
   values.Map value = 1;
   string error = 2;
+  CapabilityResponseMetadata metadata = 3;
+}
+
+message CapabilityResponseMetadata {
```
pavel-raykov, line 85: let's just call it ResponseMetadata to be in sync with other names like RequestMetadata

patrickhuie19 (approved):

pavel-raykov (approved):

PR #1081, opened 2025-03-25
Adding ByzantineQuorum utility

patrickhuie19:
Sets up the work to change the incorrect F + 1 aggregators in the Don 2 Don layer to a true Byzantine Quorum

### DIFF
--- a/pkg/aggregation/quorum.go
+++ b/pkg/aggregation/quorum.go
@@ -1,0 +1,16 @@
+package aggregation
+
+// ByzantineQuorum is a utility to calculated the number of responses from a
+// DON before Byzantine tolerant quorum is achieved.
+// NOTE: Typical usage is when N >= 3f + 1
+//
+// The formula is floor( (N+F)/2 ) + 1.
+// F should be >= 1.
+// If N = 1, a quorum size of 1 is returned.
+// Double check usage for small N and F if N < 3f + 1.
+func ByzantineQuorum(F int, N int) (Q int) {
+	if N == 1 {
+		return 1
+	}
+	return (N+F)/2 + 1
+}
--- a/pkg/aggregation/quorum.go
+++ b/pkg/aggregation/quorum.go
@@ -8,7 +8,7 @@ package aggregation
 // F should be >= 1.
 // If N = 1, a quorum size of 1 is returned.
 // Double check usage for small N and F if N < 3f + 1.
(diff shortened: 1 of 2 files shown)


bolekk (commented):
Can we use this? https://github.com/smartcontractkit/libocr/blob/master/quorumhelper/quorumhelper.go

pkg/aggregation/quorum.go
```diff
@@ -0,0 +1,16 @@
...
+
+// ByzantineQuorum is a utility to calculated the number of responses from a
+// DON before Byzantine tolerant quorum is achieved.
+// NOTE: Typical usage is when N >= 3f + 1
+//
+// The formula is floor( (N+F)/2 ) + 1.
+// F should be >= 1.
+// If N = 1, a quorum size of 1 is returned.
+// Double check usage for small N and F if N < 3f + 1.
+func ByzantineQuorum(F int, N int) (Q int) {
```
bolekk, line 11: Swap N and F.

patrickhuie19: 03917d5787c3e6b9dec471e9c8618d0af2cc41e5

patrickhuie19:
> Can we use this? https://github.com/smartcontractkit/libocr/blob/master/quorumhelper/quorumhelper.go

Name isn't great for our use case, and does not handle the case where N=1, F=1.

bolekk (approved):

vyzaldysanchez (approved):

pkg/aggregation/quorum.go
```diff
@@ -0,0 +1,16 @@
...
+
+// ByzantineQuorum is a utility to calculated the number of responses from a
+// DON before Byzantine tolerant quorum is achieved.
+// NOTE: Typical usage is when N >= 3f + 1
+//
+// The formula is floor( (N+F)/2 ) + 1.
+// F should be >= 1.
+// If N = 1, a quorum size of 1 is returned.
+// Double check usage for small N and F if N < 3f + 1.
+func ByzantineQuorum(N int, F int) (Q int) {
```
pavel-raykov, line 11: Is there any reason you decided not to import https://github.com/smartcontractkit/libocr/blob/f3b940c4f298f8652c03cd5cd0d0496eec93432a/internal/byzquorum/byzquorum.go#L1
and write an own function for that?

PR #1084, opened 2025-03-26
replay status

jmank88:


pkg/types/relayer.go
```diff
@@ -68,33 +69,30 @@ type MercuryCredentials struct {
...
+}
+type ReplayRequest struct {
+	FromBlock string
+	Args      map[string]any
+}
 
-func (rs ReplayStatus) String() string {
-	switch rs {
-	case ReplayStatusNoRequest:
+func (rs *ReplayStatus) Status() string {
```
jmank88, line 87: This should be removed. Just illustrating the equivalency.

PR #1085, opened 2025-03-26
removing global state for beholder tester

patrickhuie19:
* Subtests could accidentally share a beholder testing impl. Assertions on logs would behave unexpectedly as underlying store of messages could contain messages from other tests.

* Global vars set during testing remain in the process for the whole test run. i.e. `go test ./...` would set the beholder tester once. I believe that's why we're observing flakes: https://github.com/smartcontractkit/chainlink/pull/16923#issuecomment-2755512442

### DIFF
--- a/pkg/utils/tests/beholder.go
+++ b/pkg/utils/tests/beholder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"reflect"
 	"sync"
-	"sync/atomic"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -14,8 +13,6 @@ import (
 
 	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
 )
-
-var testEmitter atomic.Pointer[assertMessageEmitter]
 
 const (
 	packageNameBeholder = "test_beholder"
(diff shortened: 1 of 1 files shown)


pkg/utils/tests/beholder.go
```diff
@@ -105,7 +98,11 @@ func Beholder(t *testing.T) BeholderTester {
 		OnClose:               func() error { return nil },
 	}
 
-	testEmitter.Store(messageEmitter)
+	//reset Beholder state after the test
+	prevClient := beholder.GetClient()
+	t.Cleanup(func() {
+		beholder.SetClient(prevClient)
+	})
```
jmank88, line 105: This only works if the tests are not parallel, right? I don't think we can enforce that like [*testing.T.SetEnv](https://pkg.go.dev/testing#T.Setenv) does, but can we document it?

patrickhuie19: Yes, that's correct.

Agreed - 710e537c14ac1462725136ba6ded262c9529b85f

jmank88 (approved):

vyzaldysanchez (approved):

PR #1098, opened 2025-03-31
Add features to wrapper and property extractor modifiers

ilija42:
- Allow hardcoder to hardcode values for primitive variables when field name is unset in config.
- Improve property extractor to handle extractin uninitialised slices and nested slices


pkg/codec/property_extractor.go
```diff
@@ -452,3 +464,70 @@ func pathAndName(field string) ([]string, string) {
...
+
+		fieldByName, ok := typ.FieldByName(p)
+		if !ok {
+			return reflect.Value{}, fmt.Errorf("field %q not found in type %s", p, typ.Name())
+		}
+
+		fieldType := fieldByName.Type
+		for fieldType.Kind() == reflect.Ptr {
+			fieldType = fieldType.Elem()
+		}
```
amit-momin, line 498: nit: noticed we dereference like this in a few places. Could we move this logic into a small helper to reduce duplication?

ilija42: fixed

amit-momin (approved):

Farber98 (approved):

PR #1099, opened 2025-03-31
metering for consensus cap

patrickhuie19:
CRE-366 adds metering for the consensus capability of type `payload` in `bytes`. It calculates the wire-size of both the input and output data going over the wire through LibOCR and sums them. The pricing of these to universal CRE credits will be done later.

### DIFF
--- a/pkg/capabilities/consensus/ocr3/capability.go
+++ b/pkg/capabilities/consensus/ocr3/capability.go
@@ -232,7 +232,7 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 			Value: response.Value,
 			Metadata: capabilities.ResponseMetadata{
 				Metering: []capabilities.MeteringNodeDetail{
-					{SpendUnit: metering.PayloadMeteringUnit.Name, SpendValue: fmt.Sprintf("%d", inputLenBytes+outputLenBytes)},
+					{SpendUnit: metering.PayloadUnit.Name, SpendValue: fmt.Sprintf("%d", inputLenBytes+outputLenBytes)},
 				},
 			},
 		}, response.Err
--- a/pkg/capabilities/consensus/ocr3/capability.go
+++ b/pkg/capabilities/consensus/ocr3/capability.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/jonboulle/clockwork"
-	"google.golang.org/protobuf/proto"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
 	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/consensus/ocr3/requests"
@@ -242,7 +241,7 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 		if err != nil {
 			return capabilities.CapabilityResponse{}, err
 		}
-		inputLenBytes := proto.Size(values.Proto(r.Inputs))
+		inputLenBytes := values.ByteSizeOfMap(r.Inputs)
 
 		config, err := o.ValidateConfig(r.Config)
 		if err != nil {
--- a/pkg/capabilities/consensus/ocr3/capability.go
+++ b/pkg/capabilities/consensus/ocr3/capability.go
@@ -62,7 +62,6 @@ type capability struct {
 
 var _ CapabilityIface = (*capability)(nil)
 var _ capabilities.ConsensusCapability = (*capability)(nil)
-var _ capabilities.MeterableCapability = (*capability)(nil)
 
 func NewCapability(s *requests.Store, clock clockwork.Clock, requestTimeout time.Duration, aggregatorFactory types.AggregatorFactory, encoderFactory types.EncoderFactory, lggr logger.Logger,
 	callbackChannelBufferSize int) *capability {
@@ -259,7 +258,7 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 			Value: response.Value,
 			Metadata: capabilities.ResponseMetadata{
 				Metering: []capabilities.MeteringNodeDetail{
-					{SpendUnit: o.GetUnit().GetName(), SpendValue: fmt.Sprintf("%d", inputLenBytes+outputLenBytes)},
+					{SpendUnit: metering.NewPayloadMeteringUnit().Name, SpendValue: fmt.Sprintf("%d", inputLenBytes+outputLenBytes)},
 				},
 			},
 		}, response.Err
@@ -311,8 +310,4 @@ func (o *capability) queueRequestForProcessing(
 
 	o.reqHandler.SendRequest(ctx, r)
 	return callbackCh, nil
-}
-
-func (o *capability) GetUnit() metering.MeteringUnit {
-	return metering.NewPayloadMeteringUnit()
 }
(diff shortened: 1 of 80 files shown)


pkg/capabilities/consensus/ocr3/capability_test.go
```diff
@@ -140,6 +140,8 @@ func TestOCR3Capability(t *testing.T) {
 			assert.NoError(t, resp.Err)
 
 			assert.Equal(t, mresp, resp.Value)
+			assert.Equal(t, "payload", resp.Metadata.Metering[0].SpendUnit)
+			assert.Equal(t, "122", resp.Metadata.Metering[0].SpendValue)
```
patrickhuie19, line 144: Note: I didn't compute this by hand, or with some other source. I used the result from the test directly. It's a tautology for this pr, but moving forward it will serve as a regression test.

pkg/values/map.go
```diff
@@ -119,6 +120,18 @@ func (m *Map) DeleteAtPath(path string) bool {
 	return false
 }
 
+// ByteSizeOfMap is a utility to get the wire-size
+// of a values.Map. It should be the default utility
+// for metering/billing where the primary spend
+func ByteSizeOfMap(m *Map) int {
```
cedric-cordenier, line 126: This feels out of place -- you can get the proto by calling values.Proto(m); and from there get the size in the application code

pkg/capabilities/consensus/ocr3/capability.go
```diff
@@ -251,8 +254,14 @@ func (o *capability) Execute(ctx context.Context, r capabilities.CapabilityReque
 		}
 
 		response := <-ch
+		outputLenBytes := values.ByteSizeOfMap(response.Value)
```
cedric-cordenier, line 257: @patrickhuie19 Cohesion is the name of the game :) Can you reduce cognitive overhead by removing the interface and just inlining what you need here?

Your interface + then calling it could just become

```
Metadata: capabilities.ResponseMetadata{
				Metering: []capabilities.MeteringNodeDetail{
					{SpendUnit: metering.NewPayloadMeteringUnit(), SpendValue: fmt.Sprintf("%d", inputLenBytes+outputLenBytes)},				
```

pkg/values/map.go
patrickhuie19: The difference to the consumer of the Map is this:

```
		inputLenBytes := values.ByteSizeOfMap(r.Inputs)
		inputLenByes = proto.Size(values.proto(r.Inputs))
		
```

Its small, but I thought consumers would benefit from seeing what the size referred to in the values wrapper is (in-memory? string? wire-size?). I want the size over the wire, and `proto.Size` doesn't tell me that immediately.

pkg/metering/units.go
```diff
@@ -0,0 +1,56 @@
...
+	payloadName = "payload"
+	payloadUnit = "bytes"
+
+	computeName = "compute"
+	computeUnit = "seconds"
+)
+
+// MeteringUnit stores the canonical values of the units
+// that capabilities will meter their spend on
+type MeteringUnit interface {
```
cedric-cordenier, line 13: Interfaces should capture behaviour, and getters don't typically provide behaviour

What about if you just had a single struct to describe the concept of a metering unit, and then two methods to create two variants of those units

eg.

type MeteringUnit struct { ... }

NewComputeMeteringUnit() *MeteringUnit { return &MeteringUnit{computeName, computeUnit} }

+ same for payload

patrickhuie19: Following https://github.com/smartcontractkit/chainlink-common/pull/1099#discussion_r2021407662, https://github.com/smartcontractkit/chainlink-common/commit/90b73ca76e82d4c3796792f1684d3acff4630123

pkg/values/map.go
cedric-cordenier: Yeah -- I'm not disputing the value of this helper, I'm just questioning whether it should be here, as opposed to a private helper in the capability? I would prefer the latter until we find ourselves doing the same operation over and over again. Atm it's not clear that anything other than the OCR3 capability would use this.

patrickhuie19: a6389601468f24f445ee3d7420cc970d636aaba6

pkg/metering/units.go
jmank88, line 1: Are there any other example usages of these? It seems odd to construct something in order to access the const string values.

patrickhuie19: These will eventually be used for every capability that meters, but many of those capabilities are in-flight or planned for development.

jmank88: Are they expected to be mutated? Or are we trying to be defensive against mutation? i.e. what about something like:
```go
var (
	PayloadMeteringUnit = MeteringUnit{Name:"payload", Unit: "bytes"}
	ComputeMeteringUnit = MeteringUnit{Name:"compute", Unit: "seconds"}
)
```

pkg/metering/units.go
```diff
@@ -0,0 +1,25 @@
...
+
+const (
+	payloadName = "payload"
+	payloadUnit = "bytes"
+
+	computeName = "compute"
+	computeUnit = "seconds"
+)
+
+type MeteringUnit struct {
```
jmank88, line 11: The linter should be questioning the redundant name: `metering.MeteringUnit`. How about just `metering.Unit`?
```suggestion
type Unit struct {
```

patrickhuie19: 3edcee8188f82fb8907ce8e4eaaafc46e6875548

pkg/metering/units.go
patrickhuie19: The goal is to be defensive to mutation and to be as discoverable as possible. Thanks for the suggestion: 3edcee8188f82fb8907ce8e4eaaafc46e6875548

pkg/metering/units.go
```diff
@@ -0,0 +1,20 @@
+package metering
+
+var (
+	PayloadMeteringUnit = unit{Name: "payload", Unit: "bytes"}
+
+	// ComputeMeteringUnit is an example. The compute cap will eventually be obsoleted
+	// by the CRE No-DAG SDK.
+	ComputeMeteringUnit = unit{Name: "compute", Unit: "seconds"}
```
jmank88, line 8: I think we can drop `metering` from var names too for the same reason:
```suggestion
	PayloadUnit = unit{Name: "payload", Unit: "bytes"}

	// ComputeUnit is an example. The compute cap will eventually be obsoleted
	// by the CRE No-DAG SDK.
	ComputeUnit = unit{Name: "compute", Unit: "seconds"}
```

jmank88: Although, either way, we have redundant suffix and field: `Unit.Unit`. Is there a better name for one of them?

patrickhuie19: > I think we can drop metering from var names too for the same reason

Agreed, 78bed2a17c5a40e7fcefb9eb918ee4470ee022da

> Is there a better name for one of them?

`unit` --> `unitType`? `meteredType`? `meteredUnit`? `unitAndName`? 

The struct name `unit` is unexported so its trivial to update later if we can't think of something now, and is the better candidate for renaming since the inner `Unit` field is exactly what the name suggests and nothing more.

jmank88 (approved):

vyzaldysanchez (approved):

PR #1101, opened 2025-04-01
DEVSVCS-1554 add NewEventEmitter to sdk

shileiwill:
Update:
After discussing with Cedric, will remove the Logger and keep only the Emitter so workflow authors can only use the Emit() function which emits to Beholder. And the Emit() will not return an error any more.

Given Logger is removed, I removed the unit tests which are on Logger(). I had to silence some unit tests since the Emit workflow doesnt return error any more, please advise if this is appropriate.

------------------------------

Ticket: https://smartcontract-it.atlassian.net/browse/DEVSVCS-1554

When writing workflows, it is easy that users pick some common logging fucntions and assume logs would show up in Beholder. Some common ways are:
 - fmt.Print()
 - runtime.Logger()..
 
But the above will not show up in beholder logs hence not visible to users. The correct way is to 
- runtime.Emitter()

To avoid users making mistakes, we have been using the [WorkflowEventEmitter](https://github.com/smartcontractkit/por-workflow-hub-test/blob/main/workflows/lei-arb-mainnet/main.go#L17-L41) in template, but this is not ideal bc people have to copy/paste and carry it in the workflows. 

In this PR, I add the EventEmitter to SDK, so people can do the following in workflow:

```
			if resp.StatusCode != 200 {
				sdk.NewEventEmitter(rt).With("URL", PublicApiURL).Emit(fmt.Sprintf("unexpected HTTP status code: %d, error: %s", resp.StatusCode, resp.ErrorMessage))
				return aggregatorOutput{}, fmt.Errorf("unexpected HTTP status code: %d, error: %s", resp.StatusCode, resp.ErrorMessage)
			} else {
				sdk.NewEventEmitter(rt).With("URL", PublicApiURL).Emit(fmt.Sprintf("HTTP 200 status code: %d, error: %s", resp.StatusCode, resp.ErrorMessage))
			}
```


pkg/workflows/sdk/runtime.go
```diff
@@ -44,3 +44,33 @@ type FetchResponse struct {
...
+	return &EventEmitter{
+		runtime: runtime,
+	}
+}
+
+// EventEmitter provides a convenient way to emit messages and log events
+// in workflows.
+type EventEmitter struct {
+	runtime Runtime
+	labels  []string
```
patrickhuie19, line 60: could you embed a labeler (https://github.com/smartcontractkit/chainlink-common/blob/3c9893acc471de81923e27ca5f2a45628839d510/pkg/custmsg/custom_message.go#L51-L50) and take advantage of the functionality of adding labels there?

pkg/workflows/sdk/runtime.go
```diff
@@ -44,3 +44,33 @@ type FetchResponse struct {
 	Headers        map[string]any `json:"headers,omitempty"`      // HTTP headers
 	Body           []byte         `json:"body,omitempty"`         // HTTP response body
 }
+
+// NewEventEmitter creates a new event emitter that wraps the runtime's emitter
+// with additional functionality for easier logging as Beholder events.
+func NewEventEmitter(runtime Runtime) *EventEmitter {
```
cedric-cordenier, line 50: @shileiwill This is a good addition, but I would add it to the Runtime object itself as a first-class citizen, especially if our guidance is going to be for ST3 users to use this rather than what the Runtime object provides directly

shileiwill: The Runtime object has both Logger() and Emitter(). The problem this NewEventEmitter solves is that workflow authors dont know which logger to use, Logger() v.s. Emitter() or fmt.Print(). This `NewEventEmitter` uses Emitter() which emits to beholder, so we can advise workflow authors to only use this `NewEventEmitter`.

```
// Runtime exposes external system calls to workflow authors.
// - `Logger` can be used to log messages
// - `Emitter` can be used to send messages to beholder
// - `Fetch` can be used to make external HTTP calls
type Runtime interface {
	Logger() logger.Logger
	Fetch(req FetchRequest) (FetchResponse, error)

	// Emitter sends the given message and labels to the configured collector.
	Emitter() MessageEmitter
}
```

pkg/workflows/sdk/runtime.go
shileiwill: Hey @patrickhuie19 , i tried embedding the Labeler to this EventEmitter but I learnt it makes thing more complicated. The [Labeler](https://github.com/smartcontractkit/chainlink-common/blob/3c9893acc471de81923e27ca5f2a45628839d510/pkg/custmsg/custom_message.go#L51-L50) returns [MessageEmitter](https://github.com/smartcontractkit/chainlink-common/blob/3c9893acc471de81923e27ca5f2a45628839d510/pkg/custmsg/custom_message.go#L14) defined in this custmsg pkg while the runtime pkg has its own [definition](https://github.com/smartcontractkit/chainlink-common/blob/11e6c9259f652d7164daaa249f3770c819e8efd9/pkg/workflows/sdk/runtime.go#L11) of it.

Even if we reuse the Labeller, we can save a With() function, but workflow authors need to `NewEventEmitter().labeler.With()` which is not as straightforward as `NewEventEmitter().With()`

Sample code below:
```
func NewEventEmitter(runtime sdk.Runtime) *EventEmitter {
	return &EventEmitter{
		runtime: runtime,
		labeler: custmsg.NewLabeler(),
	}
}

type EventEmitter struct {
	labeler custmsg.Labeler
	runtime sdk.Runtime
}
```

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -293,7 +234,7 @@ func Test_Compute_Emit(t *testing.T) {
 		assert.Nil(t, err)
 	})
 
-	t.Run("failure on emit writes to error chain and logs", func(t *testing.T) {
+	t.Run("failure on emit does not throw error chain but logs the error", func(t *testing.T) {
```
MStreet3, line 237: this test hasn't been properly fixed, the mock is returning an error but the signature should change so that nothing is returned

pkg/workflows/sdk/runtime.go
```diff
@@ -2,27 +2,24 @@ package sdk
...
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 )
 
 // BreakErr can be used inside the compute capability function to stop the execution of the workflow.
 var BreakErr = capabilities.ErrStopExecution
 
 type MessageEmitter interface {
 	// Emit sends a message to the labeler's destination.
-	Emit(string) error
+	Emit(string)
```
MStreet3, line 12: is the `ModuleConfig.MessageEmitter` also changing? This is the interface that the host (e.g., core node) implements and provides to the wasm runtime.

pkg/workflows/wasm/sdk.go
```diff
@@ -73,8 +69,11 @@ func newWasmGuestEmitter(emitFn func(string, map[string]string) error) wasmGuest
 	}
 }
 
-func (w wasmGuestEmitter) Emit(msg string) error {
-	return w.emitFn(msg, w.labels)
+func (w wasmGuestEmitter) Emit(msg string) {
+	if err := w.emitFn(msg, w.labels); err != nil {
+		l, _ := logger.New()
+		l.Errorf("error emitting message: %s", err.Error())
```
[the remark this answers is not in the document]
MStreet3: I would inject the `*Runtime.logger` into the `wasmGuestEmitter` to log this rather than create a new logger.

MStreet3 (approved):

pkg/workflows/wasm/host/wasm_test.go
```diff
@@ -319,24 +260,28 @@ func Test_Compute_Emit(t *testing.T) {
...
+			{
+				Level:   zapcore.ErrorLevel,
+				Message: fmt.Sprintf("error emitting message* failed to create emission* %s", assert.AnError),
 			},
 		}
 		for i := range expectedEntries {
-			assert.Equal(t, expectedEntries[i].Log.Level, logs.AllUntimed()[i].Entry.Level)
-			assert.Equal(t, expectedEntries[i].Log.Message, logs.AllUntimed()[i].Entry.Message)
+			assert.Equal(t, expectedEntries[i].Level, logs.AllUntimed()[i].Entry.Level)
+			assert.Equal(t, expectedEntries[i].Message, logs.AllUntimed()[i].Entry.Message)
```
MStreet3, line 279:
```suggestion
			assert.Equal(t, expectedEntries[i].Message, logs.AllUntimed()[i].Entry.Message)
			assert.Contains(t, expectedEntries[i].Message, "error emitting message")
```

MStreet3 (approved):

justinkaseman (approved):

PR #1104, opened 2025-04-01
Fix values Wrap(v any) null Interface{} bug

cawthorne:
Null check to prevent calling Wrap on possible null value.

The existing nil check in the switch statement only catches untyped nil interface values, but not interface values holding typed nil pointers.


karen-stepanyan (dismissed):

mxiao-cll (approved):

samsondav (dismissed):

krehermann (changes requested):
Tests please.

i think you need a test here with a nil struct ptr [smartcontractkit/chainlink-common@d68a079/pkg/values/value_test.go#L154](https://github.com/smartcontractkit/chainlink-common/blob/d68a079c09d16d0b3391b4906d79c0470341e878/pkg/values/value_test.go#L154)

and another here with a nil struct pointer [smartcontractkit/chainlink-common@d68a079/pkg/values/value_test.go#L335](https://github.com/smartcontractkit/chainlink-common/blob/d68a079c09d16d0b3391b4906d79c0470341e878/pkg/values/value_test.go#L335)

cawthorne:
When I run the tests with `go test ./...` extra changes are generated in `capabilities/triggers/testdata/fixtures/ondemand/schema.json`

https://github.com/smartcontractkit/chainlink-common/pull/1104/commits/12a841e80ba154f72a03d6f387d485029f7226b7

This happens if I remove my changes too, and then run `go test ./...`.

Unsure if I should include these or not in the PR.

pkg/values/value.go
```diff
@@ -146,7 +146,10 @@ func Wrap(v any) (Value, error) {
 	case reflect.Struct:
 		return CreateMapFromStruct(v)
 	case reflect.Pointer:
-		// pointer can't be null or the switch statement above would catch it.
+		if val.IsNil() {
+			return nil, nil
+		}
+		// pointer can't be null or the statement above would catch it.
```
krehermann, line 152: rm this comment given your new change

krehermann (approved):

pkg/values/value_test.go
```diff
@@ -166,6 +177,27 @@ func Test_Value(t *testing.T) {
...
 		},
+		{
+			name: "nilStructPointer",
+			newValue: func() (any, Value, error) {
+				var v LLOStreamsTriggerEvent
+				var d1 = make([]byte, 4)
+				binary.BigEndian.PutUint32(d1, 11)
+				var d3 []byte
+				// This fails the equivalence check if uint32 is used for StreamID.
+				// TODO: Is this a blocker, do these need to pass type equivalnce too?
```
cedric-cordenier, line 188: @cawthorne Are you directing this question at me or someone else?

cedric-cordenier (approved):

pkg/values/value_test.go
cawthorne: @cedric-cordenier It was to anyone who is more familiar with this code than I.

cawthorne: My guess is that is is fine.

cedric-cordenier (approved):

PR #1106, opened 2025-04-01
INFOPLAT-2071 Beholder emits to chip ingress

hendoxc:
[INFOPLAT-2071](https://smartcontract-it.atlassian.net/browse/INFOPLAT-2071)

## What

- Adding support for beholder clients to optionally emit to `chip-ingress`. initial changes will support sending to both at the same time via `Emit`. And eventually migrate to only using chip ingress as a means to Emit custom events
- `ChipIngressEmitterEnabled` needs to be set to `true` along with the grpc endpoint to send to. If `true` the client will continue to send custom evens via the OTEL message emitter that is currently in use, but additionally wrap the payload in a Cloud Event and transmit it to `chip-ingress`

## Why

- Working towards Beholder emitting custom events exclusively through `chip-ingress`

[INFOPLAT-2071]: https://smartcontract-it.atlassian.net/browse/INFOPLAT-2071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/beholder/dual_source_emitter.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (d *DualSourceEmitter) Emit(ctx context.Context, body []byte, attrKVs ...any) error {
+
+	// Emit via OTLP first
+	if err := d.otelCollectorEmitter.Emit(ctx, body, attrKVs...); err != nil {
+		return err
+	}
+
+	if err := d.chipIngressEmitter.Emit(ctx, body, attrKVs...); err != nil {
+		return err
```
pkcll, line 31: Maybe log this error for the first iteration and make the `d.chipIngressEmitter.Emit` call async?

pkg/beholder/dual_source_emitter.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (d *DualSourceEmitter) Emit(ctx context.Context, body []byte, attrKVs ...any) error {
+
+	// Emit via OTLP first
+	if err := d.otelCollectorEmitter.Emit(ctx, body, attrKVs...); err != nil {
+		return err
+	}
+
+	if err := d.chipIngressEmitter.Emit(ctx, body, attrKVs...); err != nil {
```
pkcll, line 30:
```suggestion
     go () {
	if err := d.chipIngressEmitter.Emit(ctx, body, attrKVs...); err != nil {
	   log.Debugf("chip-ingress error: %v", err)
	 } 
      }()
```

pkcll (dismissed):

pkg/beholder/chip_ingress_emitter.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type ChipIngressEmitter struct {
+	client chipingress.ChipIngressClient
+}
+
+func NewChipIngressEmitter(client chipingress.ChipIngressClient) Emitter {
+
+	if client == nil {
+		panic("chip ingress emitter client cannot be nil")
```
patrickhuie19, line 18: We construct this client in loops and in the core node on startup, so we can't panic. Can we return an error instead?

pkg/beholder/chip_ingress_emitter.go
```diff
@@ -0,0 +1,97 @@
...
+	_, err = c.client.Publish(ctx, event)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ExtractSourceAndType extracts source domain and entity from the attributes
+func ExtractSourceAndType(attrKVs ...any) (string, string, error) {
```
patrickhuie19, line 64: this seems useful for the current custom message emitter as well? Can it be placed in a more reusable location in the beholder pkg?

pkg/beholder/client_test.go
```diff
@@ -174,132 +174,6 @@ func TestClient(t *testing.T) {
 	}
 }
 
-func TestEmitterMessageValidation(t *testing.T) {
```
patrickhuie19, line 177: While we still have both emitters, should we keep these tests?

pkg/beholder/dual_source_emitter.go
patrickhuie19, line 1: We can continue offline, but I think this dual source emitter has tradeoffs of when consumers take on complexity that I'd like to understand better

pkg/beholder/chip_ingress_emitter.go
pkcll: The way the code is currently written should not hit this code path but definitely it should not panic 
https://github.com/smartcontractkit/chainlink-common/pull/1106/files#diff-74ce7f3d5474f67956a5d9ba5779b478b715235a0d33e8826eb99b146664ae4dR210-R217

pkg/beholder/client_test.go
hendoxc: I moved those into `message_emitter_test.go`

pkcll (dismissed):

pkg/chipingress/pb/chip_ingress_write.proto
```diff
@@ -0,0 +1,36 @@
...
+
+import "github.com/cloudevents/sdk-go/binding/format/protobuf/v2/pb/cloudevent.proto";
+
+package pb;
+
+option go_package = "./pb";
+
+// ChipIngress service provides a way for senders to produce CloudEvents messages to Kafka
+service ChipIngress {
+    rpc Publish (io.cloudevents.v1.CloudEvent) returns (PublishResponse) {}
```
patrickhuie19, line 11: Should we wrap this type (`message CloudEvent{`) to allow us to pass custom fields/metadata?

pkg/beholder/dual_source_emitter.go
```diff
@@ -0,0 +1,51 @@
...
+	if err := d.otelCollectorEmitter.Emit(ctx, body, attrKVs...); err != nil {
+		return err
+	}
+
+	// Emit via chip ingress async
+	go func() {
+		if err := d.chipIngressEmitter.Emit(ctx, body, attrKVs...); err != nil {
+			// If the chip ingress emitter fails, we log the error but do not return it
+			// because we still want to send the data to the OTLP collector and not cause disruption
+			log.Printf("Error emitting message to chip ingress: %v", err)
```
patrickhuie19, line 46: Can we use the logger pkg in common?

patrickhuie19 (dismissed):

pkg/beholder/dual_source_emitter.go
```diff
@@ -0,0 +1,51 @@
...
+	"log"
+)
+
+// dualSourceEmitter emits both to chip ingress and to the otel collector
+// this is to help transition from sending custom messages via OTLP to instead use chip-ingress
+// we want to send to both during the transition period, then cutover to using
+// chipIngressEmitter only
+type DualSourceEmitter struct {
+	chipIngressEmitter   Emitter
+	otelCollectorEmitter Emitter
```
shileiwill, line 15: @patrickhuie19 @hendoxc : can you educate me on what are the differences btw those two?  Here is my understanding:

- ChipIngress is for the new CRE events, and it will land to ChipIngress then expose as Kafka topic.
- OtelCollector is the existing way, it goes to beholder, it is what we see in Grafana, e.g. https://grafana.ops.prod.cldev.sh/goto/FIr25GAHR?orgId=1 ?

pkg/beholder/dual_source_emitter.go
```diff
@@ -0,0 +1,51 @@
...
+	// Emit via OTLP first
+	if err := d.otelCollectorEmitter.Emit(ctx, body, attrKVs...); err != nil {
+		return err
+	}
+
+	// Emit via chip ingress async
+	go func() {
+		if err := d.chipIngressEmitter.Emit(ctx, body, attrKVs...); err != nil {
+			// If the chip ingress emitter fails, we log the error but do not return it
+			// because we still want to send the data to the OTLP collector and not cause disruption
```
shileiwill, line 45: Seems OtelCollectorEmitter has high priority. Why is that?

pkg/chipingress/pb/chip_ingress_write.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+import "github.com/cloudevents/sdk-go/binding/format/protobuf/v2/pb/cloudevent.proto";
+
+package pb;
+
+option go_package = "./pb";
+
+// ChipIngress service provides a way for senders to produce CloudEvents messages to Kafka
```
shileiwill, line 9: n00b, where is the kafka instance and kafka topic defined? 

is it the address in `NewChipIngressClient(address string, opts ...Opt)`?

shileiwill (commented):
LGTM. Very good reading for me to understand how we emit messages, thank you!

pkg/beholder/dual_source_emitter.go
hendoxc: - Yes. key difference is that Chip Ingress provides delivery guarantees, where as the current OTeL pipeline is fire and forget, since its telemetry.
- Yes.

pkg/beholder/dual_source_emitter.go
hendoxc: Purely for during the transition period.

pkg/chipingress/pb/chip_ingress_write.proto
hendoxc: This is determined on the server

pkg/chipingress/pb/chip_ingress_write.proto
hendoxc: Those should be added in the payload that the Cloud Event wraps

shileiwill (approved):

engnke (approved):

PR #1107, opened 2025-04-02
Extracted Generalized Aptos monitoring packages to common

silaslenihan:
### Supports

- https://github.com/smartcontractkit/chainlink-framework/pull/36
- https://github.com/smartcontractkit/chainlink/pull/17118
- https://github.com/smartcontractkit/chainlink-aptos/pull/75


dhaidashenko (dismissed):

dhaidashenko (approved):

pkg/beholder/monitor/beholder.go
patrickhuie19, line 1: The naming and some of the duplication of usage around Beholder is going to be confusing to new consumers. Until we take the time to de-duplicate a lot of this duplicated logic between the engine and aptos, can we stick this in a more domain specific package than `Beholder`? 

maybe `pkg/aptos/monitor`? 

cc @krebernisak

patrickhuie19 (changes requested):

pkg/beholder/monitor/beholder.go
silaslenihan: This logic is no longer specific to Aptos, and will be used for any chain family that utilizes the generalized WT, starting with EVM: https://github.com/smartcontractkit/chainlink/pull/17118

patrickhuie19: Cool, then `pkg/write-chain/monitor`? We usually don't want to bleed chain/product details into common, but we shouldn't ship these changes wholesale into top level packages w/o first consolidating all the utilities/wrappers around beholder + observability that exist in the engine + capabilities + aptos.

silaslenihan: ended up moving all the write-target specific beholder code to [chainlink-framework](https://github.com/smartcontractkit/chainlink-framework/pull/36)

pkg/beholder/utils/metadata.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"go.opentelemetry.io/otel/attribute"
+)
+
+const (
+	// WorkflowExecutionIDShortLen is the length of the short version of the WorkflowExecutionId (label)
+	WorkflowExecutionIDShortLen = 3 // first 3 characters, 16^3 = 4.096 possibilities (mid-high cardinality)
+)
+
+// TODO: Refactor as a proto referenced from the other proto files (telemetry messages)
```
jmank88, line 14: For now or later? If later, is there a ticket to link?

pkg/utils/retry/retry.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/google/uuid"
+	"github.com/jpillora/backoff"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+// CtxKeyTracingID is the context key for tracing ID
+type ctxKey string
+
+const CtxKeyTracingID ctxKey = "tracingID"
```
jmank88, line 17:
```suggestion
type ctxKey string

// CtxKeyTracingID is the context key for tracing ID
const CtxKeyTracingID ctxKey = "tracingID"
```

pkg/utils/retry/retry.go
```diff
@@ -0,0 +1,61 @@
...
+	numRetries := 0
+	for {
+		result, err := fn(ctx)
+		if err == nil {
+			return result, nil
+		}
+
+		wait := strategy.Duration()
+		message := fmt.Sprintf("Failed to execute function, retrying in %s ...", wait)
+		lggr.Warnw(message, "wait", wait, "numRetries", numRetries, "tracingID", tracingID, "err", err)
```
jmank88, line 46: I don't think we should be overloading the term `trace id`. What about retry ID?
```suggestion
		lggr.Warnw(message, "wait", wait, "numRetries", numRetries, "retryID", retryID, "err", err)
```

pkg/utils/retry/retry.go
```diff
@@ -0,0 +1,61 @@
...
+	// Generate a new tracing ID if not present, used to track retries
+	tracingID, ok := ctx.Value(CtxKeyTracingID).(string)
+	if !ok {
+		tracingID = uuid.New().String()
+		// Add the generated tracing ID to the context (as it was not already present)
+		ctx = context.WithValue(ctx, CtxKeyTracingID, tracingID)
+	}
+
+	// Track the number of retries
+	numRetries := 0
```
jmank88, line 37: You can ask the `backoff.Backoff` for the attempt: `strategy.Attempt()`

pkg/utils/retry/retry.go
jmank88, line 1: Can we come up with a different name for the package or the funcs? Only one or the other needs to say `retry` - not both.

pkg/beholder/utils/metadata.go
silaslenihan: This TODO was carried over from the Aptos repo, and thus I think is out of scope for this PR. cc: @krebernisak if you have more context

pkg/utils/retry/retry.go
silaslenihan: removed Retry from func names

pkg/beholder/utils/common.go
```diff
@@ -0,0 +1,98 @@
+package utils
```
jmank88, line 1: Why a new package? Can these just be part of `package beholder`?

silaslenihan: moved them into `package beholder`

pkg/utils/retry/retry.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/google/uuid"
+	"github.com/jpillora/backoff"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+type ctxKey string
+
+// CtxKeyRetryID is the context key for tracing ID
+const CtxKeyRetryID ctxKey = "retryID"
```
jmank88, line 17: Can probably drop `Retry` from here too :thinking: 
```suggestion
// CtxKey is the context key for retry ID
const CtxKey ctxKey = "retryID"
```
Actually, a helper for setting the value might be more useful, in order to make sure the correct type (string) is used. The current code would overwrite the value if it was the wrong type.
```suggestion
// CtxWithID sets the retry ID context key for logging.
func CtxWithID(ctx context.Context, retryID string) context.Context {
	return context.WithValue(ctx, ctxKey, retryID)
}
```

silaslenihan: refactored this.

pkg/utils/retry/retry.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/google/uuid"
+	"github.com/jpillora/backoff"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+type ctxKey string
+
+// CtxKeyRetryID is the context key for tracing ID
```
jmank88, line 16:
```suggestion
```
You lost the func comment too

pkg/beholder/monitor/beholder.go
krebernisak: I think that's ok to speed up the generalization work, and not block on aligning on common.

But there are definitely common stuff in this/that PR that we can extract in a follow up.

PR #1111, opened 2025-04-07
beholder client - custom msg - add support for source/type attribute

engnke:
For https://smartcontract-it.atlassian.net/browse/INFOPLAT-2072
For custom messages being emitted to chip ingress, we check for source/domain as attribute names as well as the legacy beholder_domain and beholder_entity.


pkg/beholder/chip_ingress_emitter.go
```diff
@@ -53,24 +53,26 @@ func ExtractSourceAndType(attrKVs ...any) (string, string, error) {
 			continue
 		}
 
-		if key == "beholder_domain" {
+		// TODO
+		// Retrieve source and type using either ChIP or legacy attribute names, prioritizing source/type
+		if key == "source" || key == "beholder_domain" && sourceDomain == "" {
```
hendoxc, line 57: `&& sourceDomain == ""`

maybe we should add an additional check/nested `if` `sourceDomain` isn't empty,   this means that the same key has been provided twice, in which case we should probably return an `error`

hendoxc, line 57: shouldn't we be prioritising checking `source` and `type` first  over `beholder_domain` & `beholder_entity` ?

engnke: it does both?

engnke: if both are provided for some reason with this check it goes with source/type

hendoxc (approved):

pkg/beholder/chip_ingress_emitter.go
```diff
@@ -53,24 +53,25 @@ func ExtractSourceAndType(attrKVs ...any) (string, string, error) {
 			continue
 		}
 
-		if key == "beholder_domain" {
+		// Retrieve source and type using either ChIP or legacy attribute names, prioritizing source/type
+		if key == "source" || (key == "beholder_domain" && sourceDomain == "") {
```
patrickhuie19, line 57: > && sourceDomain == ""

Could you help me understand why we need the 2nd part of this conditional? sourceDomain will always be empty until we populate it, right?

pkg/beholder/chip_ingress_emitter.go
```diff
@@ -53,24 +53,25 @@ func ExtractSourceAndType(attrKVs ...any) (string, string, error) {
...
 			if val, ok := attrKVs[i+1].(string); ok {
 				sourceDomain = val
 			}
 		}
-		if key == "beholder_entity" {
+		if key == "type" || (key == "beholder_entity" && entityType == "") {
```
patrickhuie19, line 62: same as comment above

PR #1116, opened 2025-04-08
feat(backoff): adds a backoff library to utils

MStreet3:
adds in a backoff library for retrying requests from [backoff/v5](https://github.com/cenkalti/backoff/tree/v5).  copied in to avoid extra dependency.


pkg/workflows/sdk/runtime.go
```diff
@@ -30,11 +30,13 @@ type Runtime interface {
...
-	Method    string            `json:"method,omitempty"`    // HTTP verb, defaults to GET.
-	Headers   map[string]string `json:"headers,omitempty"`   // HTTP headers, defaults to empty.
-	Body      []byte            `json:"body,omitempty"`      // HTTP request body
-	TimeoutMs uint32            `json:"timeoutMs,omitempty"` // Timeout in milliseconds
+	URL         string            `json:"url"`                   // URL to query, only http and https protocols are supported.
+	Method      string            `json:"method,omitempty"`      // HTTP verb, defaults to GET.
+	Headers     map[string]string `json:"headers,omitempty"`     // HTTP headers, defaults to empty.
+	Body        []byte            `json:"body,omitempty"`        // HTTP request body
+	TimeoutMs   uint32            `json:"timeoutMs,omitempty"`   // Timeout in milliseconds
+	WithRetries *bool             `json:"withRetries,omitempty"` // Set to true to enable retry logic for requests
```
justinkaseman, line 38: Could this be `retries uint` where the default is 0 which represents do not retry?

justinkaseman, line 38: Do these changes also need to be introduced to the protobuf?

cedric-cordenier: Both to the protobuf and to the host.FetchRequest model

pkg/workflows/sdk/runtime.go
```diff
@@ -30,11 +30,13 @@ type Runtime interface {
...
-	URL       string            `json:"url"`                 // URL to query, only http and https protocols are supported.
-	Method    string            `json:"method,omitempty"`    // HTTP verb, defaults to GET.
-	Headers   map[string]string `json:"headers,omitempty"`   // HTTP headers, defaults to empty.
-	Body      []byte            `json:"body,omitempty"`      // HTTP request body
-	TimeoutMs uint32            `json:"timeoutMs,omitempty"` // Timeout in milliseconds
+	URL         string            `json:"url"`                   // URL to query, only http and https protocols are supported.
+	Method      string            `json:"method,omitempty"`      // HTTP verb, defaults to GET.
+	Headers     map[string]string `json:"headers,omitempty"`     // HTTP headers, defaults to empty.
+	Body        []byte            `json:"body,omitempty"`        // HTTP request body
+	TimeoutMs   uint32            `json:"timeoutMs,omitempty"`   // Timeout in milliseconds
```
cedric-cordenier, line 37: Things we need to think about with this change:
- Does TimeoutMs apply to all the requests including timeout?
- How do we know how long to backoff for (in the server side)?

pkg/workflows/sdk/runtime.go
cedric-cordenier: I agree that there's some duplication here, I think we can get away with just having MaxRetries below rather than both WithRetries and MaxRetries

Also as Justin points out, we could use 0 rather than a nullable type

pkg/workflows/sdk/runtime.go
```diff
@@ -30,11 +30,13 @@ type Runtime interface {
...
+	WithRetries *bool             `json:"withRetries,omitempty"` // Set to true to enable retry logic for requests
+	MaxTries    *uint             `json:"maxTries,omitempty"`    // Max number of times to retry a request, 0 implies no retries
```
cedric-cordenier, line 39: Can we clarify whether this number includes the first attempt or not -- based on the comment I think not but it's good to be explicit about this.

pkg/utils/backoff/backoff.go
```diff
@@ -0,0 +1,66 @@
+// Package backoff implements backoff algorithms for retrying operations.
```
cedric-cordenier, line 1: 🤔 I'm not 100% wild about copying here to avoid the dependency -- it exists regardless of whether we copy it or not; the downside to the copy is that we now need to maintain it ourselves

justinkaseman (approved):

justinkaseman (dismissed):

shileiwill (dismissed):

pkg/utils/backoff/retry.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"context"
+	"time"
+)
+
+// DefaultMaxElapsedTime sets a default limit for the total retry duration.
+const DefaultMaxElapsedTime = 1 * time.Minute
+
+// Operation is a function that attempts an operation and may be retried.
+type Operation[T any] func() (T, error)
```
jmank88, line 12: Is this important to define as a separate named type? It requires callers to cast in some situations

pkg/utils/backoff/backoff.go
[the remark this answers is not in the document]
MStreet3: Yes, we had a chat on this in private channel.

```
https://github.com/jpillora/backoff - implements a struct that outputs time.Duration that follow an exponential backoff

https://github.com/cenkalti/backoff/tree/v5 - exposes a Retry API that accepts a context, a function to retry and a backoff policy definition.  Exposes several backoff policies (e.g., exponential, constant, zero) that can be used in the Retry directly.
```

Basically this included library takes the second step of exposing a function for consistently retrying any function of the type func() (T, error) while the first only gives you a counter that a developer can use however they want to do some operation.

I had three options:
- write some new util that relies on https://github.com/jpillora/backoff to create a new Retry function
- Do nothing, roll it all from scratch similar to [here](https://github.com/smartcontractkit/chainlink/pull/17009)
- use a version of `backoff/v5` slimmed down to my needs

I chose the third way.

pkg/utils/backoff/retry.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Retry attempts the operation until success, a permanent error, or backoff completion.
+// It ensures the operation is executed at least once.
+//
+// Returns the operation result or error if retries are exhausted or context is cancelled.
+func Retry[T any](ctx context.Context, operation Operation[T], opts ...RetryOption) (T, error) {
+	// Initialize default retry options.
+	args := &retryOptions{
+		BackOff:        NewExponentialBackOff(),
+		Timer:          &defaultTimer{},
```
jmank88, line 60: Why is there a separate timer? Can't we just use `time.After`? The standard pattern we use with jpillora/backoff is much less verbose.

pkg/utils/backoff/retry.go
```diff
@@ -0,0 +1,80 @@
...
+	return func(args *retryOptions) {
+		args.MaxTries = n
+	}
+}
+
+// Retry attempts the operation until success, a permanent error, or backoff completion.
+// It ensures the operation is executed at least once.
+//
+// Returns the operation result or error if retries are exhausted or context is cancelled.
+func Retry[T any](ctx context.Context, operation func() (T, error), opts ...RetryOption) (T, error) {
```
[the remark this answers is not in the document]
silaslenihan: https://github.com/smartcontractkit/chainlink-framework/blob/0647c811e8e34635171517e64571650c59402d6a/capabilities/writetarget/retry/retry.go#L1-L65

PR #1123, opened 2025-04-10
beholder: config options for chip-ingress

pkcll:
beholder: config options for chip-ingress


patrickhuie19 (approved):

pkg/loop/config.go
```diff
@@ -242,11 +242,15 @@ func (e *EnvConfig) parse() error {
 	// Chip Ingress
 	e.ChipIngressEnabled, err = getBool(envChipIngressEnabled)
 	if err != nil {
-		return fmt.Errorf("failed to parse %s: %w", envChipIngressEnabled, err)
+		e.ChipIngressEnabled = false
```
pkcll, line 245: Make it backwards compatible with chainlink core

patrickhuie19 (dismissed):

pkg/loop/config.go
```diff
@@ -43,6 +43,9 @@ const (
 	envTelemetryEmitterExportInterval     = "CL_TELEMETRY_EMITTER_EXPORT_INTERVAL"
 	envTelemetryEmitterExportMaxBatchSize = "CL_TELEMETRY_EMITTER_EXPORT_MAX_BATCH_SIZE"
 	envTelemetryEmitterMaxQueueSize       = "CL_TELEMETRY_EMITTER_MAX_QUEUE_SIZE"
+
+	envChipIngressEnabled  = "CL_CHIP_INGRESS_ENABLED"
+	envChipIngressEndpoint = "CL_CHIP_INGRESS_ENDPOINT"
```
jmank88, line 48: What if we removed the enabled field, and just considered a missing/empty endpoint to mean disabled? Is there some reason that we need to be able to both set the endpoint, but also disable it?

pkg/loop/config_test.go
```diff
@@ -137,6 +146,149 @@ func TestEnvConfig_parse(t *testing.T) {
...
+		{
+			name: "ChipIngress enabled with invalid boolean",
+			envVars: map[string]string{
+				envDatabaseURL:                    "postgres://user:password@localhost:5432/db",
+				envDatabaseIdleInTxSessionTimeout: "42s",
+				envDatabaseLockTimeout:            "8m",
+				envDatabaseQueryTimeout:           "7s",
+				envDatabaseLogSQL:                 "true",
+				envDatabaseMaxOpenConns:           "9999",
+				envDatabaseMaxIdleConns:           "8080",
```
jmank88, line 165: Do all of the unrelated fields need to be set? If so, some kind of helper would help with readability and to focus on the interesting fields.

pkg/loop/config.go
pkcll: Addressed here https://github.com/smartcontractkit/chainlink-common/pull/1123/commits/7d32c5d04c75792ed391c8cd2b5b6b5568ff8c4a

pkg/loop/config_test.go
pkcll: Removed https://github.com/smartcontractkit/chainlink-common/commit/7d32c5d04c75792ed391c8cd2b5b6b5568ff8c4a#diff-73511a683a798e90c7aacb1b52ed4af31e04d5b95bf785e7cdccb7818e6992dcL149-L291

jmank88 (approved):

patrickhuie19 (approved):

PR #1125, opened 2025-04-11
Migrate over loopinstall for installing loop plugins

chainchad:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,16 +1,19 @@
 # CODEOWNERS Best Practices
+
 # 1. Per Github docs: "Order is important; the last matching pattern takes the most precedence."
+
 # Please define less specific codeowner paths before more specific codeowner paths in order for the more specific rule to have priority
 
-* @smartcontractkit/foundations
+- @smartcontractkit/foundations
 
 /observability-lib/ @Atrax1
 
+/pkg/beholder/ @smartcontractkit/realtime
 /pkg/capabilities/ @smartcontractkit/keystone @smartcontractkit/capabilities-team
-/pkg/codec/ @smartcontractkit/bix-framework @smartcontractkit/foundations @nolag 
+/pkg/codec/ @smartcontractkit/bix-framework @smartcontractkit/foundations @nolag
+/pkg/loop/loopinstall @smartcontractkit/devex-cicd
 /pkg/monitoring/ @topliceanu @pedrovsg
+/pkg/types/ccip @smartcontractkit/ccip-offchain
+/pkg/types/ccipocr3 @smartcontractkit/ccip-offchain
 /pkg/values/ @smartcontractkit/keystone @smartcontractkit/capabilities-team
 /pkg/workflows/ @smartcontractkit/keystone
-/pkg/beholder/ @smartcontractkit/realtime
-/pkg/types/ccipocr3 @smartcontractkit/ccip-offchain
-/pkg/types/ccip @smartcontractkit/ccip-offchain
--- a/pkg/loop/loopinstall/README.md
+++ b/pkg/loop/loopinstall/README.md
@@ -1,0 +1,79 @@
+# loopinstall
+
+A tool for installing Chainlink LOOP plugins from YAML configuration files.
+
+## Installation
+
+To install from source:
+
+```bash
+go install github.com/smartcontractkit/chainlink-common/pkg/loop/loopinstall@latest
+```
+
+Or clone and install locally:
+
+```bash
+git clone https://github.com/smartcontractkit/chainlink-common.git
+cd chainlink-common
+go install ./pkg/loop/loopinstall
+```
+
+## Usage
+
+```bash
+loopinstall [options] <plugin-config-file> [<plugin-config-file>...]
+```
+
+## Configuration
+
+_(See `plugins.example.yaml` for a complete example configuration.)_
+
+Example configuration file structure:
+
+```yaml
+defaults:
+  goflags: "-ldflags='-s -w'" # Default Go build flags
+  goprivate: "github.com/myorg/myrepo,github.com/private-org/*" # Comma-separated list of private repo patterns
+
+plugins:
+  cosmos:
+    - name: "default"
+      enabled: true
+      moduleURI: "github.com/smartcontractkit/chainlink-cosmos"
+      gitRef: "f740e9ae54e79762991bdaf8ad6b50363261c056"
+      installPath: "github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/cmd/chainlink-cosmos"
+      additionalFiles:
+        - src: "/go/pkg/mod/github.com/!cosm!wasm/wasmvm@v*/internal/api/libwasmvm.*.so"
+          dest: "/usr/lib/"
+```
+
+## Private Repository Access
+
+To install plugins from private repositories:
+
+1. Set your GitHub token:
+
+   ```bash
+   export GITHUB_TOKEN=your_token_here
+   ```
+
+   > **Note**: You can use `GITHUB_TOKEN=$(gh auth token)` if you have the GitHub CLI installed.
+
+2. **Optional for CI/CD only**: Configure Git to use HTTPS with token authentication:
+
+   ```bash
+   git config --global url."https://${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
+   ```
+
+3. Configure GOPRIVATE using one of these methods:
+   - In your config file under `defaults`:
+     ```yaml
+     defaults:
+       goprivate: "github.com/myorg/*,github.com/private-org/*"
+     ```
+   - Or via environment variable:
+     ```bash
+     export GOPRIVATE=github.com/myorg/*,github.com/another-org/*
+     ```
+
+The `goprivate` setting supports glob patterns (e.g., `github.com/myorg/*`) and tells Go to bypass the public module proxy for these repositories.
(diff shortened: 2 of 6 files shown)


.github/CODEOWNERS
```diff
@@ -1,16 +1,19 @@
...
 
 * @smartcontractkit/foundations
 
 /observability-lib/ @Atrax1
 
+/pkg/beholder/ @smartcontractkit/realtime
 /pkg/capabilities/ @smartcontractkit/keystone @smartcontractkit/capabilities-team
-/pkg/codec/ @smartcontractkit/bix-framework @smartcontractkit/foundations @nolag 
+/pkg/codec/ @smartcontractkit/bix-framework @smartcontractkit/foundations @nolag
+/pkg/loop/loopinstall @smartcontractkit/devex-cicd
```
chainchad, line 14: Will get this team write access.

pkg/loop/loopinstall/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Configuration
+
+_(See `plugins.example.yaml` for a complete example configuration.)_
+
+Example configuration file structure:
+
+```yaml
+defaults:
+  goflags: "-ldflags='-s -w'" # Default Go build flags
```
jmank88, line 36: Why do we set these flags?
- `-s    disable symbol table`
- `-w    disable DWARF generation`

jmank88: Oh those were old docs. These seem important, but if we truly need `-s` then it looks like `-w` can be left off:
```
-s
	Omit the symbol table and debug information.
	Implies the -w flag, which can be negated with -w=0.
-w
	Omit the DWARF symbol table.
```

pkg/loop/loopinstall/main.go
```diff
@@ -0,0 +1,621 @@
...
+		if err != nil {
+			return "", fmt.Errorf("failed to get working directory: %w", err)
+		}
+		outputFile = filepath.Join(wd, outputFile)
+	}
+
+	return outputFile, nil
+}
+
+func main() {
```
jmank88, line 540: In the interest of making the flags more visible, WDYT about moving this to the top of the file? (or alternatively splitting the file up so `main.go` is only this func?)

pkg/loop/loopinstall/README.md
```diff
@@ -0,0 +1,75 @@
...
+  goprivate: "github.com/myorg/myrepo,github.com/private-org/*" # Comma-separated list of private repo patterns
+
+plugins:
+  cosmos:
+    - name: "default"
+      enabled: true
+      moduleURI: "github.com/smartcontractkit/chainlink-cosmos"
+      gitRef: "f740e9ae54e79762991bdaf8ad6b50363261c056"
```
krebernisak, line 44: nit: hash (to be consistent with Nix :)

pkg/loop/loopinstall/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```yaml
+defaults:
+  goflags: "-ldflags='-s -w'" # Default Go build flags
+  goprivate: "github.com/myorg/myrepo,github.com/private-org/*" # Comma-separated list of private repo patterns
+
+plugins:
+  cosmos:
+    - name: "default"
+      enabled: true
```
krebernisak, line 42: Not sure why this tool needs an `enabled` flag?

pkg/loop/loopinstall/main.go
```diff
@@ -0,0 +1,621 @@
...
+		if err := execCommand(cmd); err != nil {
+			return fmt.Errorf("failed to install plugin %s[%d]: %w", pluginType, pluginIdx, err)
+		}
+	}
+
+	return nil
+}
+
+// writeBuildManifest writes installation artifacts to the specified file
+func writeBuildManifest(tasks []PluginInstallTask, outputFile string) error {
```
krebernisak, line 283: What's the point of the output `BuildManifest` file - isn't this the same data declared on the input as build metadata?

pkg/loop/loopinstall/main.go
krebernisak: nit: What about also following this convention for app entrypoints: https://github.com/golang-standards/project-layout/tree/master/cmd

chainchad: > In the interest of making the flags more visible, WDYT about moving this to the top of the file? (or alternatively splitting the file up so main.go is only main&printHelp?)

Updated in 84e7036288665c64590d27987b22c04531010249.

> nit: What about also following this convention for app entrypoints: https://github.com/golang-standards/project-layout/tree/master/cmd

Good thoughts. Would it look like this then?

```
chainlink-common/
├── cmd/
│   └── loopinstall/
│       └── main.go                 # Entry point
├── pkg/
│   └── loop/
│       └── internal/
│           └── loopinstall/
│               ├── config.go
│               ├── installer.go
│               ├── ...
```

pkg/loop/loopinstall/README.md
chainchad: Technically git ref is more accurate :) I imagine we'd only want to install from hashes for releases (with a comment to indicate the `v...` version), so maybe on repos that use this for distributable releases we would fail if it's pointed to an e.g. branch? I think locally it might be nice just to use a branch though.

pkg/loop/loopinstall/main.go
chainchad: Moved into `pkg/loop/cmd/loopinstall` in a8f498dba69ad87cdf6fe67861e4e77bda0bb91a.

It sounds like @krebernisak was suggesting to have everything in `cmd/loopinstall` though?

jmank88: I think a root level cmd/ directory for a module like this will be confusing, compared to keeping the command co-located with the relevant package :shrug:

pkg/loop/loopinstall/plugins.example.yaml
```diff
@@ -0,0 +1,28 @@
+defaults:
+  # Can be overridden with CL_PLUGIN_GOFLAGS environment variable
+  goflags: "-ldflags='-s -w'"
+  goprivate: "github.com/smartcontractkit/example-private-repo"
```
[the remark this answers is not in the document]
jmank88: Hmm I'm not sure about this one :thinking:

pkg/loop/cmd/loopinstall/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Configuration
+
+_(See `plugins.example.yaml` for a complete example configuration.)_
+
+Example configuration file structure:
+
+```yaml
+defaults:
+  goflags: "-ldflags='-s'" # Default Go build flags
```
jmank88, line 36: Can we include a comment about why `-s` is used?

pkg/loop/cmd/loopinstall/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+   ```bash
+   git config --global url."https://${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
+   ```
+
+3. Configure GOPRIVATE using one of these methods:
+   - In your config file under `defaults`:
+     ```yaml
+     defaults:
+       goprivate: "github.com/myorg/*,github.com/private-org/*"
```
jmank88, line 67: If we expect CSV, would a list make more sense?

pkg/loop/loopinstall/plugins.example.yaml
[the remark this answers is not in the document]
chainchad: Removed in setting `defaults.goprivate` from the yaml in 073c02a91e2c95b269ab4b5933be05b2961f6f16. Now just the environment variable `GOPRIVATE` will affect this behavior.

pkg/loop/cmd/loopinstall/README.md
chainchad: Added in 2e35011cf7353bd950fb7ac696c41ca142873217

pkg/loop/cmd/loopinstall/README.md
chainchad: Updated this it 073c02a91e2c95b269ab4b5933be05b2961f6f16 to remove `defaults.goprivate` altogether.

cedric-cordenier (approved):

krehermann (approved):

krehermann (approved):

pkg/loop/cmd/loopinstall/main.go
```diff
@@ -0,0 +1,133 @@
...
+	fmt.Println("  defaults:")
+	fmt.Println("    goflags: \"\"     # Default Go build flags")
+	fmt.Println("")
+	fmt.Println("  plugins:")
+	fmt.Println("    plugin-type:")
+	fmt.Println("      - moduleURI: \"github.com/example/module\"")
+	fmt.Println("        gitRef: \"v1.0.0\"")
+	fmt.Println("        installPath: \"./cmd/example\"")
+	fmt.Println("        libs: [\"/path/to/libs/*.so\"]  # Optional library paths (can include glob patterns)")
+	fmt.Println("        # enabled: false  # Optional, defaults to true if omitted")
```
jmank88, line 132: nit/ this would b easier to read/edit quoted with `

chainchad: Will get a fix in 👍

chainchad: https://github.com/smartcontractkit/chainlink-common/pull/1138

PR #1129, opened 2025-04-11
(feat): Add mode quorum 'all' to reduce aggregator of OCR3 capability

justinkaseman:


pkg/capabilities/consensus/ocr3/aggregators/reduce_aggregator.go
```diff
@@ -37,11 +37,12 @@ const (
 	REPORT_FORMAT_ARRAY     = "array"
 	REPORT_FORMAT_VALUE     = "value"
 	MODE_QUORUM_OCR         = "ocr"
+	MODE_QUORUM_ALL         = "all"
 	MODE_QUORUM_ANY         = "any"
 
 	DEFAULT_REPORT_FORMAT     = REPORT_FORMAT_MAP
 	DEFAULT_OUTPUT_FIELD_NAME = "Reports"
-	DEFAULT_MODE_QUORUM       = MODE_QUORUM_ANY
```
justinkaseman, line 44: Was unused, changing to correct default mode and used on line 595

PR #1131, opened 2025-04-15
Rtinianov local nodag merged triggers actions

MStreet3:


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/bytecodealliance/wasmtime-go/v28 v28.0.0
 	github.com/confluentinc/confluent-kafka-go/v2 v2.3.0
 	github.com/dominikbraun/graph v0.23.0
+	github.com/ethereum/go-ethereum v1.13.8
```
jmank88, line 12: We can't pollute this module with chain-specific imports. If there is some reason to introduce chain specifics to this repo, then it must be in separate modules.

nolag: This was for the example workflow. The branch isn't meant to be checked in, it's demonstrative for how to use the SDK.

PR #1133, opened 2025-04-15
Capability Error Reporting

ettec:
Common changes required to ensure capabilities can report meaningful errors back to the workflow/beholder https://smartcontract-it.atlassian.net/browse/CAPPL-648

3 Parts to the change:

1) Flatten the contract reader into the `relayer` returned from the `relayerset` so that the returned `contractreader` is no longer a loop proxy, this is consistent with the relayer implementation and resolves an issue with error reporting when creating a new contract reader.

2) Add a `RemoteReportableError` that can be used in capability code to indicate that the details of an error that occurs in a  remote capability can be safely reported back to the requesting nodes (ie workflow nodes) and update the capability loop code to correctly handle this error type.

3) A small update to the BeholderTester to support better querying of beholder logs in test code.

Note, there are corresponding core (https://github.com/smartcontractkit/chainlink/pull/17282) and capability changes including an integration test to verify the error reported in an end to end flow that requires all changes in this PR.


cedric-cordenier (dismissed):

pkg/capabilities/remote_reportable_error.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+const remoteReportableErrorIdentifier = "RemoteReportableError:"
+
+// RemoteReportableError wraps an error an is used to indicates that the error does contain any node specific
+// information and is safe to report remotely between nodes.
```
bolekk, line 11: nit: maybe add that error type is identified by the special prefix above

bolekk (dismissed):

george-dorin (approved):

pavel-raykov (approved):

PR #1134, opened 2025-04-15
Add get transaction fee api for ChainWriter

Unheilbar:
solana ref: 3c3afd1c423d4e111883400c16f00884aad623cd
core ref: 00c74213d692e7bf15e28ea0a23a18d706da0c50

Context:
For future billing needs for CRE  Chain Writer capability will need to provide transaction cost on the network. This batch of PRs extends Receipt model in database with a field of EffectiveGasPrice and extends TxStore interface with Get method for getting this receipt.

[JIRA](https://smartcontract-it.atlassian.net/browse/BCFR-1235)

[Core PR](https://github.com/smartcontractkit/chainlink/pull/17243)
[solana PR](https://github.com/smartcontractkit/chainlink-solana/pull/1189)

[chainlink-evm PR](https://github.com/smartcontractkit/chainlink-evm/pull/25)
[chainlink-framework PR](https://github.com/smartcontractkit/chainlink-framework/pull/39)


pkg/loop/internal/pb/contract_writer.proto
```diff
@@ -34,6 +35,18 @@ message GetTransactionStatusRequest {
...
 
+// GetTransactionFeeRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ContractWriter.GetTransactionFee].
+message GetTransactionFeeRequest {
+    string transaction_id = 1;
+}
+
+// GetTransactionFeeReply has return arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ContractWriter.GetTransactionFee].
+message GetTransactionFeeReply {
+    TransactionStatus transaction_status = 1;
+    BigInt transation_fee = 2;
```
patrickhuie19, line 46: /nit pls add comment on the units here. I'm assuming its gas units because the type is BigInt, but from experience it's aspirational to expect consumers of this API to understand that

pkg/types/contract_writer.go
```diff
@@ -49,6 +52,11 @@ const (
 	Fatal
 )
 
+type TransactionFee struct {
+	TransactionFee    *big.Int
+	TransactionStatus TransactionStatus
```
patrickhuie19, line 57: should this be a pointer too?

pkg/loop/internal/relayer/pluginprovider/contractwriter/contract_writer.go
```diff
@@ -143,6 +155,18 @@ func (s *Server) GetTransactionStatus(ctx context.Context, req *pb.GetTransactio
...
 }
 
+func (s *Server) GetTransactionFee(ctx context.Context, req *pb.GetTransactionFeeRequest) (*pb.GetTransactionFeeReply, error) {
+	reply, err := s.impl.GetTransactionFee(ctx, req.TransactionId)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pb.GetTransactionFeeReply{
+		TransationFee:     pb.NewBigIntFromInt(reply.TransactionFee),
```
patrickhuie19, line 165: What are the implications of this int --> BigInt conversion here? Do we need the BigInt type if we can deserialize the value as an int once it has crossed the wire?

pkg/loop/internal/pb/contract_writer.proto
Unheilbar: transaction fee is execution cost in native currency. Sure ill put comment there

pkg/loop/internal/relayer/pluginprovider/contractwriter/contract_writer.go
Unheilbar: its not int -> bigInt conversion its bigInt conversion from math Big to protobuf big. The same as here:
```
go func (s *Server) GetFeeComponents(ctx context.Context, _ *emptypb.Empty) (*pb.GetFeeComponentsReply, error) {
	feeComponents, err := s.impl.GetFeeComponents(ctx)
	if err != nil {
		return nil, err
	}
	return &pb.GetFeeComponentsReply{
		ExecutionFee:        pb.NewBigIntFromInt(feeComponents.ExecutionFee),
		DataAvailabilityFee: pb.NewBigIntFromInt(feeComponents.DataAvailabilityFee),
	}, nil
}
```

pkg/types/contract_writer.go
Unheilbar: wdym? pointer for TransactionFee value?

patrickhuie19: TransactionStatus

pkg/types/contract_writer.go
```diff
@@ -25,6 +25,9 @@ type ContractWriter interface {
 
 	// GetFeeComponents retrieves the associated gas costs for executing a transaction.
 	GetFeeComponents(ctx context.Context) (*ChainFeeComponents, error)
+
+	// GetTransactionFee retrieves the fee of a transaction in the underlying chain's TXM
+	GetTransactionFee(ctx context.Context, transactionID string) (*TransactionFee, error)
```
patrickhuie19, line 30: You will return an error if the status is not complete?

patrickhuie19 (approved):

pkg/types/contract_writer.go
Unheilbar: nope, an error will be returned in case if transaction not found by idempotency key, or if transactions status is failed/fatal. The same logic as in GetTransactionStatus call

pkg/types/contract_writer.go
Unheilbar: Oh, I don't think it it should. It's an enum so zero value is status unknown

pkg/types/contract_writer.go
Unheilbar: Actually, lmk if you prefer to get transaction status in GetransactionFee reply, or I can remove it and return transactionFee only if status is complete and an error otherwise

patrickhuie19: I think we can separate the concerns. The consumer using these APIs will have the knowledge to understand what a transaction status means for a particular tx. It's reasonable to expect that consumer to first poll for a finalized status, and then get the fee.

PR #1137, opened 2025-04-17
pkg/utils: add CL_RUN_FLAKEY to run skipped flakey tests

jmank88:


pkg/utils/tests/tests.go
```diff
@@ -77,5 +78,7 @@ func SkipShort(tb testing.TB, why string) {
 }
 
 func SkipFlakey(t *testing.T, ticketURL string) {
-	t.Skip("Flakey", ticketURL)
+	if _, ok := os.LookupEnv("CL_RUN_FLAKEY"); !ok {
```
jmank88, line 81: Any alternative names? Or should it be an exported var that an imported could override?
```suggestion
	if _, ok := os.LookupEnv(RunFlakeyEnv); !ok {
```
```go
var RunFlakeyEnv = "CL_RUN_FLAKEY"
```

Right now it accepts any value. Should we require it to be `true`? (or case insensitive true?) Or is there some other way to parameterize this? I'm thinking we might have more options than just skip/run in the future, but I'm not sure exactly what they would be :thinking:

jmank88: Now it uses https://pkg.go.dev/strconv#ParseBool, which is reasonable permissive with casing, integers, and single chars.

Atrax1 (approved):

PR #1140, opened 2025-04-17
Evm chain service

ilija42:


pkg/loop/internal/pb/evm_chain_service.proto
```diff
@@ -0,0 +1,24 @@
...
+  bytes encoded_params = 2;
+}
+
+// Example
+message ReadContractReply {
+  bytes result = 1;
+}
+
+// EVM chain service
+service EVMChainService {
```
jmank88, line 21: Is `Service` important to say? These are all services :shrug: 
```suggestion
service EVMChain {
```

ilija42: agree

pkg/types/relayer.go
```diff
@@ -99,6 +106,8 @@ type ChainService interface {
 type Relayer interface {
 	ChainService
 
+	NewEVMChain(ctx context.Context) (EVMChain, error)
```
jmank88, line 109: Do we need to construct instances of these? Or could the Relayer have the methods embedded? 
```suggestion
	EVMChain
```

PR #1142, opened 2025-04-17
Add GettransactionFee to EVM chain service

Unheilbar:
[JIRA](https://smartcontract-it.atlassian.net/browse/BCFR-1240)
[solana PR](https://github.com/smartcontractkit/chainlink-solana/pull/1196)
[core PR](https://github.com/smartcontractkit/chainlink/pull/17336)
[chainlink-framework PR](https://github.com/smartcontractkit/chainlink-framework/pull/39)
[chainlink-evm PR](https://github.com/smartcontractkit/chainlink-evm/pull/25)

solana ref: 66752a5dffb7bdf91d7eaa61f28e97fe8b44118b
core ref: b715d575b344fc3cda6de006c68104d658b5cdd2
starknet ref: 342e8571b8845a577d5eb0259eaef65ccc490d8e


pkg/loop/internal/relayer/chains/evm_chain.go
```diff
@@ -0,0 +1,130 @@
+package chains
```
patrickhuie19, line 1: if this package name is `chains` will other chains soon conflict with the exported `Client` here? 

Rename pkg to `chains-evm`?

pkg/types/contract_writer.go
```diff
@@ -38,7 +38,7 @@ type TxMeta struct {
 }
 
 // TransactionStatus are the status we expect every TXM to support and that can be returned by StatusForUUID.
-type TransactionStatus int
+type TransactionStatus int32
```
patrickhuie19, line 41: Will this break any consumers?

Unheilbar: consumers get it through pb where transaction status is int32.  I changed to make linter happy it reported potential overflow (int -> int32)

pkg/loop/internal/relayer/chains/evm_chain.go
Unheilbar: Lets move it on demand when we add new chain family

patrickhuie19: If we envision changing it, better to do so now imo. Current status of builds across repos is fairly stable, I recommend we take advantage of it while we have it.

pkg/loop/internal/pb/evm_chain.proto
```diff
@@ -0,0 +1,40 @@
...
+
+
+// GetTransactionFeeRequest has arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ContractWriter.GetTransactionFee].
+message GetTransactionFeeRequest {
+    string transaction_id = 1; // idempotency key used when transaction was submitted
+}
+
+// GetTransactionFeeReply has return arguments for [github.com/smartcontractkit/chainlink-common/pkg/types.ContractWriter.GetTransactionFee].
+message GetTransactionFeeReply {
+    TransactionStatus transaction_status = 1;
```
silaslenihan, line 36: Why do we bundle TransactionStatus into this method?

pkg/loop/internal/pb/relayerset/relayerset.proto
```diff
@@ -58,6 +58,14 @@ message NewPluginProviderResponse {
   uint32 plugin_provider_id = 1;
 }
 
+message NewEVMChainRequest {
+  RelayerId relayerId = 1;
```
silaslenihan, line 62: nit: Some linters complain about Id instead of ID
```suggestion
  RelayerID relayerID = 1;
```

pkg/loop/internal/relayerset/client.go
```diff
@@ -140,6 +140,17 @@ func (k *Client) NewPluginProvider(ctx context.Context, relayID types.RelayID, r
 	return resp.PluginProviderId, nil
 }
 
+func (k *Client) NewEVMChain(ctx context.Context, relayID types.RelayID) (uint32, error) {
+	req := &relayerset.NewEVMChainRequest{
+		RelayerId: &relayerset.RelayerId{ChainId: relayID.ChainID, Network: relayID.Network},
+	}
+	resp, err := k.relayerSetClient.NewEVMChain(ctx, req)
+	if err != nil {
+		return 0, fmt.Errorf("error getting new aptos chain service: %w", err)
```
silaslenihan, line 149: why aptos?

pkg/types/relayer.go
```diff
@@ -78,6 +80,14 @@ type NodeStatus struct {
 	State   string
 }
 
+type EVMChain interface {
+	services.Service
+	ReadContract(ctx context.Context, method string, encodedParams []byte) ([]byte, error)
+
+	// GetTransactionFee retrieves the fee of a transaction in the underlying chain's TXM
```
silaslenihan, line 87: nit
```suggestion
	// GetTransactionFee retrieves the fee and transaction status of a transaction in the underlying chain's TXM
```

silaslenihan (dismissed):
left a few nits, otherwise LGTM

silaslenihan (approved):

pkg/loop/internal/pb/evm_chain.proto
```diff
@@ -0,0 +1,38 @@
...
+
+package loop;
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb";
+
+import "contract_writer.proto";
+import "relayer.proto";
+
+// Example
+message ReadContractRequest {
```
patrickhuie19, line 11: /nit ReadContract has been axed from this PR - is there some value in leaving the Request/Reply's behind or can we axe these too?

patrickhuie19 (approved):

silaslenihan (dismissed):

patrickhuie19 (approved):

amit-momin (approved):

pkg/types/relayer.go
```diff
@@ -78,6 +80,13 @@ type NodeStatus struct {
 	State   string
 }
 
+type EVMChain interface {
+	services.Service
+
+	// GetTransactionFee retrieves the fee of a transaction in the underlying chain's TXM
```
ilija42, line 86: Can we elaborate a bit on what is this fee? For e.g. denoted in wei (18 decimals) Base Fee + Priority fee

ilija42: also add a comment on the fee structure itself, also why is the fee in a struct when its just a big int value?

silaslenihan (approved):

pkg/types/relayer.go
Unheilbar: Wrapper is just safer long term in case we decide to return something in addition to fee

pkg/loop/internal/pb/relayer.proto
```diff
@@ -71,6 +72,11 @@ message PluginArgs {
   bytes pluginConfig = 2;
 }
 
+// NewEVMChain has reply parameters for [github.com/smartcontractkit/chainlink-common/pkg/loop.Relayer.NewEVMChain].
+message NewEVMChainReply {
+  uint32 EVMChainID = 1;
```
jmank88, line 77: We shouldn't be overloading this name

jmank88: Well if we call this an `EVMRelayer` instead of a `Chain`, then the problem goes away.

jmank88: But I think avoiding the extra indirection and client/server pair should be considered too. We don't really need a new thing - we just want to use a specialized subset of services

PR #1145, opened 2025-04-17
feat(retry): adds a permanent error

MStreet3:
defines a PermanentError that indicates that a retry strategy should stop


pkg/utils/retry/error.go
```diff
@@ -0,0 +1,26 @@
+package retry
+
+// PermanentError signals that the operation should not be retried.
+type PermanentError struct {
+	Err error
+}
+
+// Permanent wraps the given err in a *PermanentError.
+func Permanent(err error) error {
```
cedric-cordenier, line 9: Semantically I think it's confusing that this returns nil if the error is nil; are we ever likely to call this with a nil error? (I would imagine not since you'll only be wrapping something you know to be an error)

MStreet3: moved the nil check to before String

pkg/utils/retry/retry.go
```diff
@@ -65,6 +66,12 @@ func (s *Strategy[R]) Do(ctx context.Context, lggr logger.Logger, fn func(ctx co
 			return result, nil
 		}
 
+		// Handle permanent errors without retrying.
+		var permanent *PermanentError
+		if errors.As(err, &permanent) {
+			return result, err
```
jmank88, line 72: Oh should it be unwrapped?
```suggestion
		var permanent *PermanentError
		if errors.As(err, &permanent) {
			return result, permanent.Unwrap()
```

jmank88 (approved):

cedric-cordenier (approved):

PR #1146, opened 2025-04-17
[DNM] Go SDK examples

nolag:
To see the test only, diff against rtinianov_feature_sdk_vf.


Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: gomods
 gomods: ## Install gomods
-	go install github.com/jmank88/gomods@v0.1.5
+	go install github.com/jmank88/gomods@v0.1.3
```
jmank88, line 3: Accidental downgrade?

jmank88: Oh the whole file actually

pkg/workflows/sdk/v2/promise.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+type Promise[T any] interface {
+	Await() (T, error)
+	CapabilityPromise
+}
+
+type CapabilityPromise interface {
+	promise()
+}
```
jmank88, line 14: Why is it important to prevent user created `Promise`s?

If it is not strictly necessary, then this sort of flexibility would be nice:
```suggestion
type Promise[T any] func() (T, error)
```

pkg/workflows/sdk/v2/promise.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewBasicPromise[T any](await func() (T, error)) Promise[T] {
+	return &basicPromise[T]{await: await}
+}
+
+func PromiseFromResult[T any](result T, err error) Promise[T] {
+	return &basicPromise[T]{resolved: result, err: err, isResolved: true}
+}
+
+type basicPromise[T any] struct {
```
jmank88, line 24: Does this behave differently than [sync.OnceValues](https://pkg.go.dev/sync#OnceValues)? It seems like we could leverage that, or even just expect callers to use it directly instead.

PR #1147, opened 2025-04-17
[CAPPL-733] Make WASM Module mockable

bolekk:
Also adding new V2 structs to make the interface complete (to be moved to a sub-package)


pkg/workflows/wasm/host/module.go
```diff
@@ -110,7 +110,15 @@ type ModuleConfig struct {
 	Determinism *DeterminismConfig
 }
 
-type Module struct {
+type Module interface {
+	Start()
+	Close()
+	IsLegacyDAG() bool
+	Run(ctx context.Context, request *wasmpb.Request) (*wasmpb.Response, error)
```
MStreet3, line 117: _nit_ docstring for the interface methods

bolekk: done

pkg/workflows/wasm/host/module.go
```diff
@@ -110,7 +110,17 @@ type ModuleConfig struct {
...
 
-type Module struct {
+type Module interface {
+	Start()
+	Close()
+	IsLegacyDAG() bool
+	// V1/Legacy API - request either the Workflow Spec or Custom-Compute execution
+	Run(ctx context.Context, request *wasmpb.Request) (*wasmpb.Response, error)
+	// V2/"NoDAG" API - request either the list of Trigger Subscriptions or launch workflow execution
+	Execute(ctx context.Context, request *wasmpb.ExecuteRequest) (*wasmpb.ExecutionResult, error)
```
cedric-cordenier, line 120: @bolekk I would recommend not exposing the protos directly at the interface level -- I started moving away from this with eg. the FetcherFn changes I made the other week.

The reason for this is that the protos are a lot more rigid and unchangeable the moment you start running actual workflows (you don't want to break them); by isolating the protos so that they're an internal detail here you can easily change the interface without introducing a breaking change at the proto level. This is a lot more difficult if you expose the protos directly.

pkg/workflows/wasm/host/module.go
```diff
@@ -110,7 +110,17 @@ type ModuleConfig struct {
 	Determinism *DeterminismConfig
 }
 
-type Module struct {
+type Module interface {
```
cedric-cordenier, line 113: @bolekk Did you consider having a ModuleV1 and ModuleV2 rather than an interface that combines the two? They aren't really interchangeable so it would make sense to rely on the type system to enforce using the right one in the right context.

You only need to know about the difference in one place (the workflow registry syncer); after that you should be guaranteed to be dealing with one or the other depending on the context.

pkg/workflows/wasm/host/module.go
bolekk: This is an interface between Engine and Host, so shouldn't be too hard to change internally IMO, even if it's a breaking change.
Can you elaborate on the recommendation? Do you mean having equivalent Go structs? That feels like unnecessary overhead converting back and forth.

pkg/workflows/wasm/host/module.go
bolekk: Makes sense, let me see if this fits well with the draft that Ryan has for V2 implementation (keeping the same underlying object).

bolekk: Done.

pkg/workflows/wasm/host/module.go
bolekk: If you don't mind, I'd rather keep it for now. We can avoid the extra overhead of converting structs and always migrate the Engine-Host part later when we see it's necessary.

justinkaseman (approved):

cedric-cordenier (approved):

PR #1148, opened 2025-04-21
V2 feature branch

MStreet3:
The WASI runner and runtime are not tested directly, only partially through running the no DAG workflow


pkg/capabilities/capabilities.go
MStreet3, line 1: what are the fields that are being added here?

pkg/capabilities/loopserver/loopserver.go
```diff
@@ -0,0 +1,31 @@
...
+package loopserver
+
+import (
+	"github.com/hashicorp/go-plugin"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop"
+)
+
+func Serve[T loop.StandardCapabilities](serviceName string, createPluginServer func(logger.Logger) T) {
```
MStreet3, line 12: what is the loop server used for?

pkg/capabilities/protoc/main.go
MStreet3, line 1: can you please add a readme to this package

pkg/capabilities/protoc/pkg/testdata/fixtures/capabilities/actionandtrigger/action_and_trigger.proto
MStreet3, line 1: add readme to package please or some notes

pkg/capabilities/triggers/testdata/fixtures/ondemand/schema.json
MStreet3, line 1: what does this schema refer to?

pkg/workflows/sdk/v2/promise.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"sync"
+)
+
+type Promise[T any] interface {
+	Await() (T, error)
+	CapabilityPromise
+}
+
+type CapabilityPromise interface {
```
MStreet3, line 12: what pattern is this/what's the value add where a private no-op method is embedded?

pkg/workflows/sdk/v2/runtime.go
MStreet3, line 1: can you please add a readme to the sdk package.  basic definition of `runner` versus `runtime` in the package context

pkg/workflows/wasm/host/module.go
```diff
@@ -92,7 +128,7 @@ type ModuleConfig struct {
 	InitialFuel               uint64
 	Logger                    logger.Logger
 	IsUncompressed            bool
-	Fetch                     func(ctx context.Context, req *wasmpb.FetchRequest) (*wasmpb.FetchResponse, error)
+	Fetch                     func(ctx context.Context, req *wasmdagpb.FetchRequest) (*wasmdagpb.FetchResponse, error)
```
MStreet3, line 131: can just be removed from struct, no?

pkg/workflows/wasm/host/module.go
```diff
@@ -326,25 +454,49 @@ func (m *Module) Run(ctx context.Context, request *wasmpb.Request) (*wasmpb.Resp
...
+
+	return runWasm[*wasmdagpb.Request, *wasmdagpb.Response](m, ctx, request, m.requestStore)
+}
+
+type idMessage interface {
+	GetId() string
+	proto.Message
+}
+
+func runWasm[I idMessage, O proto.Message](
```
MStreet3, line 469: what is runWasm for?

pkg/workflows/wasm/host/wasm.go
```diff
@@ -5,27 +5,89 @@ import (
...
-	wasmpb "github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/pb"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities"
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+	legacySdk "github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
+	legacywasmpb "github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/pb"
+	wasmpb "github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/v2/pb"
 )
 
-func GetWorkflowSpec(ctx context.Context, modCfg *ModuleConfig, binary []byte, config []byte) (*sdk.WorkflowSpec, error) {
+func GetTriggersSpec(ctx context.Context, modCfg *ModuleConfig, binary []byte, config []byte) ([]*wasmpb.TriggerSubscriptionRequest, error) {
```
MStreet3, line 19: why the name change from workflow spec to trigger spec?

pkg/workflows/wasm/host/wasm.go
```diff
@@ -5,27 +5,89 @@ import (
...
+		Request: &wasmpb.ExecuteRequest_Subscribe{Subscribe: &emptypb.Empty{}},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	switch r := execResult.Result.(type) {
+	case *wasmpb.ExecutionResult_Value:
+		v, err := values.FromProto(r.Value)
```
MStreet3, line 39: I thought that we were no longer using values, but strong proto types?

pkg/workflows/wasm/host/test/nodag/cmd/main.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/protoc/pkg/testdata/fixtures/capabilities/basicaction"
+	"github.com/smartcontractkit/chainlink-common/pkg/capabilities/protoc/pkg/testdata/fixtures/capabilities/basictrigger"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk/v2"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/v2"
+)
+
+func main() {
+	runner := wasm.NewDonRunner()
+	basic := &basictrigger.Basic{}
+	sdk.SubscribeToDonTrigger(runner, basic.Trigger(&basictrigger.Config{
```
MStreet3, line 15: what kind of trigger is this analogous to IRL?

pkg/workflows/wasm/v2/runner_wasip1.go
```diff
@@ -0,0 +1,161 @@
...
+	runtime    T
+	setRuntime func(id string, config []byte)
+	config     []byte
+}
+
+var _ sdk.DonRunner = &runner[sdk.DonRuntime]{}
+var _ sdk.NodeRunner = &runner[sdk.NodeRuntime]{}
+
+// TODO callbacks to setup a trigger...
+// TODO can't subscribe to a trigger more than once and differentiate the return value.
```
MStreet3, line 53: this TODO and the one below, are they related?  They seem almost at odds with each other

pkg/workflows/wasm/v2/runtime_wasip1.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil, err
+		}
+
+		mptr, mlen, err := bufferToPointerLen(m)
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO make this configurable?
+		response := make([]byte, 2048)
```
MStreet3, line 62: where does this length come from?

workflows/wasm/v2/pb/wasm.pb.go
MStreet3, line 1: is this file misplaced?

values/pb/values.pb.go
MStreet3, line 1: is this file misplaced?

pkg/capabilities/triggers/testdata/fixtures/ondemand/schema.json
nolag: This is a merge conflict, I guess I accidentally reverted this file when I checked out a folder or something.

pkg/workflows/wasm/host/module.go
nolag: Not if we want backwards compatibility. We need to support the DAG code until nothing uses it.

pkg/workflows/wasm/host/module.go
nolag: I refactored the Run method that's used for legacy into a helper method so that the DAG and no DAG share common code.

pkg/workflows/wasm/host/wasm.go
nolag: It's not renamed; GetWorkflowSpec is lower in the file. I don't know why git didn't realize this was added before GetWorkflowSpec. I can put it lower in the file; maybe that'll reduce the diff by a line.

pkg/workflows/wasm/host/wasm.go
nolag: For capability calls, yes. For workflow return values and consensus, I left the values.Value package. We don't know the type the user needs for those cases.

There may be something better we can use, @cedric-cordenier mentioned an alternative, but I didn't understand it.

pkg/workflows/wasm/host/test/nodag/cmd/main.go
nolag: It could be any DON trigger. ChainEvent, Cron, HTTP (if we go with a trusted Gateway, or signed payload). All the triggers we are talking about for SmartCon would be real examples.

pkg/workflows/wasm/v2/runner_wasip1.go
nolag: They are not directly related. 

The first isn't needed for SmartCon since we don't plan on having any node mode triggers, but some of them will need to call the WASM during initialization. For example, a WebSocket trigger when the Nodes all subscribe to the incoming events directly, the capability will need to start the WASM to ask it to send any required payload before upgrading from HTTP to WebSocket. The same capability will need to call the WASM when the WebSocket is established to send any required subscriptions.  I can remove this TODO, since it's not for SmartCon and it's not designed in detail.

The second TODO is to allow multiple subscriptions to the same trigger. For example, subscribing to two events on the same blockchain won't work as expected. In the code here, I don't have an ID for the trigger's subscription; instead, I only use the trigger's ID. It should use an ID for the subscription. I'll sync with you on this in our meeting later today.

pkg/workflows/wasm/v2/runtime_wasip1.go
nolag: It started as a PoC so I picked a number. I may have borrowed it from the current runtime, I don't remember.

pkg/capabilities/capabilities.go
nolag: I'm allowing use of proto's any over values.Value in the new SDK.

pkg/capabilities/loopserver/loopserver.go
nolag: This allows you to run a standard capability inside a standard loop. I borrowed a lot of it from the capabilities repo. It works with the generated code for the capability developer

workflows/wasm/v2/pb/wasm.pb.go
nolag: No, but I'm unsure why the proto file is missing

pkg/workflows/wasm/host/module.go
MStreet3: had we considered creating a `host/v2/module` and redefine the `Module` to be v2 only?  All legacy code would use the legacy struct

PR #1150, opened 2025-04-22
add evm relayer service

Unheilbar:
core ref: 3a975c7ef81a2a19747a6596c2b982184e07c1aa
solana ref: 87e7241a5dd95b86c9f7e760c47c433f9d44a1d9
starknet ref: 32ced002d4eaf3482bcc343d7bfffbffa65a9ad3


pkg/types/relayer.go
```diff
@@ -80,9 +80,9 @@ type NodeStatus struct {
 	State   string
 }
 
-type EVMChain interface {
+type EVMRelayer interface {
 	services.Service
```
jmank88, line 84: Can drop this since we get it from the Relayer
```suggestion
```

pkg/types/relayer.go
```diff
@@ -108,8 +108,6 @@ type ChainService interface {
 type Relayer interface {
 	ChainService
 
-	NewEVMChain(ctx context.Context) (EVMChain, error)
```
jmank88, line 111: I think we want to swap in `AsEVMRelayer` here, right?

pkg/loop/internal/types/types.go
```diff
@@ -60,6 +59,12 @@ type Relayer interface {
 	NewLLOProvider(context.Context, types.RelayArgs, types.PluginArgs) (types.LLOProvider, error)
 }
 
+type EVMRelayer interface {
+	Relayer
```
ilija42, line 63: Having Relayer embedded here messes with type aliasing `type Relayer = looptypes.Relayer`, if some code implements the type alias its AsEVMRelayer will also expect the return type to be the type aliased EVMRelayer from the loop package which means that we can't inject the relayer properly.

Can we just drop the Relayer from here and just have services.Service?

jmank88: I was hoping to allow users to just "upgrade" their single Relayer reference, but if we need to separate the two method sets as exposed through these interfaces, that wouldn't be the worst.

What alternative do you propose? Without the method, do callers have to cast? Or use a helper func?

jmank88: I suppose the separate namespace for the interfaces has benefits, since we can effectively "overload" generic methods with family specific ones by the same name.

ilija42: We don't have to remove the AsEVM() method, just remove the Relayer type from EVMRelayer, right? Not sure if that is what you are refering to

ilija42: I think we can leave the Relayer here by adding an adapter in core like Vladimir did in the core PR. Then if someone wants to reduce the feature set to just EVM its trivial

ilija42 (dismissed):

jmank88 (approved):

ilija42 (approved):

PR #1152, opened 2025-04-23
Add the ability to call capabilities with an any instead of a values.Value

nolag:
More context at https://github.com/smartcontractkit/chainlink-common/pull/1146/files


bolekk (dismissed):

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	if any == nil {
+		if value == nil {
+			return false, errors.New("neither value nor any provided")
+		}
+		err := value.UnwrapTo(into)
+		return false, err
+	}
+
+	err := any.UnmarshalTo(into)
+	return true, err
```
MStreet3, line 66: _nit_ docstring on func.  what does `true` mean in the case that err != nil?

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func Execute[I, C, O proto.Message](
+	ctx context.Context,
+	request CapabilityRequest,
+	input I,
+	config C,
+	exec func(context.Context, RequestMetadata, I, C) (O, error)) (CapabilityResponse, error) {
```
MStreet3, line 74:
```suggestion
	exec func(context.Context, RequestMetadata, I, C) (O, error),
) (CapabilityResponse, error) {
```

_nit_ return types were getting confusing

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	request TriggerRegistrationRequest,
+	message C,
+	fn func(context.Context, RequestMetadata, C) (<-chan TriggerAndId[O], error)) (<-chan TriggerResponse, error) {
+	migrated, err := FromValueOrAny(request.Config, request.Request, message)
+	if err != nil {
+		return nil, fmt.Errorf("error when unwrapping request: %w", err)
+	}
+
+	// TODO size?
+	response := make(chan TriggerResponse, 100)
```
MStreet3, line 111: only needs to be `1`, no?  There's one read then close.

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	respCh, err := fn(ctx, request.Metadata, message)
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		for {
```
MStreet3, line 118: _nit_ `defer close(response)`

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return response, nil
+}
+
+type TriggerAndId[T proto.Message] struct {
+	Trigger T
+	Id      string
+}
+
+func RegisterTrigger[C, O proto.Message](
```
MStreet3, line 99: what is `C` supposed to be? instead of say `M` (message) or `I` (input)

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	"errors"
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+)
+
+func UnwrapRequest(request CapabilityRequest, config proto.Message, value proto.Message) (bool, error) {
```
MStreet3, line 14:
```suggestion
// UnwrapRequest unwraps the request into value and its config into config returning true if the unwrapped values
// were derived from the V2 No DAG fields.
func UnwrapRequest(request CapabilityRequest, config proto.Message, value proto.Message) (bool, error) {
```

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+	wrapped, err := values.WrapMap(value)
+	if err != nil {
+		return err
+	}
+
+	response.Value = wrapped
+	return nil
+}
+
+func FromValueOrAny(value values.Value, any *anypb.Any, into proto.Message) (bool, error) {
```
MStreet3, line 56:
```suggestion
// FromValueOrAny unwraps either value or any into a single message, returns true if any is used as source.
func FromValueOrAny(value values.Value, any *anypb.Any, into proto.Message) (bool, error) {
```

pkg/capabilities/utils.go
nolag: You should generally ignore results when err != nil.

MStreet3: agreed, just didn't understand that the `bool` meant `migrated` or not.

pkg/capabilities/utils.go
```diff
@@ -0,0 +1,145 @@
...
+				if migrated {
+					wrapped, err := anypb.New(resp.Trigger)
+					tr.Err = err
+					tr.Event.Value = wrapped
+				} else {
+					wrapped, err := values.WrapMap(resp.Trigger)
+					tr.Err = err
+					tr.Event.Outputs = wrapped
+				}
+				response <- tr
```
MStreet3, line 136: write should be wrapped in select with route for `<-ctx.Done()` to finish

pkg/capabilities/utils.go
nolag: Config type because of what the name was in the DAG world. Renamed to I so it's I and O like the capability request.

pkg/capabilities/utils.go
nolag: Triggers continually run. They send on the channel whenever a workflow should start. I removed the TODO but left it at 100 for now.

bolekk (approved):

MStreet3 (approved):

pkg/capabilities/capabilities.go
```diff
@@ -92,8 +97,21 @@ type RegistrationMetadata struct {
 // CapabilityRequest is a struct for the Execute request of a capability.
 type CapabilityRequest struct {
 	Metadata RequestMetadata
-	Config   *values.Map
-	Inputs   *values.Map
+
```
cedric-cordenier, line 100: :( We should try to avoid this if we can: I can see an argument for having the v1 and v2 args at the proto layer (for compatibility), but we should be able to avoid this in the Go structs since we should always know which kind of call we're dealing with

The current proposal is going to make things pretty messy IMO -- we should try to make these changes in such a way that we have a clean path towards structs and types that are version specific.

nolag: How would you have compatibility when we're supporting both? Would you make it an interface and use methods to get the values?

Longer term, once DAG is dead, we'll remove the old ones.

nolag: A breaking change that would be clean would be to make CapabilityRequest and CapabilityResponse generic and take their type. Then have the helper methods convert to the pbany.Any or values.Value.

cedric-cordenier: (It looks like you deal with that latter problem below though!)

pkg/capabilities/pb/capabilities_helpers.go
```diff
@@ -63,8 +63,11 @@ func CapabilityRequestToProto(req capabilities.CapabilityRequest) *CapabilityReq
 			ReferenceId:              req.Metadata.ReferenceID,
 			DecodedWorkflowName:      req.Metadata.DecodedWorkflowName,
 		},
-		Inputs: values.ProtoMap(inputs),
-		Config: values.ProtoMap(config),
+		Inputs:        values.ProtoMap(inputs),
```
cedric-cordenier, line 66: I'd add the same comment here -- we should know what type we're dealing with; either v1 or v2.

pkg/capabilities/capabilities.go
bolekk: I wouldn't mind having Go structs V1/V2 which use a single underlying proto, like Cedric is suggesting.

nolag: I was worried that it would make it a lot harder to re-use any V1 components. I thought that there are components you can re-use from today's engine.

If we have different higher level representations, why bother sharing the proto? As far as I can tell, nothing interacts directly with the protos. 

If we go that route, I would rather just directly use the new or current protos, and have the helpers decode the payload for you. Thoughts?

cedric-cordenier: >If we go that route, I would rather just directly use the new or current protos, and have the helpers decode the payload for you. Thoughts?

This is 100% my preferred option, it implies a bit more work in terms of the loopp grpc wrappers though (but IMO it would be work that worthwhile doing)

nolag: Summary of my sync with Cedric:

I'm going to use a new proto for the v2 capabilities, but I won't make new wrappers. The code generation I have for the server will be changed to reference those directly, if we decide we want to wrap them later we can change the code generator. 

The main argument for the thin wrapper is that we can change the wire format independently from the capability call. One thing that should help with this is that the generated code decodes the payload for you, so we won't be tied to pbany.Any, as the capability author never sees it. The metadata fields message would be tied to the proto message (but not the encoding, because it doesn't have `Any` in it), and the user's input/output type.

The main argument against the wrapper is the duplication of code, leading to more things to change when you add fields etc.

PR #1153, opened 2025-04-23
Create the interfaces for the CRE v2

nolag:


pkg/workflows/sdk/v2/promise.go
```diff
@@ -0,0 +1,58 @@
+package sdk
+
+import (
+	"sync"
+)
+
+type Promise[T any] interface {
+	Await() (T, error)
```
nolag, line 8: There's discussion around the name of Await. Summarizing other potentials:

- Get()
-() (just a function, this interface goes away. Then isn't needed, we just wraps the function directly)
- Recv() (concern is that it may imply multiple results, and it's unclear what we should do in that case.

pkg/workflows/sdk/v2/handler.go
```diff
@@ -0,0 +1,70 @@
+package sdk
+
+import (
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+type Handler[T any] interface {
```
MStreet3, line 8: _nit_ docstring on Handler

pkg/workflows/sdk/v2/handler.go
```diff
@@ -0,0 +1,70 @@
...
+	Id() string
+	Method() string
+	TriggerCfg() *anypb.Any
+	Callback() func(runtime T, triggerOutputs *anypb.Any) (any, error)
+}
+
+func NewDonHandler[T proto.Message, O any](trigger DonTrigger[T], callback func(runtime DonRuntime, triggerOutputs T) (O, error)) Handler[DonRuntime] {
```
MStreet3, line 15: `triggerOutputs` to `triggerPayload`?

pkg/workflows/sdk/v2/handler.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		return nil, callback(runtime, unwrappedTrigger)
+	}
+
+	return &handler[R, M]{Trigger: trigger, fn: wrapped}
+}
+
+type handler[R any, T proto.Message] struct {
+	Trigger[T]
```
MStreet3, line 58: Where is `Trigger` defined?

pkg/workflows/sdk/v2/runner.go
```diff
@@ -0,0 +1,44 @@
...
+	Config() []byte
+}
+
+type DonRunner interface {
+	RunnerBase
+
+	Run(args *WorkflowArgs[DonRuntime])
+}
+
+// NOTE that some triggers in node mode need callbacks into the WASM when setting up (like WebSocket trigger).
```
MStreet3, line 21: do you have an example in a doc?  not sure I follow

pkg/workflows/sdk/v2/handler.go
nolag: In runner.go. I can move it.

pkg/workflows/sdk/v2/runtime.go
```diff
@@ -0,0 +1,95 @@
...
+		if err == nil {
+			wrapped, err := values.Wrap(result)
+			if err != nil {
+				consensusRequest.Observation = &pb.BuiltInConsensusRequest_Error{Error: err.Error()}
+			} else {
+				consensusRequest.Observation = &pb.BuiltInConsensusRequest_Value{Value: values.Proto(wrapped)}
+			}
+		} else {
+			consensusRequest.Observation = &pb.BuiltInConsensusRequest_Error{Error: err.Error()}
+		}
```
MStreet3, line 84:
```suggestion
		result, err := fn(nodeRuntime)
		if err != nil {
		    consensusRequest.Observation = &pb.BuiltInConsensusRequest_Error{Error: err.Error()}
		    return consensusRequest
		}
		
		wrapped, err := values.Wrap(result)
	        if err != nil {
		        consensusRequest.Observation = &pb.BuiltInConsensusRequest_Error{Error: err.Error()}
		        return consensusRequest
	        } 
		consensusRequest.Observation = &pb.BuiltInConsensusRequest_Value{Value: values.Proto(wrapped)}
```

_nit_ reduce nesting

pkg/workflows/sdk/v2/runtime.go
```diff
@@ -0,0 +1,95 @@
...
+func DonModeCallInNodeMode() error {
+	return donModeCallInNodeMode
+}
+
+func RunInNodeMode[T any, C BuiltInConsensus[T]](runtime DonRuntime, fn func(nodeRuntime NodeRuntime) (T, error), consensus C) Promise[T] {
+	observationFn := func(nodeRuntime NodeRuntime) *pb.BuiltInConsensusRequest {
+		var primitiveConsensus *pb.PrimitiveConsensus
+		var defaultValue values.Value
+		var err error
+		switch c := any(consensus).(type) {
```
MStreet3, line 56: is it necessary to cast to `any` here?

pkg/workflows/sdk/v2/runtime.go
```diff
@@ -0,0 +1,95 @@
...
+		return consensusRequest
+	}
+
+	return Then(runtime.RunInNodeMode(observationFn), func(v values.Value) (T, error) {
+		// TODO this is wrong, but good enough for now...
+		var t T
+		err := v.UnwrapTo(&t)
+		return t, err
+	})
+}
```
MStreet3, line 95: I don't follow, what's happening here?

pkg/workflows/sdk/v2/trigger.go
```diff
@@ -0,0 +1,22 @@
+package sdk
+
+import (
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+type Trigger[T proto.Message] interface {
+	NewT() T
+	Id() string
```
MStreet3, line 10:
```suggestion
        // Id uniquely identifies a Trigger (something more?)
	Id() string
```

pkg/workflows/sdk/v2/trigger.go
```diff
@@ -0,0 +1,22 @@
...
+	ConfigAsAny() *anypb.Any
+	Method() string
```
MStreet3, line 12:
```suggestion
        // Method.. (what is method in this context?)
	Method() string
```

pkg/workflows/sdk/v2/trigger.go
nolag: Id is pretty self-describing. Not everything needs a comment, and the interface does not provide methods that a workflow author is expected to call.

pkg/workflows/sdk/v2/runtime.go
nolag: I mean to remove the comment. This is unwrapping the values.Value into your type. Users don't need to worry about how we get consensus on data structures we don't know AOT.

@cedric-cordenier you mentioned that there may be a better way to do consensus (and return values from workflows) on unknown data without values.Value. I don't remember your suggestion, but if it's type-safe (won't break if you unwrap and the number you unwrap to looses precision) I'm open to changing this.

pkg/workflows/sdk/v2/runtime.go
nolag: Yes, otherwise there's a compiler error:

`Cannot use type switch on type parameter value consensus`

MStreet3 (approved):

MStreet3 (dismissed):

pkg/workflows/sdk/v2/pb/sdk.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
+
+package cre.sdk.v2;
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk/v2/pb";
+
+import "google/protobuf/any.proto";
+import "values/pb/values.proto";
+
+message CapabilityRequest {
```
bolekk, line 10: Can we use protos from Capability API directly? (pkg/capabilities/pb/capabilities.proto)

pkg/workflows/sdk/v2/const.go
```diff
@@ -0,0 +1,3 @@
+package sdk
+
+const IdLen = 24
```
bolekk, line 3: Why 24? Add a comment with explanation?

nolag: I don't remember why I picked 24... I'll change it to be 16 bytes so it's the size of a UUID.

pkg/workflows/sdk/v2/pb/sdk.proto
nolag: No, it leaks metadata and old values.Value fields that the workflow author doesn't fill in. I want a clean separation of what the workflow author can do (even if it's inside the SDK) and what the host or engine will do.

bolekk (approved):

MStreet3 (approved):

pkg/workflows/sdk/v2/promise.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewBasicPromise[T any](await func() (T, error)) Promise[T] {
+	return &basicPromise[T]{await: await}
+}
+
+func PromiseFromResult[T any](result T, err error) Promise[T] {
+	return &basicPromise[T]{resolved: result, err: err, isResolved: true}
+}
+
+type basicPromise[T any] struct {
```
jmank88, line 24: From the other draft: Is this different from [sync.OnceValues](https://pkg.go.dev/sync#OnceValues), or could that be used, either here or directly by callers?

nolag: I don't want to use it directly because we'll likely add await (or awaitN) all at some point and we may need to add some methods to promise to support it. That's also why I made the interface private, so I can add methods and know all implementations have them.

sdk.Await2(...) (T1, T2, err)
sdk.Await3(...)(T1, T2, T3, err)

PR #1157, opened 2025-04-24
Add a client generator for CRE SDK v2.

nolag:
Note that `testutils` will test the generated code in the next PR. They will serve two purposes:

To verify `testutils` work correctly 
To verify that code generated functions correctly.

I generated them in this PR to ensure that the generated code can compile and so reviewers can see what they do.

See [this example](https://github.com/smartcontractkit/chainlink-common/pull/1146/files#diff-ee9a3cd72be77b43a3d185008ba2620a78e33c33b347391e2bddca6c38d7b07b) for how generated structs will be used.


pkg/capabilities/v2/protoc/pkg/templates/client.go.tmpl
```diff
@@ -0,0 +1,45 @@
...
+
+    {{- range allimports }}
+    {{.}}
+    {{- end }}
+)
+
+{{- range .Services}}
+{{ $service := . }}
+type {{.GoName}} struct {
+    // TODO config types (optional)
```
bolekk, line 25: You wanna keep those for now?

bolekk (approved):

pkg/capabilities/v2/protoc/pkg/templates/client.go.tmpl
nolag: Forgot to link the tickets, I'll do that now.

bolekk (approved):

pkg/capabilities/v2/protoc/pkg/pb/cre_metadata.proto
```diff
@@ -0,0 +1,18 @@
...
+
+import "workflows/sdk/v2/pb/sdk.proto";
+
+message CapabilityMetadata {
+  cre.sdk.v2.Mode mode = 1;
+  string capability_id = 2;
+}
+
+extend google.protobuf.ServiceOptions {
+  CapabilityMetadata capability = 50001;
```
patrickhuie19, line 17: why 50001?

pkg/capabilities/v2/protoc/pkg/test_capabilities/basicaction/basic_action.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
```
patrickhuie19, line 1: I'm not aware of a mechanism for protoc to compose protos cross-repo, so having these protos defined here means only other protos in common can benefit from them. I'm not saying that's a bad thing - it probably makes the most sense in this case. 

In the off chance you had ideas around other consumers being able to compose from these, `chainlink-protos` repo is an option.

pkg/capabilities/v2/protoc/pkg/test_capabilities/nodeaction/node_action.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+package basic_test_node_action.v1;
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/capabilities/v2/protoc/test_capabilities/nodeaction";
+
+import "capabilities/v2/protoc/pkg/pb/cre_metadata.proto";
+
+message Config {
+  string camel_case_in_schema_for_testing = 1;
```
patrickhuie19, line 10: /nit this is still snake?

pkg/capabilities/v2/protoc/pkg/test_capabilities/nodetrigger/node_trigger.pb.go
```diff
@@ -0,0 +1,184 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
patrickhuie19, line 1: What would consumers pull in from this generated file?

pkg/capabilities/v2/protoc/pkg/pb/cre_metadata.proto
nolag: It's the first one that's left for users to define, and ChatGPT picket it.

pkg/capabilities/v2/protoc/pkg/test_capabilities/nodeaction/node_action.proto
nolag: snake case is the preferred naming for protos. I'll update all of them, I had copied from the old json schema and forgot that I had tested different casing in that generator. That testing doesn't apply to protoc because it gets the go name.

pkg/capabilities/v2/protoc/pkg/test_capabilities/nodetrigger/node_trigger.pb.go
nolag: This file has the message generation; it's the default generator for Go protos. The service in the proto file uses these types for input, output, and config. 

My protoc generator creates the SDK (maybe I should rename client generator to SDK Generator and the output files as well...?) for use with the CRE, the methods are defined in a service. That service references the messages. You can use input and output types from other proto files if you have shared messages.

patrickhuie19: > maybe I should rename client generator to SDK Generator and the output files as well...?

i understood the original intent, but if you're taking another pass on this anyways, i think it's worth it

patrickhuie19 (dismissed):

pkg/capabilities/v2/protoc/pkg/test_capabilities/basicaction/basic_action.proto
nolag: I didn't know about chainlink-protos, but I hesitate to put the generated capabilities there for a few reasons.

- The generated code from my protoc generator references chainlink-common
- The code generator itself is in chainlink-common (not as big a deal)
- The ones in test_capabilities are tests for the components in chainlik-common
- The other protos these ones reference are in chainlink-common

The one that will be referenced across repos would be pkg/capabilities/v2/protoc/pkg/pb/cre_metadata.proto. protoc can go across anything you want. It takes paths for reference. Usually, people either create sub-repositories to get your protos, or copy them. We can decide how to expose this proto later.

We may also add additional helpers, like a go.mod for the client, and whatever other languages use when we add clients for those languages. That way the modules are independent.

patrickhuie19: > protoc can go across anything you want. It takes paths for reference. Usually, people either create sub-repositories to get your protos, or copy them. We can decide how to expose this proto later.

We can discuss async to this PR, but I haven't seen a good mechanism for doing cross-repo. Within a repo works fine, but requires some patience with the relative paths you feed protoc so it can discover what it needs.

patrickhuie19 (approved):

bolekk (approved):

PR #1158, opened 2025-04-24
Add mock generation to the capability code generator and test runners and runtimes for workflows

nolag:


pkg/workflows/sdk/v2/testutils/registry.go
```diff
@@ -0,0 +1,52 @@
...
+
+	r := &Registry{}
+	testRegistries[tb] = r
+	tb.Cleanup(func() {
+		delete(testRegistries, tb)
+	})
+	return r
+}
+
+type Registry struct {
```
cedric-cordenier, line 28: Can we avoid reimplementing a registry here? We did this with the engine and it ended up being a pain since we had to maintain two implementations. The existing implementation is totally in-memory so it should be easy to do this.

pkg/workflows/sdk/v2/testutils/runtime_test.go
```diff
@@ -0,0 +1,324 @@
...
+	action1, err := basicactionmock.NewBasicActionCapability(t)
+	require.NoError(t, err)
+	action1.PerformAction = func(_ context.Context, input *basicaction.Inputs) (*basicaction.Outputs, error) {
+		<-ch
+		return &basicaction.Outputs{AdaptedThing: anyResult1}, nil
+	}
+
+	anyResult2 := "ok2"
+	action2, err := actionandtriggermock.NewBasicCapability(t)
+	action2.Action = func(ctx context.Context, input *actionandtrigger.Input) (*actionandtrigger.Output, error) {
```
bolekk, line 48: I feel like we should leverage Mockery here instead of building it ourselves...

bolekk (approved):
Approving to unblock but I'm not sure if building mocks ourselves is the best approach.

pkg/workflows/sdk/v2/testutils/capability.go
```diff
@@ -0,0 +1,29 @@
...
+
+var _ error = NoTriggerStub("")
+
+type NoCapability string
+
+func (n NoCapability) Error() string {
+	return "Capability not found: " + string(n)
+}
+
+var _ error = NoCapability("")
```
cedric-cordenier, line 29: Idiomatically errors should start with `Err`

pkg/workflows/sdk/v2/testutils/registry.go
nolag: I'm close to having the registry in common (and simplifying it by removing the methods we discussed) and using it in core. Before completing that work, I wanted to capture some thoughts.

I realized why I was concerned about this earlier and want to capture it here while I remember, even though I'm still sold on sharing the registry. I'll take it one step further and pull some of the code that I use for the runtime and runner for WASI into this PR so that that code can be reused as much as possible, too. If we do that, we can help capture errors with large payloads, etc.

With the current iteration of the PR, we can update anything beyond the host without updating anything in a workflow. There's a clear line that this interface respects. If we use the underlying call communication between engine <-> capability in the mocks, then that's no longer true, and we're breaking that abstraction to reuse a map.  Implementing only guest <-> host communication for mocks reduces the surface area required for each SDK.

On the plus side of re-use, I want to enable fake use alongside mocks in tests.  While mocks have their purposes, like simplifying state  (it's sometimes not feasible or even possible to reproduce a 'bad' state, or testing edge cases with very complex setups), fakes are far superior. The real registry will be significantly easier for fakes since the API uses the same methods.

nolag: I'm tink I'm going too far down the rabbit hole for reusability when I decided to also redo the runner and runtime to use common code with the WASI one. Using the capability registry behind the scenes should capture almost everything, except payload sizes, but I can add tests for that.

pkg/workflows/sdk/v2/testutils/runtime_test.go
nolag: It does more when you call new. I also like that a single tool is used for everything instead of two. Let's leave this decision for the Go Guild to discuss more. I showed them this before, and I didn't get any objections.

bolekk (approved):

MStreet3 (approved):

PR #1161, opened 2025-04-29
BCFR-1330

Unheilbar:
[Ticket](https://smartcontract-it.atlassian.net/browse/BCFR-1328)

- EVMService expanded with evm-specific functionalities

1. [core](https://github.com/smartcontractkit/chainlink/pull/17496)
2. [chainlink-evm](https://github.com/smartcontractkit/chainlink-evm/pull/52)
3. [chainlink-framework](https://github.com/smartcontractkit/chainlink-framework/pull/44)

QueryTrackedLogs is coming in a [separate PR](https://github.com/smartcontractkit/chainlink-common/pull/1179)
starknet ref: c7b7169940f7d9d355b4cb9fc4326e3ee7be2d21
core ref: fb5a9eb5264efc2f9f1ec99c397629c0d5f2f813


pkg/loop/internal/example-relay/main.go
```diff
@@ -73,7 +73,7 @@ type relayer struct {
 
 func (r *relayer) Name() string { return r.lggr.Name() }
 
-func (r *relayer) AsEVMRelayer() (loop.EVMRelayer, error) {
+func (r *relayer) EVM() (types.EVMService, error) {
```
ilija42, line 76: Can we make this change in a separate PR so that its easier to get all the other repos merged more easily

pkg/loop/internal/pb/evm.proto
```diff
@@ -0,0 +1,182 @@
...
+  string block_hash = 4; // common.Hash represented as string
+  bytes data = 5; //
+  bytes eventSig = 6;
+  BigInt block_number = 7; 
+  uint32 tx_index = 8; 
+  uint32 index = 9; 
+  bool removed = 10; 
+}
+
+message EVMFilterQuery {
```
ilija42, line 51: wdyt about adding chains services folder and evm folders under pb and drop the EVM prefix

pkg/loop/internal/pb/evm.proto
```diff
@@ -0,0 +1,182 @@
...
+  repeated Log logs = 1;
+}
+
+message SubscribeLogTriggerRequest {
+  repeated Expression expression = 1;
+}
+
+message LogTriggerReply {
+  Log log = 1;
+}
```
ilija42, line 182: Lets just drop log trigger from the interface and proto for now, needs to be defined better

pkg/loop/internal/pb/evm.proto
```diff
@@ -0,0 +1,182 @@
...
+
+// ----- Request/Reply Wrappers -----
+message LatestAndFinalizedHeadReply {
+  Head latest = 1;
+  Head finalized = 2;
+}
+
+message CallContractRequest {
+  CallMsg call = 1;
+  Confidence confidenceLevel = 2; // confidenceLevel will be handled by evm loop to difine finalty
```
ilija42, line 102: typo

pkg/loop/internal/relayer/evm_service.go
```diff
@@ -0,0 +1,531 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// evm service client
+type evmRelayerClient struct {
+	cl pb.EVMClient
+}
+
+// Direct calls
```
ilija42, line 25: No need to expose implementation details on the grpc client

pkg/loop/internal/relayer/evm_service.go
```diff
@@ -0,0 +1,531 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/chains/evm"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// evm service client
+type evmRelayerClient struct {
```
ilija42, line 21: should this be called evmService now, or just evm?

pkg/types/relayer.go
```diff
@@ -46,7 +49,7 @@ type PluginArgs struct {
 }
 
 // RelayArgs are the args required to create relayer.
-// The are common to all relayer implementations.
+// The are common to all xelayer implementations.
```
ilija42, line 52: lint

pkg/types/relayer.go
```diff
@@ -101,11 +102,34 @@ type ChainService interface {
 	Replay(ctx context.Context, fromBlock string, args map[string]any) error
 }
 
+type EVMService interface {
+	// -- ChainService
+	// Direct Calls
```
ilija42, line 107: no need to expose implementation detail on the interface

pkg/loop/internal/pb/evm.proto
```diff
@@ -0,0 +1,182 @@
...
+message BalanceAtReply {
+  BigInt balance = 1;
+}
+
+message EstimateGasRequest {
+  CallMsg msg = 1;
+}
+
+message EstimateGasReply {
+  uint64 gas = 1;
```
ilija42, line 141: Add a comment for gas format, is it gwei, wei?

patrickhuie19 (dismissed):

pkg/loop/internal/relayer/evm_test.go
jmank88, line 1: Can we avoid mocks here?
Our pattern for testing these has been to make a small "static" implementation, with fixed or parameterized values.

jmank88 (changes requested):

pkg/types/chains/evm/evm.go
```diff
@@ -0,0 +1,104 @@
+package evm
+
+import (
+	"math/big"
+	"time"
+)
+
+// represents evm-style address 40-character hexadecimal string prefixed by 0x making it 42 characters total
+// lower case, upper case or checksummed case (EIP-55) are allowed.
```
jmank88, line 9: So it is OK to be permissible with casing on input params since we can translate them to bytes anyways, but we need to be careful about returning things in mixed formats, because without a canonical format equality comparison breaks downs and they can't be used as map keys, etc.

pkg/types/chains/evm/evm.go
```diff
@@ -0,0 +1,104 @@
...
+type FilterQuery struct {
+	BlockHash Hash      // for filter by exact block, if not empty can't use from/to
+	FromBlock *big.Int  // start block range
+	ToBlock   *big.Int  // end block range
+	Addresses []Address // contract(s) to filter logs from
+
+	Topics []Hash // filter log by event sigs and indexed args
+}
+
+// matches LP-filter
```
jmank88, line 42: What is `LP-filter`?

pkg/types/relayer.go
```diff
@@ -100,10 +98,29 @@ type ChainService interface {
...
+	CallContract(ctx context.Context, msg *evm.CallMsg, confidence primitives.ConfidenceLevel) ([]byte, error)
+	GetLogs(ctx context.Context, filterQuery evm.FilterQuery) ([]*evm.Log, error)
+	BalanceAt(ctx context.Context, account string, blockNumber *big.Int) (*big.Int, error)
+	EstimateGas(ctx context.Context, call *evm.CallMsg) (uint64, error)
+	TransactionByHash(ctx context.Context, hash string) (*evm.Transaction, error)
+	TransactionReceipt(ctx context.Context, txHash string) (*evm.Receipt, error)
+
+	// GetTransactionFee retrieves the fee of a transaction in wei from the underlying chain's TXM
+	GetTransactionFee(ctx context.Context, transactionID string) (*TransactionFee, error)
+	LatestAndFinalizedHead(ctx context.Context) (latest Head, finalized Head, err error)
```
jmank88, line 111: We already have `ChainService.LatestHead`. Should this one be on the generic `ChainService` too? Or alternatively, will this one diverge, and so should it be returning a custom evm head type instead of the shared one?

pkg/types/relayer.go
```diff
@@ -100,10 +98,29 @@ type ChainService interface {
 	Replay(ctx context.Context, fromBlock string, args map[string]any) error
 }
 
+type EVMService interface {
+	CallContract(ctx context.Context, msg *evm.CallMsg, confidence primitives.ConfidenceLevel) ([]byte, error)
+	GetLogs(ctx context.Context, filterQuery evm.FilterQuery) ([]*evm.Log, error)
+	BalanceAt(ctx context.Context, account string, blockNumber *big.Int) (*big.Int, error)
+	EstimateGas(ctx context.Context, call *evm.CallMsg) (uint64, error)
+	TransactionByHash(ctx context.Context, hash string) (*evm.Transaction, error)
+	TransactionReceipt(ctx context.Context, txHash string) (*evm.Receipt, error)
```
jmank88, line 107:
```suggestion
	TransactionByHash(ctx context.Context, hash evm.Hash) (*evm.Transaction, error)
	TransactionReceipt(ctx context.Context, txHash evm.Hash) (*evm.Receipt, error)
```

pkg/types/relayer.go
```diff
@@ -100,10 +98,29 @@ type ChainService interface {
 	Replay(ctx context.Context, fromBlock string, args map[string]any) error
 }
 
+type EVMService interface {
+	CallContract(ctx context.Context, msg *evm.CallMsg, confidence primitives.ConfidenceLevel) ([]byte, error)
+	GetLogs(ctx context.Context, filterQuery evm.FilterQuery) ([]*evm.Log, error)
+	BalanceAt(ctx context.Context, account string, blockNumber *big.Int) (*big.Int, error)
```
jmank88, line 104:
```suggestion
	BalanceAt(ctx context.Context, account evm.Address, blockNumber *big.Int) (*big.Int, error)
```

pkg/types/relayer.go
```diff
@@ -100,10 +98,29 @@ type ChainService interface {
...
+	TransactionReceipt(ctx context.Context, txHash string) (*evm.Receipt, error)
+
+	// GetTransactionFee retrieves the fee of a transaction in wei from the underlying chain's TXM
+	GetTransactionFee(ctx context.Context, transactionID string) (*TransactionFee, error)
+	LatestAndFinalizedHead(ctx context.Context) (latest Head, finalized Head, err error)
+	QueryLogsFromCache(ctx context.Context, filterQuery []query.Expression,
+		limitAndSort query.LimitAndSort, confidenceLevel primitives.ConfidenceLevel) ([]*evm.Log, error)
+	RegisterLogTracking(ctx context.Context, filter evm.LPFilterQuery) error
+	UnregisterLogTracking(ctx context.Context, filterName string) error
+	GetTransactionStatus(ctx context.Context, transactionID string) (TransactionStatus, error)
```
jmank88, line 116: Do we have duplicate `TransactionStatus`s?
```suggestion
	GetTransactionStatus(ctx context.Context, transactionID string) (evm.TransactionStatus, error)
```

pkg/types/relayer.go
```diff
@@ -100,10 +98,29 @@ type ChainService interface {
...
+type EVMService interface {
+	CallContract(ctx context.Context, msg *evm.CallMsg, confidence primitives.ConfidenceLevel) ([]byte, error)
+	GetLogs(ctx context.Context, filterQuery evm.FilterQuery) ([]*evm.Log, error)
+	BalanceAt(ctx context.Context, account string, blockNumber *big.Int) (*big.Int, error)
+	EstimateGas(ctx context.Context, call *evm.CallMsg) (uint64, error)
+	TransactionByHash(ctx context.Context, hash string) (*evm.Transaction, error)
+	TransactionReceipt(ctx context.Context, txHash string) (*evm.Receipt, error)
+
+	// GetTransactionFee retrieves the fee of a transaction in wei from the underlying chain's TXM
+	GetTransactionFee(ctx context.Context, transactionID string) (*TransactionFee, error)
```
jmank88, line 110: What is `transactionID`? Not the hash?

pkg/types/chains/evm/evm.go
```diff
@@ -0,0 +1,104 @@
...
+	ContractAddress   Address  // Address of the contract if one was created by this transaction
+	GasUsed           uint64   // actual gas used during execution in gas units
+	BlockHash         Hash     // hash of the block containing this receipt
+	BlockNumber       *big.Int // number of the block containing this receipt
+	TransactionIndex  uint64   // index of the transaction inside of the block
+	EffectiveGasPrice *big.Int // actual price in wei paid per gas unit
+}
+
+// matches simplified evm-style head
+type Head struct {
```
jmank88, line 87: Is this used?

pkg/types/chains/evm/evm.go
Unheilbar: Filter for Cache (Log Poller) Its slightly different from evm-style one

pkg/types/relayer.go
Unheilbar: Nope, it's idempotency key generated by caller

pkg/types/relayer.go
Unheilbar: I should have put a TODO for another PR that will deduplicate TransactionStatus and TransactionFee, since now they are used by TxManager

pkg/types/relayer.go
jmank88: This API must be stable by changing only in compatible ways, so a decision needs to be made now in this PR. If the intention is for this to be an evm specific method that returns an evm specific head, then it must start that way now.

pkg/types/relayer.go
jmank88: This API must be evolved in compatible ways. If these are duplicates, then either one must be removed, or one must be defined in terms of the other, e.g. as an alias.

pkg/types/relayer.go
jmank88: This should be documented somehow. Maybe an alias type?

pkg/types/chains/evm/evm.go
jmank88: Please expand this comment

pkg/types/relayer.go
Unheilbar: It's now used in chain_writer as a string, I was making this signature consistent with the rest of the codebase

pkg/types/relayer.go
```diff
@@ -100,10 +98,30 @@ type ChainService interface {
...
+
+	// GetTransactionFee retrieves the fee of a transaction in wei from the underlying chain's TXM
+	// If transaction is not finalized returns error
+	GetTransactionFee(ctx context.Context, transactionID IdempotencyKey) (*evm.TransactionFee, error)
+	LatestAndFinalizedHead(ctx context.Context) (latest evm.Head, finalized evm.Head, err error)
+	QueryLogsFromCache(ctx context.Context, filterQuery []query.Expression,
+		limitAndSort query.LimitAndSort, confidenceLevel primitives.ConfidenceLevel) ([]*evm.Log, error)
+	RegisterLogTracking(ctx context.Context, filter evm.LPFilterQuery) error
+	UnregisterLogTracking(ctx context.Context, filterName string) error
+	GetTransactionStatus(ctx context.Context, transactionID IdempotencyKey) (evm.TransactionStatus, error)
```
jmank88, line 117: What is the advantage of duplicating `GetTransactionStatus` and `TransactionStatus`?

pkg/types/relayer.go
```diff
@@ -100,10 +98,30 @@ type ChainService interface {
 	Replay(ctx context.Context, fromBlock string, args map[string]any) error
 }
 
+type EVMService interface {
```
jmank88, line 101: What is the rationale for when to include a `Get` prefix in the method name?

Unheilbar: GetTransactionStatus and GetTransactionsFee arent part of geth api, the rest of calls repeat geth api

Unheilbar: GetLogs is aligned with[geth](https://github.com/ethereum/go-ethereum/blob/d6655cb4506bf2e838eed5199b84bfc63bf4515a/eth/filters/api.go#L340)
It accepts evm-style filter query and will be converted into geth rpc call.
QueryLogs will query logs from  LogPoller's database
I'll add more comments, but as it's been pointed out above we shouldn't expose implementations details in this interface

pkg/types/relayer.go
```diff
@@ -100,10 +98,30 @@ type ChainService interface {
...
+	CallContract(ctx context.Context, msg *evm.CallMsg, confidence primitives.ConfidenceLevel) ([]byte, error)
+	GetLogs(ctx context.Context, filterQuery evm.FilterQuery) ([]*evm.Log, error)
```
ilija42, line 103: Lets make this FilterLogs to be more aligned with our internal evm  client 
https://github.com/smartcontractkit/chainlink-evm/blob/d3266955d38c4afa18602bda6e3ebf8b647d9900/pkg/client/chain_client.go#L80

pkg/types/relayer.go
ilija42: GetTransactionStatus shouldn't be included anyway since we aren't allowing the workflow authors to send txs in this way.

pkg/loop/internal/pb/evm/evm.proto
```diff
@@ -0,0 +1,229 @@
...
+  rpc GetLogs(GetLogsRequest) returns (GetLogsReply);
+  rpc BalanceAt(BalanceAtRequest) returns (BalanceAtReply);
+  rpc EstimateGas(EstimateGasRequest) returns (EstimateGasReply);
+  rpc GetTransactionByHash(GetTransactionByHashRequest) returns (GetTransactionByHashReply);
+  rpc GetTransactionReceipt(GetReceiptRequest) returns (GetReceiptReply);
+  rpc LatestAndFinalizedHead(google.protobuf.Empty) returns (LatestAndFinalizedHeadReply);
+  rpc QueryLogsFromCache(QueryLogsFromCacheRequest) returns (QueryLogsFromCacheReply);
+  rpc RegisterLogTracking(RegisterLogTrackingRequest) returns (google.protobuf.Empty);
+  rpc UnregisterLogTracking(UnregisterLogTrackingRequest) returns (google.protobuf.Empty);
+  rpc GetTransactionStatus(GetTransactionStatusRequest) returns (GetTransactionStatusReply);
```
ilija42, line 24: This is crossed out in the DD as the workflow author won't be able to send a tx in this way, at least for now

pkg/types/relayer.go
ilija42: My bad, we do need this on the relayer, but the capability interface will hide this

pkg/types/relayer.go
ilija42: Most of the calls are exposing our internal evm client where naming should mostly be aligned with geth https://github.com/smartcontractkit/chainlink-evm/blob/ee990aefea6c083b1fdd713160cab1ba2e3ebfa8/pkg/client/chain_client.go#L29

Everything else is calling into log poller or headtracker `LatestAndFinalizedHead `, `QueryLogsFromCache`, `RegisterLogTracking `, `UnregisterLogTracking`

pkg/types/relayer.go
```diff
@@ -100,10 +98,30 @@ type ChainService interface {
...
+
+	// GetTransactionFee retrieves the fee of a transaction in wei from the underlying chain's TXM
+	// If transaction is not finalized returns error
+	GetTransactionFee(ctx context.Context, transactionID IdempotencyKey) (*evm.TransactionFee, error)
+	QueryLogsFromCache(ctx context.Context, filterQuery []query.Expression,
+		limitAndSort query.LimitAndSort, confidenceLevel primitives.ConfidenceLevel) ([]*evm.Log, error)
+	RegisterLogTracking(ctx context.Context, filter evm.LPFilterQuery) error
+	UnregisterLogTracking(ctx context.Context, filterName string) error
+	GetTransactionStatus(ctx context.Context, transactionID IdempotencyKey) (TransactionStatus, error)
+}
```
ilija42, line 118: Lets also add `BatchCallContext(ctx context.Context, b []BatchElem) error`, its also defined on the evm client

pkg/types/relayer.go
jmank88: > where naming should mostly be aligned with geth

Why do you say that? I think internal consistency within this interface is important. The layer below is an implementation detail.

pkg/types/relayer.go
ilija42: Lets discuss this further and open another PR

pkg/types/relayer.go
ilija42: I think you misunderstood, I'm just saying that the layer below is mostly consistent with the geth client for context.
The `EVMService` layer should be as consistent with the EVMClient as it can, only the functionalities that don't exist on the EVMClient should diverge (which are the chain service functionalities like LP).

I didn't make a clear enough distinction between the EVM Chain Capability and the EVM Relayer Service in the DD. The ask for the capability API is to more closely follow geth client for UX, but this shouldn't affect the EVM Relayer.

jmank88: Why should anything diverge at all?

Regardless, this shouldn't be a secret. We need documentation that makes the situation clear to the users, whatever it is.

jmank88: If you literally intend to adapt the geth client (sub)interfaces, then I think we should rethink our field types, since we'll be doing conversions to and from strings that aren't even user facing.

ilija42: > Why should anything diverge at all?
The capability will diverge for CRE UX purposes, changing the evm client to match the naming is too much work

> Regardless, this shouldn't be a secret. We need documentation that makes the situation clear to the users, whatever it is.
The capability will have its own documentation, as for the EVM relayer I don't like exposing the implementation detail, what would you document here?

jmank88: If there are two distinct sets of methods here with different concerns and users, then I would define those interfaces separately (or at least the client set), and have this compose them:
```go
type EVMService interface {
	EVMClient
	...
```
Then each method set can be internally consistent. Perhaps the client methods don't need individual docs if the interface type explains that they are just proxying, but the other methods should say something. For example, when do I need to (un)register log tracking? Does it persist?

Unheilbar: Okay, to summarize this convo:

1. the methods exposing our EVMClient component will be separated under EVMClient interface{} inside EVMService. Their signature will be as consistent with [client component](https://github.com/smartcontractkit/chainlink-evm/blob/21113ff284bd29b5becfc2fb1a74e5d832a2dec5/pkg/client/chain_client.go#L29) as possible.

2.  methods exposing TXM and LogPoller remain as is with clarifying comments

3. chain-specific fields in types will be more explicit. Hashes/Addresses will be represented as bytes in .proto and [32]bytes/[20]bytes in types/chains/evm

pkg/types/relayer.go
```diff
@@ -100,10 +98,51 @@ type ChainService interface {
...
+
+	// UnregisterLogTracking removes a previously registered log filter by its name.
+	// After removal, logs matching this filter will no longer be collected or cached.
+	// noop guaranteed when filterName doesn't exist
+	UnregisterLogTracking(ctx context.Context, filterName string) error
+
+	// QueryLogsFromCache retrieves logs from the  log cache based on the provided
+	// query expression, sorting, and confidence level. It only returns logs that were
+	// collected through previously registered log filters.
+	QueryLogsFromCache(ctx context.Context, filterQuery []query.Expression,
```
ilija42, line 127: wdyt about calling this QueryTrackedLogs? Its not really a cache, its more of a log storage

pkg/types/relayer.go
```diff
@@ -100,10 +98,51 @@ type ChainService interface {
 	Replay(ctx context.Context, fromBlock string, args map[string]any) error
 }
 
+type EVMClient interface {
```
jmank88, line 101:
```suggestion
// GethClient is the subset of go-ethereum client methods implemented by EVMService.
type GethClient interface {
```

Unheilbar: Regular GethClient doesn't guarantee support of any evm-family chain, should I clarify that these methods are any evm-family chain compatible then?

jmank88: I'm not sure that I follow, but I suppose it could be declared in `package evm` so then `evm.GethClient` from here? Would that help?

jmank88 (approved):

ilija42 (approved):

PR #1164, opened 2025-04-29
Add capability server helper generation

nolag:


pkg/capabilities/capabilities.go
```diff
@@ -37,6 +37,9 @@ const (
 	CapabilityTypeAction    CapabilityType = "action"
 	CapabilityTypeConsensus CapabilityType = "consensus"
 	CapabilityTypeTarget    CapabilityType = "target"
+
+	// CapabilityTypeV2 allows capabilities to offer both trigger and executable types.
+	CapabilityTypeV2 CapabilityType = "v2"
```
pavel-raykov, line 42: Sorry, it is not clear what is going on.

First of all, it seems that you add a new type of Capability which allows to be both trigger and executable - why is it needed? (also "executable" capability is not defined - what do you mean?)

Second, v2 is a strange name (has nothing to do with the previous naming convention based on the capability function). Why did you settle on it?

nolag: As for why it's needed, the advantages of grouping related actions into one capability are that versioning is much simpler and easier to track. For example, if we update an RPC client used by a capability, you wouldn't want to have a potential world where the read and trigger use a different version of the client.

pavel-raykov: sorry, so what is the plan for combined capabilities: will they all be marked as CapabilityTypeV2 ? 

what about the interfaces? Will we have a separate interface for each combination of capabilities? (e.g., something that supports capabilities A, B and C must declare and implement interface AandBandC)

nolag: We renamed to Combined in another PR, this file won't show up here once I merge with main.

The notion that you only can do one thing with a capability won't be true anymore. If you have A, B, and C, they will be methods on your capability. See the (now merged) PR I linked to earlier.

ettec (dismissed):

ettec (approved):

PR #1173, opened 2025-05-05
Add mismatched package naming handling to protoc template generator

ilija42:


nolag (dismissed):

nolag (commented):
Please add an example that would fail to the test_capabilities folder so that this cannot be regressed.

nolag (approved):

jmank88 (approved):

PR #1174, opened 2025-05-05
Plex 131 evm capability Part 1

ilija42:
80% of the PR is re-generated grpc because of changed imports, which is why its split into two parts, where this part is mostly imports and the actual changes are in part 2 https://github.com/smartcontractkit/chainlink-common/pull/1202 . This was done to allow the chain capabilities to import the chain sdk and all the dependant code

The capability implementation is here
[EVM Capability implementation in capabilities repo](https://github.com/smartcontractkit/capabilities/pull/101)


pkg/types/relayer.go
```diff
@@ -100,10 +98,30 @@ type ChainService interface {
 	Replay(ctx context.Context, fromBlock string, args map[string]any) error
 }
 
+type EVMService interface {
+	CallContract(ctx context.Context, msg *evm.CallMsg, confidence primitives.ConfidenceLevel) ([]byte, error)
+	GetLogs(ctx context.Context, filterQuery evm.FilterQuery) ([]*evm.Log, error)
+	BalanceAt(ctx context.Context, account evm.Address, blockNumber *big.Int) (*big.Int, error)
+	EstimateGas(ctx context.Context, call *evm.CallMsg) (uint64, error)
+	TransactionByHash(ctx context.Context, hash evm.Hash) (*evm.Transaction, error)
```
dhaidashenko, line 106: How can a user obtain the transaction hash to use this method? 
Shouldn't it be TransactionByIdempotencyKey?

ilija42: This method isn't related to txm, its just a regular tx by hash call

pkg/loop/internal/pb/relayerset/relayerset.proto
```diff
@@ -128,6 +125,19 @@ service RelayerSet {
...
+  rpc EVMFilterLogs(EVMFilterLogsRequest) returns (loop.chain_capabilities.evm.FilterLogsReply);
+  rpc EVMBalanceAt(EVMBalanceAtRequest) returns (loop.chain_capabilities.evm.BalanceAtReply);
+  rpc EVMEstimateGas(EVMEstimateGasRequest) returns (loop.chain_capabilities.evm.EstimateGasReply);
+  rpc EVMTransactionByHash(EVMTransactionByHashRequest) returns (loop.chain_capabilities.evm.TransactionByHashReply);
+  rpc EVMTransactionReceipt(EVMReceiptRequest) returns (loop.chain_capabilities.evm.TransactionReceiptReply);
+  rpc EVMLatestAndFinalizedHead(LatestHeadRequest) returns (loop.chain_capabilities.evm.LatestAndFinalizedHeadReply);
+  rpc EVMQueryTrackedLogs(EVMQueryTrackedLogsRequest) returns (loop.chain_capabilities.evm.QueryTrackedLogsReply);
+  rpc EVMRegisterLogTracking(EVMRegisterLogTrackingRequest) returns (google.protobuf.Empty);
+  rpc EVMUnregisterLogTracking(EVMUnregisterLogTrackingRequest) returns (google.protobuf.Empty);
+  rpc EVMGetTransactionStatus(EVMGetTransactionStatusRequest) returns (GetTransactionStatusReply);
```
jmank88, line 140: Why don't we define these as a separate `EVMServiceSet` service, so we can drop all the `EVM` prefixes

pkg/loop/internal/pb/ccip/generate.go
```diff
@@ -1,12 +1,12 @@
-//go:generate protoc --proto_path=.:.. --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative commitstore.proto
+//go:generate protoc --proto_path=../../../../ --go_out=../../../../ --go_opt=paths=source_relative --go-grpc_out=../../../../ --go-grpc_opt=paths=source_relative loop/internal/pb/ccip/commitstore.proto
+//go:generate protoc --proto_path=../../../../ --go_out=../../../../ --go_opt=paths=source_relative --go-grpc_out=../../../../ --go-grpc_opt=paths=source_relative loop/internal/pb/ccip/gaspriceestimator.proto
```
ilija42, line 2: Chain capability can't import evm service without having the relative path start from pkg

pkg/capabilities/v2/chain-capabilities/evm/capability/capability.proto
```diff
@@ -0,0 +1,28 @@
...
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/chain-capabilities/evmcappb;evmcappb";
+
+package cre.sdk.v2.evmcap;
+
+import "google/protobuf/empty.proto";
+import "capabilities/v2/protoc/pkg/pb/cre_metadata.proto";
+import "capabilities/v2/chain-capabilities/evm/chain-service/evm.proto";
+
+service EVM {
```
ilija42, line 11: The evm capability defines a subset of method from the evm relayer service

pkg/loop/chain-common/proto_helpers.go
```diff
@@ -0,0 +1,388 @@
...
+	JSONEncodingVersion2
+	CBOREncodingVersion
+	ValuesEncodingVersion
+)
+
+const DefaultEncodingVersion = CBOREncodingVersion
+
+type ValueEncoder func(value any) (*VersionedBytes, error)
+
+func EncodeVersionedBytes(data any, version EncodingVersion) (*VersionedBytes, error) {
```
ilija42, line 40: I don't really like this chain-common package, but I'm out of ideas at the moment. I had to move the evm service out of internal package which created a bunch of cyclical and some internal imports, pulling codec and query out of internal fixes this as both contract reader and evm service depend on these (and a couple other files)

pkg/loop/internal/relayerset/evm.go
```diff
@@ -0,0 +1,372 @@
...
+
+	evmpb "github.com/smartcontractkit/chainlink-common/pkg/capabilities/v2/chain-capabilities/evm/chain-service"
+	chaincommonpb "github.com/smartcontractkit/chainlink-common/pkg/loop/chain-common"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop/internal/pb/relayerset"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+	valuespb "github.com/smartcontractkit/chainlink-common/pkg/values/pb"
+)
+
+// eVMClient wraps the EVMRelayerSetClient to attach RelayerID to EVMClient request.
+type eVMClient struct {
```
jmank88, line 18:
```suggestion
// evmClient wraps the EVMRelayerSetClient to attach RelayerID to EVMClient request.
type evmClient struct {
```

pkg/loop/internal/relayerset/client.go
```diff
@@ -22,12 +23,13 @@ type Client struct {
 
 	log logger.Logger
 
-	relayerSetClient relayerset.RelayerSetClient
+	relayerSetClient    relayerset.RelayerSetClient
+	eVMRelayerSetClient relayerset.EVMRelayerSetClient
```
jmank88, line 27:
```suggestion
	evmRelayerSetClient relayerset.EVMRelayerSetClient
```

Unheilbar (approved):

pkg/loop/chain-common/proto_helpers.go
jmank88: Do you mean that this package is only exported for other packages within the module to access? If so, we can still use an internal/ dir, it just has to be rooted high enough for everyone to access, e.g. `chainlink-common/internal/chains` or `chainlink-common/pkg/internal/chains`.

Unheilbar (approved):

pkg/loop/chain-common/proto_helpers.go
ilija42: The capability implementation has to import these too, so this has to be pulled out of internal. I don't like this too much because there is some codec stuff that doesn't have to be exported (at least for now), but moving it back creates cyclical import which means that I would have to split this up further

ilija42: I guess some of this could be moved to chainlink-common/internal/chains or chainlink-common/pkg/internal/chains.

ilija42: Made changes, now we are exporting only what we have to

jmank88 (dismissed):

Unheilbar (dismissed):

Unheilbar (approved):

jmank88 (dismissed):

jmank88 (approved):

Unheilbar (approved):

PR #1175, opened 2025-05-05
Add WASI runner and runtime for the CRE v2 SDK

nolag:


pkg/workflows/internal/v2/sdkimpl/runtime_test.go
```diff
@@ -0,0 +1,332 @@
...
+		require.NoError(t, err)
+		assert.True(t, ran)
+		assert.Equal(t, anyResult1+anyResult2, result)
+	})
+
+	t.Run("call capability errors", func(t *testing.T) {
+		test := func(rt sdk.DonRuntime, _ *basictrigger.Outputs) (string, error) {
+			workflowAction1 := &basicaction.BasicAction{}
+			call := workflowAction1.PerformAction(rt, &basicaction.Inputs{InputThing: true})
+			_, err := call.Await()
```
MStreet3, line 73: why does this error?

nolag: The capability isn't in the registry, so call capability returns an error. I'm testing that callback errors are handled correctly.

MStreet3 (approved):

pkg/workflows/sdk/v2/const.go
```diff
@@ -1,4 +1,7 @@
 package sdk
 
-// IdLen is 16 bytes to match a UUID's byte length
-const IdLen = 16
+const (
+	IdLen                       = 36              // IdLen is 36 bytes to match a UUID's string length
+	DefaultMaxResponseSizeBytes = 5 * 1024 * 1024 // 5 MB
+	ResponseBufferToSmall       = "response buffer too small"
```
bolekk, line 6: typo: "ToSmall" -> "TooSmall". Also could it be an error type not a string?

pkg/workflows/wasm/host/wasip1.go
```diff
@@ -22,6 +22,7 @@ func newWasiLinker(modCfg *ModuleConfig, engine *wasmtime.Engine) (*wasmtime.Lin
 	linker.AllowShadowing(true)
 
 	err := linker.DefineWasi()
+
```
bolekk, line 25: nit: remove to reduce noise?

pkg/workflows/wasm/pb/wasm.proto
```diff
@@ -24,6 +24,7 @@ message Request {
     ComputeRequest computeRequest = 3;
     google.protobuf.Empty specRequest = 4;
   }
+  string triggerId = 5;
```
bolekk, line 27: Why do we need this? This proto is not used by NoDAG, is it...?

pkg/workflows/wasm/v2/syncer_wasip1.go
```diff
@@ -0,0 +1,8 @@
+package wasm
```
bolekk, line 1: Why is this file named "syncer"?

pkg/workflows/wasm/v2/writer.go
```diff
@@ -0,0 +1,19 @@
+package wasm
```
bolekk, line 1: nit: you could probably combine utils, writer and syncer files in this package into one, to reduce noise.

pkg/workflows/internal/v2/sdkimpl/runtime.go
```diff
@@ -0,0 +1,110 @@
...
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/values"
+	valuespb "github.com/smartcontractkit/chainlink-common/pkg/values/pb"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk/v2"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk/v2/pb"
+)
+
+type CallCapabilityFn func(request *pb.CapabilityRequest) ([]byte, error)
```
bolekk, line 16: nit: name return types "(id []byte, err error)". It's not obvious that the byte array is an id.

pkg/workflows/internal/v2/sdkimpl/runtime.go
```diff
@@ -0,0 +1,110 @@
...
+type AwaitCapabilitiesFn func(request *pb.AwaitCapabilitiesRequest, maxResponseSize uint64) (*pb.AwaitCapabilitiesResponse, error)
```
bolekk, line 17: Does it make sense for the guest to always provide max response size? Why not enforce it in the host with a global default? We don't let guest provide other limits like timeout...

pkg/workflows/internal/v2/sdkimpl/runtime.go
```diff
@@ -0,0 +1,110 @@
...
+		awaitRequest := &pb.AwaitCapabilitiesRequest{
+			ExecId: r.ExecId,
+			Ids:    []string{string(id)},
+		}
+		awaitResponse, err := r.Await(awaitRequest, r.MaxResponseSize)
+		if err != nil {
+			return nil, err
+		}
+
+		capResponse, ok := awaitResponse.Responses[string(id)]
```
bolekk, line 51: Do we always expect len(Responses) == 1 here? Can you add a comment?

pkg/workflows/internal/v2/sdkimpl/runtime.go
```diff
@@ -0,0 +1,110 @@
...
+	d.modeErr = sdk.DonModeCallInNodeMode()
+	observation := fn(nrt)
+	nrt.modeErr = sdk.NodeModeCallInDonMode()
+	d.modeErr = nil
+	wrapped, _ := anypb.New(observation)
+
+	// TODO https: //smartcontract-it.atlassian.net/browse/CAPPL-816 use the generated consensus code
+	capabilityRequest := &pb.CapabilityRequest{
+		ExecutionId: d.ExecId,
+		Id:          "consensus@1.0.0",
```
bolekk, line 83: The ID is not right but I'm assuming that the TODO will handle that too.

pkg/workflows/wasm/pb/wasm.proto
```diff
@@ -24,6 +24,7 @@ message Request {
...
 }
 
 message ComputeResponse { capabilities.CapabilityResponse response = 1; }
```
bolekk, line 30: I think you can get rid of the protos down below under the comment:
`// TODO: v2 structs to be moved to a different package`

You added equivalents inside v2 directories and converted Module V2 interface.

pkg/workflows/wasm/host/module.go
```diff
@@ -295,13 +301,82 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
...
+
+		cfg:         modCfg,
+		stopCh:      make(chan struct{}),
+		isLegacyDAG: isLegacyDAG,
+	}
+
+	return m, nil
+}
+
+func linkNoDAG(linker *wasmtime.Linker, modCfg *ModuleConfig, execStore *store[*wasmpb.ExecutionResult]) error {
```
bolekk, line 329: We're losing "emit" - are you sure we don't need it? How was it used in DAG?

pkg/workflows/wasm/host/wasm_nodag_test.go
```diff
@@ -0,0 +1,140 @@
...
+	config := &basictrigger.Config{}
+	require.NoError(t, configProto.UnmarshalTo(config))
+	require.Equal(t, "name", config.Name)
+	require.Equal(t, int32(100), config.Number)
+
+	m, err := NewModule(mc, binary)
+	require.NoError(t, err)
+	m.Start()
+
+	// When a TriggerEvent occurs, Engine calls Run() with that Event.
```
bolekk, line 52: nit: not Run()

pkg/workflows/wasm/host/wasm.go
```diff
@@ -41,5 +42,33 @@ func GetWorkflowSpec(ctx context.Context, modCfg *ModuleConfig, binary []byte, c
 
 	m.Close()
 
-	return wasmpb.ProtoToWorkflowSpec(sr)
+	return legacywasmpb.ProtoToWorkflowSpec(sr)
+}
+
+func GetTriggersSpec(ctx context.Context, modCfg *ModuleConfig, binary []byte, config []byte) (*sdkpb.TriggerSubscriptionRequest, error) {
```
bolekk, line 48: This is only called from a test. Do we need it? Shouldn't the test just call Execute() to get triggers...? Why expose this "backdoor"?

pkg/workflows/wasm/v2/runner.go
```diff
@@ -0,0 +1,170 @@
...
+
+var _ sdk.DonRunner = &runner[sdk.DonRuntime]{}
+var _ sdk.NodeRunner = &runner[sdk.NodeRuntime]{}
+
+func (d *runner[T]) Run(args *sdk.WorkflowArgs[T]) {
+	// used to ensure that the export isn't optimized away
+	versionV2()
+	for _, handler := range args.Handlers {
+		// TODO: https://smartcontract-it.atlassian.net/browse/CAPPL-809 multiple of the same trigger registered
+		// The ID field could be changed to tirger-# and we can use the index or similar.
```
bolekk, line 55: nit: typo "tirger"

pkg/workflows/wasm/v2/runner.go
```diff
@@ -0,0 +1,170 @@
...
+				wrapped, err := values.Wrap(response)
+				if err != nil {
+					execResponse.Result = &pb.ExecutionResult_Error{Error: err.Error()}
+				} else {
+					execResponse.Result = &pb.ExecutionResult_Value{Value: values.Proto(wrapped)}
+				}
+			} else {
+				execResponse.Result = &pb.ExecutionResult_Error{Error: err.Error()}
+			}
+			marshalled, _ := proto.Marshal(execResponse)
```
bolekk, line 69: nit: still handle those errors? (and a few more places below)

pkg/workflows/wasm/v2/runner_test_hooks.go
```diff
@@ -0,0 +1,46 @@
...
+	t.Cleanup(func() {
+		lock.Lock()
+		defer lock.Unlock()
+		testTb = nil
+		executionId = ""
+		outstandingCalls = nil
+	})
+}
+
+var lock sync.Mutex
```
bolekk, line 33: Why are these all globals?

pkg/workflows/internal/v2/sdkimpl/runtime.go
nolag: In this case we do, but it's possible that we refactor once we add await helpers that let you await multiple values. r.Await in those cases could return a cached response. The only expectation is that the id you request is in the response.

pkg/workflows/internal/v2/sdkimpl/runtime.go
nolag: Agreed, I put something in place so that the tests could pass. Once we have capability generating the SDK, this will fix itself.

pkg/workflows/internal/v2/sdkimpl/runtime_test.go
nolag: I added a comment

pkg/workflows/wasm/pb/wasm.proto
nolag: To know which trigger fired. I'll add a TODO so we can support multiple of the same trigger. Currently, I'm using the id of the trigger to know which callback to make.

nolag: I'll revert, I allowed you to wrap V2 calls as V1 calls, but we're not going that route. I forgot to revert the file.

pkg/workflows/wasm/v2/writer.go
nolag: copy/pasted and modified from V1. I forgot to rename the file.

pkg/workflows/wasm/v2/syncer_wasip1.go
nolag: I copy/pasted from V1, but renamed and modified a lot. I missed renaming one file.

pkg/workflows/wasm/v2/writer.go
nolag: After the rename, the intent is clearer. X.go is a file that always compiles, X_test_hooks.go is meant to allow us to test, since we won't have the host, and X_wasip1 is the host callback(s).

pkg/workflows/wasm/host/wasm.go
nolag: Up to you. I thought it would be helpful for the engine, since it returns the correct types. It wasn't meant to be a backdoor. It calls Execute and decodes the response for you.

pkg/workflows/wasm/v2/runner.go
nolag: I don't know what we or the user would do with the error. It shouldn't be possible, given that we made the protos, so we know it'll work with marshal.

pkg/workflows/wasm/v2/runner_test_hooks.go
nolag: The goal of the test hooks are to allow testing to be as close as possible to the WASI calls.  To achieve this, I wrote a test method for each WASI call. The WASI calls are all exposed like C calls so they are global. Since they are global, I needed the global test locks.

pkg/workflows/internal/v2/sdkimpl/runtime.go
bolekk: nit: It would be cool to have a comment inline with what you just said :)

pkg/workflows/wasm/host/module.go
bolekk: probably OK for now

bolekk (approved):

MStreet3 (dismissed):

bolekk (approved):

PR #1176, opened 2025-05-05
migrate port allocator

krehermann:
by inspection, is it possible to rm [GetRandomPort](https://github.com/search?q=org%3Asmartcontractkit+%22GetRandomPort%22&type=code)


pkg/utils/ports.go
```diff
@@ -28,16 +18,9 @@ func IsPortOpen(t *testing.T, port string) bool {
...
-
-		// check port if port is open
-		if IsPortOpen(t, port) {
-			t.Log("found open port: " + port)
-			return port
-		}
-	}
-
+	p := freeport.GetOne(t)
+	strconv.Itoa(p)
```
jmank88, line 24:
```suggestion
	return strconv.Itoa(p)
```
And remove the panic below - I guess it being there is why this lack of return wasn't flagged?

george-dorin (approved):

jmank88 (approved):

PR #1177, opened 2025-05-05
CapabilityExecutor interface for Module/Host

bolekk:
Also adding a new field: CapabilityRequest.CapabilityId, which is now needed to be passed to the Engine.

### DIFF
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -19,6 +19,7 @@ import (
 	"github.com/bytecodealliance/wasmtime-go/v28"
 	"google.golang.org/protobuf/proto"
 
+	cappb "github.com/smartcontractkit/chainlink-common/pkg/capabilities/pb"
 	"github.com/smartcontractkit/chainlink-common/pkg/custmsg"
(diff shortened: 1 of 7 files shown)


MStreet3 (approved):

pkg/workflows/wasm/host/module.go
```diff
@@ -470,6 +478,10 @@ func (m *module) Execute(ctx context.Context, request *wasmpb.ExecuteRequest) (*
 	return nil, errors.New("not implemented")
 }
 
+func (m *module) SetCapabilityExecutor(handler CapabilityExecutor) {
```
patrickhuie19, line 481: What consumer benefits from this API? Consider an error return val?

`func (m *module) SetCapabilityExecutor(handler CapabilityExecutor) error {`

bolekk: The Engine calls it. Sure, we can return an error.

patrickhuie19 (dismissed):

MStreet3 (approved):

MStreet3 (approved):

nolag (approved):

patrickhuie19 (approved):

PR #1179, opened 2025-05-06
PLEX-500

Unheilbar:
Extension of https://github.com/smartcontractkit/chainlink-common/pull/1161
Wires QueryTrackedLogs and adds EVM-specific primitives

1.  https://github.com/smartcontractkit/chainlink-evm/pull/54
2. https://github.com/smartcontractkit/chainlink/pull/17660

starknet ref: c7b7169940f7d9d355b4cb9fc4326e3ee7be2d21


pkg/loop/internal/pb/evm/evm.proto
```diff
@@ -125,6 +125,73 @@ message Head {
   Hash parent_hash = 4;
 }
 
+
+
+message Expression {
```
ilija42, line 130: This overloading approach makes sense  👍

pkg/loop/internal/pb/evm/evm.proto
```diff
@@ -123,6 +123,74 @@ message Head {
...
+}
+
+message EventByTopicFilter {
+  uint64 topic = 1;
+  repeated HashValueComparator hashed_value_comparers = 2;
+}
+
+// evm-style primitives
+// can be used in combination with general primnitives for querying tracked logs
+message EVMPrimitive {
```
jmank88, line 185: Let's drop the prefix since it's in `package evm`:
```suggestion
message Primitive {
```

jmank88: Actually, it seems like we need some other qualifier since these are not general primitives at all. What are we trying to capture here? A union of query filters?

pkg/types/query/evm/primitives.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/types/chains/evm"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
+)
+
+type Address = evm.Address
+type Hash = evm.Hash
+
+type EVMVisitor interface {
```
jmank88, line 12: `evm.Visitor`
```suggestion
type Visitor interface {
```

pkg/types/query/evm/primitives.go
jmank88, line 1: Let's find a better name for this file. These types are not primitives.

jmank88: Ah they do implement the `Primitive` interface apparently, since we use that term as part of a cyclical type definition :thinking: 
Regardless, single file packages are typically names for the package since there is nothing else to distinguish: `evm.go`
Although, I would consider moving the whole package under `types/query/primitives/evm`

pkg/loop/internal/pb/evm/evm.proto
jmank88: > I can't use Primitive name since it will override general Primitive from contract_reader

Names outside of this package are not a collision because they are package qualified

pkg/types/query/evm/primitives.go
Unheilbar: hm.. I think ideal solution would be to keep Visitor and Primitive chain-agnostic
`type Visitor interface {
	Visit(Primitive)
}

type Primitive interface {
	Accept(v Visitor)
}` 
the parser can handle the types it supports by switching primitive type:
` Visit(p primitives.Primitive) {
	switch f := p.(type) {
	case *evm.AddressFilter:
		// convert to SQL filter or internal query model
	case *evm.EventSig:
		// ...
	default:
`
But it leads to broad changes across all the parsers
@ilija42 wdyt?

pkg/loop/internal/pb/evm/evm.proto
```diff
@@ -123,6 +123,74 @@ message Head {
...
+  repeated Expression expression = 2;
+}
+
+message And {
+  repeated Expression expr = 1;
+}
+
+message Or {
+  repeated Expression expr = 1;
+}
```
ilija42, line 149: why can't we import BooleanExpression from contract_reader.proto?

ilija42: Also moving the query proto defs from contract_reader.proto to query.proto probably makes sense

Unheilbar: > why can't we import BooleanExpression from contract_reader.proto?

Because Expression is overloaded in evm.proto so everything that contains Expression has to be overloaded as well.

ilija42 (approved):

pkg/loop/internal/pb/evm/evm.proto
```diff
@@ -123,6 +123,69 @@ message Head {
...
+
+
+// EXTPrimitive extends general primitives with evm-specific primitives 
+message EXTPrimitive {
```
jmank88, line 148: What does `EXT` stand for? If it is just shorthand for `Extended`, then go idiom is to use `Ext`
```suggestion
// ExtPrimitive extends general primitives with evm-specific primitives 
message ExtPrimitive {
```

jmank88: I was trying to come up with a better name for this, and now I am wondering, does it actually need to exist? Or could this type and `Primitive` be combined? As in:
```diff
message Primitive {
  oneof primitive {
+    loop.Primitive general_primitive = 1;
    ContractAddress contract_address = 2;
    EventSig event_sig = 3;
    EventByWord event_by_word = 4;
    EventByTopic event_by_topic = 5;
  }
}
```

jmank88 (approved):

reductionista (approved):

ilija42 (approved):

PR #1180, opened 2025-05-06
Register multiple triggers

MStreet3:
Previously the Trigger ID was just the Capability ID, this is insufficient to match trigger execution requests to the correct handler because the same capability can be configured for different callbacks.  This bug is resolved by requiring triggers to be identified by a monotocially increasing `uint64` corresponding to the index of the trigger in the subscription list.

This PR:
- refactors `Trigger` interface to expose a `CapabilityID() string` method
- matches incoming trigger executions to handlers by handler index in runner
- adds test of multiple handlers with the same `CapabilityID() string`


pkg/workflows/sdk/v2/handler.go
```diff
@@ -9,7 +9,8 @@ import (
 //
 //	The interface represents what trigger to listen to and how to handle its invocations
 type Handler[T any] interface {
-	Id() string
+	Id() uint64
```
bolekk, line 12: uint16 should be more than enough? Or uint32 if you prefer

pkg/workflows/wasm/host/wasm_nodag_test.go
```diff
@@ -90,6 +99,89 @@ func Test_NoDag_Run(t *testing.T) {
...
+		binary       = createTestBinary(nodagMultiTriggerBinaryCmd, nodagMultiTriggerBinaryLocation, true, t)
+	)
+
+	m, err := NewModule(mc, binary)
+	require.NoError(t, err)
+
+	m.Start()
+	defer m.Close()
+
+	t.Run("Subscribe to triggers", func(t *testing.T) {
```
MStreet3, line 118:
```suggestion
	t.Run("Subscribe to triggers with identical capability IDs", func(t *testing.T) {
```

pkg/workflows/internal/v2/testhelpers/test_workflow.go
```diff
@@ -17,6 +17,31 @@ func RunTestWorkflow(runner sdk.DonRunner) {
...
+	runner.Run(&sdk.WorkflowArgs[sdk.DonRuntime]{
+		Handlers: []sdk.Handler[sdk.DonRuntime]{
+			sdk.NewDonHandler(
+				basic.Trigger(TestWorkflowTriggerConfig()),
+				onTrigger,
+			),
+			sdk.NewDonHandler(
+				basic.Trigger(&basictrigger.Config{
+					Name:   "second-trigger",
+					Number: 200,
```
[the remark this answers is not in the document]
nolag: I see this was resolved, but I still see the number in the trigger.

pkg/workflows/wasm/host/wasm_nodag_test.go
```diff
@@ -23,17 +23,23 @@ import (
...
-	mc := createNoDagMc(t)
-	triggerID := "basic-test-trigger@1.0.0"
-	binary := createTestBinary(nodagBinaryCmd, nodagBinaryLocation, true, t)
+
+	var (
+		mc           = createNoDagMc(t)
+		triggerIndex = int(0)
+		capID        = (&basictrigger.Basic{}).Trigger(&basictrigger.Config{}).CapabilityID()
+		binary       = createTestBinary(nodagBinaryCmd, nodagBinaryLocation, true, t)
+	)
```
nolag, line 42: nit: just do mc := etc. It's more common.

nolag (dismissed):

nolag (approved):

jmank88 (approved):

PR #1181, opened 2025-05-07
migrate cron capability to v2 api

ettec:
Part of: https://smartcontract-it.atlassian.net/browse/CAPPL-792


pkg/capabilities/v2/protoc/pkg/pb/cre_metadata.proto
```diff
@@ -11,6 +11,7 @@ import "workflows/sdk/v2/pb/sdk.proto";
 message CapabilityMetadata {
   cre.sdk.v2.Mode mode = 1;
   string capability_id = 2;
+  bool supports_legacy_trigger_api = 3;
```
nolag, line 14: I was thinking this could be another option that lives on the method called `leagacy_default` or something similar

pkg/capabilities/v2/protoc/pkg/templates/server.go.tmpl
```diff
@@ -118,6 +124,11 @@ func (c *{{.GoName|LowerFirst}}Capability) RegisterTrigger(ctx context.Context,
                 input := &{{name .Input.GoIdent ""}}{}
                 return capabilities.RegisterTrigger(ctx, "{{(CapabilityId $service)}}", request, input, c.{{$service.GoName}}Capability.Register{{.GoName}})
         {{- end }}
+        {{- end }}
+        {{- if (SupportsLegacyTriggerApi $service) }}
+            case "":
+                // The capability supports the legacy trigger API, in the case of an empty method name invoke the legacy trigger API
+                return c.CronCapability.RegisterLegacyTrigger(ctx, request)
```
nolag, line 131: See my prior comment, then we don't hard code this for cron only. I'm also hoping that we can re-use the callback we already have for Trigger and use the same type (I'm assuming no workflow uses the actual execution time).

pkg/capabilities/v2/protoc/pkg/pb/cre_metadata.proto
```diff
@@ -16,3 +16,11 @@ message CapabilityMetadata {
 extend google.protobuf.ServiceOptions {
   CapabilityMetadata capability = 50000; // 50000-99999 is the range for custom options.
 }
+
+message CapabilityMethodMetadata {
+  bool map_to_untyped_trigger_api = 1;
```
nolag, line 21: Can you call this map to legacy and plug in into both trigger and action?

nolag (commented):
Something you updated broke complication for for the servers below. You may need to run make generate so they get regenerated.

pkg/capabilities/v2/protoc/pkg/pb/cre_metadata.proto
ettec: moved it to 'map_to_untyped_api'

nolag (approved):

nolag (approved):

MStreet3 (approved):

MStreet3 (approved):

PR #1184, opened 2025-05-08
wip set capability executor

MStreet3:

### DIFF
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -17,9 +18,9 @@ import (
 
 	"github.com/andybalholm/brotli"
 	"github.com/bytecodealliance/wasmtime-go/v28"
+	"github.com/google/uuid"
 	"google.golang.org/protobuf/proto"
 
-	cappb "github.com/smartcontractkit/chainlink-common/pkg/capabilities/pb"
 	"github.com/smartcontractkit/chainlink-common/pkg/custmsg"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
@@ -140,7 +141,7 @@ type ModuleV2 interface {
 // Implemented by the Engine
 type CapabilityExecutor interface {
 	// blocking call to the Engine
-	CallCapability(ctx context.Context, request *cappb.CapabilityRequest) (*cappb.CapabilityResponse, error)
+	CallCapability(ctx context.Context, request *sdkpb.CapabilityRequest) (*sdkpb.CapabilityResponse, error)
 }
 
 type module struct {
@@ -149,8 +150,9 @@ type module struct {
 	linker  *wasmtime.Linker
 	wconfig *wasmtime.Config
 
-	requestStore *store[*wasmdagpb.Response]
-	executeStore *store[*wasmpb.ExecutionResult]
+	requestStore    *store[*wasmdagpb.Response]
+	executeStore    *store[*wasmpb.ExecutionResult]
+	capabilityStore *store[*sdkpb.CapabilityResponse]
 
 	cfg *ModuleConfig
 
@@ -299,6 +301,10 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 		m: map[string]*RequestData[*wasmpb.ExecutionResult]{},
 	}
 
+	capStore := &store[*sdkpb.CapabilityResponse]{
+		m: map[string]*RequestData[*sdkpb.CapabilityResponse]{},
+	}
+
 	logger := modCfg.Logger
 	err = linker.FuncWrap(
 		"env",
@@ -323,8 +329,9 @@ func NewModule(modCfg *ModuleConfig, binary []byte, opts ...func(*ModuleConfig))
 		linker:  linker,
 		wconfig: cfg,
 
-		requestStore: requestStore,
-		executeStore: execStore,
+		requestStore:    requestStore,
+		executeStore:    execStore,
+		capabilityStore: capStore,
 
 		cfg:         modCfg,
 		stopCh:      make(chan struct{}),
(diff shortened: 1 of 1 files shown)


pkg/workflows/wasm/host/module.go
```diff
@@ -580,9 +579,46 @@ func runWasm[I idMessage, O proto.Message](
...
+		return fmt.Errorf("error wrapping callcap func: %w", err)
+	}
 	return errors.New("not implemented")
 }
 
+func toCallCapAsync(
+	capStore *store[*sdkpb.CapabilityResponse],
+	handler CapabilityExecutor,
+) func(ctx context.Context, req *sdkpb.CapabilityRequest) ([sdk.IdLen]byte, error) {
+	return func(ctx context.Context, req *sdkpb.CapabilityRequest) ([36]byte, error) {
```
nolag, line 597: nit: sdk.IdLen

pkg/workflows/wasm/host/module.go
```diff
@@ -580,9 +579,46 @@ func runWasm[I idMessage, O proto.Message](
...
+		callId := uuid.New()
+		idBuffer := make([]byte, sdk.IdLen)
+		copy(idBuffer, callId.String())
```
nolag, line 600: I would read the comment in sdk.IdLen. IIRC, I made it the size of uuid.NewString()

pkg/workflows/wasm/host/module.go
```diff
@@ -580,9 +579,46 @@ func runWasm[I idMessage, O proto.Message](
...
+			if err != nil {
+				resp = &sdkpb.CapabilityResponse{
+					Response: &sdkpb.CapabilityResponse_Error{
+						Error: err.Error(),
+					},
+				}
+			}
+			// TODO: Delete id on await
+			capStore.add(
+				hex.EncodeToString(idBuffer[:]),
```
nolag, line 613: same comment about how to string a UUID

pkg/workflows/wasm/host/module.go
```diff
@@ -580,9 +579,46 @@ func runWasm[I idMessage, O proto.Message](
...
+				&RequestData[*sdkpb.CapabilityResponse]{response: resp},
+			)
+		}()
+
+		return [36]byte(idBuffer), nil
```
nolag, line 618: nit: sdk.IdLen

pkg/workflows/wasm/host/module.go
```diff
@@ -580,9 +579,46 @@ func runWasm[I idMessage, O proto.Message](
...
+			resp, err := handler.CallCapability(ctx, req)
+			if err != nil {
+				resp = &sdkpb.CapabilityResponse{
+					Response: &sdkpb.CapabilityResponse_Error{
+						Error: err.Error(),
+					},
+				}
+			}
+			// TODO: Delete id on await
+			capStore.add(
```
nolag, line 612: why do you need a capstore? You don't need to get context at runtime. Can/should you use a channel so you can read the response async later?

pkg/workflows/wasm/host/module.go
```diff
@@ -580,7 +587,49 @@ func runWasm[I idMessage, O proto.Message](
...
+	}
+
+	// shadow any existing implementation with the one passed by the call
+	if err := m.linker.FuncWrap(
+		"env",
+		"call_capability",
+		createCallCapFn(m.cfg.Logger, m.executeStore, callCapAsync),
+	); err != nil {
+		return fmt.Errorf("error wrapping callcap func: %w", err)
+	}
```
nolag, line 602: As an FYI, it shouldn't change executables that are already running.

Why not have the linker take the module instead and set the handler on the module. Then you can link aginst the module and the callback will always use the latest executor.

pkg/workflows/wasm/host/module.go
```diff
@@ -580,7 +587,49 @@ func runWasm[I idMessage, O proto.Message](
...
+			ch := make(chan *sdkpb.CapabilityResponse, 1)
+			resp, err := handler.CallCapability(ctx, req)
+			if err != nil {
+				resp = &sdkpb.CapabilityResponse{
+					Response: &sdkpb.CapabilityResponse_Error{
+						Error: err.Error(),
+					},
+				}
+			}
+			// TODO: Delete id on await
```
nolag, line 626: nit: TODO should have a jira

pkg/workflows/wasm/host/module.go
```diff
@@ -580,7 +587,49 @@ func runWasm[I idMessage, O proto.Message](
...
+			ch <- resp
```
nolag, line 627: You should select  in case the context is cancelled. For example, if the executable stops running, we don't need to be blocked anymore.

pkg/workflows/wasm/host/module.go
```diff
@@ -580,7 +587,49 @@ func runWasm[I idMessage, O proto.Message](
 }
 
 func (m *module) SetCapabilityExecutor(handler CapabilityExecutor) error {
```
nolag, line 589: Something seems off to me. Why is it in the mod config and also something you can set later?

jmank88 (approved):

pkg/workflows/wasm/host/module.go
```diff
@@ -57,13 +57,14 @@ func (r *store[T]) add(id string, req *RequestData[T]) error {
 	return nil
 }
 
-func (r *store[T]) get(id string) (*RequestData[T], error) {
+func (r *store[T]) get(id string) (T, error) {
 	r.mu.Lock()
```
ettec, line 61: Presumably this should be using the read lock?

pkg/workflows/wasm/host/module.go
```diff
@@ -57,13 +57,14 @@ func (r *store[T]) add(id string, req *RequestData[T]) error {
 	return nil
 }
 
-func (r *store[T]) get(id string) (*RequestData[T], error) {
+func (r *store[T]) get(id string) (T, error) {
```
ettec, line 60: not a fan of using error to indicate not found, better to follow map semantics and return a bool

pkg/workflows/wasm/host/module.go
```diff
@@ -580,7 +603,73 @@ func runWasm[I idMessage, O proto.Message](
...
+		if err != nil {
+			resp = &sdkpb.CapabilityResponse{
+				Response: &sdkpb.CapabilityResponse_Error{
+					Error: err.Error(),
+				},
+			}
+		}
+
+		select {
+		case <-ctx.Done():
```
ettec, line 641: as the channel has a buffer of size 1, select stmt is unnecessary

pkg/workflows/wasm/host/module.go
```diff
@@ -580,7 +603,73 @@ func runWasm[I idMessage, O proto.Message](
...
+
+	for _, callId := range acr.Ids {
+		ch, err := m.capabilityCallStore.get(callId)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get call from store : %w", err)
+		}
+
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("failed to wait for capability response %s : %w", callId, ctx.Err())
```
ettec, line 662: on its own this message will not indicate which capability was timed out, maybe add the request to the item put in the capabilityCallStore so it can be add to this message (the engine could arguably do this as well, but here would be eaiser I think)

PR #1193, opened 2025-05-15
chore: adds unit tests and clearer error message on failure

MStreet3:
passing a nil concrete type to the transform did not previously return the expected error.  this PR names the expected error, checks for the case of a nil `*values.Map` and adds context to the error in other failure modes.

### DIFF
--- a/pkg/capabilities/utils.go
+++ b/pkg/capabilities/utils.go
@@ -11,6 +11,8 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/values"
 )
 
+var ErrNeitherValueNorAny = errors.New("neither value nor any provided")
+
 // UnwrapRequest extracts the input and config from the request, returning true if they were migrated to use pbany.Any values.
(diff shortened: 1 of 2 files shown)


pkg/capabilities/utils.go
```diff
@@ -59,15 +61,19 @@ func SetResponse(response *CapabilityResponse, migrated bool, value proto.Messag
 // FromValueOrAny extracts the value from either a values.Value or an anypb.Any, returning true if the value was migrated to use pbany.Any.
 func FromValueOrAny(value values.Value, any *anypb.Any, into proto.Message) (bool, error) {
 	if any == nil {
+		// Check if the underlying concrete value is nil
+		if v, ok := value.(*values.Map); ok && v == nil {
```
MStreet3, line 65: this isn't exhaustive, but covers the case that an empty config values map is passed in.

justinkaseman (approved):

nolag (approved):

justinkaseman (approved):

PR #1197, opened 2025-05-15
fix: ensures every call can be found by an await

MStreet3:
fixes an issue where some awaits could not find the matching initial call.  this was caused by putting the write of the call to the map into a goroutine.  fixed by ensuring the channel exists in the map before calling capability.

### DIFF
--- a/pkg/utils/matches/matches.go
+++ b/pkg/utils/matches/matches.go
@@ -1,0 +1,21 @@
+package matches
+
+import (
+	"context"
+
+	"github.com/stretchr/testify/mock"
+)
+
+func anyContext(_ context.Context) bool {
+	return true
+}
+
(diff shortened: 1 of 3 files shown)


nolag (dismissed):

pkg/utils/matches/matches.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+
+	"github.com/stretchr/testify/mock"
+)
+
+func anyContext(_ context.Context) bool {
+	return true
+}
+
+func anyString(_ string) bool {
```
krehermann, line 13: unused afaict. preemptive or cruft or oversight?

krehermann (approved):

justinkaseman (approved):

PR #1198, opened 2025-05-16
fix(capabilities): enable long running trigger response transform

MStreet3:
The `RegisterTrigger` function calls the typed `RegisterTrigger` method of a capability and then consumes the returned source channel of trigger events and transforms it into a channel of `TriggerResponse` events.

There was a bug in the `RegisterTrigger` transform in that canceling the parent context also canceled the goroutine transforming the source channel.

This PR addresses the issue by separating the request context from the goroutine by passing a stop channel, which is closed when the server is shutdown.  Adds a stop channel to the generated server, which is closed when the server is closed.

### DIFF
--- a/pkg/capabilities/utils.go
+++ b/pkg/capabilities/utils.go
@@ -119,6 +119,7 @@ type TriggerAndId[T proto.Message] struct {
 // while adhering to the standard capability interface.
 func RegisterTrigger[I, O proto.Message](
 	ctx context.Context,
+	stop <-chan struct{},
 	triggerType string,
 	request TriggerRegistrationRequest,
 	message I,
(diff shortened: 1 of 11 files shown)


pkg/capabilities/utils.go
```diff
@@ -161,10 +162,10 @@ func RegisterTrigger[I, O proto.Message](
 				}
 				select {
 				case response <- tr:
-				case <-ctx.Done():
+				case <-stop:
```
ettec, line 165: Is it not sufficient to rely on the open/closed state of the source channel?  Do we need this as well?

pkg/capabilities/utils.go
```diff
@@ -119,6 +119,7 @@ type TriggerAndId[T proto.Message] struct {
 // while adhering to the standard capability interface.
 func RegisterTrigger[I, O proto.Message](
 	ctx context.Context,
+	stop <-chan struct{},
```
jmank88, line 122:
```suggestion
```
I think adding a return arg, like a `close func()`, would be a more idiomatic option. Makes things a little more intuitive for the callers. Although I can see that not every caller would be simplified so :shrug:

nolag (approved):

PR #1200, opened 2025-05-21
CAPPL-881: Make capability call IDs deterministic

nolag:
[Core change](https://github.com/smartcontractkit/chainlink/pull/17851)


pkg/workflows/internal/v2/sdkimpl/runtime.go
```diff
@@ -11,44 +11,51 @@ import (
...
 
-	modeErr error
+	modeErr    error
+	Mode       pb.Mode
+	nextCallId int32
 }
 
 func (r *RuntimeBase) CallCapability(request *pb.CapabilityRequest) sdk.Promise[*pb.CapabilityResponse] {
+	if r.Mode == pb.Mode_DON {
+		r.nextCallId++
```
bolekk, line 31: Is this really safe? If a user writes a multithreaded workflow, is it guaranteed that a single-threaded WASM sandbox will always execute it in the same way?

pkg/workflows/sdk/v2/pb/sdk.proto
```diff
@@ -35,10 +35,10 @@ message ConsensusDescriptor {
 }
 
 message CapabilityRequest {
-  string executionId = 1;
```
bolekk, line 38: Why are you dropping the execution ID? Engine needs it to distinguish executions because it has only 1 host object that it receives capability calls from.

bolekk (changes requested):

pkg/workflows/internal/v2/sdkimpl/runtime.go
nolag: WASM is single threaded. If/when they allow multi-threading, you would need to enable it on the host. 

Adding locks alone wouldn't be sufficient if it wasn't, because if two threads raced and we couldn't guarantee order anymore and the ID wouldn't match.

pkg/workflows/sdk/v2/pb/sdk.proto
nolag: It was blindly passed from the WASM. I'll look at doing it differently.

bolekk: There's a TODO to use it in the Engine, I'm not sure where the host plugs it in before your change...
I agree that it has to be done in the host, not guest.

Requiring a separate callback object per execution also sounds OK to me but it's a larger change. Why not keep things as they were?

pkg/workflows/internal/v2/sdkimpl/runtime.go
nolag: Go routines seem very deterministic in WASM, they don't yield unless you do it explicitly or wait on a channel that doesn't have anything.

Given we're seeding random and time, I don't know where languages would be able to get non-determinism for routine/thread selection. Point taken though, I'm open to other ideas.

bolekk: You're probably right about the sources of non-determinism. I not super worried  right now but it would be cool to explore how exactly this behaves one day... Maybe leave a comment with a warning for now?

pkg/workflows/sdk/v2/pb/sdk.proto
nolag: It's a lot cleaner for the engine to be in control of the ID. It's the one that's aware of why it's used (to correlate requests across instances).

bolekk: I think there's value in having the execution ID in the host as well. Even if only for debugging purposes (we usually grep logs for the execution ID and get nice history of everything that happened). If you can, please keep it there.

nolag: We didn't use it for that yet. The right approach then is to return the ID in any errors in core. We can also have the logger be execution specific and decorate it. I'm a huge fan of separation of concerns. When we run workflows in TEEs, the less they need to understand the easier it'll be to isolate. We'll also keep user debugging cleaner in the simulator if it doesn't have bloat.

bolekk: SGTM. I don't mind modifying module.Execute() to take a callback handler, thanks.

bolekk (approved):

pkg/workflows/wasm/host/execution.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type execution[T any] struct {
+	fetchRequestsCounter     int
+	response                 T
+	ctx                      context.Context
+	capabilityResponses      map[int32]<-chan *sdkpb.CapabilityResponse
+	lock                     sync.RWMutex
+	module                   *module
+	internalToExternalCallId map[int32]string
```
MStreet3, line 18: what is this used for?

bolekk (approved):

MStreet3 (approved):

PR #1209, opened 2025-05-23
pkg/loop/cmd/loopinstall: don't clean filepath

jmank88:
Leading `./` prefix is meaningful with go. When it is trimmed, the path is interpreted as part of the standard library instead of the relative to the working directory in the current module.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/17867


chainchad:
Related https://github.com/smartcontractkit/chainlink-common/pull/1210

PR #1210, opened 2025-05-23
Fix loopinstall to install via local relative path

chainchad:
- Removes `filepath.Clean()` and env var expansion to simplify things.
- Prefer to use the relative local path for the `installPath` yaml input but will extract it if the full path (with moduleURI) is included for backwards compatibility.

### DIFF
--- a/pkg/loop/cmd/loopinstall/README.md
+++ b/pkg/loop/cmd/loopinstall/README.md
@@ -44,10 +44,15 @@ plugins:
     - name: "default"
       moduleURI: "github.com/smartcontractkit/chainlink-cosmos"
       gitRef: "f740e9ae54e79762991bdaf8ad6b50363261c056"
-      installPath: "github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/cmd/chainlink-cosmos"
+      installPath: "./pkg/cosmos/cmd/chainlink-cosmos"
       libs:
         - "/go/pkg/mod/github.com/!cosm!wasm/wasmvm@v*/internal/api/libwasmvm.*.so"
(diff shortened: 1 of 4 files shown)


pkg/loop/cmd/loopinstall/README.md
```diff
@@ -44,11 +44,16 @@ plugins:
...
       gitRef: "f740e9ae54e79762991bdaf8ad6b50363261c056"
-      installPath: "github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/cmd/chainlink-cosmos"
+      installPath: "./pkg/cosmos/cmd/chainlink-cosmos"
       libs:
         - "/go/pkg/mod/github.com/!cosm!wasm/wasmvm@v*/internal/api/libwasmvm.*.so"
 ```
 
+The `installPath` is relative to the root of the downloaded module but it can also be an absolute path like: `github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/cmd/chainlink-cosmos`. The absolute path will be stripped to the relative path (it is supported for backwards compatibility). The path is cleaned using Go's `filepath.Clean` function before use. This process normalizes the path by:
+
+- Resolving `.` (current directory) and `..` (parent directory) components (e.g., `path/to/../plugin` becomes `path/plugin`).
```
erikburt, line 54: Does the `../` path prefix, work? I don't see a test for it.

chainchad: Whoops, old comment. Thanks.

Updated in e278fadd

erikburt (approved):

PR #1220, opened 2025-05-29
fix(services): do not cancel context for fn

MStreet3:
`Engine.Go` launches a go routine to call `fn(ctx)` but immediately cancels the context once `fn` returns.  This is not correct as `fn` itself may launch go routines that assume they will only be canceled when the stop channel is closed.  The fix is to remove `defer cancel()` and only cancel the passed context when the service's stop channel is closed.

### DIFF
--- a/pkg/services/service.go
+++ b/pkg/services/service.go
@@ -69,8 +69,7 @@ func (e *Engine) Go(fn func(context.Context)) {
 	e.wg.Add(1)
 	go func() {
 		defer e.wg.Done()
-		ctx, cancel := e.StopChan.NewCtx()
-		defer cancel()
+		ctx, _ := e.StopChan.NewCtx()
 		fn(ctx)
 	}()
 }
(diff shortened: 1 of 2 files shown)


pkg/services/service.go
```diff
@@ -69,8 +69,7 @@ func (e *Engine) Go(fn func(context.Context)) {
 	e.wg.Add(1)
 	go func() {
 		defer e.wg.Done()
-		ctx, cancel := e.StopChan.NewCtx()
-		defer cancel()
+		ctx, _ := e.StopChan.NewCtx()
```
MStreet3, line 72: `StopChan` handles calling `cancel()`

jmank88: The parent cancellation is different. We also want to cancel this instance when the goroutine is complete. What are we trying to achieve here?

jmank88 (changes requested):

PR #1221, opened 2025-05-29
fix(capabilities): internally stop forwarding responses on unregister

MStreet3:
calling RegisterTrigger creates a new go routine that should be cancelled when a client calls UnregisterTrigger.  Currently the goroutine only lives as long as the request context, which is opinionated about how long the context should last.  This fix ties the lifecycle of the routine to the expected lifecycle of the trigger.


jmank88 (approved):

jmank88 (changes requested):
send on closed channel

krehermann (approved):

PR #1226, opened 2025-06-02
CRE-362: dynamic batching based on observation sizes

mchain0:
[CRE-362](https://smartcontract-it.atlassian.net/browse/CRE-362)

[CRE-362]: https://smartcontract-it.atlassian.net/browse/CRE-362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/capabilities/consensus/ocr3/reporting_plugin.go
+++ b/pkg/capabilities/consensus/ocr3/reporting_plugin.go
@@ -66,6 +66,7 @@ func (r *reportingPlugin) Query(_ context.Context, _ ocr3types.OutcomeContext) (
 	allExecutionIDs, serializedBatch, err := packToSizeLimit(
 		r.lggr,
 		QueriesSerializable{all},
+		ocr3types.MaxMaxQueryLength,
 	)
 	if err != nil {
 		r.lggr.Errorw("could not serialize query batch", "error", err)
@@ -104,6 +105,7 @@ func (r *reportingPlugin) Observation(ctx context.Context, outctx ocr3types.Outc
 	allExecutionIDs, obs, err := packToSizeLimit(
 		r.lggr,
 		ObservationSerializable{reqMap, registeredWorkflowIDs},
+		ocr3types.MaxMaxObservationLength,
 	)
 	if err != nil {
 		r.lggr.Errorw("could not serialize observations batch", "error", err)
@@ -302,6 +304,7 @@ func (r *reportingPlugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeC
 	allExecutionIDs, rawOutcome, err := packToSizeLimit(
 		r.lggr,
 		outcomeSerializable,
+		ocr3types.MaxMaxOutcomeLength,
 	)
 
 	previousOutcome.Outcomes = outcomeSerializable.previousOutcome.Outcomes
--- a/pkg/capabilities/consensus/ocr3/reporting_plugin_batching.go
+++ b/pkg/capabilities/consensus/ocr3/reporting_plugin_batching.go
@@ -208,7 +208,7 @@ func (o OutcomeSerializable) Mid(mid int) Serializable {
 			make(map[string]*pbtypes.AggregationOutcome),
 		}
 	}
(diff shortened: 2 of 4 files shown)


pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -57,15 +57,81 @@ func NewReportingPlugin(s *requests.Store, r CapabilityIface, batchSize int, con
...
+			mid = 1 // poor man's ceil
+		}
+		candidate := all[:mid]
+		executionIDs, serialized, err := r.serializeQueries(candidate)
+		if err != nil {
+			return nil, nil, err
+		}
+		size := len(serialized)
+
+		r.lggr.Debugw(
```
bolekk, line 90: nit: I'd avoid logging on each iteration, might be too noisy

pkg/capabilities/consensus/ocr3/reporting_plugin.go
```diff
@@ -80,10 +146,22 @@ func (r *reportingPlugin) Query(ctx context.Context, outctx ocr3types.OutcomeCon
...
+		r.lggr.Errorw("could not retrieve batch", "error", err)
+		return nil, err
+	}
+
+	r.lggr.Debugw("Query complete", "len", len(serializedBatch), "allExecutionIDs", allExecutionIDs)
+
+	return serializedBatch, nil
 }
 
 func (r *reportingPlugin) Observation(ctx context.Context, outctx ocr3types.OutcomeContext, query types.Query) (types.Observation, error) {
```
bolekk, line 167: apply also to observation and outcome

pkg/capabilities/consensus/ocr3/reporting_plugin_batching.go
```diff
@@ -0,0 +1,297 @@
...
+		weids,
+		newOutcomes,
+	}
+}
+
+// packToSizeLimit function maximizes the number of requests being included in a batch.
+// It finds the best utilization of space with the protobuf-marshalled structures using logarithmic (binary search)
+// approach to identify the optimal number of Requests that can be serialized without exceeding
+// the limit (defaultBatchSizeMiB).
+func packToSizeLimit(lggr logger.Logger, all Serializable) ([]string, []byte, error) {
```
bolekk, line 240: I'm starting to doubt that bin-search is actually useful.
  - Query elements are all of the same size so we can simply divide max by that size (minus some buffer).
  - Observations can be less predictable so that's the only place where bin search makes sense.
  - Outcome already does a relatively expensive computation for each request (aggregation). I can't imagine proto marshaling being meaningful compared to that. So we can simply marshal them one-by-one as we process and stop early if we happen to hit the limit (again, with some buffer).
  
As much as I appreciate the generalization of the approach behind a Serializable interface, I think it causes a lot of unnecessary churn in the existing code.

How about we handle Query and Outcome in a simpler way? If you want, you can still use bin-search for observations (maybe inline for simplicity). Or we could also run an experiment to measure how expensive it is to proto-marshal one-by-one until we hit the limit, compared to bin-search. If marshaling cost is proportional to input size then we can do the simple thing. Or maybe the overhead of every call is super high?

pkg/capabilities/consensus/ocr3/reporting_plugin_batching.go
```diff
@@ -0,0 +1,297 @@
...
+
+	optRound := 0
+
+	for low < high {
+		mid := (low + high) / 2
+		if mid == 0 {
+			mid = 1 // poor man's ceil
+		}
+		candidate := all.Mid(mid)
+		executionIDs, serialized, err := candidate.Serialize(lggr)
```
bolekk, line 261: We should probably be extra careful here in case the leader goes crazy and sends a very large query. Maybe a maxmax const for protection?

I also realized that we don't have deduplication of request IDs inside query - something we could add.

bolekk (changes requested):

PR #1229, opened 2025-06-02
[CAPPL-736] Beholder Logger

bolekk:
A new logger object that combines logging and sending events to Beholder Client, following the existing custmsg methods.


pkg/custmsg/mocks/message_emitter.go
jmank88, line 1: Is this only used from this module? If so, could keep it unexported, in a _test file

pkg/custmsg/logger.go
```diff
@@ -0,0 +1,151 @@
...
+	b.underlying.Errorw(msg, keysAndValues...)
+}
+
+func (b *beholderLogger) Panicw(msg string, keysAndValues ...any) {
+	_ = b.emitter.With(LogLevelKey, "panic").WithMapLabels(getLabelMap(keysAndValues...)).Emit(context.Background(), msg)
+	b.underlying.Panicw(msg, keysAndValues...)
+}
+
+func (b *beholderLogger) Fatalw(msg string, keysAndValues ...any) {
+	_ = b.emitter.With(LogLevelKey, "fatal").WithMapLabels(getLabelMap(keysAndValues...)).Emit(context.Background(), msg)
```
patrickhuie19, line 126: For other readers, `WithMapLabels` is additive: https://github.com/smartcontractkit/chainlink-common/blob/8aa3f7dc56eaf05a74000464f73a38f82e9fa8cd/pkg/custmsg/custom_message.go#L38

PR #1233, opened 2025-06-04
make gatewayConnector into a gRPC service

jinhoonbang:
- make gatewayConnector into a gRPC service
- move common Gateway code from core for re-use in capabilities repo

https://smartcontract-it.atlassian.net/browse/PRODCRE-409


pkg/loop/standard_capabilities.go
```diff
@@ -69,6 +69,7 @@ type StandardCapabilities interface {
 		pipelineRunner core.PipelineRunnerService,
 		relayerSet core.RelayerSet,
 		oracleFactory core.OracleFactory,
+		gatewayConnector core.GatewayConnector,
```
justinkaseman, line 72: This change will need to be carried through to each capability. Is there a fast follow planned for that?

pkg/loop/internal/pb/gatewayconnector/gateway_common.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+option go_package = "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/gatewayconnector";
+
+package loop;
+
+import "google/protobuf/empty.proto";
+
+message MessageBody {
```
bolekk, line 9: I don't like that the Message struct is duplicated as a proto. It was supposed to be an internal representation. Over the wire we send JSON-RPC-encoded messages. If @DeividasK adds his own version of Message, we would need to duplicate it too. 

Another option to consider is to send raw bytes in the gRPC service and decode in the Handler using our own JSON-RPC codec. WDYT? cc @cedric-cordenier 

Longer term, when we move to don2don framework for Gateways, we'll be passing raw bytes anyway.

pkg/loop/internal/pb/gatewayconnector/gateway_common.proto
```diff
@@ -0,0 +1,26 @@
...
+}
+
+message Message {
+  string signature = 1;
+  MessageBody body = 2;
+}
+
+message SendMessageRequest {
+  string gateway_id = 1;
+  Message message = 2;
```
bolekk, line 25: We could put a byte array here (see comment above).

pkg/loop/internal/pb/gatewayconnector/gateway_connector.proto
```diff
@@ -0,0 +1,41 @@
...
+  string don_id = 1;
+}
+
+message AddHandlerRequest {
+  repeated string methods = 1;
+  uint32 handler_id = 2;
+}
+
+service GatewayConnector {
+  rpc Start(google.protobuf.Empty) returns (google.protobuf.Empty);
```
bolekk, line 33: We don't need Start and Close, do we? Those are called locally (not over gRPC).

pkg/types/core/gateway_connector.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/gateway"
+)
+
+// GatewayConnector is a component run by Nodes to connect to a set of Gateways.
+type GatewayConnector interface {
+	// Start starts the GatewayConnector
+	Start(context.Context) error
```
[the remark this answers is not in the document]
jinhoonbang: I was under the impression that this interface is used in both core and capabilities. Are you suggesting we keep this interface to a minimum and define a broader interface like below? That sounds totally feasible too: 

```
type GatewayConnector interface {
	services.Service
	network.ConnectionInitiator
    core.GatewayConnector // narrow interface defined in chainlink-common for shared functionalities 
                                              // between core and capabilities repo
}
```

pkg/loop/standard_capabilities.go
jinhoonbang: yes can update all capabilities after this PR is merged

pkg/types/core/gateway_connector.go
```diff
@@ -0,0 +1,38 @@
+package core
+
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/gateway"
+)
+
+// GatewayConnector is a component run by Nodes to connect to a set of Gateways.
+type GatewayConnector interface {
```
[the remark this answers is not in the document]
jinhoonbang: draft PR: https://github.com/smartcontractkit/chainlink/pull/17892
ticket : https://smartcontract-it.atlassian.net/browse/PRODCRE-462

PR #1244, opened 2025-06-09
Allow consensus on time type, fix identical consensus big int, and rename the tag

nolag:


pkg/workflows/sdk/v2/consensus_aggregators.go
```diff
@@ -63,6 +64,8 @@ func ConsensusAggregationFromTags[T any]() ConsensusAggregation[T] {
 }
 
 var bigIntType = reflect.TypeOf((*big.Int)(nil))
+var timeType = reflect.TypeOf(time.Time{})
```
jmank88, line 67: Where do the comparison actually happen? Can you add a test for the case where unix nanos are equal but monotonic clock is different?

nolag: In the OCR plugin, this simply sends the data to the plugin. I can update the values.Value package to use a nano time and ensure it's encoded correctly :).

nolag: Done :)

jmank88: Do you mean that the comparison is done on the Value nanos, not with the `time.Time` type?

jmank88: Anything done on the unix nanos is fine. Just should avoid `==` and `Compare` with `time.Time`, which has an `Equals` method (and the extra nuance of monotonic time).

nolag: I'm a bit confused. The OCR plugin isn't done yet, we can compare for identical or sort however you recommend. This is just capturing the description of what the workflow author wants, and allowing them to ask for identical or median on time.

PR #1245, opened 2025-06-09
Add relayerset mock and extra validation to limitAndSort proto helpers

ilija42:


jmank88 (approved):

jmank88:
Where is the mock used? It is usually cleaner to generate them locally. Our habit of exporting mocks is not a great one, especially with the name collisions from using just `package mocks`. Nbd in this case since there is local precedent already though :shrug:

ilija42:
> Where is the mock used? It is usually cleaner to generate them locally. Our habit of exporting mocks is not a great one, especially with the name collisions from using just `package mocks`. Nbd in this case since there is local precedent already though 🤷

Using it for mocking the relayer in capabilities, could've added a mocks package there I guess

PR #1257, opened 2025-06-12
move common gateway code from core

jinhoonbang:
- move common gateway code from core for reuse in capabilities repo

https://smartcontract-it.atlassian.net/browse/PRODCRE-409


pkg/types/jsonrpc/codec.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+type Codec struct {
+}
+
+func (*Codec) DecodeRequest(requestBytes []byte, jwtTokenFromHeader string) (Request, error) {
```
bolekk, line 12: I don't fully follow - is this the new code that @DeividasK proposed in a draft PR?
Why aren't you moving more of existing Gateway types?

bolekk (approved):

pkg/types/gateway/ratelimiter.go
```diff
@@ -0,0 +1,69 @@
+package gateway
```
bolekk, line 1: nit (after seeing the core PR): maybe we can put it in some "lib"/"shared"/"utils" package instead?

jinhoonbang: makes sense. can definitely be used outside of gateway. how about a new top-level package? `pkg/ratelimit`

bolekk (approved):

jinhoonbang:
@jmank88 agreed. Seems like `pkg/ratelimit` is specific to rate limiting whereas `pkg/settings` would evolve to support different types of CRE settings in the near future? can converge later if needed.

pkg/types/gateway/round_robin_selector.go
```diff
@@ -0,0 +1,55 @@
+package gateway
+
+import (
+	"math/rand"
+	"sync"
+
+	"github.com/pkg/errors"
```
jmank88, line 7:
```suggestion
	"errors"
```

bolekk (approved):

bolekk (approved):

jmank88 (approved):

PR #1264, opened 2025-06-13
PLEX-250 - WriteReport initial implementation

pablolagreca:


pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,50 @@ type Head struct {
...
+
+type SignedReport struct {
+	RawReport     []byte
+	ReportContext []byte
+	Signatures    [][]byte
+	ID            []byte
+}
+
+// TransactionStatus is the result of the transaction sent to the chain
+type TransactionStatus int
```
jmank88, line 126: Do we need a new type for this? Other (evm) methods use a different type.

pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,50 @@ type Head struct {
...
+	TxID string
+}
+
+func (e *TxError) Error() string {
+	return fmt.Sprintf("Fail processing Transaction with internal TxID: %s", e.TxID)
+}
+
+type TransactionResult struct {
+	TxStatus TransactionStatus
+	TxHash   Hash
```
jmank88, line 148: The idea of a tx result containing a tx hash doesn't fit my mental model. Normally a tx hash is available before even sending, and the result or status has to do with inclusion. In this case the caller can't produce the hash since we construct the transaction, so I understand that we need to return it, but is this payload more like just a `Transaction` rather than a `Result` per se?
```suggestion
type Transaction struct {
	Status TransactionStatus
	Hash   Hash
```

pkg/types/chains/evm/evm.go
pablolagreca: We want a different abstraction that the one we have. TransactionResult is what's expected from a client in EVMService while the other TransactionStatus we have are: 
 - In contract_writer, which is specific to that implementation which will deprecate
 - In TXManager which has more states than what we should expose outside EVMService.

pkg/types/relayer.go
```diff
@@ -101,22 +101,15 @@ type ChainService interface {
 
 // GethClient is the subset of go-ethereum client methods implemented by EVMService.
 type GethClient interface {
```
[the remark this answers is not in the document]
pablolagreca: I'll revert this change since it's unrelated to this PR. Please check with ilija why he removed everything from this interface.

pkg/chains/evm/evm.proto
```diff
@@ -238,12 +240,40 @@ message QueryTrackedLogsReply {
...
+
+message SubmitTransactionRequest {
+  bytes to  = 1;
+  bytes data = 2;
+  optional GasConfig gas_config = 3;
+}
+
+message SubmitTransactionReply {
+  bytes txHash = 1;
+  TxStatus txStatus = 2;
```
jmank88, line 276:
```suggestion
  bytes tx_hash = 1;
  TxStatus tx_status = 2;
```

pkg/types/chains/evm/evm.go
jmank88: A quick follow up is fine if it's not affecting anyone yet, but keeping this API stable is important to reduce friction.

pkg/chains/evm/evm.proto
```diff
@@ -238,12 +240,40 @@ message QueryTrackedLogsReply {
   repeated Log logs = 1;
 }
 
-message GetTransactionStatusRequest {
-  string transaction_id = 1;
+enum TxStatus {
+  TX_SUCCESS = 0;
+  TX_REVERTED = 1;
```
silaslenihan, line 245: why do we need to distinguish between reverted and fatal here?

pkg/types/contract_writer.go
```diff
@@ -51,6 +51,7 @@ const (
 	Unknown TransactionStatus = iota
 	Pending
 	Unconfirmed
+	Confirmed
```
silaslenihan, line 54: Why do we need both Unconfirmed and Confirmed?

jmank88: Yeah if `Unconfirmed` means mined but not finalized, then I don't see what `Confirmed` is for :thinking:

pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,51 @@ type Head struct {
...
+	Signatures    [][]byte
+	ID            []byte
+}
+
+// TransactionStatus is the result of the transaction sent to the chain
+type TransactionStatus int
+
+const (
+	//Transaction was sent successfully to the chain and successfully executed
+	TxSuccess TransactionStatus = iota
```
jmank88, line 130: It might be risky to have Success be the default zero value. Is there a better option? What is the plan for Pending? That would be an improvement.

pkg/capabilities/v2/chain-capabilities/evm/capability.proto
```diff
@@ -39,4 +40,36 @@ service Client {
...
+message GasConfig {
+  uint64 gas_limit = 1;
+}
+
+message WriteReportReply {
+  loop.evm.TxStatus tx_status = 1;
+  optional ReceiverContractExecutionStatus receiver_contract_execution_status = 2;
+  optional bytes tx_hash = 3;
+  optional values.BigInt transaction_fee = 4;
+  optional string error_message = 5;
```
ilija42, line 74: you don't have to keep error message in the proto

pkg/chains/evm/evm.proto
```diff
@@ -238,12 +240,40 @@ message QueryTrackedLogsReply {
   repeated Log logs = 1;
 }
 
-message GetTransactionStatusRequest {
-  string transaction_id = 1;
+enum TxStatus {
```
ilija42, line 243: From my understanding we want to hide Pending from the workflow author, but this proto here is also used for the evm relayer

pkg/types/chains/evm/evm.go
```diff
@@ -84,6 +85,11 @@ type Transaction struct {
 	Value    *big.Int   // amount of eth sent in wei
 }
 
+type ReceiptGasInfo struct {
+	GasUsed           uint64   // actual gas used during execution in gas units
+	EffectiveGasPrice *big.Int // actual price in wei paid per gas unit
+}
```
ilija42, line 91: would be nice to have actual units in the comment

pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,51 @@ type Head struct {
...
+// TransactionStatus is the result of the transaction sent to the chain
+type TransactionStatus int
+
+const (
+	//Transaction was sent successfully to the chain and successfully executed
+	TxFatal TransactionStatus = iota
+	//Transaction was sent successfully to the chain but the smart contract execution reverted
+	TxReverted
+	//Transaction was not sent successfully to the chain
+	TxSuccess
```
ilija42, line 134:
```suggestion
	// Transaction was sent successfully to the chain and successfully executed
	TxFatal TransactionStatus = iota
	// Transaction was sent successfully to the chain but the smart contract execution reverted
	TxReverted
	// Transaction was not sent successfully to the chain
	TxSuccess
```

pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,51 @@ type Head struct {
...
+	Signatures    [][]byte
+	ID            []byte
+}
+
+// TransactionStatus is the result of the transaction sent to the chain
+type TransactionStatus int
+
+const (
+	//Transaction was sent successfully to the chain and successfully executed
+	TxFatal TransactionStatus = iota
```
ilija42, line 130: comment doesn't sound right

pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,51 @@ type Head struct {
...
+	//Transaction was sent successfully to the chain but the smart contract execution reverted
+	TxReverted
+	//Transaction was not sent successfully to the chain
```
ilija42, line 133: This comment also doesn't make sense to me

pkg/types/chains/evm/evm.go
ilija42: > FWIW geth JSON and GO clients return transaction types with extra fields set conditionally depending on status, so there is precedent for that 

Just want to +1 this, we should keep this familiar with geth ux

pkg/types/relayer.go
```diff
@@ -141,6 +140,12 @@ type EVMService interface {
 	// GetTransactionFee retrieves the fee of a transaction in wei from the underlying chain
 	GetTransactionFee(ctx context.Context, transactionID IdempotencyKey) (*evm.TransactionFee, error)
 
+	// WriteReport writes a transaction against a keystone forwarder contract
+	SubmitTransaction(ctx context.Context, txRequest evm.SubmitTransactionRequest) (*evm.TransactionResult, error)
```
ilija42, line 144:
```suggestion
	// SubmitTransaction writes a transaction against a keystone forwarder contract
	SubmitTransaction(ctx context.Context, txRequest evm.SubmitTransactionRequest) (*evm.TransactionResult, error)
```
Also is this comment even necessary?

pkg/types/chains/evm/evm.go
jmank88: It looks like success and fatal swapped spots but have each other's comments.
I think they still need to be more precise though. Are we waiting for finalization?

pkg/loop/internal/pb/relayerset/relayerset.proto
```diff
@@ -135,6 +135,8 @@ service EVMRelayerSet {
   rpc RegisterLogTracking(RegisterLogTrackingRequest) returns (google.protobuf.Empty);
   rpc UnregisterLogTracking(UnregisterLogTrackingRequest) returns (google.protobuf.Empty);
   rpc GetTransactionStatus(GetTransactionStatusRequest) returns (loop.evm.GetTransactionStatusReply);
+  rpc SubmitTransaction(SubmitTransactionRequest) returns (loop.evm.SubmitTransactionReply);
```
silaslenihan, line 138: It might be useful to have some docs on how to add a new chain capability operation. A list of all the APIs that need to be updated would be helpful if someone lacking context needs to make some changes.

silaslenihan (approved):

dhaidashenko (approved):

pkg/types/relayer.go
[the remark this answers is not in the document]
pablolagreca: I just reverted this change. it wasn't intentional.

pkg/capabilities/v2/chain-capabilities/evm/capability.proto
```diff
@@ -46,5 +47,37 @@ service Client {
   rpc QueryTrackedLogs(loop.evm.QueryTrackedLogsRequest) returns (loop.evm.QueryTrackedLogsReply);
   rpc RegisterLogTracking(loop.evm.RegisterLogTrackingRequest) returns (google.protobuf.Empty);
   rpc UnregisterLogTracking(loop.evm.UnregisterLogTrackingRequest) returns (google.protobuf.Empty);
-  rpc LogTrigger(FilterLogTriggerRequest) returns (stream loop.evm.Log);
+  rpc LogTrigger(FilterLogTriggerRequest) returns (stream loop.evm.Log);;
+  rpc WriteReport(WriteReportRequest) returns (WriteReportReply);
+
+}
+
+message SignedReport {
```
bolekk, line 55: FYI @ettec

pkg/types/chains/evm/evm.go
```diff
@@ -108,3 +114,51 @@ type Head struct {
 type TransactionFee struct {
 	TransactionFee *big.Int // Cost of transaction in wei
 }
+
+type SignedReport struct {
```
bolekk, line 118: Where is this used? We already have a proto, why do we need a Go struct as well?

bolekk (approved):

PR #1265, opened 2025-06-13
pkg/settings/limits: add package limits

jmank88:
https://smartcontract-it.atlassian.net/browse/CAPPL-931

This PR introduces `package limits` for enforcing limits. A `Limiter` can optionally be scoped to the Organization, User, or Workflow level, and any necessary metadata is obtained from the `context.Context`. Metrics are supported via otel gauges. Limits can be adjusted dynamically via an adapter to any source.

### DIFF
--- a/pkg/contexts/contexts.go
+++ b/pkg/contexts/contexts.go
@@ -1,0 +1,45 @@
+package contexts
+
+import "context"
+
+// Value gets a value from the context and casts it to T.
+func Value[T any](ctx context.Context, key any) T {
+	v := ctx.Value(key)
+	if v == nil {
+		var zero T
+		return zero
+	}
+	return v.(T)
+}
+
+type key string
+
+const (
+	keyOrg      key = "org"
+	keyUser     key = "user"
+	keyWorkflow key = "workflow"
+)
+
+func WithOrg(ctx context.Context, org string) context.Context {
+	return context.WithValue(ctx, keyOrg, org)
+}
+
+func OrgValue(ctx context.Context) (org string) {
+	return Value[string](ctx, keyOrg)
+}
+
+func WithUser(ctx context.Context, user string) context.Context {
+	return context.WithValue(ctx, keyUser, user)
+}
+
+func UserValue(ctx context.Context) (org string) {
+	return Value[string](ctx, keyUser)
+}
+
+func WithWorkflow(ctx context.Context, workflow string) context.Context {
+	return context.WithValue(ctx, keyWorkflow, workflow)
+}
+
+func WorkflowValue(ctx context.Context) (org string) {
+	return Value[string](ctx, keyWorkflow)
+}
--- a/pkg/settings/limits/limits.go
+++ b/pkg/settings/limits/limits.go
@@ -1,0 +1,261 @@
+// Package limits helps enforce request-scoped, multi-scope limits.
+package limits
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"golang.org/x/exp/constraints"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/settings"
+)
+
+// Number includes all integer and float types, although metrics will be emitted either as int64 or float64.
+type Number interface {
+	constraints.Integer | constraints.Float
+}
+
+// A Limiter limits usage of a resource by tracking allocations and ensuring that the threshold is not exceeded.
+// TODO Allocator.Alloc?
+// TODO Resource.Take? Return?
+type Limiter[N Number] interface {
+	// Use reserves amount of resources, or returns an error. The free function must be
+	// called to release resources, and should be deferred immediately when possible.
+	// Execution is typically blocked until resources are available (or context expires), but may short-circuit.
+	// ErrorLimitReached is returned when a limit is reached.
+	Use(ctx context.Context, amount N) (free func(context.Context), err error)
+}
+
+// Config holds optional configuration fields for Limiters.
+type Config[N Number] struct {
+	// GetLimit should return the current limit, for the given settings.Scope scope, or an error if none is set.
+	GetLimit settings.GetScoped[N]
+
+	// GaugeFn is an optional way to emit limit and usage metrics.
+	// - resource.*.limit
+	// - resource.*.used
+	GaugeFn func(string) (Gauge[N], error)
+
+	// If ShortCircuit is true, Limiter.Use will return immediately when there are not enough resources available,
+	// instead of waiting for them to become free.
+	ShortCircuit bool
+}
+
+// UnscopedLimiter returns an unscoped Limiter with default options.
+// See Config.UnscopedLimiter for dynamic limits, metering, and more.
+func UnscopedLimiter[N Number](key string, defaultLimit N) (Limiter[N], error) {
+	return Config[N]{}.UnscopedLimiter(key, defaultLimit)
+}
+
+// UnscopedLimiter returns an unscoped Limiter with the given Configuration.
+func (c Config[N]) UnscopedLimiter(key string, defaultLimit N) (Limiter[N], error) {
+	l := &unscopedLimiter[N]{
+		limiter: limiter[N]{
+			key:          key,
+			defaultLimit: defaultLimit,
+			shortCircuit: c.ShortCircuit,
+			getLimitFn: func(ctx context.Context) (N, error) {
+				return c.GetLimit(ctx, settings.ScopeGlobal, key)
+			},
+		},
+	}
+	return l, l.init(c.GaugeFn)
+}
+
+type limiter[N Number] struct {
+	key          string // required
+	defaultLimit N
+	shortCircuit bool
+	getLimitFn   func(context.Context) (N, error)
+
+	recordUsage Gauge[N]
+	recordLimit Gauge[N]
+
+	mu   sync.RWMutex
+	cond sync.Cond
+}
+
+func (l *limiter[N]) init(factoryFn func(string) (Gauge[N], error)) error {
+	if factoryFn == nil {
+		return nil
+	}
+	var err error
+	l.recordLimit, err = factoryFn("resource." + l.key + ".limit")
+	if err != nil {
+		return err
+	}
+	l.recordUsage, err = factoryFn("resource." + l.key + ".used")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (l *limiter[N]) getLimit(ctx context.Context) N {
+	limit, err := l.getLimitFn(ctx)
+	if err != nil {
+		//TODO log about fallback to default?
+		limit = l.defaultLimit
+	}
+	l.recordLimit(ctx, limit) //TODO can't from here w/o attributes....
+	return limit
+}
+
+type unscopedLimiter[N Number] struct {
+	limiter[N]
+
+	used N
+}
+
+func (g *unscopedLimiter[N]) Use(ctx context.Context, amount N) (func(context.Context), error) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	limit := g.getLimit(ctx)
+
+	for g.used+amount > limit {
+		if g.shortCircuit {
+			return nil, ErrorLimitReached[N]{
+				key:    g.key,
+				used:   g.used,
+				limit:  limit,
+				amount: amount,
+			}
+		}
+		g.cond.Wait()
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+	}
+	g.used += amount
+	return func(ctx context.Context) { g.free(ctx, amount) }, nil
+}
+
+func (g *unscopedLimiter[N]) free(ctx context.Context, amount N) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	g.used -= amount
+	g.recordUsage(ctx, g.used)
+	//TODO sanity check >0?
+
+	//TODO cache limit instead? (avoids risk of wrongly scoped context)
+	if g.used < g.getLimit(ctx) {
+		g.cond.Broadcast()
+	}
+	return
+}
+
+func OrgLimiter[N Number](key string, defaultValue N) (Limiter[N], error) {
+	return Config[N]{}.newScopedLimiter(settings.ScopeOrg, key, defaultValue)
+}
+
+func (c Config[N]) OrgLimiter(key string, defaultValue N) (Limiter[N], error) {
+	return c.newScopedLimiter(settings.ScopeOrg, key, defaultValue)
+}
+
+func UserLimiter[N Number](key string, defaultValue N) (Limiter[N], error) {
+	return Config[N]{}.newScopedLimiter(settings.ScopeUser, key, defaultValue)
+}
+
+func (c Config[N]) UserLimiter(key string, defaultValue N) (Limiter[N], error) {
+	return c.newScopedLimiter(settings.ScopeUser, key, defaultValue)
+}
+
+func WorkflowLimiter[N Number](key string, defaultValue N) (Limiter[N], error) {
+	return Config[N]{}.newScopedLimiter(settings.ScopeWorkflow, key, defaultValue)
+}
+
+func (c Config[N]) WorkflowLimiter(key string, defaultValue N) (Limiter[N], error) {
+	return c.newScopedLimiter(settings.ScopeWorkflow, key, defaultValue)
+}
+
+// TODO export?
+func (c Config[N]) newScopedLimiter(scope settings.Scope, key string, defaultValue N) (*scopedLimiter[N], error) {
+	l := &scopedLimiter[N]{
+		limiter: limiter[N]{
+			key:          key,
+			defaultLimit: defaultValue,
+			shortCircuit: c.ShortCircuit,
+			getLimitFn: func(ctx context.Context) (N, error) {
+				return c.GetLimit(ctx, scope, key)
+			},
+		},
+		scope: scope,
+	}
+	return l, l.init(c.GaugeFn)
+}
+
+// scopedLimiter extends limiter with a settings.Scope and enforces limits for each tenant separately.
+type scopedLimiter[N Number] struct {
+	limiter[N]
+	scope settings.Scope
+
+	used map[string]N
+}
+
+func (l *scopedLimiter[N]) Use(ctx context.Context, amount N) (func(context.Context), error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	limit := l.getLimit(ctx)
+
+	tenant := l.scope.Value(ctx)
+	if tenant == "" {
+		//TODO fallback to global?
+		return nil, fmt.Errorf("missing scope: %s", l.scope)
+	}
+
+	used := l.used[tenant]
+	for used+amount > limit {
+		if l.shortCircuit {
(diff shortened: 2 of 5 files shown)


pkg/contexts/contexts.go
jmank88, line 1: @patrickhuie19 This is all that I depend on directly, but we'll need GRPC support for these to be available everywhere.

pkg/settings/limits/limits.go
```diff
@@ -0,0 +1,308 @@
...
+	GetLimit settings.GetScoped[N]
+
+	// GaugeFn is an optional way to emit limit and usage metrics.
+	// - resource.*.limit
+	// - resource.*.used
+	GaugeFn func(string) (Gauge[N], error)
+
+	// If ShortCircuit is true, Limiter.Use will return immediately when there are not enough resources available,
+	// instead of waiting for them to become free.
+	ShortCircuit bool
```
pavel-raykov, line 44: would it make sense to use the given deadline from the context and abort if it reached instead of the ShortCircuit option?

jmank88: That is the default behavior when `ShortCircuit` is false

pkg/contexts/contexts.go
```diff
@@ -0,0 +1,45 @@
...
+
+type key string
+
+const (
+	keyOrg      key = "org"
+	keyUser     key = "user"
+	keyWorkflow key = "workflow"
+)
+
+func WithOrg(ctx context.Context, org string) context.Context {
```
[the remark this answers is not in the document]
jmank88: There is a hierarchical nature to the Scope that doesn't generalize to arbitrary keys, but we could export the parameterized method.

pkg/settings/scope.go
```diff
@@ -0,0 +1,53 @@
...
+	ScopeGlobal Scope = iota //TODO better name?
+	ScopeOrg
+	ScopeUser
+	ScopeWorkflow
+)
+
+func (s Scope) String() string {
+	switch s {
+	case ScopeGlobal:
+		return "none"
```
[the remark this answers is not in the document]
jmank88: Yeah, this was "unscoped" before I switched back to global. I'm still not sure that I like "global" though. I was avoiding it since I thought it could be misleading.

pkg/settings/limits/limits.go
jmank88: `ShortCircuit` means to give up (immeadiately) before the context deadline. The default behavior is to wait until the context deadline.

jmank88: Alternatively, we could provide a method variant e.g. `Try` like [Mutex.TryLock](https://pkg.go.dev/sync#Mutex.TryLock)

jmank88: I don't think that deliberately passing a context with an expired deadline ever makes sense, but in this case it would short circuit before even attempting to look up the limit.

jmank88: Another use case could be to calibrate resource allocations dynamically based on current usage. For example, imagine an org with concurrency limits. A single workflow related task might want to allocate all/most of that per-org capacity if possible, but will settle for less if the system is busy. In that case, you could `TryUse()` decreasing amounts until some min, which you settle for by blocking via `Use()` instead.

jmank88: We can leave it out, but I want to be forward thinking so we don't have to make breaking changes later.

jmank88: It's also much easier to test and demonstrate examples via `TryUse` :shrug:

pavel-raykov (dismissed):

pkg/contexts/grpc.go
jmank88, line 1: @patrickhuie19 Now it includes the GRPC `stats.Handler` too

pkg/loop/server.go
```diff
@@ -60,21 +61,19 @@ type Server struct {
 	DataSource      sqlutil.DataSource // optional
 	promServer      *PromServer
 	checker         *services.HealthChecker
+	LimitsFactory   limits.Factory
```
jmank88, line 64: LOOP Plugins can construct Limits from this factory, which is wired up to beholder metrics, and will have dynamic updates plugged in seamlessly later on.

pkg/settings/cresettings/settings.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+func init() {
+	err := InitKeys(&Config)
+	if err != nil {
+		log.Fatalf("failed to initialize keys: %v", err)
+	}
+}
+
+var Config = cfg{
```
jmank88, line 21: This is the entrypoint for using a pre-defined setting:
```go
limit := cresettings.Config.PerWorkflow.TriggerRateLimit
limiter, _ := limitsFactory.NewRateLimiter(limit)
```

pkg/settings/settings.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Setting holds a key, default value, and parsing function for a particular setting.
+// Use Setting.GetOrDefault with a Getter to look up settings.
+// Use Setting.Subscribe with a Registry to have updates pushed over a channel.
+type Setting[T any] struct {
+	Key          string
+	DefaultValue T
+	Scope        Scope
+	Parse        func(string) (T, error) `json:"-" toml:"-"`
```
jmank88, line 39: Using plain string values below this layer is simpler and more portable. However, depending on the file format, it can be difficult to get access to the raw string (e.g. TOML numbers), so currently all fields in the files are strings to avoid the issue. If we choose JSON, we should be able to use `json.Number` to support parsing the raw value. On the other hand, many types are not just simple numbers but instead formatted as strings with suffixes, so perhaps the quoted numbers in the TOML are acceptable.

pkg/settings/testdata/combined.toml
jmank88, line 1: This is an example of a combined set of files, with global overrides, and per org, owner, & workflow as well. This is what get's distributed and loaded in to memory.

pkg/settings/cresettings/defaults.json
jmank88, line 1: This is the defaults in JSON format (TOML option below). Each override file (global or per org/owner/workflow) would contain a subset of these fields as required (and permitted). For each level of scoping in the config, overrides can be set at that level and above. In other words:
- top level settings can only be overridden globally
- per-org settings can be overridden globally or for a particular org
- per-owner settings can be overridden globally, for an org, or for an owner
- per-workflow can be overridden globally, by org, by owner, or by workflow.
See `func rawKeys` for the logic to resolve these overrides.

pkg/settings/limits/rate.go
```diff
@@ -0,0 +1,641 @@
...
+			if s.usageCounter != nil {
+				s.usageCounter.Add(ctx, incr, withScope(ctx, s.scope))
+			}
+		},
+	}
+}
+
+func (s *scopedRateLimiter) Allow(ctx context.Context) bool {
+	_, l, err := s.get(ctx)
+	if err != nil {
```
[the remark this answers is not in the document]
jmank88: I don't think we have the same option of failing open like billing since it would overload the system, but we always fall back to the static default values anyways. The only other error path here is programmer error from missing context values.

pkg/settings/scope.go
```diff
@@ -0,0 +1,82 @@
...
+		return cre.Owner
+	case ScopeWorkflow:
+		return cre.Workflow
+	default:
+		return ""
+	}
+}
+
+// RoundCRE returns a modified CRE with out-of-scope tenants omitted.
+func (s Scope) RoundCRE(c contexts.CRE) contexts.CRE {
```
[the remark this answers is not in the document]
jmank88: When a setting is scoped to a specific level, in some situations we only want to compare tenants at that level and above, while ignoring the more more specific ones. So imagine a limiter scoped per Owner. The contexts passing through may contain workflow id, but we only care about detecting changes in org id, so we "round off" the more specific tenant(s), as in https://pkg.go.dev/time#Time.Round :shrug:

pkg/config/bytes_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	tests := []struct {
+		input    string
+		expected Size
+		valid    bool
+	}{
+		// valid
+		{"0", Size(0), true},
+		{"0.0", Size(0), true},
+		{"1.12345", Size(1), true},
```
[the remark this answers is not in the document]
jmank88: I'm surprised that it doesn't reject this. Should we reject values that we can't represent exactly too?

pkg/settings/limits/resource.go
```diff
@@ -0,0 +1,495 @@
...
+type ResourceLimiter[N Number] interface {
+	Limiter
+
+	// Use reserves amount of resources, or returns an error. The free function must be
+	// called to release resources, and should be deferred immediately when possible.
+	// Blocks until resources are available, or context has expired.
+	Use(ctx context.Context, amount N) (free func(), err error)
+
+	// TryUse is like Use, but returns immediately with ErrorResourceLimited when resources are not available.
+	TryUse(ctx context.Context, amount N) (free func(), err error)
```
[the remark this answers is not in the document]
jmank88: Perhaps verbose names are the most straightforward: e.g. Persistent vs Ephemeral Resource? Temporary?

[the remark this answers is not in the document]
jmank88: It could make sense to align this with the `RateLimiter` better too, which has explicit `Wait` semantics rather than `Try` variants.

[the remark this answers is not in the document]
jmank88: I think I like the simplicity of this shape (one method expecting negative values like an otel counter; extending the base interface), though I'm still open to alternative names:
```go
type PersistentResourceLimiter[N Number] interface {
	// Add adjusts the resource count by amount, or returns an error if the limit is reached.
	// It does not block. Use an ephemeralResource for blocking semantics.
	Add(ctx context.Context, amount N) error
}

type EphemeralResourceLimiter[N Number] interface {
	PersistentResourceLimiter[N]
	// Wait is like Add, but blocks until resources are available, or context has expired.
	// The free func must be called and should be deferred immediately when possible. It effectively calls Add(-N) to
	// release resources.
	Wait(context.Context, N) (free func(), err error)
}
```

pkg/services/waitgroup.go
```diff
@@ -0,0 +1,49 @@
...
+func (t *WaitGroup) Done() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.count--
+	if t.count == 0 {
+		t.cond.Signal()
+	}
+}
+
+// Wait halts futures additions and then blocks until all Done calls complete.
```
pavel-raykov, line 41: future

pkg/settings/limits/limits.go
jmank88: The std lib rate limiter uses an explicit `Wait` method for the blocking case, so I switched this over to align.

pkg/contexts/contexts.go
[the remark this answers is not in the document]
jmank88: These are grouped together on one `contexts.CRE` struct now.

pkg/settings/limits/time.go
```diff
@@ -0,0 +1,200 @@
...
+
+type timeLimiter struct {
+	*updater[time.Duration]
+	defaultTimeout time.Duration
+
+	key   string // optional
+	scope settings.Scope
+
+	limitGauge     metric.Int64Gauge     // optional
+	runtimeGauge   metric.Int64Histogram // optional
```
jmank88, line 97: I will switch these from int nanoseconds to float seconds

pkg/settings/limits/rate_test.go
```diff
@@ -0,0 +1,272 @@
...
+			assert.Error(t, rl.AllowNErr(ctx, time.Now(), 101))
+
+			time.Sleep(2 * pollPeriod) // ensure at least one update
+
+			attrs := attribute.NewSet(kvsFromScope(ctx, tt.scope)...)
+			ms := mc.lastResourceFirstScopeMetric(t)
+			require.Equal(t, metrics{
+				{
+					Name: "rate.foo.bar.limit",
+					Data: metricdata.Gauge[float64]{
```
jmank88, line 208: I will fill in unit and description too:
```suggestion
					Name: "rate.foo.bar.limit",
					Unit: "",
					Description "",
					Data: metricdata.Gauge[float64]{
```

pkg/settings/cresettings/settings.go
jmank88, line 1: Do these units make sense? https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

pkg/config/bytes.go
```diff
@@ -0,0 +1,92 @@
+package config
```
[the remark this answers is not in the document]
pavel-raykov: Sure, this seems like a pretty simple implementation, so it is fine to keep our own,

pavel-raykov (approved):

pavel-raykov (approved):

pavel-raykov (dismissed):

pkg/contexts/grpc.go
```diff
@@ -0,0 +1,101 @@
...
+// CREStreamInterceptor is a [grpc.StreamInterceptor] that converts CRE context values to GRPC metadata.
+func CREStreamInterceptor(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	ctx = appendOutgoingMetadata(ctx)
+	return streamer(ctx, desc, cc, method, opts...)
+}
+
+func appendOutgoingMetadata(ctx context.Context) context.Context {
+	cre := CREValue(ctx)
+	var kvs []string
+	if cre.Org != "" {
```
krehermann, line 33: comment on CRE struct says that Org might be empty if others are present.

does that mean we if `cre` is not zero value then both Owner and Workflow will be populated here?

pkg/contexts/grpc.go
```diff
@@ -0,0 +1,101 @@
...
+func (i *CREServerInterceptor) extractIncomingMetadata(ctx context.Context) context.Context {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return ctx
+	}
+	var cre CRE
+	if vs := md.Get(headerKeyOrg); len(vs) == 1 {
+		cre.Org = vs[0]
+	} else if len(vs) > 1 {
+		i.lggr.Errorw("GRPC header contains multiple orgs", "orgs", vs)
```
krehermann, line 80: this seems critically wrong? 

and we need an otel metric to alert on this too

pkg/contexts/grpc.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if vs := md.Get(headerKeyOwner); len(vs) == 1 {
+		cre.Owner = vs[0]
+	} else if len(vs) > 1 {
+		i.lggr.Errorw("GRPC header contains multiple owners", "owners", vs)
```
krehermann, line 85: similar

pkg/loop/telem.go
```diff
@@ -148,7 +159,7 @@ func (config TracingConfig) NewSpanExporter() (sdktrace.SpanExporter, error) {
 var grpcpromBuckets = []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120}
 
 // dialOptions returns [grpc.DialOption]s to intercept and reports telemetry.
-func dialOptions(r prometheus.Registerer) []grpc.DialOption {
+func dialOptions(lggr logger.Logger, r prometheus.Registerer, ) []grpc.DialOption {
```
krehermann, line 162: am i losing my mind? how is this compiling with the trailing comma? `func dialOptions(lggr logger.Logger, r prometheus.Registerer, )`

pkg/services/waitgroup.go
```diff
@@ -0,0 +1,54 @@
+package services
+
+import (
+	"errors"
+	"sync"
+)
+
+// WaitGroup is like [sync.WaitGroup], but TryAdd may be called after Wait and will return error rather than cause a race.
```
krehermann, line 8: what are the use cases for want or needing this?

add more doc comments so internal devs know if they ought to default to this impl or the standard one

pkg/settings/cresettings/defaults.json
```diff
@@ -0,0 +1,68 @@
...
+		"TriggerEventQueueLimit": "1000",
+		"TriggerEventQueueTimeout": "10m0s",
+		"ZeroBalancePruningTimeout": "24h0m0s"
+	},
+	"PerOwner": {
+		"ExecutionConcurrencyLimit": "50"
+	},
+	"PerWorkflow": {
+		"TriggerLimit": "10",
+		"TriggerRateLimit": "every30s,3",
```
krehermann, line 18: /nit i was confused about the embedded comma. this would have been obvious to me `every30s:3` or `3@every30s`

pkg/settings/cresettings/defaults.json
```diff
@@ -0,0 +1,68 @@
...
+		"HTTPTrigger": {
+			"RateLimit": "every30s,3",
+			"AuthRateLimit": "1rps,-1",
+			"IncomingPayloadSizeLimit": "10kb",
+			"OutgoingPayloadSizeLimit": "-1b"
+		},
+		"LogTrigger": {
+			"RateLimit": "every10s,-1",
+			"Limit": "5",
+			"EventRateLimit": "every-1s,-1",
```
krehermann, line 44: what is `every-1s`? compared to other entries it appears as a negative time.
what does `,-1` mean?

pkg/settings/settings.go
```diff
@@ -0,0 +1,258 @@
...
+		stop()
+		close(stopped)
+	}
+}
+
+// InitConfig accepts a pointer to a config struct and iterates over all the fields, initializing any Setting fields
+// with a full-qualified, dot-separated key.
+// Every field must either be a Setting or a nested struct following the same rules.
+func InitConfig(a any) error {
+	return initConfig(ScopeGlobal, "", a)
```
krehermann, line 207: does there need to be a lock around the scope? iiuc it looks like parallel calls to this could result in mangle config

pkg/loop/telem.go
jmank88: huh, I guess funcs are different somehow? It's not even a lack of formatting - it does not get simplified away :shrug:
Will remove though ofc

pkg/contexts/grpc.go
jmank88: As in just log critical level? I didn't want to fail the call since it might not depend on org.

pkg/contexts/grpc.go
jmank88: No I don't think we should be that strict. I was trying to deflect the intuition that all of the inferable information would be included. I think it would be fine to have only Org or Org+Owner in certain contexts. But I'll document that explicitly.

pkg/settings/settings.go
jmank88: Oh but scope is not modified. Reordered the args for clarity

pkg/contexts/grpc.go
krehermann: thx, yes i did not mean change the code, being flexible here is a good idea. was  curious about how it ought to work under normal circumstances

pavel-raykov (approved):

pkg/contexts/grpc.go
jmank88: I guess also consider that since we just drop them in this case, anything that depended on org would fail, and the failures should be associated already on some level.

PR #1266, opened 2025-06-16
Update SDK per the mental model

nolag:


pkg/workflows/wasm/v2/runner_wasip1.go
```diff
@@ -17,14 +16,8 @@ func versionV2()
...
-func NewDonRunner() sdk.DonRunner {
-	switchModes((int32)(sdkpb.Mode_DON))
-	return newDonRunner(runnerInternalsImpl{}, runtimeInternalsImpl{})
-}
-
-func NewNodeRunner() sdk.NodeRunner {
-	switchModes((int32)(sdkpb.Mode_Node))
-	return newNodeRunner(runnerInternalsImpl{}, runtimeInternalsImpl{})
+func NewRunner[T any](parse func(configBytes []byte) (T, error)) sdk.Runner[T] {
+	return newRunner[T](parse, runnerInternalsImpl{}, runtimeInternalsImpl{})
```
jmank88, line 20: nit/ could drop the name from the signature and just have the param say more, and would `C` be more intuitive for a config type than `T`? :shrug: 
```suggestion
func NewRunner[C any](parseConfig func([]byte) (C, error)) sdk.Runner[C] {
	return newRunner[C](parseConfig, runnerInternalsImpl{}, runtimeInternalsImpl{})
```
We could even use a decorative type (alias?) if we wanted to be super readable, but maybe that's unnecessary:
```suggestion
// Config is....
type Config any

func NewRunner[C Config](parseConfig func([]byte) (C, error)) sdk.Runner[C] {
	return newRunner[C](parseConfig, runnerInternalsImpl{}, runtimeInternalsImpl{})
```

nolag: Will for the first part. I realized I have more changes we discussed that I didn't add yet. I'll do that in the next commit with those changes.

pkg/workflows/sdk/v2/parsers.go
```diff
@@ -0,0 +1,11 @@
+package sdk
+
+import (
+	"encoding/json"
+)
+
+func ParseJson[T any](bytes []byte) (*T, error) {
```
jmank88, line 7:
```suggestion
func ParseJSON[T any](bytes []byte) (*T, error) {
```

jmank88 (approved):

PR #1279, opened 2025-06-19
generic jsonrpc2 library

prashantkumar1982:
Basic implementation of jsonrpc2 spec.

Taken from another PR https://github.com/smartcontractkit/chainlink/pull/18018

### DIFF
--- a/pkg/jsonrpc2/handler.go
+++ b/pkg/jsonrpc2/handler.go
@@ -1,0 +1,71 @@
+package jsonrpc2
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+type Handler struct {
+}
+
+func (*Handler) DecodeRequest(requestBytes []byte, jwtTokenFromHeader string) (Request, error) {
+	var request Request
+	err := json.Unmarshal(requestBytes, &request)
+	if err != nil {
+		return Request{}, err
+	}
+	if request.Version != JsonRpcVersion {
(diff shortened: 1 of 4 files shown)


bolekk (approved):

pkg/jsonrpc2/handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (*Handler) DecodeResponse(responseBytes []byte) (Response, error) {
+	var response Response
+	err := json.Unmarshal(responseBytes, &response)
+	if err != nil {
+		return Response{}, err
+	}
+	if response.Error != nil {
+		return Response{}, fmt.Errorf("received non-empty error field: %v", response.Error)
+	}
```
jinhoonbang, line 53: do we need this? I think this is saying DON nodes (Vault nodes, workflow nodes etc) shouldn't communicate JSONRPC errors to gateway nodes.

pkg/jsonrpc2/handler.go
```diff
@@ -0,0 +1,71 @@
...
+	if request.Params == nil {
+		return Request{}, ErrInvalidParams
+	}
+	if request.Auth != "" {
+		return request, nil
+	}
+
+	if jwtTokenFromHeader == "" {
+		return request, errors.New("missing auth token")
+	}
```
jinhoonbang, line 34: I wouldn't treat this as an error. For communication between gateway nodes and DON nodes, Messages without JWT should be considered valid

PR #1280, opened 2025-06-19
Improve Beholder Logger

bolekk:
1. Don't drop errors
2. Protect against blocking calls


pkg/custmsg/logger.go
```diff
@@ -3,28 +3,37 @@ package custmsg
...
+	emitter     MessageEmitter
+	emitTimeout time.Duration
 }
 
-func NewBeholderLogger(underlying logger.Logger, emitter MessageEmitter) *beholderLogger {
+func NewBeholderLogger(underlying logger.Logger, emitter MessageEmitter, emitTimeout time.Duration) *beholderLogger {
+	if emitTimeout == 0 {
+		emitTimeout = defaultBeholderEmitTimeoutMs * time.Millisecond
+	}
 	return &beholderLogger{
```
jmank88, line 26: Another option is to export the field for the caller to set themselves:
```go
type BeholderLogger struct {
	underlying  logger.Logger
	emitter     MessageEmitter
	EmitTimeout time.Duration
}

func NewBeholderLogger(underlying logger.Logger, emitter MessageEmitter) *BeholderLogger {
	return &beholderLogger{
```
Does introduce a bit of risk though since they need to set it before they start reading it concurrently, and then can't mutate it after that point.

jmank88: nbd but isn't it injected as an interface? This would only be callable on the returned exported type.

jmank88: Another option that doesn't change this signature is a `WithEmitTimeout` method. That would be a safe way to make it available downstream too.

bolekk: @jmank88 done

jmank88 (approved):

pkg/custmsg/logger.go
```diff
@@ -146,3 +253,8 @@ func getLabelMap(keysAndValues ...any) map[string]string {
 	}
 	return labels
 }
+
+func getBeholderCallContext() (context.Context, context.CancelFunc) {
```
krehermann, line 257: i thought the entire point of chip was to guarantee delivery of beholder logs.

i understand the technical reason for a timeout.

but, i am concerned this is shortcutting the product requirements and expectations.

what am i missing?

jmank88: IIUC otel treats logs and events as one kind of thing on the same stream

bolekk: Based on what I understood from @hendoxc and Mathew, Beholder Client will select a different endpoint to send the data to, under the hood.

patrickhuie19: All logs use the `BaseMessage` proto format and go over the custom message endpoint.

bolekk: So is it OK to merge? :)

nolag (approved):

PR #1282, opened 2025-06-20
Remove CRE Capability Aggregator scaling

cawthorne:
**Description:**
The EAs which feed data to the LLO Asset DON can provide answers pre-scaled by different amounts.

The LLO Asset DON can scale these values again to normalise them, but with the aggregator always scaling up by 1e18, the Asset DON would have to scale all feeds down to a non-scaled value. Typically we only scale upwards, not downwards to prevent precisions loss. Although this is a non-issue within the BigInt type.

Current behaviour is that the LLO Asset DON passes on the same scaling the EAs output when going through the `capability_trigger` report type. The reason that this behaviour has worked so far, is that the feeds we have used up to this point use EAs which do not send values pre-scaled.

We now are needing to deploy feeds where the EAs output values pre-scaled by e.g. 1e18 already. So this aggregator logic would over scale the end result.

We are merging in this PR:
https://github.com/smartcontractkit/chainlink/pull/18463

To allow us to multiply Streams and the Channel Definition stage instead.

We will now provide all LLO Stream scaled to 18 decimals before they leave the Asset DON.

**Rollout plan:**
LLO is not in Production Mainnet yet, so impact is limited. It is on the DF Prod Mainnet Asset DON, but not yet being used by any Prod Mainnet Workflow job.

LLO is on Prod Testnet, but is operating alongside V03.

Steps to migrate to the new behaviour would be be:

1. Merge this PR in to a core node image (not deployed yet).
2. Disable all LLO Workflow jobs on the currently targeted environment.
3. Apply the core node update to all DONs in that environment.
4. Apply Channel Definition scaling to ensure all LLO Streams are outputting at 18 decimal.
6. Re-enable LLO Workflow DON jobs.

Since V03 is still running on Prod Testnet and Prod Mainnet, there will not be customer facing down time. And we will leave V03 running until the LLO migration is fully complete.


pkg/capabilities/consensus/ocr3/datafeeds/llo_aggregator.go
```diff
@@ -275,15 +275,12 @@ type EVMEncodableStreamUpdate struct {
 	RemappedID []byte
 }
 
-// decimalShift is the number of decimal places to shift the decimal.Decimal to convert it to a big.Int in convert an llo price to a big.Int.
-const decimalShift = 18
```
bolekk, line 279: I don't get it. Isn't stuff going to break? What's the rollout plan?

cawthorne: LLO isn't in Production Mainnet yet, so impact is limited. It is on the DF Prod Mainnet Asset DON, but not yet being used by any Prod Mainnet Workflow job.

Steps would roughly be:
1. Get this change in a core node image (not deployed yet).
2. Edit CLO to read in scaling for LLO at the stream RDD template level, and do the same type of jobspec scaling as V03. (Alternatively we could edit the LLO `capability_trigger` channel type to allow per steam scaling. Probably this is better, but this also would take longer).
3. Disable all LLO Workflow jobs on the currently targeted environment.
4. Apply the core node update to all DONs in that environment.
5. Update that environments LLO Asset DON Stream job instances to have the jobspec scaling.
6. Re-enable LLO workflows.

If we went with adding per StreamID scaling to LLO `capability_trigger` channel type, then step 5 would be replaced with just updating the channel definition .json file to specify the scaling factors.

alecgard (dismissed):

cawthorne:
This PR is ready to merge.

Just waiting for the CI issues to be resolved.

mxiao-cll (approved):

justinkaseman (approved):

PR #1284, opened 2025-06-20
Create a standard test suite that all languages can run against to verify host-guest interactions

nolag:
Common tests directly using the callbacks to ensure the host works

### DIFF
--- a/pkg/chains/evm/evm.pb.go
+++ b/pkg/chains/evm/evm.pb.go
@@ -12,6 +12,7 @@ import (
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	emptypb "google.golang.org/protobuf/types/known/emptypb"
+	_ "google.golang.org/protobuf/types/known/timestamppb"
 	reflect "reflect"
 	sync "sync"
 	unsafe "unsafe"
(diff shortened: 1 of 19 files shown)


pkg/chains/evm/evm.pb.go
nolag, line 1: I'll revert this.

pkg/workflows/wasm/host/standard_test.go
```diff
@@ -0,0 +1,469 @@
...
+func runWithBasicTrigger(t *testing.T, executor ExecutionHelper) *pb.ExecutionResult {
+	trigger := &basictrigger.Outputs{CoolOutput: anyTestTriggerValue}
+	executeRequest := triggerExecuteRequest(t, 0, trigger)
+	m := makeTestModule(t)
+	response, err := m.Execute(t.Context(), executeRequest, executor)
+	require.NoError(t, err)
+	return response
+}
+
+func makeTestModule(t *testing.T) *module {
```
cedric-cordenier, line 361: @nolag Can you document how this works in the README? I was very confused and had to hunt around the code to figure out what was happening

pkg/workflows/wasm/host/standard_tests/internal/rawsdk/helpers_wasip1.go
```diff
@@ -0,0 +1,215 @@
+package rawsdk
```
cedric-cordenier, line 1: @nolag I'm not super keen on this -- I think this will just create a situation where the reference sdk will diverge from the ones we actually use

What are you trying to achieve by doing this?

pkg/workflows/wasm/host/standard_tests/Makefile
```diff
@@ -0,0 +1,7 @@
+# Declare all matching test.wasm targets as phony
+.PHONY: FORCE
+FORCE:
```
jmank88, line 3: Does this have an effect?

cedric-cordenier (dismissed):

pkg/workflows/wasm/host/standard_tests/Makefile
nolag: I couldn't get it to respect %/test.wasm as . PHONY and I wanted to ensure the go build decides when to rebuild instead of attempting to list all files.

ChatGPT came up with this solution. It tried a lot of other things to say that anything with .wasm should always be generated, but none of them worked.

jmank88 (commented):
Looks good after aligning Make/Do

mathewdgardner (approved):

PR #1285, opened 2025-06-20
Keystore capability service

vreff:
- Adds the keystore LOOPP service as a standard capability service.
- Adds a new implementation of keystore: `SingleAccountSigner` that accepts a single signer.
- Related to: https://github.com/smartcontractkit/chainlink/pull/18274


pkg/capabilities/pb/capabilities.proto
```diff
@@ -165,6 +165,7 @@ message InitialiseRequest {
   uint32 relayer_set_id = 7;
   uint32 oracle_factory_id = 8;
   uint32 gateway_connector_id = 9;
+  uint32 keyStore_id = 10;
```
cedric-cordenier, line 168: key_store_id :)

pkg/types/core/keystore.go
```diff
@@ -70,3 +70,30 @@ func (s *Ed25519Signer) Sign(r io.Reader, digest []byte, opts crypto.SignerOpts)
 	}
 	return s.signFn(ctx, s.account, digest)
 }
+
+// multiAccountSigner implements Keystore for multiple accounts. If a signer is not
+// found for a requested account, an error is returned.
+type multiAccountSigner struct {
+	accounts []string
```
cedric-cordenier, line 77: As I said in the other thread; I'm curious as to why you need this -- do you mind explaining?

Aside from that -- you're making an implicit assumption that accounts and signers are in the same order. Can you add a comment making this explicit?

vreff: Nice callout, will add a comment.

Requoting from the other PR:
```
The existing [Keystore](https://github.com/smartcontractkit/chainlink-common/blob/PRIV-2-capability-sign-service/pkg/types/core/keystore.go#L16-L21) interface I am using supports multiple accounts. While this PR only adds one account (P2P key), using an underlying struct that supports multiple accounts may easily permit future capability authors to use other keys.

That being said, the Keystore interface itself already allows this so really I don't need the implementation of that interface to overtly support multiple accounts at present.
```

Proceed in whichever thread you prefer. I am happy to define a "singleAccountSigner" implementation of the Keystore interface rather than a "multiAccountSigner" which would come at the benefit of no ambiguity around account/signer ordering. If we wanted more accounts for the capability to access later down the road, it would involve converting `singleAccountSigner` to `multiAccountSigner`, which has no backwards compatibility issues but is a _tad_ bit more work.

pkg/loop/internal/core/services/keystore/keystore.go
jmank88, line 1: Don't we already have an implementation of this?

pkg/loop/internal/core/services/capability/standard/standard_capabilities.go
```diff
@@ -33,7 +34,7 @@ type StandardCapabilities interface {
 	Initialise(ctx context.Context, config string, telemetryService core.TelemetryService, store core.KeyValueStore,
 		capabilityRegistry core.CapabilitiesRegistry, errorLog core.ErrorLog,
 		pipelineRunner core.PipelineRunnerService, relayerSet core.RelayerSet, oracleFactory core.OracleFactory,
-		gatewayConnector core.GatewayConnector) error
+		gatewayConnector core.GatewayConnector, keyStore core.Keystore) error
```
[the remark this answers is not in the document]
vreff: The implementation upstream is using the p2p key so I can call it `p2pKeystore`.

pkg/loop/internal/core/services/keystore/keystore.go
jmank88: I think it is just buried: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/loop/internal/relayer/relayer.go#L161

Should be fine to move though since it is all internal/.

vreff: Done: https://github.com/smartcontractkit/chainlink-common/pull/1285/commits/596c63136797d770012fe98940b5c31d9c68c415

jmank88 (approved):

cedric-cordenier (approved):

PR #1288, opened 2025-06-23
DON Time OCR Plugin

DylanTinianov:
Implements an OCR3 plugin for supporting [time in the CRE](https://docs.google.com/document/d/1Azo7e4niuCLxJxweg6z2j41sdVa6xbmY0bGNcdf69oA/edit?tab=t.0#heading=h.a0gpwirbwpmf).

[Jira Ticket](https://smartcontract-it.atlassian.net/browse/CAPPL-902)


pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	newInvalidRequestError := func(requestSeqNum int64, id string, currSeqNum int) error {
+		return fmt.Errorf("request number %d for id %s is greater than the number of observed don times %d", requestSeqNum, id, currSeqNum)
+	}
+
+	// A DON time beyond the expected number of requests is invalid, as there was never consensus on the prior request, which should be blocking.
+	var err error
+	for id, requestedSeqNumber := range observation.Requests {
+		if _, ok := priorOutcome.ObservedDonTimes[id]; !ok {
```
nolag, line 132: Get the value from the map once and re-use it on line 138

nolag (approved):

pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,270 @@
...
+
+		for id, currSeqNum := range observation.Requests {
+			if err := ctx.Err(); err != nil {
+				return nil, err
+			}
+			if _, ok := prevOutcome.ObservedDonTimes[id]; !ok {
+				prevOutcome.ObservedDonTimes[id] = &pb.ObservedDonTimes{}
+			}
+			if currSeqNum == int64(len(prevOutcome.ObservedDonTimes[id].Timestamps)) {
+				NumFinishedRequests[id]++
```
bolekk, line 180: It took me a very long time to understand this logic :) Please add a comment explaining the we only count requests for the next sequence number and ignore all other ones.

pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,270 @@
...
+			outcome.RemovedExecutionIDs[id] = true
+			delete(outcome.FinishedExecutionRemovalTimes, id)
+		}
+	}
+
+	return proto.Marshal(outcome)
+}
+
+func (p *workflowLibPlugin) Reports(ctx context.Context, seqNr uint64, outcome ocr3types.Outcome) ([]ocr3types.ReportPlus[struct{}], error) {
+	return nil, nil
```
bolekk, line 257: If you return nil here then nothing is going to be passed to the Transmitter. You need to pass unchanged outcome.

pkg/workflows/wasm/workflowLib/transmitter.go
```diff
@@ -0,0 +1,74 @@
...
+			request.SendResponse(ctx, DonTimeResponse{
+				WorkflowExecutionID: id,
+				SeqNum:              request.SeqNum,
+				Timestamp:           donTime,
+				Err:                 nil,
+			})
+		}
+	}
+
+	for id := range outcome.RemovedExecutionIDs {
```
bolekk, line 65: I don't think you need that. Just remove the ones that either expired or Engine reported as finished.

pkg/workflows/wasm/workflowLib/transmitter.go
```diff
@@ -0,0 +1,74 @@
...
+	if err := proto.Unmarshal(r.Report, outcome); err != nil {
+		return err
+	}
+
+	for id, observedDonTimes := range outcome.ObservedDonTimes {
+		t.store.SetDonTimes(id, observedDonTimes.Timestamps)
+	}
+	t.store.SetLastObservedDonTime(outcome.Timestamp)
+
+	requests, err := t.store.Requests.FirstN(t.batchSize)
```
bolekk, line 43: This is not right, you should iterate over what you have in your outcome, not over FirstN again.

pkg/workflows/wasm/workflowLib/factory.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/chainlink-common/pkg/workflows/wasm/workflowLib/pb"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+)
+
+const (
+	defaultMaxPhaseOutputBytes     = 1000000 // 1 MB
+	defaultMaxReportCount          = 20
```
bolekk, line 18: You always use a single report. This could be set to 1.

pkg/workflows/wasm/workflowLib/factory.go
```diff
@@ -0,0 +1,106 @@
...
+	defaultBatchSize               = 1000
```
bolekk, line 19: This can be much larger, maybe 10000? Each request is a 64 byte string and a number.

pkg/workflows/wasm/workflowLib/factory.go
```diff
@@ -0,0 +1,106 @@
...
+	defaultOutcomePruningThreshold = 3600
```
bolekk, line 20: unused? you don't do any outcome pruning

pkg/workflows/wasm/workflowLib/pb/dontime.proto
```diff
@@ -0,0 +1,25 @@
...
+
+message ObservedDonTimes {
+repeated int64 timestamps = 1;
+}
+
+message Outcome {
+int64 timestamp = 1;
+map<string, ObservedDonTimes> observed_don_times = 2;
+map<string, int64> finished_execution_removal_times = 3;
+map<string, bool> removedExecutionIDs = 4;
```
bolekk, line 24: I don't think you need either (3) or (4) - see my comments in plugin and transmitter.

bolekk (changes requested):

pkg/workflows/wasm/workflowLib/dontime_request.go
```diff
@@ -0,0 +1,66 @@
+package workflowLib
```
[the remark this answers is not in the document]
DylanTinianov: Sounds good, I chose the name since we discussed making it more generic and extendable. However, I agree keeping the plugin for DonTime makes sense and is the more simple approach.

pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,270 @@
...
+	var times []int64
+
+	for _, ao := range aos {
+		observation := &pb.Observation{}
+		if err := proto.Unmarshal(ao.Observation, observation); err != nil {
+			return nil, err
+		}
+
+		prevOutcome := &pb.Outcome{}
+		if err := proto.Unmarshal(outctx.PreviousOutcome, prevOutcome); err != nil {
```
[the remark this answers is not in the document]
DylanTinianov: Nice catch!

pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,270 @@
...
+func (p *workflowLibPlugin) ObservationQuorum(_ context.Context, _ ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation) (quorumReached bool, err error) {
+	return quorumhelper.ObservationCountReachesObservationQuorum(quorumhelper.QuorumTwoFPlusOne, p.config.N, p.config.F, aos), nil
+}
+
+func (p *workflowLibPlugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation) (ocr3types.Outcome, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	NumFinishedRequests := map[string]int64{} // counts how many nodes reported where a new DON timestamp might be needed
```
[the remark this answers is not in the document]
DylanTinianov: Yeah that's a better name, I just kept the one from Ryan's PoC

pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,270 @@
...
+			observedDonTimes, ok := outcome.ObservedDonTimes[id]
+			if !ok {
+				observedDonTimes = &pb.ObservedDonTimes{}
+			}
+			observedDonTimes.Timestamps = append(observedDonTimes.Timestamps, donTime)
+			outcome.ObservedDonTimes[id] = observedDonTimes
+		}
+	}
+
+	// Check if consensus is reached on the workflow execution being finished
```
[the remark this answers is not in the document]
DylanTinianov: We want consensus on when to delete an executionID since some nodes may be behind and still making time calls for a workflow. We still want them to be able to complete without errors from missing the observed time.

pkg/workflows/wasm/workflowLib/plugin.go
```diff
@@ -0,0 +1,270 @@
...
+			}
+			continue
+		}
+		seqNum := len(times.Timestamps)
+		if requestedSeqNumber > int64(seqNum) {
+			err = errors.Join(err, newInvalidRequestError(requestedSeqNumber, id, seqNum))
+		}
+	}
+
+	return err
```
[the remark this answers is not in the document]
DylanTinianov: Good point. I think we should actually verify this in Observation() by delivering an error and not including it in the request batch right away.

pkg/workflows/dontime/request.go
```diff
@@ -0,0 +1,66 @@
...
+func (r *Request) SendResponse(ctx context.Context, resp DonTimeResponse) {
+	select {
+	case <-ctx.Done():
+		return
+	case r.CallbackCh <- resp:
+		close(r.CallbackCh)
+	}
+}
+
+func (r *Request) SendTimeout(ctx context.Context) {
```
bolekk, line 37: Can we make sure that SendTimeout and SendResponse are non-blocking and remove context from them? We generally shouldn't be calling anything blocking from inside OCR plugin. In this case we simply need to make the channel of size 1 (which it already is) and make sure that we don't respond more than once (which I'm almost sure we do...?).

pkg/workflows/dontime/transmitter.go
```diff
@@ -0,0 +1,62 @@
...
+	for executionID, donTimes := range outcome.ObservedDonTimes {
+		request := t.store.GetRequest(executionID)
+		if request == nil {
+			continue
+		}
+
+		if len(donTimes.Timestamps) > request.SeqNum {
+			donTime := donTimes.Timestamps[request.SeqNum]
+			t.store.requests.Evict(executionID) // Make space for next request before delivering
+			request.SendResponse(ctx, DonTimeResponse{
```
bolekk, line 48: One downside of doing it here is that if a node is behind on multiple requests, it will need to wait for a full OCR round for every single one. There is an alternative solution where you cache all future times locally to avoid that. I don't think it's necessarily worth the complexity but maybe leave a comment explaining that for future readers?

pkg/workflows/dontime/plugin.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (p *Plugin) Outcome(_ context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation) (ocr3types.Outcome, error) {
+	observationCounts := map[string]int64{} // counts how many nodes reported where a new DON timestamp might be needed
+	finishedNodes := map[string]int64{}     // counts number of nodes finished with the workflow for executionID
+	var times []int64
+
+	prevOutcome := &pb.Outcome{}
+	if err := proto.Unmarshal(outctx.PreviousOutcome, prevOutcome); err != nil {
+		return nil, err
```
bolekk, line 147: same as above

pkg/workflows/dontime/plugin.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	for _, ao := range aos {
+		observation := &pb.Observation{}
+		if err := proto.Unmarshal(ao.Observation, observation); err != nil {
+			return nil, err
```
bolekk, line 153: This is even more problematic - if one malicious notes sends us crap, we fail the whole plugin. We have to ignore the error here and continue.

pkg/workflows/dontime/plugin.go
```diff
@@ -0,0 +1,256 @@
...
+				observedDonTimes = &pb.ObservedDonTimes{}
+			}
+			observedDonTimes.Timestamps = append(observedDonTimes.Timestamps, donTime)
+			outcome.ObservedDonTimes[id] = observedDonTimes
+		}
+	}
+
+	// Check if consensus is reached on the workflow execution being finished
+	for id, numFinished := range finishedNodes {
+		if numFinished >= int64(p.config.F) {
```
bolekk, line 216: F+1

pkg/workflows/dontime/plugin.go
```diff
@@ -0,0 +1,256 @@
...
+			if _, ok := outcome.FinishedExecutionRemovalTimes[id]; ok {
+				continue
+			}
+			outcome.FinishedExecutionRemovalTimes[id] = donTime + p.offChainConfig.ExecutionRemovalTime.AsDuration().Milliseconds()
```
bolekk, line 220: I still don't think that reaching consensus on finished requests is a great idea. For example, if a whole DON gets restarted, request store gets wiped but previous outcome stays the same (it's persisted in the DB). So some request can stay there forever, essentially causing a memory leak.

My preference would be to go over all requests in previous outcome here and remove those that are older than configured ExecutionRemovalTime. Same in transmitter - evict everything from the store that is expired (alternatively could be a separate goroutine that does that async). It will also simplify the plugin and store.

One downside of my solution is that we will always keep requests inside outcome for max time while in practice they will be usually finished earlier. But let's say we have some crazy load of 100 RPS (sustained), 10 mins expiry - that's 100*60*10 = 60000 requests in the state, less than ~80 bytes per request that gives us < 5MB state size, which fits within limits.

I guess to get the best of both worlds you could implement both (i.e. what you have now + removal from state based on initial time + ExecutionRemovalTime). Up to you.

pkg/workflows/dontime/plugin.go
```diff
@@ -0,0 +1,256 @@
...
+			unscheduledFinishedExecutionIDs = append(unscheduledFinishedExecutionIDs, id)
+		}
+	}
+
+	// Collect up to batchSize unexpired requests
+	requests := map[string]int64{} // Maps executionID --> seqNum
+	for batchOffset := 0; batchOffset < p.store.requests.Len() && len(requests) < p.batchSize; {
+		batch, err := p.store.requests.RangeN(batchOffset, p.batchSize)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get request batch: %w", err)
```
bolekk, line 82: fail open

bolekk (changes requested):

pkg/workflows/dontime/plugin.go
DylanTinianov: Changed to remove executions after `ExecutionRemovalTime` from the first generated DON Time for its ID.

The observation phase deletes expired requests from the store while batching, so I don't think we need to have the transmitter also delete these?

pkg/workflows/dontime/request.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (r *Request) ExpiryTime() time.Time {
+	return r.ExpiresAt
+}
+
+func (r *Request) SendResponse(_ context.Context, resp Response) {
+	select {
+	case r.CallbackCh <- resp:
+		close(r.CallbackCh)
+	default: // Don't block trying to send
```
bolekk, line 32: nit: this should never happen as we need to guarantee the channel is large enough - maybe add an Error log?

bolekk (approved):

pkg/workflows/dontime/pb/config.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
DylanTinianov: I don't think so in this case, given these are only used internally by the plugin.

bolekk (approved):

PR #1297, opened 2025-06-25
Update CODEOWNERS to include bix-framework for capabilities and chains

poopoothegorilla:


justinkaseman (changes requested):
What is the goal of this change? Is it to be aware of changes or is it to get changes through faster?

By adding BIX to `/capabilities` that would mean that changes could go into the Capabilities API without Keystone/Capabilities team's review.

If it's to get changes in faster, can the scope be narrowed further?

fernandezlautaro:
@justinkaseman updated the /capabilities folder and added others I see PLEX team is usually working one, could you please re-review?

pablolagreca (approved):

PR #1298, opened 2025-06-25
feat: add TONService to Relayer interface

briansztamfater:
[NONEVM-2045](https://smartcontract-it.atlassian.net/browse/NONEVM-2045)

This PR introduces a `TONService` interface to encapsulate TON-specific functionality and updates the `Relayer` interface to return a `TONService` implementation.

[NONEVM-2045]: https://smartcontract-it.atlassian.net/browse/NONEVM-2045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/loop/internal/relayer/relayer.go
```diff
@@ -148,6 +149,9 @@ func (p *pluginRelayerServer) NewRelayer(ctx context.Context, request *pb.NewRel
 		if evmService, ok := r.(types.EVMService); ok {
 			evmpb.RegisterEVMServer(s, newEVMServer(evmService, p.BrokerExt))
 		}
+		if tonService, ok := r.(types.TONService); ok {
+			tonpb.RegisterTONServer(s, newTONServer(tonService, p.BrokerExt))
+		}
```
ogtownsend, line 154: 👍

ogtownsend (approved):

pkg/types/chains/ton/ton.go
```diff
@@ -0,0 +1,65 @@
...
+	ID          int64
+	Name        string
+	Address     string
+	EventName   string
+	EventTopic  uint64
+	StartingSeq uint32
+	Retention   time.Duration
+}
+
+type Log struct {
```
jadepark-dev, line 41: QQ: Would it be possible to import from logpoller pkg once it has solid types?

pkg/types/relayer.go
```diff
@@ -151,12 +152,28 @@ type EVMService interface {
 	GetTransactionStatus(ctx context.Context, transactionID IdempotencyKey) (TransactionStatus, error)
 }
 
+type TONService interface {
```
jadepark-dev, line 155: I would suggest to scope a ticket for updating this check(once we have a buildable TON relayer) as well
https://github.com/smartcontractkit/chainlink-common/blob/main/.github/workflows/build_external.yml

pkg/loop/internal/relayerset/server.go
```diff
@@ -50,6 +52,7 @@ type Server struct {
 
 var _ relayerset.RelayerSetServer = (*Server)(nil)
 var _ evmpb.EVMServer = (*Server)(nil)
+var _ tonpb.TONServer = (*Server)(nil)
```
briansztamfater, line 55: As we add `TONServer`, `EVMServer`, etc., backed by the same `relayerset.Server`, we should consider refactoring this to avoid method name collisions. It makes sense to split each protocol handler into its own server implementation to keep things modular and prevent conflicts as the API surface grows.

pkg/loop/internal/relayerset/ton.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+// tonClient wraps the TONRelayerSetClient by attaching a RelayerID to TONClient requests.
+// The attached RelayerID is stored in the context metadata.
+type tonClient struct {
+	relayID types.RelayID
+	client  tonpb.TONClient
+}
+
+var _ tonpb.TONClient = (*tonClient)(nil)
```
ilija42, line 24: if you pull in main, this boilerplate is no longer required

pkg/chains/ton/ton.proto
```diff
@@ -0,0 +1,130 @@
...
+
+// ---- Transaction Management ----
+
+message Message {
+  uint32 mode = 1;
+  string to_address = 2;
+  string amount_nano = 3;
+  bool bounce = 4;
+  bytes body = 5;                  // BOC
+  optional bytes state_init = 6;  // optional BOC
```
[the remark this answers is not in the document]
briansztamfater: In TON, you don’t set a gas limit in the transaction like in Ethereum. Instead, you attach some TON to the message (see `AmountNano`), which covers gas and fees. If it’s not enough, the transaction fails or halts early. You can also set a mode to control how gas is paid, from the contract’s balance, the attached TON, or with refunds if the message bounces. In any case, this is just an abstraction of TON's TXM Request struct, in case you want to dive deeper.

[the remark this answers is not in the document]
huangzhen1997: Thank you

pkg/types/chains/ton/ton.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type ExitCode = int32
+
+type BOC = []byte
+
+type BlockIDExt struct {
+	Workchain int32
+	Shard     uint64
+	Seqno     uint32
```
jmank88, line 16:
```suggestion
	SeqNo     uint32
```

pkg/loop/internal/relayerset/evm.go
```diff
@@ -27,60 +26,65 @@ type evmClient struct {
 
 var _ evmpb.EVMClient = (*evmClient)(nil)
 
+var evmMetadata = relayerMetadata{
+	chain:   metadataEVMChain,
+	network: metadataEVMNetwork,
+}
```
jmank88, line 32: What is this?

pkg/loop/internal/relayerset/metadata.go
jmank88, line 1: Why is this being refactored?

jmank88: I think I see how it fits together now, but why don't we just use `network` and `chain-id` for everything :shrug:

briansztamfater: Sounds good, let me know if this makes sense! [3ccbb60b6423f431732acba7810140dab2dd8e93](https://github.com/smartcontractkit/chainlink-common/pull/1298/commits/3ccbb60b6423f431732acba7810140dab2dd8e93)!

pkg/chains/ton/ton.proto
```diff
@@ -0,0 +1,130 @@
...
+  rpc RegisterFilter(RegisterFilterRequest) returns (google.protobuf.Empty);
+  rpc UnregisterFilter(UnregisterFilterRequest) returns (google.protobuf.Empty);
+}
+
+// ---- LiteClient Types ----
+
+message BlockIDExt {
+  int32 workchain = 1;
+  uint64 shard = 2;
+  uint32 seqno = 3;
```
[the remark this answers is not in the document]
briansztamfater: Updated

pkg/types/chains/ton/ton.go
briansztamfater: Updated

ilija42 (commented):
Needs tests in `relayerset_test.go` and `pkg/loop/internal/relayer/test/relayer.go` to confirm that everything is wired properly, looks good otherwise

jmank88 (approved):

pkg/types/chains/ton/ton.go
```diff
@@ -0,0 +1,65 @@
...
+	Workchain int32
+	Shard     uint64
+	SeqNo     uint32
+}
+type Block struct {
+	GlobalID int32 // Represents EVM-equivalent ChainID for TVM
+}
+
+type Message struct {
+	Mode       uint8  // TON send mode
```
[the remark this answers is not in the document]
briansztamfater: Good question! I think we should keep it as uint8 as the Mode field represents a set of bitwise flags that can be combined, so an enum wouldn’t be appropriate here.

pkg/loop/internal/pb/relayerset/relayerset.proto
```diff
@@ -9,6 +9,7 @@ import "loop/internal/pb/contract_reader.proto";
 import "loop/internal/pb/contract_writer.proto";
 import "loop/internal/pb/relayer.proto";
 import "chains/evm/evm.proto";
+import "chains/ton/ton.proto";
```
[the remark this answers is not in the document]
briansztamfater: Removed!

pkg/types/chains/ton/ton.go
```diff
@@ -0,0 +1,65 @@
...
+	ToAddress  string // TON address (raw or user-friendly)
```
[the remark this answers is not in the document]
briansztamfater: TON addresses are a bit different, they can be raw or base64 (user-friendly), include metadata like the workchain ID and bounce flag, and aren't just fixed-length byte arrays. I'd kept it as a string for now to keep things simple and interoperable with the existing TON Go SDK, but we could introduce a dedicated Address type later if we want stricter typing.

briansztamfater:
> Needs tests in `relayerset_test.go` and `pkg/loop/internal/relayer/test/relayer.go` to confirm that everything is wired properly, looks good otherwise

Thank you, just updated with more coverage 👍

pkg/chains/ton/proto_helpers.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func ConvertBlockFromProto(pb *Block) *types.Block {
+	if pb == nil {
+		return nil
+	}
+	return &types.Block{
+		GlobalID: pb.GlobalId,
+	}
+}
```
jmank88, line 48: nit/ if we can find the right terminology, these could be expressed more succinctly as constructors and methods:
```suggestion
func NewBlock(block *types.Block) *Block {
	if block == nil {
		return nil
	}
	return &Block{
		GlobalId: block.GlobalID,
	}
}

func (pb *Block) AsBlock() *types.Block {
	if pb == nil {
		return nil
	}
	return &types.Block{
		GlobalID: pb.GlobalId,
	}
}
```
It is just an awkward situation with the modified type (proto) being the one local to this package :thinking:

pkg/types/chains/ton/ton.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"context"
+	"math/big"
+	"time"
+)
+
+const AddressLength = 20
+
+// represents evm-style address
```
ilija42, line 11: evm copy paste? I don't see this being used, address is mainly treated as a string in the api, should we fix this definition and use it?

pkg/types/chains/ton/ton.go
```diff
@@ -0,0 +1,70 @@
...
+
+type BOC = []byte
+
+type BlockIDExt struct {
+	Workchain int32
+	Shard     int64
+	SeqNo     uint32
+}
+type Block struct {
+	GlobalID int32 // Represents EVM-equivalent ChainID for TVM
```
ilija42, line 24: Why does block data store chainID, I'm not too familiar with TON?

ilija42 (commented):
looks good, left a nit comment. Also to add context, which go TON API is the LiteClient following?

jmank88 (approved):

pkg/types/chains/ton/ton.go
briansztamfater: Yeah, some copy-pasta leftover 🤦‍♂️  Just removed it

ilija42 (approved):

PR #1302, opened 2025-06-26
DonTime module hooks

DylanTinianov:
Enabled DON Time module hooks.
- Creates custom Plugin Provider for delegate initialization.
- Implements and links clockTimeGet and pollOneoff with wasm calls
- Adds a timeFetcher for safely retrieving DON or Node time within trap handlers.
- Also some small plugin updates.

[Jira Ticket](https://smartcontract-it.atlassian.net/browse/CAPPL-903)

### Requires
[WorkflowLib Plugin](https://github.com/smartcontractkit/chainlink-common/pull/1288
)

### DIFF
--- a/pkg/values/installer/pkg/consts.go
+++ b/pkg/values/installer/pkg/consts.go
@@ -1,3 +1,3 @@
 package pkg
 
-const chainlinkProtosVersion = "cre-sdk/v1alpha.6"
+const chainlinkProtosVersion = "cre-sdk/v1alpha.7"


pkg/workflows/wasm/host/execution.go
```diff
@@ -18,6 +20,7 @@ type execution[T any] struct {
 	lock                 sync.RWMutex
 	module               *module
 	executor             ExecutionHelper
+	timeFetcher          *timeFetcher
```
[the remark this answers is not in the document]
DylanTinianov: Not sure what you mean. I see hooks in the `module`, but we can't access the `ExecutionHelper` from there which is needed for the time calls.

pkg/workflows/wasm/host/standard_test.go
```diff
@@ -68,6 +71,7 @@ func TestStandardCapabilityCallsAreAsync(t *testing.T) {
 	t.Parallel()
 	mockExecutionHelper := NewMockExecutionHelper(t)
 	mockExecutionHelper.EXPECT().GetWorkflowExecutionID().Return("id")
+	mockExecutionHelper.EXPECT().GetNodeTime().Return(time.Now())
```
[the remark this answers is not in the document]
DylanTinianov: I think it's a Go wasm initialization thing, I'll make it optional though,

pkg/workflows/wasm/host/standard_test.go
```diff
@@ -138,10 +143,37 @@ func TestStandardModeSwitch(t *testing.T) {
 
 		m := makeTestModule(t)
 		request := triggerExecuteRequest(t, 0, &basictrigger.Outputs{CoolOutput: anyTestTriggerValue})
-
 		result := executeWithResult[string](t, m, request, mockExecutionHelper)
-
 		require.Equal(t, "test556", result)
+
+		// Verify time call sequence is Node --> DON --> Node --> DON
```
nolag, line 149: This may not always be true, it's language dependant. Go does it during initialization. Before the first DON mode, we can just ignore it.

nolag: I wonder if the easiest thing to do is to have three flags below initialized to false.

donCall1
nodeCall
donCall2

In the don time call, if nodeCall is true, set donCall2 true. Otherwise, set donCall1 true

In node call, if donCall1 is true, set nodeCall to true. Here, you can leave a comment about initialization. 

At the end, check all three are true.

pkg/workflows/wasm/host/standard_test.go
```diff
@@ -105,6 +117,24 @@ func TestStandardModeSwitch(t *testing.T) {
...
+		var donCall2 bool
+		mockExecutionHelper.EXPECT().GetNodeTime().RunAndReturn(func() time.Time {
+			if donCall1 {
+				nodeCall = true
+			}
+			return time.Now()
+		})
+		// We want to make sure time.Sleep() is called at least twice in DON mode and once in node Mode
+		mockExecutionHelper.EXPECT().GetDONTime(mock.Anything).RunAndReturn(func(ctx context.Context) (time.Time, error) {
+			donCall1 = true
```
nolag, line 132: set this in an else, otherwise, if they only call in don mode after they will both be set.

pkg/workflows/wasm/host/time.go
```diff
@@ -0,0 +1,87 @@
...
+				donTime = t.executor.GetNodeTime()
+			}
+
+			if err != nil {
+				select {
+				case t.timeErrChan <- err:
+				case <-t.ctx.Done():
+					return
+				}
+			} else {
```
MStreet3, line 78: unnecessary else

pkg/workflows/wasm/host/time.go
```diff
@@ -0,0 +1,87 @@
...
+	case donTime := <-t.timeResultChan:
+		return donTime, nil
+	case err := <-t.timeErrChan:
+		return time.Time{}, err
+	case <-t.ctx.Done():
+		return time.Time{}, t.ctx.Err()
+	}
+}
+
+func (t *timeFetcher) Start() {
```
MStreet3, line 52: how does this get cleaned up?

pkg/workflows/wasm/host/time_test.go
```diff
@@ -0,0 +1,87 @@
...
+func TestTimeFetcher_GetTime_NODE(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	mockExec := NewMockExecutionHelper(t)
+	expected := time.Now()
+	mockExec.EXPECT().GetNodeTime().Return(expected)
+
+	tf := newTimeFetcher(ctx, mockExec)
+	tf.Start()
```
MStreet3, line 24: fetcher should create a new child context it can cancel to expose a Close method

pkg/workflows/dontime/provider.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/services"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+)
+
+// Provider wraps an existing OCR3 plugin provider (from the relayer)
+// and overrides the plugin factory and contract transmitter for DonTime.
+type Provider struct {
```
MStreet3, line 12: why?

pkg/workflows/wasm/host/time.go
nolag: Actually, it may be a bug in the error condition. It will send the error to the channel, but not return.

pkg/workflows/dontime/provider.go
DylanTinianov: It's needed for [initializing the plugin](https://github.com/smartcontractkit/chainlink/pull/18355/files#diff-ee42cd8d3c2284e2aa914b4f7797b308ef4bf9f3e73a5692da84685f3bf9f542R673-R678)

pkg/workflows/wasm/host/time.go
MStreet3: should add a `continue` after sending the error

pkg/workflows/wasm/host/time.go
MStreet3: yes, owner should cancel the context.  Should be fine.

MStreet3 (approved):

PR #1314, opened 2025-07-01
Add messages for CRUDL operations

cedric-cordenier:

### DIFF
--- a/pkg/capabilities/actions/vault/messages.proto
+++ b/pkg/capabilities/actions/vault/messages.proto
@@ -4,11 +4,15 @@ option go_package = "github.com/smartcontractkit/chainlink-common/pkg/capabiliti
 
 package vault;
 
-message SecretRequest {
-    string id = 1;
+message SecretIdentifier {
+    string key = 1;
     string namespace = 2;
     string owner = 3;
-    repeated string encryption_keys = 4;
+}
+
+message SecretRequest {
+    SecretIdentifier id = 1;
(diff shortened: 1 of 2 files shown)


pkg/capabilities/actions/vault/messages.proto
```diff
@@ -4,24 +4,26 @@ option go_package = "github.com/smartcontractkit/chainlink-common/pkg/capabiliti
...
-    string id = 1;
-    string namespace = 2;
-    string owner = 3;
-    string encrypted_value = 4;
-    repeated EncryptedShares encrypted_decryption_key_shares = 5;
-    string error = 6;
+    SecretIdentifier id = 1;
+    string encrypted_value = 2;
+    repeated EncryptedShares encrypted_decryption_key_shares = 3;
+    string error = 4;
```
prashantkumar1982, line 26: Optional: proto allows us to tag some fields as optional, to indicate they can be omitted, and easily checked in code with logic.

Example :
message SecretResponse {
    .
    .
    optional string encrypted_value = 2;
    optional string error = 4;
}

Then in go code:
secretResponse.has_error() OR
secretResponse.has_encryptedValue()

This is instead of directly checking if these are empty strings.

cedric-cordenier: I went with the oneof :)

ettec (approved):

george-dorin (approved):

PR #1322, opened 2025-07-01
CCIP - move CAL interface and required types into chainlink-common

ogtownsend:
NONEVM-2081

### DIFF
--- a/pkg/types/ccipocr3/chainaccessor.go
+++ b/pkg/types/ccipocr3/chainaccessor.go
@@ -113,8 +113,7 @@ type DestinationAccessor interface {
 	// Access Type: Method(NextSeqNum)
 	// Contract: OffRamp
 	// Confidence: Unconfirmed
-	NextSeqNum(ctx context.Context, sources []ChainSelector) (
-		seqNum map[ChainSelector]SeqNum, err error)
+	NextSeqNum(ctx context.Context, sources []ChainSelector) (map[ChainSelector]SeqNum, error)
 
 	// Nonces for all provided selector/address pairs. Addresses must be encoded
 	// according to the source chain requirements by using the AddressCodec.
--- a/pkg/types/ccipocr3/chainaccessor.go
+++ b/pkg/types/ccipocr3/chainaccessor.go
@@ -218,9 +218,9 @@ type SourceAccessor interface {
 }
 
 type RMNAccessor interface {
-	// GetRmnCurseInfo returns rmn curse/pausing information about the provided chains
+	// GetRMNCurseInfo returns rmn curse/pausing information about the provided chains
 	// from the destination chain RMN remote contract. Caller should be able to access destination.
-	GetRmnCurseInfo(ctx context.Context) (CurseInfo, error)
+	GetRMNCurseInfo(ctx context.Context) (CurseInfo, error)
 }
 
 ////////////////////////////////////////////////////////////////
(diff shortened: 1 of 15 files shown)


pkg/types/ccipocr3/chainaccessor.go
```diff
@@ -0,0 +1,351 @@
...
+	GetFeeQuoterDestChainConfig(
+		ctx context.Context,
+		dest ChainSelector,
+	) (FeeQuoterDestChainConfig, error)
+}
+
+type RMNAccessor interface {
+	// GetRmnCurseInfo returns rmn curse/pausing information about the provided chains
+	// from the destination chain RMN remote contract. Caller should be able to access destination.
+	GetRmnCurseInfo(ctx context.Context) (CurseInfo, error)
```
jmank88, line 223:
```suggestion
	GetRMNCurseInfo(ctx context.Context) (CurseInfo, error)
```

pkg/types/ccipocr3/common_types.go
```diff
@@ -7,6 +7,48 @@ import (
...
+	for _, b := range a {
+		if b != 0 {
+			return false // zero
+		}
+	}
+	return true
+}
+
+// UnknownEncodedAddress represents an encoded address with an unknown encoding.
+type UnknownEncodedAddress string
```
jmank88, line 50: Is this fundamentally different from `Account`? Or could that be used instead?

makramkd: Is this referring to `libocr.Account` or some other type defined in chainlink-common?

makramkd (dismissed):

pkg/types/ccipocr3/common_types.go
```diff
@@ -5,8 +5,52 @@ import (
...
+	for _, b := range a {
+		if b != 0 {
+			return false // zero
+		}
+	}
+	return true
+}
+
+// UnknownEncodedAddress represents an encoded address with an unknown encoding.
+type UnknownEncodedAddress libocr.Account
```
jmank88, line 52: Why define another type? Can we just use Account directly? If the package import is obnoxious, then we can use an alias:
```suggestion
type Account = libocr.Account
```

ogtownsend: A lot of these types are duplicated in `chainlink-ccip` so after this PR is merged, I was hoping to delete all of the duplicate types in `chainlink-ccip`. `UnknownEncodedAddress` is used heavily in chainlink-ccip in addition to being used in this CAL interface, but I guess we could separate the deletion of `UnknownEncodedAddress` from the rest of the cleanup

jmank88: If we want to preserve the name, that is fine, but an alias still seems like the simplest way to preserve compatibility :shrug:

ogtownsend: yeah sounds good, in interest of this whole effort of moving towards compatibility and reusability across CRE and beyond, I'll update this to `Account` here and handle the cleanup of `UnknownEncodedAddress ` in chainlink-ccip separately

pkg/types/ccipocr3/common_types.go
ogtownsend: Discussed here: https://github.com/smartcontractkit/chainlink-common/pull/1322#discussion_r2193047276

pkg/types/ccipocr3/common_types.go
jmank88: I mean this: `type UnknownEncodedAddress = libocr.Account`

jmank88 (approved):

winder (approved):

PR #1339, opened 2025-07-04
PLEX-1560 header by number

dhaidashenko:
* Replaced LatestAndFinalized with HeaderByNumber to support safe tag.
* Added confidence level to all calls lockable to a block to provide a guarantee that returned data was finalized/safe at the specified height.
Requires: https://github.com/smartcontractkit/chainlink-protos/pull/108

### DIFF
--- a/pkg/billing/workflow_client.go
+++ b/pkg/billing/workflow_client.go
@@ -53,11 +53,9 @@ type workflowConfig struct {
 type WorkflowClientOpt func(*workflowConfig)
 
 func defaultWorkflowConfig() workflowConfig {
-	loggerInst, _ := logger.New()
 	// By default, no JWT manager is set and we fallback to insecure creds.
 	return workflowConfig{
 		transportCredentials: insecure.NewCredentials(),
-		log:                  loggerInst,
 		tlsCert:              "",
 		// Default to "localhost" if not overridden.
 		serverName:   "localhost",
@@ -68,12 +66,6 @@ func defaultWorkflowConfig() workflowConfig {
 func WithWorkflowTransportCredentials(creds credentials.TransportCredentials) WorkflowClientOpt {
 	return func(cfg *workflowConfig) {
 		cfg.transportCredentials = creds
-	}
-}
-
-func WithWorkflowLogger(l logger.Logger) WorkflowClientOpt {
-	return func(cfg *workflowConfig) {
-		cfg.log = l
 	}
 }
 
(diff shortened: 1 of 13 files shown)


pkg/types/relayer.go
```diff
@@ -131,6 +122,54 @@ type GethClient interface {
...
+	//
+	// blockNumber - defines block at which call will be executed:
+	//  nil - execute at latest block of specified confidence level (e.g. latest finalized, latest safe, etc.);
+	//  positive value - execute as specified height with confidence level;
+	//
+	// confidenceLevel - defines at which confidence level request will be executed:
+	//    Unconfirmed - returns the most recent data
+	//    Finalized - returned data was finalized;
+	//    Safe - returned data is highly unlikely to be reorged;
+	BalanceAtWithConfidence(ctx context.Context, account evm.Address, blockNumber *big.Int, confidenceLevel primitives.ConfidenceLevel) (*big.Int, error)
```
ilija42, line 135: Didn't we settle for passing in confidence level like this? : 

```
// CallContract reads a contract as specified in the call message at a block height defined by blockNumber where:
	// blockNumber :
	//   nil (default) or (-2) → use the latest mined block (“latest”)
	//   FinalizedBlockNumber(-3) → last finalized block (“finalized”)
	//
	// Any positive value is treated as an explicit block height.
```

jmank88: I agree there is no need to reinvent this system. https://github.com/smartcontractkit/chainlink-evm/pull/138#discussion_r2197580513

ilija42 (approved):

pablolagreca (approved):

pkg/types/relayer.go
```diff
@@ -131,6 +122,59 @@ type GethClient interface {
...
+	//   - -3: finalized block
+	//   - -4: safe block
+	//   - positive value: specific block at that height
+	//
+	// request.ConfidenceLevel - determines if additional verification is required (only applicable for positive blockNumber values):
+	//   - "Unconfirmed" or empty string: no additional verification
+	//   - "Finalized": returns error if requested is not finalized
+	//   - "Safe": returns error if requested block is not safe
+	HeaderByNumber(ctx context.Context, request evm.HeaderByNumberRequest) (*evm.HeaderByNumberReply, error)
+
```
jmank88, line 177: Would these make sense on `GethClient` now that they are unified?

ilija42 (approved):

ilija42 (approved):

DylanTinianov (approved):

PR #1359, opened 2025-07-09
INFOPLAT-1035: logger implement otelzap

kirqz23:
This PR introduces `OtelZapCore`, a custom `zapcore.Core` implementation that emits structured logs to OpenTelemetry using the OTLP protocol.

It also integrates this core into the existing `logger` package, allowing all logs to be exported via OTEL while preserving existing logging interfaces.


pkg/logger/logger.go
```diff
@@ -102,6 +104,11 @@ func NewWithSync(w io.Writer) Logger {
 	return &logger{zap.New(core).Sugar()}
 }
 
+// NewWithOtelZapCore returns a new Logger using an OpenTelemetry Zap Core.
+func NewWithOtelZapCore(loggerProvider otellog.LoggerProvider) Logger {
+	return &logger{zap.New(NewOtelZapCore(loggerProvider)).Sugar()}
+}
+
```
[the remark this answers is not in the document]
kirqz23: There is one issue with importing beholder.GetClient().LoggerProvider  directly in logger packages as it generates import cycle due to the logger imported in beholder [here](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/beholder/dual_source_emitter.go#L7) and [here](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/beholder/proto_emitter.go#L10)

pkcll (approved):

pkg/logger/zap_otel.go
jmank88, line 1: I am surprised that we have to implement this ourselves - none of the packages in the wild will work?

pkg/logger/zap_otel.go
```diff
@@ -0,0 +1,266 @@
...
+	logRecord.SetTimestamp(entry.Time)
+	logRecord.SetObservedTimestamp(time.Now())
+	logRecord.AddAttributes(otelAttrs...)
+
+	o.logger.Emit(context.Background(), logRecord)
+
+	// If a zapCore is set, delegate the write to Zap as well
+	if o.zapCore != nil {
+		return o.zapCore.Write(entry, fields)
+	}
```
jmank88, line 168: This seem really odd. Zap already has mechanisms for writing to multiple cores: https://pkg.go.dev/go.uber.org/zap/zapcore#NewTee

Layering like this doesn't make sense because they each own their own Enabled() and Check() implementations

pkg/logger/logger.go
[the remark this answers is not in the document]
jmank88: Let's separate these concerns. All that this package API needs to do is allow a caller to pass in a zap.Core. It doesn't need to know about otel at all. 
```go
// NewWithCore returns a new Logger with a given zapcore.Core.
func NewWithCore(core zapcore.Core) Logger {
	return &logger{zap.New(core).Sugar()}
}
```
Perhaps we would rather accept variadic cores than just one, but either way they should be combined via https://pkg.go.dev/go.uber.org/zap/zapcore#NewTee

The more interesting question is whether we _always_ want to include the default core or provide a way for the caller to get it so it's optional.
```go 
lggr := logger.NewWithCores(otellogger.NewCore())
```
vs.
```go 
lggr := logger.NewWithCores(logger.NewCore(), otellogger.NewCore())
```

pkg/logger/otellogger/zap_otel.go
pkcll, line 1: nit: maybe worth making the file to the same name as the package 
maybe rename `pkg/logger/otellogger/zap_otel.go` to  `pkg/logger/otelzap/otelzap.go`

pkg/logger/otelzap/otelzap.go
```diff
@@ -0,0 +1,249 @@
...
+	logger       otellog.Logger
+	fields       []zapcore.Field
+	levelEnabler zapcore.LevelEnabler
+}
+
+type Option func(c *OtelZapCore)
+
+// NOTE: [beholder.SetGlobalOtelProviders](https://github.com/smartcontractkit/chainlink-common/blob/27faefc9ce454c8aa2b1b7484e377ea3e8996bba/pkg/beholder/global.go#L50)
+// must be called before using `NewOtelLogger` otherwise otelglobal.GetLoggerProvider() returns noop provider
+func NewOtelLogger(name string) otellog.Logger {
```
jmank88, line 30: What is the expected use of this function? We did a very deliberate conversion away from global logging a long time ago, and we shouldn't encourage that pattern.

pkg/logger/otelzap/otelzap.go
```diff
@@ -0,0 +1,249 @@
...
+	return otelglobal.GetLoggerProvider().Logger(name)
+}
+
+// NewOtelCore initializes an OpenTelemetry Core for exporting logs in OTLP format
+func NewOtelZapCore(logger otellog.Logger, opts ...Option) zapcore.Core {
```
jmank88, line 35: No need to repeat `OtelZap` since we are already in `package otelzap`:
```suggestion
func NewCore(logger otellog.Logger, opts ...Option) zapcore.Core {
```

pkg/logger/logger.go
```diff
@@ -102,6 +102,21 @@ func NewWithSync(w io.Writer) Logger {
...
+	var core zapcore.Core
+	switch len(cores) {
+	case 0:
+		core = zapcore.NewNopCore()
+	case 1:
+		core = cores[0]
+	default:
+		core = zapcore.NewTee(cores...)
+	}
+	return &logger{zap.New(core).Sugar()}
```
jmank88, line 117: The `NewTee` implementation already contains these optimizations:
```suggestion
	return &logger{zap.New(zapcore.NewTee(cores...)).Sugar()}
```

jmank88 (changes requested):

pkg/logger/otelzap/otelzap.go
kirqz23: Removed

pkg/logger/otellogger/zap_otel.go
kirqz23: Done

pkcll (approved):

jmank88 (approved):

PR #1391, opened 2025-07-16
Bump pkg/values 

dhaidashenko:


go.mod
```diff
@@ -142,3 +142,5 @@ require (
 	golang.org/x/xerrors v0.0.0-20240903120638-7835f813f4da // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20250218202821-56aae31c358a // indirect
 )
+
+replace github.com/smartcontractkit/chainlink-common/pkg/values => ./pkg/values
```
jmank88, line 146: This repo is not equipped to handle the complexity of local replaces in library modules.
```suggestion
```

jmank88 (changes requested):

go.mod
dhaidashenko: Could you elaborate on what complexity do you have in mind? Isn't it working as is?
The reason I made this change is that it seems redundant to produce two PRs when we need to make any changes in the pkg/values module.

PR #1392, opened 2025-07-16
migrate golangci-lint to v2; swap out gomodguard for gomoddirectives

jmank88:
I noticed that `gomodguard` was not catching local replaces, and it turns out it is not enabled by default. However, when enabled, it doesn't actually behave like we expect. Rather than flag local replace directives in this project's go.mod file(s), it is blocking use of dependencies that have local replaces in their go.mod files. Swapping for `gomoddirectives` instead, which does what we want, and by default.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/shopspring/decimal v1.4.0
 	github.com/smartcontractkit/chain-selectors v1.0.62
 	github.com/smartcontractkit/chainlink-common/pkg/chipingress v0.0.1
-	github.com/smartcontractkit/chainlink-common/pkg/values v0.0.0-20250715171446-42e31e2bef64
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -25,132 +24,137 @@ linters:
...
-            desc: Use the standard library instead, for example https://pkg.go.dev/errors#Join
-          - pkg: "gopkg.in/guregu/null.v1"
-            desc: Use gopkg.in/guregu/null.v4 instead
-          - pkg: "gopkg.in/guregu/null.v2"
-            desc: Use gopkg.in/guregu/null.v4 instead
-          - pkg: "gopkg.in/guregu/null.v3"
-            desc: Use gopkg.in/guregu/null.v4 instead
-  gomodguard:
-    blocked:
-      local_replace_directives: true
```
jmank88, line 137: Removed

chainchad (approved):

PR #1411, opened 2025-07-17
In LLO Aggregation use the stream old price when Observation is absent

cawthorne:


cawthorne:
Actually, we do just want to skip over writing the value at all, not just use the old price. Will update.

cawthorne:
Adding test.

cawthorne:
If it is an old price (that has been onchain) and a heartbeat update is triggered this round, then there isn't much point in writing it onchain.

But as we can't actually tell when the last price got onchain from this Aggregator, we should just write the last received price ever heartbeat (with the same timestamp each time too).

cawthorne:
If we update the previous timestamp of the missing feed in the current state we will write the wrong timestamp for that price onchain, If we don't update it we will trigger hearbeats every time once hb is reached.

This could be solvable with extra state, but...

In hindsight it is better to not write the price at all to avoid this, even if it means we had a fresher price after the last heartbeat, but before the stream job errored, and we don't get to write that onchain.

Also it is unconventional/potentially incorrect behaviour to write an price onchain significantly after it was observed at the next heartbeat interval.

cawthorne:
PR is ready for review @smartcontractkit/foundations

mxiao-cll (approved):

DylanTinianov (approved):

PR #1412, opened 2025-07-17
feat(ci): build cache

erikburt:


jmank88 (approved):

chudilka1 (approved):

.github/workflows/go-mod-cache.yml
```diff
@@ -0,0 +1,53 @@
+name: Go Module Cache
```
chainchad, line 1: Since we're also using this on /chainlink, we should we move this to the /.github repo?

chainchad (approved):

.github/workflows/go-mod-cache.yml
erikburt: That's a good idea. It would require us consolidating our `setup-go` actions, because the one here, and in /chainlink, and in .github have different behaviours. Two of which are tailored to the repo, and the setup-golang in .github is very general.

I'll create a backlog task.

erikburt:
@jmank88 or @chudilka1 - can someone merge this for me? I don't have write access to the repository.

PR #1413, opened 2025-07-17
fix(ci): build chainlink job

erikburt:
Fixes build chainlink job. This rips out the whole image build and solely builds the binary, I think this validates it well enough. If not, I can fix the workflow to build the image, it's just much harder.


erikburt:
Going to let @chainchad take this one in #1463.

PR #1415, opened 2025-07-18
CCIP - make required config types public and update legacy snapshot f…

ogtownsend:
…unction args

### DIFF
--- a/pkg/types/ccipocr3/chainaccessor.go
+++ b/pkg/types/ccipocr3/chainaccessor.go
@@ -62,7 +62,11 @@ type AllAccessors interface {
 	// Access Type: Method(many, see code)
 	// Contract: Many
 	// Confidence: Unconfirmed
-	GetAllConfigLegacySnapshot(ctx context.Context) (ChainConfigSnapshot, error)
+	GetAllConfigLegacySnapshot(
+		ctx context.Context,
+		destChainSelector ChainSelector,
+		sourceChainSelectors []ChainSelector,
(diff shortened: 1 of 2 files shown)


pkg/types/ccipocr3/chainaccessor.go
```diff
@@ -62,7 +62,11 @@ type AllAccessors interface {
 	// Access Type: Method(many, see code)
 	// Contract: Many
 	// Confidence: Unconfirmed
-	GetAllConfigLegacySnapshot(ctx context.Context) (ChainConfigSnapshot, error)
+	GetAllConfigLegacySnapshot(
+		ctx context.Context,
+		destChainSelector ChainSelector,
+		sourceChainSelectors []ChainSelector,
```
winder, line 68: Isn't `destChainSelector` implicit? Thats the only chain the accessor is able to read from.

ogtownsend: @winder it's for this check: https://github.com/smartcontractkit/chainlink-ccip/blob/main/pkg/reader/ccip.go#L1224

the accessor needs to know which configs it should fetch (unless want to remove that check and attempt to fetch them all)

ogtownsend: No, that check would be within the accessor since that function is very specific to the chain reader, here is a WIP: https://github.com/smartcontractkit/chainlink-ccip/compare/cal-implement-GetAllConfigLegacySnapshot?expand=1

PR #1424, opened 2025-07-21
pkg/loop/internal/relayer/pluginprovider: switch to JSON with type hints for ChainReader/ChainWriter

cfal:


pkg/loop/internal/relayer/pluginprovider/json/type_hints.go
jmank88, line 1: By design, this level of type detail granularity must not matter. The API speaks (language agnostic) encoded JSON/CBOR, and implementations need to be written to work with those general formats. Note, this does not limit the caller's ability to use any specific type (built-in, std lib, or user-defined) on their side when deserializing.

jmank88 (changes requested):
This API has been stable for over a year. Breaking it is not an option. What is the goal of this PR?

PR #1427, opened 2025-07-23
feat: add storage-service workflow client

anirudhwarrier:


pkg/storage/workflow_client.go
```diff
@@ -0,0 +1,182 @@
...
+	grpcOpts := []grpc.DialOption{
+		grpc.WithTransportCredentials(cfg.transportCredentials),
+	}
+
+	conn, err := grpc.NewClient(address, grpcOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg.log.Infow("connected to storage service (WorkflowClient)", "address", address)
```
leeyikjiun, line 152: There's no I/O performed upon `grpc.NewClient`. So there's no guarantees that it is "connected"

pkg/storage/workflow_client.go
```diff
@@ -0,0 +1,182 @@
...
+	if n.jwtGenerator != nil {
+		var err error
+		ctx, err = n.injectToken(ctx, req)
+		if err != nil {
+			return err
+		}
+	}
+	n.log.Infow("DownloadArtifactStream RPC called", "id", req.Id, "type", req.Type.String(), "environment", req.Environment.String())
+
+	stream, err := n.client.DownloadArtifact(ctx, req)
```
leeyikjiun, line 61: How about replacing this entire part with
```go
stream, err := n.DownloadArtifact(ctx, req)
```

pkg/storage/workflow_client.go
```diff
@@ -0,0 +1,182 @@
...
+// It returns a WorkflowClient which can DownloadArtifacts
+func NewNodeClient(ctx context.Context, address string, opts ...NodeClientOpt) (WorkflowClient, error) {
+	cfg := defaultNodeConfig()
+
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
+	grpcOpts := []grpc.DialOption{
+		grpc.WithTransportCredentials(cfg.transportCredentials),
```
justinkaseman, line 144: For this to work in local development environments on a localhost URL, it is going to need insecure credentials.
See billing client [here](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/billing/workflow_client.go#L58).

pkg/storage/workflow_client.go
anirudhwarrier: While it is true that there is no I/O done, but it does validate that the address is reachable and so there is some connection happening here. Looked at other clients and they have similar log events.

leeyikjiun: Nope.

```go
_, err := grpc.NewClient("asdasd",
	grpc.WithTransportCredentials(insecure.NewCredentials()))
fmt.Printf("err: %v\n", err)
```
```
err: <nil>
```

PR #1428, opened 2025-07-23
Migrate core plugin codec types to common

huangzhen1997:


pkg/types/ccipocr3/mocks/source_chain_extra_data_codec.go
```diff
@@ -0,0 +1,151 @@
+// Code generated by mockery v2.53.0. DO NOT EDIT.
+
+package mocks
```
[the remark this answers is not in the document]
huangzhen1997: We have some test for the executecodec that would require mocking, that's why I added mock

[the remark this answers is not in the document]
huangzhen1997: I'm not sure if this matters, I noticed some these interfaces are also mocked in chainlink-common:
```
  github.com/smartcontractkit/chainlink-common/pkg/types:
    interfaces:
      ContractReader:
      EVMService:
      TONService:
```

pkg/types/ccipocr3/plugincodec.go
```diff
@@ -0,0 +1,80 @@
+package ccipocr3
```
huangzhen1997, line 1: Those interfaces come from https://github.com/smartcontractkit/chainlink/tree/develop/core/capabilities/ccip/common

.mockery.yaml
```diff
@@ -17,6 +17,10 @@ packages:
       OffRampReader:
         config:
           filename: offramp_reader_mock.go
+  github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3:
```
winder, line 20: If you aren't using these here, you can mock them in your other repo.

winder (changes requested):
I don't think you need to define the mock objects here.

.mockery.yaml
huangzhen1997: Removed

jmank88 (approved):

winder (approved):

PR #1433, opened 2025-07-28
first pass at implementing a zap log handler for slog

EasterTheBunny:

### DIFF
--- a/pkg/logger/v2/logger.go
+++ b/pkg/logger/v2/logger.go
@@ -6,9 +6,12 @@ import (
 	"log/slog"
 	"runtime"
 	"strings"
+	"testing"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
+	"go.uber.org/zap/zaptest/observer"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/config/build"
 )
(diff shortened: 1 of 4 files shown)


pkg/logger/v2/logger.go
```diff
@@ -0,0 +1,169 @@
...
+	if hndlr, ok := handler.(ZapHandler); ok {
+		return slog.New(hndlr.WithName(name))
+	}
+
+	return logger
+}
+
+type ZapHandler interface {
+	WithName(string) slog.Handler
+}
```
jmank88, line 59: I think we could inline this and maybe keep it unexported until there is a need to expose it? I was worried more about multiple internal types than external.
```suggestion
	if hndlr, ok := handler.(interface { withName(string) slog.Handler }); ok {
		return slog.New(hndlr.WithName(name))
	}

	return logger
}
```

pkg/logger/v2/logger.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/config/build"
+)
+
+type Config struct {
+	Name  string
+	Level slog.Leveler
+
+	// (optional) Logger helps convert existing zap.Logger to slog.Logger.
+	Logger *zap.Logger
```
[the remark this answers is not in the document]
EasterTheBunny: The only reason I used this was to allow users to drop in an existing logger. I thought it might make the conversion easier.

pkg/logger/v2/logger.go
EasterTheBunny: That works. I had it this way to run some tests, but it's not necessary.

PR #1437, opened 2025-07-30
[CRE-595] Fix context handling in BeholderLogger

bolekk:
Push to a channel serviced by the caller (ideally in a separate goroutine)

### DIFF
--- a/pkg/custmsg/logger.go
+++ b/pkg/custmsg/logger.go
@@ -8,227 +8,228 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 )
 
-const (
-	LogLevelKey                  = "log_level"
-	defaultBeholderEmitTimeoutMs = 100
-)
+const LogLevelKey = "log_level"
 
-type beholderLogger struct {
-	underlying  logger.Logger
-	emitter     MessageEmitter
-	emitTimeout time.Duration
+type LogTaskProcessor interface {
+	Enqueue(func(context.Context))
 }
 
-func NewBeholderLogger(underlying logger.Logger, emitter MessageEmitter) *beholderLogger {
+type beholderLogger struct {
+	underlying logger.Logger
+	emitter    MessageEmitter
+	emitQueue  LogTaskProcessor
+}
+
+func NewBeholderLogger(underlying logger.Logger, emitter MessageEmitter, emitQueue LogTaskProcessor) *beholderLogger {
 	return &beholderLogger{
-		underlying:  logger.Helper(underlying, 1),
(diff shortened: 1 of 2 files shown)


pkg/custmsg/logger.go
```diff
@@ -254,7 +251,10 @@ func getLabelMap(keysAndValues ...any) map[string]string {
...
 }
 
-func getBeholderCallContext() (context.Context, context.CancelFunc) {
-	// Beholder Emit() should not block for logs wrapped in a custom event but let's be safe
-	return context.WithTimeout(context.Background(), defaultBeholderEmitTimeoutMs*time.Millisecond)
+func (b *beholderLogger) enqueue(fn func(context.Context)) {
+	select {
+	case b.emitQueue <- fn:
+	default:
+		b.underlying.Errorw("emit queue is full, dropping log", "log_level", "debug")
```
krehermann, line 258: can we have a msgId or some other identifier logged here so we know which message was dropped?

iucc we will see the underlying log, and this message, but we can't put them to together to uniquely determine which log is dropped.

as it stands,iiuc,  are we at least able to know which wf execution dropped the message bc the underlying logger is annotated. this that right?

i think we should we have an metric and an alert for this kind of dropping behavior. user's will be effected by the mis-behavior

bolekk: What is the benefit of knowing which logs were dropped? If some are dropped we know we have a problem with Beholder but it will be very hard to recover them if that's what you're thinking about.

Also user logs (i.e. the ones coming out of WASM) are handled outside of this Logger (see emitUserLogs in engine.go).

I could move pushing to channel into Engine to add a metric more easily ..?

bolekk: I moved this to the Engine.

jmank88 (approved):

pkg/custmsg/logger.go
```diff
@@ -8,227 +8,228 @@ import (
...
+	Enqueue(func(context.Context))
+}
 
 type beholderLogger struct {
-	underlying  logger.Logger
-	emitter     MessageEmitter
-	emitTimeout time.Duration
+	underlying logger.Logger
+	emitter    MessageEmitter
+	emitQueue  LogTaskProcessor
```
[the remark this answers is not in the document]
bolekk: Yes, the problem is that Beholder Client puts Emit() requests onto a separate goroutine while keeping the context passed by us. So the fixed 100ms that I set is failing easily if that goroutine slows down :( 

There is a Slack thread where we agreed to push it asynchronously on our side as well (I'll tag you there).

PR #1448, opened 2025-08-05
DBConfig: add EnableTracing flag

pkcll:
DBConfig: add EnableTracing flag (default false)


patrickhuie19:
> Can we keep the default to true? Escape hatch is great 👍 , but ideally we're shipping nodes with a large amount of tracing OOTB.

NVM - better to handle in core.

bukata-sa (approved):

PR #1457, opened 2025-08-06
Refactor DonTimeStore

DylanTinianov:
Removes batching and consensus store dependency, instead sends all requests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,8 +37,8 @@ require (
 	github.com/shopspring/decimal v1.4.0
 	github.com/smartcontractkit/chain-selectors v1.0.62
 	github.com/smartcontractkit/chainlink-common/pkg/chipingress v0.0.1
-	github.com/smartcontractkit/chainlink-common/pkg/values v0.0.0-20250718143957-41236f9ef8b4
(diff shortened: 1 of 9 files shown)


pkg/workflows/dontime/plugin.go
```diff
@@ -66,52 +66,38 @@ func (p *Plugin) Observation(_ context.Context, outctx ocr3types.OutcomeContext,
...
 
-	// Collect up to batchSize unexpired requests
 	requests := map[string]int64{} // Maps executionID --> seqNum
-	for batchOffset := 0; batchOffset < p.store.requests.Len() && len(requests) < p.batchSize; {
-		batch, err := p.store.requests.RangeN(batchOffset, p.batchSize)
-		if err != nil {
-			p.lggr.Errorf("failed to get request batch at offset %d: %v", batchOffset, err)
-			batchOffset += p.batchSize
+	timeoutCheck := time.Now()
+	for _, req := range p.store.GetRequests() {
```
bolekk, line 71: I'd like to highlight the risk again - if this ever gets larger than max Query size, then the plugin could get stuck forever (until node restart). Request expiry should prevent that from happening.

bolekk (approved):

PR #1463, opened 2025-08-07
Fix broken external chainlink builds in CI

chainchad:
Builds now in: https://github.com/smartcontractkit/chainlink-common/actions/runs/16815947249/job/47632688098?pr=1463.

Updates:
- Used github-script to extract the ref's from the PR body with validation.
- Converted chainlink builds from Docker to just Go to test compatibility.

core ref: 01facb75d9af98873fda0c8a0e0707e67f259f2f

### DIFF
--- a/.github/scripts/get-refs-from-pr-body.js
+++ b/.github/scripts/get-refs-from-pr-body.js
@@ -1,0 +1,215 @@
+/**
+ * Configuration for repositories and their default refs
+ * Add new repos here to automatically include them in the ref extraction
+ */
+const REPO_CONFIG = {
+  core: {
+    pattern: /core ref:\s*(\S+)/i,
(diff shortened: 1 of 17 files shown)


chainchad:
@erikburt I ended up going back to build the image to leave this as it was in case there were plans to use it at some point.

@Tofel Erik brought up a really good point. We could shift this around to just do a `go build` of the chainlink core node (instead of a Docker build) and it would speed it up. Would that be more preferable or do we think we should keep it build a Docker image?

chainchad:
@jmank88 updated this to Go build now.

chainchad:
Once this gets one more approval, I'll need someone to click the merge button
> You're not authorized to push to this branch.

PR #1470, opened 2025-08-12
DF-21518 Remove seq nr validation in secure mint aggregator 

cll-gg:
### Requires

Nothing

### Supports

https://github.com/smartcontractkit/chainlink/pull/18926


mxiao-cll:
Why remove this?

cll-gg:
> Why remove this?

As I understand it, this is done in the consensus capability itself

PR #1488, opened 2025-08-19
INFOPLAT-2449: loop/server:  use otelzap logger

pkcll:
loop/server:  use otelzap logger

follow up for https://github.com/smartcontractkit/chainlink-common/pull/1486


pkg/loop/logger.go
```diff
@@ -171,6 +173,17 @@ func NewLogger() (logger.Logger, error) {
 	})
 }
 
+func NewOtelLogger(otelLogger otellog.Logger) (logger.Logger, error) {
+	cfgFn := func(cfg *zap.Config) {
+		cfg.Level.SetLevel(zap.DebugLevel)
+		cfg.EncoderConfig.LevelKey = "@level"
+		cfg.EncoderConfig.MessageKey = "@message"
+		cfg.EncoderConfig.TimeKey = "@timestamp"
+		cfg.EncoderConfig.EncodeTime = zapcore.TimeEncoderOfLayout("2006-01-02T15:04:05.000000Z07:00")
```
jmank88, line 182: Does this change the keys for both cores? I'm not sure that is safe to do.

pkcll: To clarify, this means ensure that the only first core is affected by this config and the second otel core is not. Also verify that the second core can properly get data from zap log entry and convert it into otel log record for this particular configuration.

PR #1494, opened 2025-08-20
pkg/services/otelhealth: add otelhealth; update promhealth

jmank88:


pkg/services/otelhealth/otelhealth.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return services.HealthCheckerConfig{}, err
+	}
+	uptimeSeconds, err := meter.Float64Gauge("uptime_seconds", metric.WithDescription("Uptime of the service"))
+	if err != nil {
+		return services.HealthCheckerConfig{}, err
+	}
+	cfg.AddUptime = func(ctx context.Context, d time.Duration) {
+		if cfg.AddUptime != nil {
+			cfg.AddUptime(ctx, d)
```
yashnevatia, line 37: will this not create an infinite loop ?
or are these functions passed by value ?

jmank88: Yeah I biffed a refactor :facepalm:   Should reference `orig` from here.

pkg/services/health.go
```diff
@@ -110,18 +111,18 @@ func (cfg HealthCheckerConfig) initVerSha() {
 	}
 }
 
-func (cfg HealthCheckerConfig) setNoopHooks() {
+func (cfg *HealthCheckerConfig) setNoopHooks() {
```
jmank88, line 114: This was leading to nil pointers

PR #1495, opened 2025-08-20
loop/server: wire up ChipIngressInsecureConnection option

pkcll:
loop/server: wire up ChipIngressInsecureConnection option

## Why? 
ChipIngress should use TLS by default, while Beholder does not use TLS for local connection to sidecar collector by default. 

Related PR https://github.com/smartcontractkit/chainlink-common/pull/1374 cc @krehermann @Tofel 

## Testing

Companion PR https://github.com/smartcontractkit/chainlink/pull/18947
Was tested on staging on cl-keystone-one


jmank88 (approved):

patrickhuie19 (approved):

krehermann (changes requested):
this will break the local env. please update all the relevant tomls like 


https://github.com/smartcontractkit/chainlink/blob/cec7743de3e74839ce92f236146ce9e320a4a076/core/scripts/cre/environment/configs/workflow-don.toml#L74

ie review all the toml artifacts in that dir and add your new config val as appropriate

pkcll:
> this will break the local env. please update all the relevant tomls like
> 
> https://github.com/smartcontractkit/chainlink/blob/cec7743de3e74839ce92f236146ce9e320a4a076/core/scripts/cre/environment/configs/workflow-don.toml#L74
> 
> ie review all the toml artifacts in that dir and add your new config val as appropriate

@krehermann,  pease see the note in the PR description "Companion PR https://github.com/smartcontractkit/chainlink/pull/18947"
I believe Its done there.

PR #1496, opened 2025-08-21
pkg/utils: add LazyLoadCtx variant

jmank88:


vyzaldysanchez (dismissed):

pkg/utils/lazy.go
```diff
@@ -34,3 +33,38 @@ func (l *LazyLoad[T]) Reset() {
 	defer l.lock.Unlock()
 	l.ok = false
 }
+
+type LazyLoadCtx[T any] struct {
```
krehermann, line 37: func comments please. i don't really understand the intention or use case

jmank88: Also added more docs

PR #1503, opened 2025-08-22
Adding TypeAndVersion to access in local registry in core

patrickhuie19:


pkg/types/core/capabilities_registry.go
```diff
@@ -52,6 +52,10 @@ func (UnimplementedCapabilitiesRegistryMetadata) ConfigForCapability(ctx context
 	return capabilities.CapabilityConfiguration{}, errors.New("ConfigForCapability not implemented")
 }
 
+func (UnimplementedCapabilitiesRegistryMetadata) TypeAndVersion(ctx context.Context) (string, string, error) {
+	return "", "", errors.New("TypeAndVersion not implemented")
```
jmank88, line 56: If you use the grpc status package, then you can include the `Unimplemented` code:
```suggestion
	return "", "", status.Errorf(codes.Unimplemented, "method TypeAndVersion not implemented")
```

PR #1504, opened 2025-08-25
[PRIV-128] Add DONForCapability to capability registry

cedric-cordenier:
- Add a new method to the capability registry
- Implement client and server methods for it
- This change is backwards compatible since all core implementations of the interface already embed the unimplemented type.

### DIFF
--- a/pkg/loop/internal/pb/capabilities_registry.proto
+++ b/pkg/loop/internal/pb/capabilities_registry.proto
@@ -11,6 +11,7 @@ service CapabilitiesRegistry {
   rpc LocalNode (google.protobuf.Empty) returns (NodeReply) {}
   rpc NodeByPeerID(NodeRequest) returns (NodeReply) {}
   rpc ConfigForCapability (ConfigForCapabilityRequest) returns (ConfigForCapabilityReply) {}
+  rpc DONForCapability (DONForCapabilityRequest) returns (DONForCapabilityReply) {}
   rpc Get (GetRequest) returns (GetReply) {}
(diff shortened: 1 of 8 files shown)


pkg/types/core/capabilities_registry.go
```diff
@@ -18,9 +18,9 @@ type CapabilitiesRegistryMetadata interface {
 	LocalNode(ctx context.Context) (capabilities.Node, error)
 	NodeByPeerID(ctx context.Context, peerID types.PeerID) (capabilities.Node, error)
 	ConfigForCapability(ctx context.Context, capabilityID string, donID uint32) (capabilities.CapabilityConfiguration, error)
+	DONForCapability(ctx context.Context, capabilityID string) (capabilities.DON, []capabilities.Node, error)
```
bolekk, line 21: DON already has a list of nodes inside, doesn't it?

pkg/loop/internal/pb/capabilities_registry.proto
```diff
@@ -112,3 +113,12 @@ message ConfigForCapabilityRequest {
 message ConfigForCapabilityReply {
   loop.CapabilityConfig capability_config = 1;
 }
+
+message DONForCapabilityRequest {
+  string capabilityID = 1;
+}
+
+message DONForCapabilityReply {
+  DON don = 1;
```
krehermann, line 122: i'm confused. can't there be multiple dons with the same capability? isn't that the point of families?

prashantkumar1982 (approved):

krehermann (approved):

PR #1506, opened 2025-08-26
DF-21546 Configure secure mint aggregator with data id

cll-gg:
So far, the data id that's used for writing the secure mint report to the DF Cache contract was constructed from the chain selector. However, to enable multiple secure mint reports on the same chain, we should use a more configurable data id. 
That's why in this PR the data id is set on the aggregator config in the workflow. 

### Supports

https://github.com/smartcontractkit/chainlink/pull/19106


Unheilbar (approved):

Atrax1 (approved):
LGTM

PR #1507, opened 2025-08-28
[PRODCRE-792] Per-method onchain config for capabilities V2

bolekk:
1. Add per-method config - each method can be either a trigger or an executable.
2. Include some extra fields that were previously hardcoded in Launcher.


pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -141,15 +141,14 @@ func (cr *capabilitiesRegistryClient) ConfigForCapability(ctx context.Context, c
...
-		remoteExecutableConfig.RegistrationRefresh = prtc.RegistrationRefresh.AsDuration()
-		remoteExecutableConfig.RegistrationExpiry = prtc.RegistrationExpiry.AsDuration()
 	}
 
 	return capabilities.CapabilityConfiguration{
 		DefaultConfig:          mc,
 		RemoteTriggerConfig:    remoteTriggerConfig,
 		RemoteTargetConfig:     remoteTargetConfig,
 		RemoteExecutableConfig: remoteExecutableConfig,
+		// TODO CapabilityMethodConfig
```
fernandezlautaro, line 151: missing something?

pkg/capabilities/pb/registry.proto
```diff
@@ -16,20 +16,54 @@ message RemoteTriggerConfig {
...
+  uint32 server_max_parallel_requests = 7;
+  RequestHasherType request_hasher_type = 8;
+}
+
+message AggregatorConfig {
+  AggregatorType aggregator_type = 1;
+}
+
+message CapabilityMethodConfig {
+  oneof remote_config {
```
fernandezlautaro, line 62: this will require a custom unmarshall for the local CRE like this one https://github.com/smartcontractkit/chainlink/pull/19207, which makes me wonder if what I did for the "other" oneOf field makes sense once we merge this

bolekk: That is only for JSON config in Local CRE, right? Can we reuse the same logic here?

pkg/loop/internal/core/services/capability/capabilities_registry.go
bolekk: yes, filling it now

pkg/capabilities/capabilities.go
```diff
@@ -546,4 +580,7 @@ type CapabilityConfiguration struct {
 	RemoteTriggerConfig    *RemoteTriggerConfig
 	RemoteTargetConfig     *RemoteTargetConfig
 	RemoteExecutableConfig *RemoteExecutableConfig
+
+	// v2 / "NoDAG" capabilities
+	CapabilityMethodConfig map[string]CapabilityMethodConfig
```
fernandezlautaro, line 585: I'll add a comment here saying the key has to match the exact literal value of the capability's [proto file](https://github.com/smartcontractkit/chainlink-protos/blob/main/cre/capabilities/blockchain/evm/v1alpha/client.proto#L259C7-L259C17) like `LogTrigger` (trigger), `WriteReport` (action/executable), `GetTransactionByHash` (action/executable) as the generated Go code for the actions will be consistent, but for the trigger will prepend the `Register`, such as `RegisterLogTrigger` and the method to be added to that map is `LogTrigger` and not `RegisterLogTrigger`)

cedric-cordenier (approved):

PR #1510, opened 2025-08-29
remove auth rate limit check and payload size from CRESettings

jinhoonbang:
- remove auth rate limit check and payload size from CRESettings. These checks are applied globally (including non-cre components) for the gateway via TOML config
- add cacheAge default for HTTP action

### DIFF
--- a/pkg/settings/cresettings/defaults.json
+++ b/pkg/settings/cresettings/defaults.json
@@ -2,9 +2,6 @@
 	"WorkflowLimit": "200",
 	"WorkflowRegistrationQueueLimit": "20",
 	"WorkflowExecutionConcurrencyLimit": "50",
-	"HTTPTrigger": {
-		"AuthRateLimit": "100rps:-1"
-	},
 	"PerOrg": {
 		"WorkflowDeploymentRateLimit": "every1m0s:1",
 		"ZeroBalancePruningTimeout": "24h0m0s"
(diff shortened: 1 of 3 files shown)


jmank88 (changes requested):
Removing settings doesn't make sense. What effect are you trying to achieve?

pkg/settings/cresettings/defaults.json
```diff
@@ -2,9 +2,6 @@
 	"WorkflowLimit": "200",
 	"WorkflowRegistrationQueueLimit": "20",
 	"WorkflowExecutionConcurrencyLimit": "50",
-	"HTTPTrigger": {
```
jinhoonbang, line 5: include unauthenticated limits here

pkg/settings/cresettings/defaults.json
```diff
@@ -38,9 +35,6 @@
 		},
 		"HTTPTrigger": {
 			"RateLimit": "every30s:3",
-			"AuthRateLimit": "1rps:-1",
-			"IncomingPayloadSizeLimit": "10kb",
```
jinhoonbang, line 42: move to global level.

jinhoonbang: gave this a little more thought and I don't see the need for every product to have a workflow-level incomingPayloadSizeLimit. I've raised this in the [doc](https://docs.google.com/document/d/1NpJNhEi6OUuQ7EQ80kun28HLPSy2dLGaLPH6U4Va4-s/edit?disco=AAABqj49kgs) and can revisit this if needed

pkg/settings/cresettings/settings.go
```diff
@@ -157,14 +155,8 @@ type Workflows struct {
 type cronTrigger struct {
 	RateLimit Setting[config.Rate]
 }
-type httpTriggerGlobal struct {
```
jinhoonbang, line 160: removing httpTriggerGlobal and instead adding `Gateway*` limits to top-level `Schema` because the gateway limits not only affect httpTrigger but also Vault

jmank88 (approved):

ilija42 (approved):

PR #1523, opened 2025-09-03
add offramp address and plugin type for the ccip provider

huangzhen1997:
Adding two more field to the CCIP provider args to support contract transmitter instantiation in chainlink core oracle creator plugin


ogtownsend:
can you update `CcipProviderArgs` usages here too? https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/loop/internal/relayer/relayer.go

ogtownsend:
and one nit if it doesn't cause too much messiness, could you update `rargs CCIPProviderArgs` to `pargs CCIPProviderArgs` or `cargs CCIPProviderArgs`: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/types/relayer.go#L258

ogtownsend (approved):

PR #1527, opened 2025-09-07
[CRE=804] - Add tx idempotency key to EVM Service SubmitTransaction reply

ilija42:
Idempotency key wasn't surfaced from SubmiTransaction which makes it impossible to retrieve some tx meta data from the evm chain capability


silaslenihan (approved):

patrickhuie19 (approved):

pkg/chains/evm/evm.proto
```diff
@@ -322,6 +322,7 @@ message SubmitTransactionRequest {
 message SubmitTransactionReply {
   bytes txHash = 1;
   TxStatus txStatus = 2;
+  string tx_idempotency_key = 3; // idempotency key generated for this transaction
```
patrickhuie19, line 325: are txHash and tx_idempotency_key related in any way? we do that for workflowExecutionIDs, and it tends to be pretty useful in scenarios when you need to forecast an execution ID

PR #1534, opened 2025-09-09
adding otlptracegrpc.WithCompressor(gzip),

yashnevatia:

### DIFF
--- a/pkg/loop/telem.go
+++ b/pkg/loop/telem.go
@@ -28,8 +28,6 @@ import (
 	loopnet "github.com/smartcontractkit/chainlink-common/pkg/loop/internal/net"
 )
 
-const DefaultCompressionType = "gzip"
-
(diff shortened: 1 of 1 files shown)


pkg/loop/telem.go
```diff
@@ -149,6 +149,7 @@ func (config TracingConfig) NewSpanExporter() (sdktrace.SpanExporter, error) {
 	traceExporter, err := otlptracegrpc.New(ctx,
 		otlptracegrpc.WithGRPCConn(conn),
 		otlptracegrpc.WithHeaders(config.AuthHeaders),
+		otlptracegrpc.WithCompressor("gzip"),
```
dhaidashenko, line 152: Use `gzip.Name` for imported artifacts. gzip package contains init that runs only if pkg is imported. Without it compression might not work

PR #1551, opened 2025-09-16
Add schema registry to beholder client

engnke:
[Jira Ticket](https://smartcontract-it.atlassian.net/browse/INFOPLAT-2884)
- Add a generic chipIngressClient with one method allowing users to register schemas
- Beholder Clients can use `client.chip.RegisterSchema(...)` to register one or more schemas


pkg/beholder/registrar.go
```diff
@@ -0,0 +1,40 @@
+package beholder
+
+import (
+	"context"
+	"fmt"
+	"github.com/smartcontractkit/chainlink-common/pkg/chipingress"
+	"github.com/smartcontractkit/chainlink-common/pkg/chipingress/pb"
+)
+
+type Registrar interface {
```
RamSinha, line 10: Can we prefix it with Chip, something like ChipSchemaRegistrar so it carries the context..

pkg/beholder/registrar.go
```diff
@@ -0,0 +1,40 @@
...
+		return nil, fmt.Errorf("chip ingress client is nil")
+	}
+
+	return &schemaRegistry{
+		client: client,
+	}, nil
+}
+
+// Register registers one or more schemas with the Chip Ingress service
+func (sr *schemaRegistry) RegisterSchema(ctx context.Context, schemas ...*pb.Schema) error {
```
engnke, line 29: @patrickhuie19 @jmank88 

Wanted to get some feedback here on the inputs for this new`RegisterSchema` endpoint that would allow a beholder client to register a schema (backend chip-ingress service)
Flow
- Create a beholder client with `ChipSchemaRegistry` enabled
- Initiate schema registrations with `client.registrar.RegisterSchema` 
-- Is the method name okay or would `Register` be preferred?
- The inputs would be a list of schemas to register, all passed in as `Schema` objects ([proto def](https://github.com/smartcontractkit/chainlink-common/blob/b41d6a9e3c7172145195c6c436c57f3b086bdcf0/pkg/chipingress/pb/chip_common.proto#L8)). 
- -Is this okay or would a different input value be preferred thats more user friendly

go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/scylladb/go-reflectx v1.0.1
 	github.com/shopspring/decimal v1.4.0
 	github.com/smartcontractkit/chain-selectors v1.0.67
-	github.com/smartcontractkit/chainlink-common/pkg/chipingress v0.0.1
+	github.com/smartcontractkit/chainlink-common/pkg/chipingress v0.0.0-20250916154016-04a897dcb361
```
jmank88, line 39: Why the downgrade?

engnke: This has beenn updated to the proper tag v0.0.4

pkg/beholder/registrar.go
jmank88: There is nothing wrong in principle with using the generated protobuf types :shrug:

jmank88: The method name seems fine to me, but see my other name comment ^ :+1:

pkg/loop/config.go
```diff
@@ -121,6 +121,7 @@ type EnvConfig struct {
 
 	ChipIngressEndpoint           string
 	ChipIngressInsecureConnection bool
+	ChipSchemaRegistryEnabled     bool
```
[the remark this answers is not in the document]
engnke: I don’t see any issues with having it always enabled. My thought in adding the feature flag was just to be explicit, but we could just default it to enabled. If there’s no real use case for disabling it, I’m fine with leaving it as always on given that the `ChipIngressEndpoint` above is provided.

pkg/beholder/chip_client.go
```diff
@@ -0,0 +1,40 @@
...
+		return nil, fmt.Errorf("chip ingress client is nil")
+	}
+
+	return &chipIngressClient{
+		client: client,
+	}, nil
+}
+
+// RegisterSchema registers one or more schemas with the Chip Ingress service
+func (sr *chipIngressClient) RegisterSchema(ctx context.Context, schemas ...*pb.Schema) error {
```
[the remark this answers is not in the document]
engnke: Fair point this is pretty minimal. I’d still lean against exposing the full chipingress.Client, since theres other functions not needed her. Also we can stay flexible if we want to evolve the inputs here later on

pkg/loop/config.go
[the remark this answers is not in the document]
jmank88: I think it's worth the simplicity to leave it out for now.

pkcll (approved):

engnke:
@jmank88 do you mind merging this for us? Thank you

PR #1552, opened 2025-09-17
CCIP - ChainAccessor add test for Sync call state persistence across restarts

ogtownsend:
- Adds a test to verify that the persisted `syncs` map contents of `ChainAccessorClient` are persisted over relayer server crashes and restarts


pkg/loop/internal/net/client.go
```diff
@@ -96,6 +96,7 @@ func (c *clientConn) NewStream(ctx context.Context, desc *grpc.StreamDesc, metho
 // refresh replaces c.cc with a new (different from orig) *grpc.ClientConn, and returns it as well.
 // It will block until a new connection is successfully dialed, or return nil if the context expires.
 func (c *clientConn) refresh(ctx context.Context, orig *grpc.ClientConn) *grpc.ClientConn {
+	c.Logger.Debugw("clientConn: refreshing connection")
```
ogtownsend, line 99: oops. missed this one, will remove

jmank88 (approved):

PR #1567, opened 2025-09-29
INFOPLAT 2962 rotating beholder headers

hendoxc:
## What
- **Adds rotating header impl**
- **`beholder.Config` has `Signer` property, which can be satisfied by passing `keystore.Keystore`

## Why
- beholder. uses auth headers that expire


pkg/beholder/client.go
```diff
@@ -91,11 +94,42 @@ func NewGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
...
 	}
+	// Initialize auth here for reuse with log, trace, and metric exporters
+	var auth Auth
+	if cfg.AuthKeySigner != nil {
+
+		if cfg.AuthPublicKeyHex == "" {
+			return nil, fmt.Errorf("auth: public key hex required when signer is set")
+		}
+		// Set the min TTL to 1 minute, avoid signing headers too often if config is misconfigured
+		if cfg.AuthHeadersTTL < 1*time.Minute {
```
pkcll, line 110: This needs to be set explicitly in chainlink core where this is wired up

pkcll: 10min seems sensible default for this

pkg/beholder/auth.go
```diff
@@ -1,30 +1,122 @@
...
+		csaPubKey:                csaPubKey,
+		signer:                   signer,
+		ttl:                      ttl,
+		headers:                  make(map[string]string),
+		lastUpdated:              time.Unix(0, 0),
+		requireTransportSecurity: requireTransportSecurity,
+	}
+}
+
+func (r *rotatingAuth) Headers(ctx context.Context) (map[string]string, error) {
```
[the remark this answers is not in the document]
hendoxc: addressed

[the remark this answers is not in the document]
jmank88: The first line accessing `r.lastupdated` needs to be behind the lock too.

pkg/beholder/auth.go
```diff
@@ -1,30 +1,147 @@
...
+	"sync/atomic"
+	"time"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/chipingress"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 )
 
 // authHeaderKey is the name of the header that the node authenticator will use to send the auth token
 var authHeaderKey = "X-Beholder-Node-Auth-Token"
```
fouadkada, line 20: I would go for a `const` instead of a `var`

pkg/beholder/auth.go
```diff
@@ -1,30 +1,147 @@
...
 
 // authHeaderVersion is the version of the auth header format
 var authHeaderVersion = "1"
```
fouadkada, line 23: same, `const` instead of a `var`

pkg/beholder/auth.go
```diff
@@ -1,30 +1,147 @@
...
+	lastUpdatedNanos         atomic.Int64
+	requireTransportSecurity bool
+	mu                       sync.Mutex
+}
+
+func NewRotatingAuth(csaPubKey ed25519.PublicKey, signer Signer, ttl time.Duration, requireTransportSecurity bool) Auth {
+	return &rotatingAuth{
+		csaPubKey:                csaPubKey,
+		signer:                   signer,
+		signerTimeout:            time.Second * 5,
```
fouadkada, line 88: should this be configurable or is it an overkill?

pkg/beholder/auth.go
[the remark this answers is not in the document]
jmank88: Ah nvm, looks like that is accounted for!

jmank88 (approved):

pkcll (approved):

pkcll (dismissed):

fouadkada (approved):

PR #1572, opened 2025-10-01
Beholder heartbeat 

pkcll:
Add beholder heartbeat 
Based of https://github.com/smartcontractkit/chainlink-common/pull/1566


pkg/beholder/heartbeat.go
```diff
@@ -0,0 +1,227 @@
...
+		h.Meter = meter
+	}
+}
+
+// WithTracer sets a custom tracer for the heartbeat
+func WithTracer(tracer trace.Tracer) HeartbeatOpt {
+	return func(h *Heartbeat) {
+		h.Tracer = tracer
+	}
+}
```
hendoxc, line 102: These will pollute the `beholder` pkg namespace. might be better to use a config struct like `heartbeatConfig`

PR #1580, opened 2025-10-02
Keystore library interfaces and admin create/delete

connorwstein:
https://smartcontract-it.atlassian.net/browse/ARCH-328


keystore/admin.go
```diff
@@ -0,0 +1,201 @@
...
+	ksCopy := maps.Clone(ks.keystore)
+	var responses []CreateKeyResponse
+	for _, keyReq := range req.Keys {
+		if _, ok := ksCopy[keyReq.KeyName]; ok {
+			return CreateKeysResponse{}, fmt.Errorf("%w: %s", ErrKeyAlreadyExists, keyReq.KeyName)
+		}
+		if err := ValidKeyName(keyReq.KeyName); err != nil {
+			return CreateKeysResponse{}, fmt.Errorf("%w: %s", ErrInvalidKeyName, err)
+		}
+		switch keyReq.KeyType {
```
connorwstein, line 121: still some more key types to add (starknet) but will do that in a follow up as this is already getting large

keystore/keystore.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+type keystore struct {
+	mu       sync.RWMutex
+	keystore map[string]key
+	storage  storage.Storage
+	enc      EncryptionParams
+}
+
+func NewKeystore(ctx context.Context, storage storage.Storage, enc EncryptionParams) (Keystore, error) {
```
connorwstein, line 141: I know that context via constructor is frowned upon, but here I'm not sure - if I add a separate Load(ctx) function I think it'd add a bunch of boilerplate to clients always loading after creating and also error handling for every method checking if loaded or not. Seems simpler this way - wdyt @jmank88

keystore/keystore.go
```diff
@@ -0,0 +1,222 @@
...
+func publicKeyFromPrivateKey(privateKeyBytes internal.Raw, keyType KeyType) ([]byte, error) {
+	switch keyType {
+	case Ed25519:
+		return ed25519.PublicKey(internal.Bytes(privateKeyBytes)), nil
+	case EcdsaSecp256k1:
+		// Here we use SEC1 (uncompressed) format for ECDSA public keys.
+		// Its commonly used and EVM addresses are derived from this format.
+		// We use the geth crypto library for secp256k1 support
+		// because stdlib doesn't support it.
+		privateKey, err := gethcrypto.ToECDSA(internal.Bytes(privateKeyBytes))
```
connorwstein, line 116: I'm not sure why the core node logic doesn't use these geth methods but instead copies the inner logic (still depends on geth so wouldn't be to remove the dep) https://github.com/smartcontractkit/chainlink/blob/develop/core/services/keystore/keys/ethkey/key_v2.go#L23. Seems cleaner to use the methods unless I'm missing something

keystore/keystore.go
```diff
@@ -0,0 +1,222 @@
...
+	enc      EncryptionParams
+}
+
+func NewKeystore(ctx context.Context, storage storage.Storage, enc EncryptionParams) (Keystore, error) {
+	ks, err := load(ctx, storage, enc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load keystore: %w", err)
+	}
+	return &keystore{
+		mu:       sync.RWMutex{},
```
jmank88, line 147: No effect:
```suggestion
```

keystore/keystore.go
jmank88: Usually the problems are deeper than this, like storing the context or launching a goroutine. This just seems like a matter of readability. We can keep concerns separate without forcing boilerplate on anyone though. In this case, the constructor is not doing anything else, so it could be eliminated in exchange for a load method:
```go
type StoredKeystore struct {
	Storage  storage.Storage
	Enc      EncryptionParams

	mu sync.RWMutex
	keystore map[string]key
}

func (k *StoredKeystore) Load(ctx context.Context) error {
```
Or something like an `Init` to avoid the exported fields:
```suggestion
func (*StoredKeystore) Init(ctx context.Context, storage storage.Storage, enc EncryptionParams) (Keystore, error) {
```

keystore/keystore.go
```diff
@@ -0,0 +1,222 @@
...
+		Metadata:  metadata,
+	}
+}
+
+type Keystore interface {
+	Admin
+	Reader
+	Signer
+	Encryptor
+}
```
jmank88, line 75: We should probably use the unexported method trick to be able to grow this interface in a compatible way:
```suggestion
type Keystore interface {
	Admin
	Reader
	Signer
	Encryptor
	
	mustEmbedUnimplementedKeystore()
}

type UnimplementedKeystore struct {}

func (UnimplementedKeystore) mustEmbedUnimplementedKeystore() {}
```

jmank88: (and perhaps the sub-interfaces? 🤷 )

keystore/keystore.go
```diff
@@ -0,0 +1,222 @@
...
+		publicKey, err := publicKeyFromPrivateKey(pkRaw, keyType)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get public key from private key: %w", err)
+		}
+		keystore[k.Name] = newKey(keyType, pkRaw, publicKey, time.Unix(k.CreatedAt, 0), k.Metadata)
+	}
+	return keystore, nil
+}
+
+func save(ctx context.Context, storage storage.Storage, enc EncryptionParams, keystore map[string]key) error {
```
jmank88, line 192: Did you consider making load/save methods on `keystore`? Or does the map copying make it odd?

connorwstein: wasn't really map copying but I think I was thinking this would be a bit more readable (esp in tests). But on second thought I think its fine I can just do that

keystore/keystore.go
connorwstein: Renamed. The concern was just about adding something like if not loaded: return ErrNotLoaded to all the ks methods to avoid potentially opaque errors on empty map

keystore/keystore.go
```diff
@@ -0,0 +1,244 @@
...
+type Keystore interface {
+	Admin
+	Reader
+	Signer
+	Encryptor
+}
+
+type Unimplemented struct{}
+
+func (Unimplemented) mustEmbedUnimplementedKey() {}
```
jmank88, line 94: If we expect external implementations of these interfaces, then they should each have their own `UnimplmentedX`, with methods satisfying `X`. In the current form, adding a new method is still a breaking change.

jmank88: The point of an exported `Unimplemented` type is to be embedded in implementations so that we may add methods to the interface in the future without breaking compatibility. If we don't want other implementations at all, then we should drop this type - but that seems unnecessarily restrictive, why make it impossible to make a simple fake/mock implementation?

connorwstein: ok misunderstood its purpose then. I was thinking of just preventing external implementations all together (even a fake/mock implementation would be supplied in this package). Can switch to the other model of allow external impls which would embed the type so they can be extended

jmank88 (approved):
LGTM although I am wondering why GRPC chose to use pointers for their `XRequest`/`XResponse` types... 🤔 Maybe just efficiency?
Request seems annoying as a pointer, since you never pass nil. On the other hand, Response is cleaner as a pointer since you can return `nil, err` instead of explicit zero values.

pavel-raykov (approved):

PR #1587, opened 2025-10-06
Implement encryptor in keystore library

connorwstein:
Note signer impl already reviewed.


keystore/serialization/keystore.proto
```diff
@@ -15,4 +15,14 @@ message Key {
   int64 created_at = 3;
   string key_type = 4;
   bytes metadata = 5;
+}
+
+// ECDH encryption envelope for hybrid encryption
+message EcdhEnvelope {
```
jmank88, line 21: Should these be caps per go naming standards?
```suggestion
message ECDHEnvelope {
```

connorwstein: sure will change

cedric-cordenier (dismissed):

PR #1588, opened 2025-10-06
[CRE-954] Extra logging limits in WASM Module

bolekk:


pkg/workflows/wasm/host/module.go
```diff
@@ -529,6 +545,7 @@ func runWasm[I, O proto.Message](
 		secretsResponses:    map[int32]<-chan *secretsResponse{},
 		module:              m,
 		executor:            helper,
+		mode:                sdkpb.Mode_MODE_DON,
```
bolekk, line 548: @nolag should this be set and to what?

krehermann (approved):

bolekk:
I will wait for @nolag's response before merging.

pkg/workflows/wasm/host/execution.go
```diff
@@ -139,6 +141,32 @@ func (e *execution[T]) awaitSecrets(ctx context.Context, acr *sdkpb.AwaitSecrets
 }
 
 func (e *execution[T]) log(caller *wasmtime.Caller, ptr int32, ptrlen int32) {
+	switch e.mode {
+	case sdkpb.Mode_MODE_DON:
+		e.donLogCount++
+		if e.donLogCount == e.module.cfg.MaxLogCountDONMode {
+			e.module.cfg.Logger.Errorf("max log count for don mode reached: %d - all subsequent logs will be dropped", e.donLogCount)
```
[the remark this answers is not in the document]
bolekk: Yes, it will be. It's valuable info to see that a workflow crossed the threshold as it might indicate a malicious one. Do you think differently?

pkg/workflows/wasm/host/module.go
bolekk: But then how do we make the new test work?

bolekk: Added manual mode switch at the beginning of the test.

nolag (approved):
Approved with a comment, I haven't been on-call so I won't block it if you think it's ok to make it an error after reading my reasoning.

pkg/workflows/wasm/host/execution.go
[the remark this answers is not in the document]
bolekk: OK, how about  I'll switch to warn. Error logs don't cause pages anyway so it wouldn't be THAT critical.

PR #1590, opened 2025-10-07
update standard cap initialise to take a struct

patrickhuie19:
Adding new services to Initialise in the standard capabilities interface is always breaking. By making it a struct, we can make updates a little less painful, and only opt-in the capabilities that need the new service(s).

Capabilities PR: https://github.com/smartcontractkit/capabilities/pull/290
Core PR: https://github.com/smartcontractkit/chainlink/pull/19768
common PR: https://github.com/smartcontractkit/chainlink-common/pull/1590

Order of operations:
[] get approvals on common, capabilities, core
[ x ] merge common
[ x ] merge core
[ x ] update capabilities to pull in the new core ref
[ x ] update core to reference the new git ref of capabilities

### DIFF
--- a/pkg/types/gateway/action.go
+++ b/pkg/types/gateway/action.go
@@ -15,8 +15,10 @@ const (
 
 // CacheSettings defines cache control options for outbound HTTP requests.
 type CacheSettings struct {
-	ReadFromCache bool  `json:"readFromCache,omitempty"` // If true, attempt to read a cached response for the request
-	MaxAgeMs      int32 `json:"maxAgeMs,omitempty"`      // Maximum age of a cached response in milliseconds.
+	MaxAgeMs int32 `json:"maxAgeMs,omitempty"` // Maximum age of a cached response in milliseconds.
+	Store    bool  `json:"store,omitempty"`    // If true, cache the response.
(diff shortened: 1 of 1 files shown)


pkg/types/core/standard_capabilities_services.go
```diff
@@ -0,0 +1,16 @@
+package core
+
+// StandardCapabilitiesServices contains all the services required for capability initialization.
+// We use a struct to evolve the interface without requiring updates to all implementors.
+type StandardCapabilitiesServices struct {
```
jmank88, line 5: How about a more general name that also avoids collisions with the `services` packages?
```suggestion
// StandardCapabilitiesDependencies contains all the dependencies injected for capability initialization.
// We use a struct to evolve the interface without requiring updates to all implementors.
type StandardCapabilitiesDependencies struct {
```

jmank88: (or `InitialiseRequest` to match the grpc layer underneath 🤷 )

patrickhuie19: ec1247e49492f16c8a062c5ec2cdd918f878c17e

jmank88 (approved):

nolag (approved):

PR #1591, opened 2025-10-08
beholder: add telemetry LogLevel config option for log streaming

pkcll:
beholder: add Telemetry.LogLevel config option for log streaming

Related PR https://github.com/smartcontractkit/chainlink/pull/19775


pkg/beholder/config.go
```diff
@@ -47,7 +48,8 @@ type Config struct {
 	LogBatchProcessor     bool // Enabled by default. Disable only for testing.
 	// Retry config for shared log exporter, used by Emitter and Logger
 	LogRetryConfig      *RetryConfig
-	LogStreamingEnabled bool // Enable logs streaming to the OTel log exporter
+	LogStreamingEnabled bool          // Enable logs streaming to the OTel log exporter
+	LogLevel            zapcore.Level // Log level for telemetry streaming
```
jmank88, line 52: We need to use this in `loop.NewOtelLogger` too, right?

PR #1593, opened 2025-10-08
pkg/loop: wait from chain family

jmank88:
The `Service` field is initialized async so access must be gated by a call to `Wait`.

### DIFF
--- a/pkg/loop/internal/goplugin/plugin_service.go
+++ b/pkg/loop/internal/goplugin/plugin_service.go
@@ -257,7 +257,17 @@ func (s *PluginService[P, S]) WaitCtx(ctx context.Context) error {
 	case <-s.serviceCh:
 		return nil
 	case <-s.stopCh:
-		return fmt.Errorf("service was stoped while waiting: %w", context.Canceled)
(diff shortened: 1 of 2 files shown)


pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -261,6 +261,16 @@ func (s *PluginService[P, S]) WaitCtx(ctx context.Context) error {
 	}
 }
 
+// Wait waits for the service to start up until it receives the stop signal.
+func (s *PluginService[P, S]) Wait() error {
+	select {
+	case <-s.serviceCh:
+		return nil
+	case <-s.stopCh:
+		return fmt.Errorf("service was stoped while waiting: %w", context.Canceled)
```
cll-gg, line 270:
```suggestion
		return fmt.Errorf("service was stopped while waiting: %w", context.Canceled)
```

vyzaldysanchez (approved):

ilija42 (approved):

pavel-raykov (approved):

PR #1862, opened 2026-02-27
chip ingress batching

thomaska:
Ticket: https://smartcontract-it.atlassian.net/browse/INFOPLAT-3436

## Summary
This PR adds chip-ingress batch delivery behind beholder, makes `beholder.Client` own the batch emitter lifecycle, and adds a service-layer escape hatch for cleanup when a lifecycle-managed service is closed before it was ever started.

## What changed
- added `ChipIngressBatchEmitterService`, which batches emitted events and publishes them through `chipingress.Client.PublishBatch`
- `beholder.Client` now owns the optional batch emitter as a sub-service and starts/stops it through its service lifecycle
- added `services.Config.CloseIfNeverStarted` so ctor-held resources can still run their close hook even when `Start()` was never called
- updated `DualSourceEmitter.Close()` so it does not directly close a service-managed batch emitter
- `pkg/chipingress/batch.Client` now records batch request metrics and flushes shutdown work independently of caller cancellation
- LOOP now starts/stops `beholder.Client` directly and no longer uses test-only indirection in `pkg/loop/server.go`
- removed the obsolete `pkg/loop/server_test.go`
- added LOOP config support for `CL_CHIP_INGRESS_BATCH_EMITTER_ENABLED`

## Data Flow
```mermaid
flowchart LR
    A["Caller or app code"] -->|"Emit(ctx, body, attrs)"| B["beholder.Client.Emitter"]
    B --> C["DualSourceEmitter"]
    C -->|"synchronous"| D["OTLP message emitter"]
    C -->|"batch enabled"| E["ChipIngressBatchEmitterService"]
    E -->|"queue event"| F["batch.Client"]
    F -->|"accumulate by size or interval"| G["PublishBatch request"]
    G --> H["chipingress.Client"]
    H --> I["Chip Ingress gRPC endpoint"]

    F --> J["batch client metrics"]
    E --> K["emitter success or drop metrics"]
    B --> L["beholder logger, tracer, and meter providers"]
```

## Dependency Diagram
```mermaid
flowchart TD
    S["loop.Server"] -->|"start and close"| BC["beholder.Client"]
    S --> CFG["loop.EnvConfig"]
    CFG --> FLAG["CL_CHIP_INGRESS_BATCH_EMITTER_ENABLED"]

    BC --> ENG1["services.Engine"]
    BC --> DSE["DualSourceEmitter"]
    BC --> CHIP["chipingress.Client"]
    BC --> OTLP["OTLP providers"]
    ENG1 --> BES["ChipIngressBatchEmitterService"]

    BES --> ENG2["services.Engine"]
    BES --> BATCH["batch.Client"]
    BATCH --> CHIP

    ENG1 --> CINS["CloseIfNeverStarted"]
    ENG2 --> CINS
    CINS --> SVC["services.Config"]
```

## Metrics
Added batch delivery metrics for:
- `chip_ingress.batch.send_requests_total`
- `chip_ingress.batch.send_failures_total`
- `chip_ingress.batch.request_size_messages`
- `chip_ingress.batch.request_size_bytes`
- `chip_ingress.batch.request_latency_ms`
- `chip_ingress.batch.config.info`
- `chip_ingress.events_sent`
- `chip_ingress.events_dropped`

## Tests
Added coverage for:
- `CloseIfNeverStarted` lifecycle behavior
- beholder client ownership of the batch emitter sub-service
- batch emitter lifecycle and success/drop metrics
- batch client metrics
- benchmark smoke coverage for batch queueing and emitter enqueue paths

### Supports
https://github.com/smartcontractkit/chainlink/pull/21327


pkg/beholder/chip_ingress_batch_emitter.go
```diff
@@ -0,0 +1,148 @@
...
+
+	e.Service, e.eng = services.Config{
+		Name:  "ChipIngressBatchEmitter",
+		Start: e.start,
+	}.NewServiceEngine(lggr)
+
+	return e, nil
+}
+
+func (e *ChipIngressBatchEmitter) start(_ context.Context) error {
```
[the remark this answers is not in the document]
thomaska: it was mostly added as a placeholder, but can be omitted as well. 
And after checking, in the core/services/workflows/syncer/v2/handler.go in `EventHandler` it's also omitted, so. probably it's more consistent.

pkg/beholder/chip_ingress_batch_emitter.go
```diff
@@ -0,0 +1,266 @@
...
+			return fmt.Errorf("failed to convert to proto: %w", err)
+		}
+
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		queueErr := worker.batchClient.QueueMessage(eventPb, func(sendErr error) {
+			if sendErr != nil {
+				e.metrics.eventsDropped.Add(context.Background(), 1, worker.metricAttrs)
```
hendoxc: why context.Background ?

thomaska:
> Blocking behavior regression in `DualSourceEmitter` when batch mode is off
> 
> In `pkg/beholder/dual_source_emitter.go`, chip-ingress emit was changed from async to inline:
> 
> ```go
> if err := d.chipIngressEmitter.Emit(ctx, body, attrKVs...); err != nil { ... }
> ```
> 
> This makes `DualSourceEmitter.Emit` block on chip-ingress in legacy mode (`ChipIngressBatchEmitterEnabled=false`, still default in `pkg/beholder/client.go`). Previously this path was fire-and-forget via goroutine, so caller latency/backpressure characteristics change materially.
> 
> Can we keep async behavior for non-batch emitters (or gate sync behavior strictly to `ChipIngressBatchEmitter`) to avoid regressing existing deployments?

that's a very good spot, thank you @pkcll 🙏

pkcll (approved):

pkcll (dismissed):

pkg/loop/server_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+func TestServerStartBeholderClientUsesBackgroundContext(t *testing.T) {
```
jmank88: What is this test evaluating?

pkg/loop/server.go
```diff
@@ -101,8 +101,16 @@ type Server struct {
...
 	LimitsFactory   limits.Factory
 	profiler        *pyroscope.Profiler
+	beholderClient  *beholder.Client
 }
 
+var (
+	newBeholderClient   = beholder.NewClient
+	startBeholderClient = func(client *beholder.Client, ctx context.Context) error {
+		return client.Start(ctx)
+	}
```
jmank88: This is extremely unusual. What is the purpose of using indirect, mutable variables?

pkg/loop/server.go
```diff
@@ -349,8 +348,34 @@ func (s *Server) MustRegister(c services.HealthReporter) {
 
 func (s *Server) Register(c services.HealthReporter) error { return s.checker.Register(c) }
 
+func (s *Server) startBeholderClient(beholderCfg beholder.Config) error {
+	beholderClient, err := newBeholderClient(beholderCfg)
+	if err != nil {
+		return fmt.Errorf("failed to create beholder client: %w", err)
+	}
+	if err := startBeholderClient(beholderClient, context.Background()); err != nil {
```
jmank88: We need to use a real context tied to the server's lifecycle

pkg/beholder/client.go
```diff
@@ -223,38 +235,101 @@ func NewGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
...
+		MessageLoggerProvider: messageLoggerProvider,
+		lazySigner:            signer,
+		OnClose:               onClose,
+		batchEmitterService:   batchEmitterService,
+	}
+	c.Service, c.eng = services.Config{
+		Name:  "BeholderClient",
+		Start: c.start,
+		Close: c.closeResources,
+	}.NewServiceEngine(pkglogger.Nop())
```
jmank88: Why are we using a no-op logger?

pkg/beholder/client.go
```diff
@@ -223,38 +235,101 @@ func NewGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
...
+		}
+		return c.Service.Close()
 	}
-	return
+	return c.closeResources()
+}
+
+func (c *Client) start(ctx context.Context) error {
+	if c.batchEmitterService != nil {
+		return c.batchEmitterService.Start(ctx)
```
jmank88: Does this need to be closed too? If pass it to the Engine instead, it will manage it for you.

pkg/beholder/client.go
```diff
@@ -223,38 +235,101 @@ func NewGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
...
-	if c.Emitter != nil {
-		err = errors.Join(err, c.Emitter.Close())
+	return []services.Service{c}
+}
+
+// Close shuts down OTel providers and the chip ingress connection.
+func (c *Client) Close() (err error) {
+	if c == nil {
+		return nil
 	}
```
jmank88:
```suggestion
It is the caller's responsibility to prevent nil receivers.
```

pkg/beholder/client.go
```diff
@@ -223,38 +235,101 @@ func NewGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
...
+	if c == nil || c.batchEmitterService == nil || c.Service == nil {
+		return nil
 	}
-	if c.Emitter != nil {
-		err = errors.Join(err, c.Emitter.Close())
+	return []services.Service{c}
+}
+
+// Close shuts down OTel providers and the chip ingress connection.
+func (c *Client) Close() (err error) {
```
jmank88: You must not override the Close() method implemented by the Engine. The func that you pass to the Config is your Close hook.

pkg/loop/server.go
pkcll: The vars were indirecting beholder.NewClient and client.Start solely to support test stubbing. Reverted that.

pkg/loop/server.go
pkcll: addressed

pkg/beholder/client.go
pkcll: Addressed

pkg/loop/server_test.go
pkcll: Removed it

pkg/beholder/client.go
pkcll: Addressed

pkg/beholder/client.go
pkcll: Adressed

pkg/chipingress/batch/client.go
```diff
@@ -230,7 +260,11 @@ func (b *Client) sendBatch(ctx context.Context, messages []*messageWithCallback)
 		for i, msg := range messages {
 			events[i] = msg.event
 		}
-		_, err := b.client.PublishBatch(ctxTimeout, &chipingress.CloudEventBatch{Events: events})
+		batchReq := &chipingress.CloudEventBatch{Events: events}
+		batchBytes := proto.Size(batchReq)
+		startedAt := time.Now()
+		_, err := b.client.PublishBatch(ctxTimeout, batchReq)
+		b.metrics.recordSend(context.Background(), len(messages), batchBytes, time.Since(startedAt), err == nil)
```
jmank88: Why background context?
```suggestion
		b.metrics.recordSend(ctxTimeout, len(messages), batchBytes, time.Since(startedAt), err == nil)
```

pkg/beholder/httpclient.go
```diff
@@ -181,13 +184,32 @@ func NewHTTPClient(cfg Config, otlploghttpNew otlploghttpFactory) (*Client, erro
...
+		MeterProvider:         meterProvider,
+		MessageLoggerProvider: messageLoggerProvider,
+		lazySigner:            nil,
+		OnClose:               onClose,
+	}
+	c.Service, c.eng = services.Config{
+		Name:                 "BeholderClient",
+		Close:                c.close,
+		CloseIfNeverStarted:  true,
+	}.NewServiceEngine(pkglogger.Nop())
```
jmank88: Why a no-op logger?

go.mod
```diff
@@ -157,3 +157,5 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20251029180050-ab9386a59fda // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/smartcontractkit/chainlink-common/pkg/chipingress => ./pkg/chipingress
```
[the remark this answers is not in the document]
jmank88:
```suggestion
```

pkg/services/service.go
```diff
@@ -167,6 +167,11 @@ type Config struct {
 	Start func(context.Context) error
 	// Close is an optional hook called before closing SubServices.
 	Close func() error
+	// CloseIfNeverStarted configures Close() to invoke the Close hook when the lifecycle
+	// state is still Unstarted, instead of returning ErrCannotStopUnstarted.
+	// Sub-services are not closed by this branch; callers should Start the parent normally
+	// if sub-service shutdown must run via the usual teardown path.
+	CloseIfNeverStarted bool
```
jmank88: I don't understand this field. What is preventing normal symmetric use of Start/Close? Why would it be necessary to execute Close if Start did not execute?

pkg/services/service.go
```diff
@@ -299,6 +309,40 @@ func (s *service) Close() error {
 	})
 }
 
+// tryCloseNeverStarted handles Close when the lifecycle never reached Started but
+// Config.Close hook must still run (used for tearing down ctor-held resources).
```
jmank88: If these resources can leak when Start is not called or fails, then the caller that owns them and calls Start is responsible for cleaning them up.

pkg/services/service.go
pkcll: It runs this method Client.close which closes internal client resources

pkcll: Will remove it

pkg/beholder/dual_source_emitter.go
```diff
@@ -51,7 +53,14 @@ func (d *DualSourceEmitter) Close() error {
 	}
 	close(d.stopCh)
 	d.wg.Wait()
-	return errors.Join(d.chipIngressEmitter.Close(), d.otelCollectorEmitter.Close())
+	// When the batch emitter is enabled its lifecycle (Start/Stop) is owned by the Client's
+	// service engine, not by DualSourceEmitter. Calling Close() on it here would return
+	// ErrCannotStopUnstarted when the client is closed without ever being started.
+	var chipErr error
+	if !d.chipIngressBatchEmitterEnabled {
+		chipErr = d.chipIngressEmitter.Close()
```
jmank88: This sounds like an implementation detail that shouldn't be tracked at this level. Can we find a way to separate ownership from implementation details?

go.mod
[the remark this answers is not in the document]
pkcll: done

pkg/services/service.go
pkcll: removed

pkg/beholder/httpclient.go
pkcll: fixed

pkg/beholder/dual_source_emitter.go
pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/1862/changes/8db77633a839a21f260ab061af965c6fc8e4764e

pkg/beholder/beholdertest/beholder.go
```diff
@@ -149,10 +151,15 @@ func NewObserver(t *testing.T) Observer {
 		MessageLoggerProvider: loggerProvider,
 		OnClose:               func() error { return nil },
 	}
+	client.Service, _ = services.Config{
+		Name: "BeholderClient",
+	}.NewServiceEngine(pkglogger.Nop())
+	require.NoError(t, client.Start(context.Background()))
```
jmank88:
```suggestion
	client.Service, client.eng = services.Config{
		Name: "BeholderClient",
	}.NewServiceEngine(pkglogger.Test(t))
	require.NoError(t, client.Start(t.Context()))
```

jmank88: Although why does this snippet exist at all? Can we call a shared constructor instead of duplicating logic like this?

pkg/beholder/noop.go
```diff
@@ -40,7 +42,24 @@ func NewNoopClient() *Client {
...
+		LoggerProvider:        loggerProvider,
+		TracerProvider:        tracerProvider,
+		MeterProvider:         meterProvider,
+		MessageLoggerProvider: loggerProvider,
+		OnClose:               noopOnClose,
+	}
+	c.Service, c.eng = services.Config{
+		Name:  "BeholderClient",
+		Close: c.close,
+	}.NewServiceEngine(logger.Nop())
```
jmank88: Ideally there should only be one instance of this code, in the lowest level constructor.

pkg/beholder/global_test.go
```diff
@@ -76,6 +76,10 @@ func TestClient_SetGlobalOtelProviders(t *testing.T) {
 	var b strings.Builder
 	client, err := beholder.NewWriterClient(&b)
 	require.NoError(t, err)
+	require.NoError(t, client.Start(t.Context()))
+	defer func() {
+		require.NoError(t, client.Close())
+	}()
```
jmank88, line 83:
```suggestion
	servicetest.Run(t, client)
```

pkg/chipingress/client.go
```diff
@@ -151,7 +155,10 @@ func (c *client) RegisterSchema(ctx context.Context, in *pb.RegisterSchemaReques
 }
 
 func (c *client) Close() error {
-	return c.conn.Close()
+	c.closeOnce.Do(func() {
+		c.closeErr = c.conn.Close()
+	})
+	return c.closeErr
```
jmank88: What is the motivation for supporting double close? That is usually a code smell for a root cause that should be fixed.

pkcll: This is just a precaution. Will remove it

pkg/beholder/global_test.go
pkcll: servicetest.Run it causes an import cycle: beholdertest → servicetest → utils/tests → beholdertest

pkg/beholder/beholdertest/beholder.go
pkcll: Reused noop client constructor here https://github.com/smartcontractkit/chainlink-common/pull/1862/changes/32ee48319d32d8685acf298012260f72e8cf29d6

pkg/beholder/noop.go
pkcll: This is just a stub for noop constructor. I reused noop constructor for NewObserver. Now there are only two places when this code is executed - beholder client constructor and noop client constructor

pkg/chipingress/client.go
pkcll: https://github.com/smartcontractkit/chainlink-common/pull/1862/changes/402a54538edf276e8445928a3513112228d01bb1

pkg/beholder/client.go
```diff
@@ -221,40 +233,90 @@ func NewGRPCClient(cfg Config, otlploggrpcNew otlploggrpcFactory) (*Client, erro
...
-	if c.Emitter != nil {
-		err = errors.Join(err, c.Emitter.Close())
-	}
-	if c.OnClose != nil {
-		err = errors.Join(err, c.OnClose())
-	}
-	return
+// close is the lifecycle Close hook: OTel/message shutdown and CHIP close.f
+func (c *Client) close() (err error) {
+	c.closeOnce.Do(func() {
```
jmank88: Why suppress errors on double close?

pkg/beholder/noop.go
jmank88: I see 4 versions of this code, with slight variations. Can we consolidate that down to just one? These are very low level utils that are not meant to be conditional and variable between instances.

pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/1862/commits/3b37b7e0394a04ba677a7f868a1bcc15e9e140a3

pkg/beholder/client.go
pkcll: https://github.com/smartcontractkit/chainlink-common/pull/1862/commits/2f33499231dad8eefca63c0569a05cdfa9c2e228

pkg/beholder/client.go
```diff
@@ -60,6 +65,27 @@ type Client struct {
...
+	// batchEmitterService is owned by the client and started/stopped via Client lifecycle.
+	batchEmitterService *ChipIngressBatchEmitterService
+}
+
+// initService wires up the services.Service lifecycle for the Client.
+// Must be called exactly once after populating the Client fields.
+func (c *Client) initService(lggr pkglogger.Logger) {
+	if lggr == nil {
+		lggr = pkglogger.Nop()
+	}
```
jmank88: Why not pass a logger in all cases?
```suggestion
// initService wires up the services.Service lifecycle for the Client.
// Must be called exactly once after populating the Client fields.
func (c *Client) initService(lggr pkglogger.Logger) {
```

pkg/beholder/dual_source_emitter.go
```diff
@@ -60,22 +62,27 @@ func (d *DualSourceEmitter) Emit(ctx context.Context, body []byte, attrKVs ...an
...
-	if err := d.wg.TryAdd(1); err != nil {
-		return err
-	}
-	go func(ctx context.Context) {
-		defer d.wg.Done()
-		var cancel context.CancelFunc
-		ctx, cancel = d.stopCh.Ctx(ctx)
-		defer cancel()
-
+	if d.chipIngressBatchEmitterEnabled {
```
jmank88: Did you consider containing this logic inside of the `chipIngressBatchEmitter` so that we could just call `Emit` from this point rather than manage two modes?

pkg/beholder/client.go
pkcll: Addressed https://github.com/smartcontractkit/chainlink-common/pull/1862/commits/30590cb075b23378399212f4798d78514820b4be

pkg/beholder/dual_source_emitter.go
pkcll: Addressed here https://github.com/smartcontractkit/chainlink-common/pull/1862/commits/a04e4ead6eb0718b9651dd06c63bcffa61c6bd4c

jmank88 (dismissed):

jmank88 (approved):

pkcll:
Split pr into two parts 
- https://github.com/smartcontractkit/chainlink-common/pull/2059
- https://github.com/smartcontractkit/chainlink-common/pull/2059

PR #1894, opened 2026-03-12
feat: add ExpectedSimulationFailureRule to ContractWriter.SubmitTransaction

Fletch153:
## Summary

- Adds `ExpectedSimulationFailureRule` struct (`ErrorContains string`) to `pkg/types/` and a new `[]ExpectedSimulationFailureRule` parameter to `ContractWriter.SubmitTransaction`
- Adds corresponding proto messages to `contract_writer.proto` (field 8) and `aptos.proto` (field 4) with regenerated Go code
- Wires gRPC client/server via `FailureRulesToProto`/`FailureRulesFromProto` converter functions, following the existing `TxMeta` pattern
- Adds `UnimplementedContractWriter` forward-compat base type
- Updates all in-repo call sites and test fakes

## Breaking Change

`ContractWriter.SubmitTransaction` signature now requires an additional `[]ExpectedSimulationFailureRule` parameter. All implementations must update their signatures. Pass `nil` to preserve existing behavior.

## Context

chainlink-aptos (`NONEVM-2618`) added `ExpectedSimulationFailureRule` to its TXM `Enqueue`, but `SubmitTransaction` (the product-agnostic interface in chainlink-common) didn't carry these rules. This change enables failure rules to flow from callers through the LOOP gRPC boundary to chain-specific implementations.

## Test plan

- [x] Converter round-trip tests (nil, empty, populated) for both directions
- [x] `go build ./...` passes
- [x] `go test ./...` passes (pre-existing `otelzap` failure unrelated)
- [ ] CI pipeline (lint, tests, gorelease API diff)


pkg/loop/internal/relayer/pluginprovider/contractwriter/contract_writer.go
```diff
@@ -45,20 +45,21 @@ func WithClientEncoding(version codecpb.EncodingVersion) ClientOpt {
 	}
 }
 
-func (c *Client) SubmitTransaction(ctx context.Context, contractName, method string, params any, transactionID, toAddress string, meta *types.TxMeta, value *big.Int) error {
+func (c *Client) SubmitTransaction(ctx context.Context, contractName, method string, params any, transactionID, toAddress string, meta *types.TxMeta, value *big.Int, expectedSimulationFailureRules []types.ExpectedSimulationFailureRule) error {
```
jmank88: You could avoid future breaking changes by switching to a single struct arg like grpc uses:
```suggestion
func (c *Client) SubmitTransaction(ctx context.Context, req SubmitTransactionRequest) error {
```

PR #1896, opened 2026-03-13
Add rebased Aptos CRE common and loop support

cawthorne:
## Summary
- rebase the Aptos CRE common changes on top of the latest default branch
- preserve the working Aptos read-capability baseline from the earlier local CRE stack
- retain only the shared compatibility needed by the write path

## What changed
- added the Aptos service and loop typing needed by the CRE integration path
- retained `LedgerVersion` support used by the Aptos read flow
- kept the shared Aptos conversion helpers and tests
- kept the transmitter-bytes compatibility used by the write path
- refreshed the generated Aptos protobuf / gRPC outputs to match the rebased interfaces

## Why
The Aptos local CRE stack needs the common Aptos loop/service surface from the earlier read-capability work, plus the shared compatibility pieces consumed by the write path. This PR keeps that layer narrow so downstream repos do not need to reach around `chainlink-common`.

## Testing
- ran:
  - `go test ./pkg/types/chains/aptos ./pkg/loop/internal/relayer ./pkg/loop/internal/relayerset ./pkg/loop/internal/core/services/capability -count=1`

## CI refs
- core ref: e2a5c22eff7c7cf8e913453387f8fefe89fff40f
- solana ref: develop
- starknet ref: develop


pkg/chains/aptos/aptos.proto
```diff
@@ -142,10 +152,10 @@ message SubmitTransactionReply {
   TxStatus tx_status = 1;
   string tx_hash = 2;
   string tx_idempotency_key = 3;
+  optional uint64 transaction_fee = 4;
```
yashnevatia: dont need this

PR #1911, opened 2026-03-19
ChIP Durable Emitter

DylanTinianov:
Creates ChIP `DurableEmitter` to persist and re-send Chip messages.
[DurableEmitter Integration Testing + ORM](https://github.com/smartcontractkit/chainlink/pull/21609)


pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+			"batch_size", len(batch), "error", err,
+			"elapsed_ms", elapsed.Milliseconds())
+		return
+	}
+
+	if d.metrics != nil {
+		d.metrics.publishBatchEvOK.Add(context.Background(), int64(len(batch)))
+	}
+
+	// Async MarkDelivered: the DB UPDATE runs in a background goroutine so
```
tarcisiozf, line 708: due to its async nature, is there a risk that a call to `ListPending` creates rework of tasks because this is in-flight?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+
+	return nil
+}
+
+// Close signals background loops to stop and waits for them to finish.
+// stopCh is closed first so workers exit before publishCh is closed after wg.Wait,
+// avoiding send vs close races on publishCh.
+func (d *DurableEmitter) Close() error {
+	close(d.stopCh)
+	if d.insertCh != nil {
```
tarcisiozf, line 451: should we drain insertCh as a graceful shutdown?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+	OnEmitInsert func(elapsed time.Duration, err error)
+	// OnBatchPublish is called after each PublishBatch RPC in the batch publish loop.
+	// batchSize is the number of events in the batch; err is nil on success.
+	OnBatchPublish func(elapsed time.Duration, batchSize int, err error)
+	// OnBatchMarkDelivered is called after MarkDeliveredBatch following a successful batch publish.
+	OnBatchMarkDelivered func(elapsed time.Duration, count int)
+	// OnRetransmitBatchPublish is not invoked (retransmit uses the same batch loop as Emit; use OnBatchPublish).
+	OnRetransmitBatchPublish func(elapsed time.Duration, eventCount int, err error)
+	// OnRetransmitBatchDeletes is not invoked (mark delivered is async per batch; hook retained for API compatibility).
+	OnRetransmitBatchDeletes func(elapsed time.Duration, markedDeliveredCount int)
```
tarcisiozf: what do you mean by "not invoked"? can we just remove then instead?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+}
+
+func DefaultDurableEmitterConfig() DurableEmitterConfig {
+	return DurableEmitterConfig{
+		RetransmitInterval:  5 * time.Second,
+		RetransmitAfter:     10 * time.Second,
+		RetransmitBatchSize: 100,
+		ExpiryInterval:      1 * time.Minute,
+		EventTTL:            24 * time.Hour,
```
patrickhuie19: please 72 hours for this default in case there is an issue with a NOP getting a node back online

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+	store  DurableEventStore
+	client chipingress.Client
+	cfg    DurableEmitterConfig
+	log    logger.Logger
+
+	metrics *durableEmitterMetrics
+
+	// rawConn is the underlying gRPC connection when the client exposes it.
+	// Non-nil enables zero-copy batch publishing (protowire + ForceCodec).
+	rawConn *grpc.ClientConn
```
patrickhuie19, line 145: Can we still fit into the interface if we consider implementing the emitter with a pool of conns?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-d.stopCh:
+			return
+		case <-ticker.C:
+			if d.metrics == nil {
+				return
+			}
+			bctx := context.Background()
```
jmank88: Why a background context?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+		insertWorkers = 4
+	}
+	if d.insertCh != nil {
+		n += insertWorkers
+	}
+	if d.metrics != nil && d.cfg.Metrics != nil {
+		n++
+	}
+	d.wg.Add(n)
+	go d.retransmitLoop(ctx)
```
jmank88: The context passed to Start is only for the duration of startup. Background routines need to derive context from the `StopCh`.

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,943 @@
...
+	// MarkDelivered, Delete, or DeleteExpired. No polling required.
+	pendingCount atomic.Int64
+	pendingMax   atomic.Int64
+
+	// publishCh buffers events for the batch publish loop.
+	publishCh chan publishWork
+
+	stopCh chan struct{}
+	wg     sync.WaitGroup
+	markWg sync.WaitGroup // tracks in-flight async MarkDelivered goroutines
```
jmank88: Why a second WaitGroup? They both wait in Close - what happens if everything shares one?

jmank88: Or one `services.WaitGroup` to shed some boilerplate

pkg/beholder/durable_emitter.go
tarcisiozf: maybe we should do a first version without raw grpc calls and leave as a future optimization option

pkg/beholder/durable_emitter.go
DylanTinianov: Yeah async mark implies ListPending can still see those rows and re-enqueue them until mark lands, which is a tradeoff. Not a big deal since we guarantee at least once delivery.

We can fix this in a follow-up PR

pkg/beholder/durable_emitter.go
DylanTinianov: That should be doable, probably not necessary for now though

pkg/beholder/durable_emitter.go
DylanTinianov: Good call, updated to ensure inserts finish before shutdown and added a test to verify.

pkg/beholder/durable_event_store.go
```diff
@@ -0,0 +1,300 @@
...
+	return &MemDurableEventStore{
+		events: make(map[int64]*DurableEvent),
+	}
+}
+
+func (m *MemDurableEventStore) Insert(_ context.Context, payload []byte) (int64, error) {
+	id := m.nextID.Add(1)
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.events[id] = &DurableEvent{
```
jmank88: So obviously `nextID` is atomic on its own, but there is still a logical race in that `m.events[id+1]` could be set in the map before `m.events[id]` for example. Is that problematic? Or is it fine to just be eventually consistent?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,962 @@
...
+
+	ticker := time.NewTicker(poll)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-d.stopCh:
+			return
+		case <-ticker.C:
+			if d.metrics == nil {
+				return
```
jmank88: Can we just check this at the start of the method?

pkg/beholder/durable_emitter.go
```diff
@@ -0,0 +1,962 @@
...
+		insertWorkers = 4
+	}
+	if d.insertCh != nil {
+		n += insertWorkers
+	}
+	if d.metrics != nil && d.cfg.Metrics != nil {
+		n++
+	}
+	d.wg.Add(n)
+	go d.retransmitLoop()
```
jmank88: You can use the new `Go` helper method to avoid the boilerplate `Add` and `Done` calls:
```suggestion
	wg.Go(d.retransmitLoop)
```

pkg/beholder/durable_event_store.go
DylanTinianov: Yeah this is fine. This event store is also only used for tests. Moving it to durable_emitter_test.go.

pkg/beholder/durable_emitter.go
DylanTinianov: Can remove this actually since it's only called after already checking in Start()

pkg/beholder/durable_emitter.go
DylanTinianov: Awesome, love that!

jmank88 (dismissed):

vyzaldysanchez (dismissed):

pkg/chipingress/client.go
```diff
@@ -274,6 +282,8 @@ func NewEvent(domain, entity string, payload []byte, attributes map[string]any)
 		attributes = make(map[string]any)
 	}
 
+	const beholderDataSchemaKey = "beholder_data_schema"
```
hendoxc: You don't need this for `chip-ingress`

hendoxc (approved):

jmank88 (approved):

PR #1923, opened 2026-03-24
Aptos: add shared capability payload conversion helpers

cawthorne:
## Summary
- add shared Aptos capability-to-domain payload conversion helpers in `pkg/chains/aptos`
- move the Aptos capability payload/type-tag conversion tests into `chainlink-common`

## Why
- keeps Aptos proto/domain conversion logic in the same place as the other chain proto helpers
- unblocks the follow-up cleanup in capabilities without duplicating conversion code


pkg/chains/aptos/capability_proto_helpers.go
```diff
@@ -0,0 +1,128 @@
...
+	typeaptos "github.com/smartcontractkit/chainlink-common/pkg/types/chains/aptos"
+)
+
+// ConvertCapabilityViewPayloadFromProto converts Aptos capability proto input into the
+// shared Aptos domain payload used by relayers and loop clients.
+func ConvertCapabilityViewPayloadFromProto(payload *capaptos.ViewPayload) (*typeaptos.ViewPayload, error) {
+	if payload == nil {
+		return nil, fmt.Errorf("viewRequest.Payload is required")
+	}
+	if payload.Module == nil {
```
ilija42: should module name be validated as well?

PR #1969, opened 2026-04-07
Trigger Delivery: Refactor BaseTrigger

DylanTinianov:
Refactors BaseTrigger to handle registrations better. Implements the following:
- Max retries to along with DB pruning loop for cleanup.
- Adds pre-ACK cache for slow nodes which receive an ACK before even seeing the trigger event
- Fixes race condition on ACK during event Delivery
- Backoffs + jitter to optimize traffic distribution
- Additional test coverage.


pkg/capabilities/base_trigger.go
```diff
@@ -479,11 +544,96 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+
+		b.mu.Lock()
+		inMemory := false
+		if evts, ok := b.pending[rec.TriggerId]; ok {
+			_, inMemory = evts[rec.EventId]
+		}
+		b.mu.Unlock()
+
+		if inMemory {
+			// Still actively tracked — scanPending will handle it (gave-up or ACK).
```
fernandezlautaro: this should technically never happen, right?

If the prune time is set to 24hs, and max attempts to 20, with a retrial of 30 seconds, that's 10m of retrials, so eventually this should have been deleted from in-mem.

I believe you should error here, or even throw a metric to show there's an inconsistence

pkg/capabilities/base_trigger.go
```diff
@@ -479,11 +544,96 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+	cutoff := time.Now().Add(-age)
+
+	recs, err := b.store.List(b.ctx)
+	if err != nil {
+		b.lggr.Errorw("prune: failed to list events from store", "capabilityID", b.capabilityId, "err", err)
+		return
+	}
+
+	for _, rec := range recs {
+		if rec.FirstAt.After(cutoff) {
```
fernandezlautaro: why `FirstAt` field and not `LastSeenAt`?
Shouldn't you remove based on the last time you modified the row in the DB instead of the first time you inserted it?

pkg/capabilities/base_trigger.go
```diff
@@ -435,13 +475,39 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+				}
+				delete(pendingForTrigger, eventID)
+				if len(pendingForTrigger) == 0 {
+					delete(b.pending, triggerID)
+				}
+				toGiveUp = append(toGiveUp, gaveUpEvent{
+					triggerID:   triggerID,
+					eventID:     eventID,
+					attempts:    rec.Attempts,
+					wasCritical: wasCritical,
```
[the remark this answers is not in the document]
DylanTinianov: Yeah let's save this for another PR

pkg/capabilities/base_trigger.go
```diff
@@ -435,13 +475,39 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+	maxAttempts := b.maxRetries(ctx)
 	warnThreshold := 1 * interval
 	critThreshold := 3 * interval
 
 	b.mu.Lock()
 	toResend := make([]PendingEvent, 0, len(b.pending))
+	var toGiveUp []gaveUpEvent
 	for triggerID, pendingForTrigger := range b.pending {
 		for eventID, rec := range pendingForTrigger {
+			if maxAttempts > 0 && rec.Attempts >= maxAttempts {
```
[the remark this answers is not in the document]
DylanTinianov: Makes sense, addressed

pkg/capabilities/base_trigger.go
```diff
@@ -479,11 +544,96 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+}
+
+func (b *BaseTriggerCapability[T]) pruneStaleEvents() {
+	age := b.pruneAge(b.ctx)
+	if age <= 0 {
+		return
+	}
+	cutoff := time.Now().Add(-age)
+
+	recs, err := b.store.List(b.ctx)
```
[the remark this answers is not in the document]
DylanTinianov: This can be another clean up PR since this changes the interface and will need to be propagated through

pkg/capabilities/base_trigger.go
```diff
@@ -148,28 +183,48 @@ func (b *BaseTriggerCapability[T]) retryInterval(ctx context.Context) time.Durat
...
+	if b.settings == nil {
+		return defaultMaxSendsPerTick
+	}
+	v, err := cresettings.Default.BaseTriggerMaxSendsPerTick.GetOrDefault(ctx, b.settings)
+	if err != nil {
+		b.lggr.Warnw("CRE settings read failed for BaseTriggerMaxSendsPerTick; using default", "err", err)
+		return defaultMaxSendsPerTick
+	}
+	if v <= 0 {
+		return defaultMaxSendsPerTick
```
fernandezlautaro, line 209: I'd print an error here

pkg/capabilities/base_trigger.go
```diff
@@ -148,28 +183,48 @@ func (b *BaseTriggerCapability[T]) retryInterval(ctx context.Context) time.Durat
...
+	}
+	return v
+}
+
+// pruneAge returns how long a pending row must exist before the pruning loop may remove it.
+func (b *BaseTriggerCapability[T]) pruneAge(ctx context.Context) time.Duration {
+	if b.settings == nil {
+		return 0
```
fernandezlautaro: could you have a defaultPruneAge as you did for the defaultMaxSendsPerTick and return that, but not 0, the current value you are using as 24hs in the pruneLoop() method, so it reduces the complexity on pruneLoop(), and leaves the data handling separated from the logic of punning out

fernandezlautaro: further review shows me that if the value from config is 0 in pruneStaleEvents() you basically are not prunning at all, could you comment that?

So if it's not defined you check every 24hs it it has been defined by configuration, and until it is you do nothing.

pkg/capabilities/base_trigger.go
```diff
@@ -261,9 +321,39 @@ func (b *BaseTriggerCapability[T]) DeliverEvent(
...
+	if pa, ok := b.preAcked[triggerID]; ok {
+		if _, wasAcked := pa[te.ID]; wasAcked {
+			delete(pa, te.ID)
+			if len(pa) == 0 {
+				delete(b.preAcked, triggerID)
+			}
+			b.mu.Unlock()
+			b.lggr.Infow("base trigger DeliverEvent skipped: event was already ACKed (pre-ACK)",
+				"capabilityID", b.capabilityId, "triggerID", triggerID, "eventID", te.ID)
+			b.metrics.IncAckMemoryOutcome("pre_ack_delivery_skipped")
```
fernandezlautaro: could you put all of this literal values `pre_ack_delivery_skipped` (or any similar for this metric) in a constant group, it will be easier later on to gather metrics looking in a single place vs scattered through the code

pkg/capabilities/base_trigger.go
```diff
@@ -357,6 +481,18 @@ func (b *BaseTriggerCapability[T]) AckEvent(ctx context.Context, triggerId strin
...
 
+	// Always record the ACK so that a later DeliverEvent for this event is
+	// skipped. This covers two scenarios:
+	//   1. (pre-ACK) Other DON nodes delivered the event first, so the ACK
+	//      arrives at this node before DeliverEvent is called.
+	//   2. (re-delivery) The upstream trigger re-delivers the same event
+	//      (e.g. EVM trigger after block finalization prunes its sent-set),
+	//      even though the event was already ACKed during a prior delivery.
+	if b.preAcked[triggerId] == nil {
+		b.preAcked[triggerId] = make(map[string]time.Time)
```
fernandezlautaro: why not doing this on the RegisterTrigger() method?, if you do it early you can avoid safe guards in the code checking wether preAcked[triggerID] is present, as it will be there with an empty array of eventIDs

pkg/capabilities/base_trigger.go
```diff
@@ -273,22 +363,56 @@ func (b *BaseTriggerCapability[T]) DeliverEvent(
...
 	}
 	b.lggr.Infow("base trigger persisted pending event for ACK tracking",
 		"capabilityID", b.capabilityId, "triggerID", triggerID, "eventID", te.ID)
 
+	// Double-check preAcked under the same lock as adding to pending.
+	// An ACK may have arrived during the store.Insert call above. Without
+	// this second check, the event would be retransmitted forever because
+	// the first preAcked check (before Insert) narrowly missed the ACK.
 	b.mu.Lock()
+	if pa, ok := b.preAcked[triggerID]; ok {
```
fernandezlautaro: I believe you could ditch the `ok` boolean here if you initialize the `preAcked` map upon registration per trigger, will reduce code complexness

pkg/capabilities/base_trigger.go
```diff
@@ -273,22 +363,56 @@ func (b *BaseTriggerCapability[T]) DeliverEvent(
...
+		if _, wasAcked := pa[te.ID]; wasAcked {
+			delete(pa, te.ID)
+			if len(pa) == 0 {
+				delete(b.preAcked, triggerID)
```
fernandezlautaro: what's the point of this deletion, you are constantly going to delete this entry and re-add it on every ack.
Being the triggerID a long lived state, I believe it makes sense to have them all properly initialized. It does make sense to remove the eventId from it tho.

pkg/capabilities/base_trigger.go
```diff
@@ -75,6 +101,14 @@ type BaseTriggerCapability[T proto.Message] struct {
 	inboxes map[string]chan<- TriggerAndId[T]   // triggerID --> registered send channel
 	pending map[string]map[string]*PendingEvent // triggerID --> eventID --> PendingEvent
 
+	// preAcked remembers ACKs that arrived before DeliverEvent was called on this node.
+	// In a multi-node capability DON, other nodes may deliver the event first, causing
+	// the workflow engine to ACK before the slower node has even persisted the event.
+	// Without this cache, the late node would persist and retransmit forever since no
+	// new ACK will arrive (other nodes already stopped retransmitting, so the subscriber
+	// can't reach aggregation quorum).
+	preAcked map[string]map[string]time.Time // triggerID -> eventID -> ackedAt
```
fernandezlautaro: you currently have inboxes, pending and preAcked per triggerID maps, wouldn't be better to have a single structure holding that, so you can actually pull from it given a triggerID, and then ask for the channel to send (inbox array), if it has pending events to send (pending array) or if check if a given event was already acked (preAcked).

pkg/capabilities/base_trigger.go
```diff
@@ -426,6 +562,56 @@ func (b *BaseTriggerCapability[T]) retransmitLoop() {
...
 
+type stoppedResendingEvent struct {
+	triggerID   string
+	eventID     string
+	attempts    int
+	wasCritical bool
+}
+
+// reachedMaxRetries returns true when the event has exhausted its allowed send attempts.
+func reachedMaxRetries(attempts, maxRetries int) bool {
```
fernandezlautaro, line 547: why did you add the new code as function and not methods? is it because they seemed to be a helper rather than part of the base trigger code?

pkg/capabilities/base_trigger.go
```diff
@@ -273,22 +363,56 @@ func (b *BaseTriggerCapability[T]) DeliverEvent(
 	}
 
 	if err := b.store.Insert(ctx, rec); err != nil {
+		if isDuplicateKeyError(err) {
```
fernandezlautaro, line 355: is this necessary or did you see this happening during your tests?

pkg/capabilities/base_trigger.go
```diff
@@ -435,52 +621,246 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+		b.undeliveredAlertStates[triggerID][eventID] = state
+	}
+
+	if warnThreshold > 0 && !state.emittedWarning && age >= warnThreshold {
+		b.metrics.EmitUndeliveredWarning(triggerID, eventID)
+		state.emittedWarning = true
+	}
+
+	if critThreshold > 0 && !state.emittedCritical && age >= critThreshold {
+		b.metrics.EmitUndeliveredCritical(triggerID, eventID)
```
fernandezlautaro: now that you are emitting also on `capabilities_base_trigger_stopped_resending_total`, is there a point in `EmitUndeliveredCritical` or even worse `EmitUndeliveredWarning`?.

pkg/capabilities/base_trigger.go
```diff
@@ -435,52 +621,246 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+		b.metrics.IncStuckEvent(triggerID, eventID)
```
fernandezlautaro: and the same goes for IncStuckEvent or DescStuckEvent, if we already have the stopped_resending meter, what's the point on this gauge going up and down?

I believe it makes the code much more complex that it needs to be, and I believe it doesn't do much for us on the o11y side (specially since this shouldn't happen at all, ideally)

pkg/capabilities/base_trigger.go
DylanTinianov: Included `defaultPruneAge` and fallback to that instead

pkg/capabilities/base_trigger.go
DylanTinianov: Yeah I saw this happening in tests sometimes which is why I added it. Although it might not be necessary anymore. I don't think it hurts to keep this.

pkg/capabilities/base_trigger.go
DylanTinianov: Good point, removed warning and crits

pkg/capabilities/base_trigger.go
DylanTinianov: Removed :)

pkg/capabilities/base_trigger.go
DylanTinianov: Yeah it's just a simple helper, shouldn't make a difference.

pkg/capabilities/base_trigger.go
```diff
@@ -13,6 +16,40 @@ import (
...
+	ackMemoryOutcomeMissNilRecord         = "miss_nil_record"
+	ackMemoryOutcomePreAckDeliverySkipped = "pre_ack_delivery_skipped"
+)
+
+// triggerReg holds all per-triggerID runtime state for a [BaseTriggerCapability].
+// triggerID remains the map key in [BaseTriggerCapability.byTrigger]; this type
+// groups inbox, pending rows, and the pre-ACK dedupe cache for that trigger.
+type triggerReg[T proto.Message] struct {
+	inbox    chan<- TriggerAndId[T]
+	pending  map[string]*PendingEvent
```
fernandezlautaro: previous comment of what the key is `eventID --> PendingEvent` was nice, might be worth adding it here and for the preAcked as well so anyone knows it's a map of eventsIDs -> something

pkg/capabilities/base_trigger.go
```diff
@@ -344,53 +457,47 @@ func (b *BaseTriggerCapability[T]) AckEvent(ctx context.Context, triggerId strin
 			found = true
 		case recOk && rec == nil:
```
fernandezlautaro: `rec == nil` will always be `true`, pls remove

pkg/capabilities/base_trigger.go
```diff
@@ -332,9 +439,15 @@ func (b *BaseTriggerCapability[T]) AckEvent(ctx context.Context, triggerId strin
...
 	b.mu.Lock()
-	eventsForTrigger, ok := b.pending[triggerId]
-	hadTriggerBucket = ok && eventsForTrigger != nil
-	if hadTriggerBucket {
+	reg := b.byTrigger[triggerId]
+	eventWasInPending := false
+	if reg != nil {
+		_, eventWasInPending = reg.pending[eventId]
+	}
+	hadTriggerBucket = reg != nil && (len(reg.pending) > 0 || eventWasInPending)
```
fernandezlautaro: shouldn't `hadTriggerBucket` be `eventWasInPending`?
If the trigger is inside `b.byTrigger` map, and then there's an entry for `reg.pending[]` for the event beign acked, isnt't that enough to determine if the `hadTriggerBucket` is actually true/false?

pkg/capabilities/base_trigger.go
```diff
@@ -332,9 +439,15 @@ func (b *BaseTriggerCapability[T]) AckEvent(ctx context.Context, triggerId strin
 	)
 
 	b.mu.Lock()
-	eventsForTrigger, ok := b.pending[triggerId]
-	hadTriggerBucket = ok && eventsForTrigger != nil
-	if hadTriggerBucket {
+	reg := b.byTrigger[triggerId]
+	eventWasInPending := false
+	if reg != nil {
```
fernandezlautaro, line 447: you have this `if reg != nil {` condition twice almost 2-3 lines apart, can you collapse this logic into the single code block for easy readability later on?

pkg/capabilities/base_trigger.go
fernandezlautaro: you have this `if found {` in line #488, why are you repeating this here? shouldnt just be moved in the same code blocks above to do a proper printing?

pkg/capabilities/base_trigger.go
```diff
@@ -344,53 +457,47 @@ func (b *BaseTriggerCapability[T]) AckEvent(ctx context.Context, triggerId strin
...
-		b.lggr.Infow("base trigger ACK: no in-memory pending bucket for trigger (not pending here; still deleting from store if row exists)",
-			"capabilityID", b.capabilityId, "triggerID", triggerId, "eventID", eventId)
+	} else {
+		b.lggr.Infow("base trigger ACK: event not found in memory (reconciling store)",
+			"capabilityID", b.capabilityId, "triggerID", triggerId, "eventID", eventId,
+			"hadTriggerBucket", hadTriggerBucket, "hadEventKey", hadEventKey,
+			"hadNilPendingRecord", hadNilPendingRecord)
 	}
 
 	if err := b.store.DeleteEvent(ctx, triggerId, eventId); err != nil {
```
fernandezlautaro, line 511: shouldn't this deletion be done only if `found==true`?

pkg/capabilities/base_trigger.go
```diff
@@ -344,53 +457,47 @@ func (b *BaseTriggerCapability[T]) AckEvent(ctx context.Context, triggerId strin
...
-			delete(b.undeliveredAlertStates, triggerId)
-		}
+	// Always record the ACK so that a later DeliverEvent for this event is
+	// skipped. This covers two scenarios:
+	//   1. (pre-ACK) Other DON nodes delivered the event first, so the ACK
+	//      arrives at this node before DeliverEvent is called.
+	//   2. (re-delivery) The upstream trigger re-delivers the same event
+	//      (e.g. EVM trigger after block finalization prunes its sent-set),
+	//      even though the event was already ACKed during a prior delivery.
+	if reg == nil {
```
fernandezlautaro, line 489: I'm failing to understand how this is possible.
A given node will get an ack for an eventID of a triggerID that has not been registered: so that means a single node is so slow others could potentially make quorum, and make the subscriber ack on 2F+1 nodes and send an ACK to all nodes, and that node receive it without registering the trigger?

fernandezlautaro: just in case, I get registering the ack, not recreating the whole structure using ` b.getOrCreateRegLocked(triggerId)`

pkg/capabilities/base_trigger.go
```diff
@@ -211,38 +268,28 @@ func (b *BaseTriggerCapability[T]) Stop() {
 
 func (b *BaseTriggerCapability[T]) RegisterTrigger(triggerID string, sendCh chan<- TriggerAndId[T]) {
 	b.mu.Lock()
-	_, existed := b.inboxes[triggerID]
-	b.inboxes[triggerID] = sendCh
+	reg := b.getOrCreateRegLocked(triggerID)
```
fernandezlautaro, line 275: if the AckEvent() method could run before this one, this initialization could eventually erase a pre-acked event, makign this node sent the event until failure (I believe this code is ok, and the problem is in the AckEvent(), but mentioning here just in case you are right in leaving the other usage of `getOrCreateRegLocked` in the AckEvent method)

pkg/capabilities/base_trigger.go
DylanTinianov: This happens when an ACK arrives on a slow node (before RegisterTrigger has even been called yet) exactly.

pkg/capabilities/base_trigger.go
DylanTinianov: The delete must run even when found == false. After a node restart, the store survives but the in-memory map starts empty and is rebuilt lazily from the store. An ACK could arrive after the restart before the store rows are loaded back into memory, leaving found = false despite a live store row. The store implementation is required to treat a missing row as a no-op.

fernandezlautaro: ty!

fernandezlautaro (dismissed):

pkg/capabilities/base_trigger.go
```diff
@@ -426,6 +536,56 @@ func (b *BaseTriggerCapability[T]) retransmitLoop() {
 	}
 }
 
+type stoppedResendingEvent struct {
+	triggerID   string
+	eventID     string
+	attempts    int
+	wasCritical bool
```
justinkaseman: What is this used for? Currently is dead code

pkg/capabilities/base_trigger.go
```diff
@@ -148,28 +179,52 @@ func (b *BaseTriggerCapability[T]) retryInterval(ctx context.Context) time.Durat
...
+	if b.settings == nil {
+		return defaultPruneAge
+	}
+	v, err := cresettings.Default.BaseTriggerPruneAge.GetOrDefault(ctx, b.settings)
+	if err != nil {
+		b.lggr.Warnw("CRE settings read failed for BaseTriggerPruneAge; falling back to default", "err", err)
+		return defaultPruneAge
+	}
+	if v <= 0 {
+		b.lggr.Errorw("found BaseTriggerPruneAge <=0 in settings; falling back to default", "defaultPruneAge", defaultPruneAge)
```
justinkaseman, line 226: Small bug here that probably wouldn't happen in the wild, but better to be defensive:
1. BaseTriggerPruneAge = "0s"
2. GetOrDefault parses it successfully and returns 0 with no error. It returns the default value only on error or empty string.
3. The v <= 0 guard catches that, logs "falling back to default", but then returns v (which is 0) instead of defaultPruneAge

pkg/capabilities/base_trigger.go
```diff
@@ -435,52 +595,198 @@ func (b *BaseTriggerCapability[T]) scanPending() {
...
+			b.pruneStaleEvents()
 		}
 	}
-	b.mu.Unlock()
+}
 
-	for _, event := range toResend {
-		b.trySend(event)
+func (b *BaseTriggerCapability[T]) pruneStaleEvents() {
+	//
```
justinkaseman: Cleanup? Looks like used to be a comment

pkg/capabilities/base_trigger_metrics.go
justinkaseman: EmitUndeliveredWarning, EmitUndeliveredCritical, IncStuckEvent, DecStuckEvent are dead. Clean them up

justinkaseman (changes requested):

justinkaseman (approved):

fernandezlautaro (approved):

PR #1976, opened 2026-04-10
fix(services): HealthChecker.Register immediately populates cache when already running

Fletch153:
## Summary

- **Root cause:** When `Register()` is called on an already-started `HealthChecker` (e.g. by the job spawner after starting a new job service), the new service's health state is absent from `IsHealthy()` / `IsReady()` for up to 15 seconds — the full polling interval — even though `Ready()` already returns nil.
- **Fix:** `Register()` now calls `IfStarted` after adding the service, immediately populating `c.ready` and `c.healthy` for that service.
- The `servicesMu` write lock is scoped to an inner closure and released *before* calling `IfStarted` to avoid a deadlock with `Start()`, which holds the `StateMachine` lock while calling `update()` (which acquires `servicesMu.RLock`).

## Changes

- `pkg/services/health.go` — `Register()` triggers an immediate single-service health check when the checker is already running
- `pkg/services/health_test.go` — two new tests covering healthy and unhealthy services registered on a running checker

## Test plan

- [ ] `go test ./pkg/services/... -run TestHealthChecker_Register` — both new tests pass
- [ ] `go test ./pkg/services/...` — no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/services/health.go
+++ b/pkg/services/health.go
@@ -47,8 +47,8 @@ func CopyHealth(dest, src map[string]error) {
 // HealthChecker is a services.Service which monitors other services and can be probed for system health.
 type HealthChecker struct {
 	StateMachine
-	chStop    chan struct{}
-	chDone    chan struct{}
+	chStop chan struct{}
+	chDone chan struct{}
 
 	cfg HealthCheckerConfig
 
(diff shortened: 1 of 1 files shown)


mchain0 (dismissed):

.github/workflows/benchmark.yml
jmank88, line 1: How is this related?

jmank88 (changes requested):

pkg/services/health.go
```diff
@@ -226,6 +230,14 @@ func (c *HealthChecker) Register(service HealthReporter) error {
 		}
 	}
 	c.services[name] = service
+
+	// Signal the run() goroutine to refresh immediately so the newly
+	// registered service is visible without waiting for the next tick.
+	select {
+	case c.chRefresh <- struct{}{}:
+	default: // refresh already pending
+	}
```
jmank88: How about inserting the new service response to c.ready and c.healthy directly?

jmank88 (dismissed):

mchain0 (approved):

jmank88 (approved):

PR #1981, opened 2026-04-14
Fix workflow syntax error

chainchad:


.github/workflows/dependabump.yml
```diff
@@ -25,12 +25,15 @@ jobs:
...
+          if git diff --cached --quiet; then
+            echo "changed=false" | tee -a "$GITHUB_OUTPUT"
+          else
+            echo "changed=true" | tee -a "$GITHUB_OUTPUT"
+          fi
         continue-on-error: true
 
       - name: Notify Fatal Error
-        if: !contains([0,1],  steps.changes.outputs.exit_code != '0' ) # Fatal exit code
+        if: steps.changes.outcome == 'failure'
```
chainchad, line 36: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context

jmank88: Oh so I definitely biffed that old logic, but the point here was to catch failures other than `1`. So I meant:

```suggestion
        if: !contains([0,1],  steps.changes.outputs.exit_code ) # Fatal exit code
```
w/o that leftover zero comparison. Does that make sense?

chainchad: Ah ok, yeah so 0 or 1 are expected and > 1 is a critical failure.

chainchad:
Will close because it doesn't seem like it's needed but feel free to re-open or merge if desired.

PR #1983, opened 2026-04-16
Parse raw bytes instead of hex over capReg grpc wire for transmitter address

yashnevatia:
- Aptos capability loop plugin calls ConfigForCapability on capability registry server
- When converting the ocr3 config for the capability to proto, the server cannot [hexDecode](https://github.com/smartcontractkit/chainlink-common/blob/a931c3f2f1624245c75d465889588c399ecefcc3/pkg/loop/internal/core/services/capability/capabilities_registry.go#L515) the transmitter address
- Hence we get `"error":"failed to get capability config: rpc error: code = Unknown desc = failed to decode transmitter: encoding/hex: invalid byte: U+0093"`

This is the flow

Pushing on chain:
   - Push ocr3 config for any capability (does not matter aptos) [here](https://github.com/smartcontractkit/chainlink/blob/0ff1113f173aec2f7782ee2b47aad2e93f0a9c14/deployment/cre/capabilities_registry/v2/changeset/add_capabilities.go#L125)
     - calls [ComputeOCR3Config](https://github.com/smartcontractkit/chainlink/blob/0ff1113f173aec2f7782ee2b47aad2e93f0a9c14/deployment/cre/capabilities_registry/v2/changeset/configure_capabilities_registry.go#L284)
     - Which ends up creating `OCR2OracleConfig` with `Transmitters as []common.Address`
     - It then calls `OCR2OracleConfigToMap` to push this data to chain which stores transmitter address as [base64](https://github.com/smartcontractkit/chainlink/blob/0ff1113f173aec2f7782ee2b47aad2e93f0a9c14/deployment/cre/ocr3/registry_config.go#L146) string
     - protojson interprets base64 strings as raw bytes, on-chain proto stores the raw 20-byte EVM address
 

Reading on chain:
- LocalRegistry.Unmarshal does ocrtypes.Account(t) where t is those raw 20 bytes → the Account string contains raw binary. So when the gRPC server sees the Account, it contains raw binary bytes (like \xab\xcd\x93...). It naturally cannot hexDecode that

### DIFF
--- a/pkg/loop/internal/core/services/capability/capabilities_registry.go
+++ b/pkg/loop/internal/core/services/capability/capabilities_registry.go
@@ -2,6 +2,7 @@ package capability
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"time"
 
@@ -511,7 +512,12 @@ func (c *capabilitiesRegistryServer) ConfigForCapability(ctx context.Context, re
 			}
 			transmitters := make([][]byte, len(cfg.Transmitters))
 			for i, t := range cfg.Transmitters {
-				transmitters[i] = []byte(t)
+				decoded, hexErr := hex.DecodeString(string(t))
+				if hexErr == nil {
+					transmitters[i] = decoded
+				} else {
+					transmitters[i] = []byte(t)
+				}
 			}
 			ccp.Ocr3Configs[key] = &capabilitiespb.OCR3Config{
 				ConfigCount:           cfg.ConfigCount,
--- a/pkg/loop/internal/core/services/capability/capabilities_registry.go
+++ b/pkg/loop/internal/core/services/capability/capabilities_registry.go
@@ -2,7 +2,6 @@ package capability
 
 import (
 	"context"
-	"encoding/hex"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mchain0 (dismissed):

Unheilbar (dismissed):

jmank88 (changes requested):
Are you sure that this works for all cases? IIRC protobufs only accept utf8 encoded strings - not arbitrary bytes like Go does. Can we add a test?

pkg/loop/internal/core/services/capability/capabilities_registry.go
```diff
@@ -512,9 +512,11 @@ func (c *capabilitiesRegistryServer) ConfigForCapability(ctx context.Context, re
 			}
 			transmitters := make([][]byte, len(cfg.Transmitters))
 			for i, t := range cfg.Transmitters {
-				transmitters[i], err = hex.DecodeString(string(t))
-				if err != nil {
-					return nil, fmt.Errorf("failed to decode transmitter: %w", err)
+				decoded, hexErr := hex.DecodeString(string(t))
```
ilija42, line 515: From the discussion that we had it sounds like `hex.DecodeString` never works because the transmitter string isn't in hex, so If the transmitter is base64 encoded lets just decode it here into a hex string and keep it in that format.

ilija42 (changes requested):

yashnevatia:
in favour of https://github.com/smartcontractkit/chainlink/pull/22062/changes#diff-77b503c8d6b1703eed193d8c7837bb9a32935e53e6aaccfb567d3f45e8cb4cd7

PR #1984, opened 2026-04-16
fix: eliminate data race in SleeperTask context lifecycle

Fletch153:
## Summary

Defense-in-depth against testify/mock's unsynchronized reflect reads of context arguments. Replaces `SleeperTask`'s context implementation with a read-only `stopChanCtx` that has no mutable internal state.

## Background

testify's mock library has a fundamentally broken code path: when a mock method is called without a matching expectation, `callString()` formats all arguments via `fmt.Sprintf("%#v", args...)`, which walks them with `reflect`. This is completely unsynchronized — any concurrent mutation of an argument (like `context.cancel()` writing to `cancelCtx` internals) causes a data race. An upstream fix was attempted but rejected.

## What This Changes

`SleeperTask.workerLoop()` previously created its context via `StopChan.NewCtx()`, which internally calls `context.WithCancel()` + spawns a detached goroutine that calls `cancel()` when the stop channel closes. That `cancel()` writes to `cancelCtx` internal fields (`sync/atomic.AddInt32`, mutex state, etc.) — racing with any concurrent `fmt.Sprintf("%#v", ctx)` from testify's error path.

The new `stopChanCtx` is a read-only `context.Context` implementation backed directly by the stop channel:

- `Done()` returns the stop channel — closing it signals cancellation
- `Err()` checks if the channel is closed
- No `cancel()` function, no detached goroutine, no mutable internal state
- Child contexts (`context.WithCancel`, `context.WithTimeout`) propagate correctly via Go's standard `Done()` mechanism

Since there's nothing to mutate, testify can `fmt.Sprintf("%#v")` this context all day without racing.

## Test Evidence

`TestSleeperTask_NoConcurrentContextRace` reproduces the exact scenario: `Work()` goroutines read the context via `fmt.Sprintf` while `Stop()` concurrently closes the stop channel.

**Old code** (NewCtx/CtxCancel): `WARNING: DATA RACE` — `reflect.typedmemmove` vs `sync/atomic.AddInt32`
**New code** (stopChanCtx): 10 runs × all tests, zero races

## Related

- Root cause fix (shutdown ordering): https://github.com/smartcontractkit/chainlink-evm/pull/425
- Belt-and-suspenders (mock expectations): https://github.com/smartcontractkit/chainlink/pull/22042
- Failing CI: https://github.com/smartcontractkit/chainlink/actions/runs/24509952973/job/71638172320

### DIFF
--- a/pkg/utils/sleeper_task.go
+++ b/pkg/utils/sleeper_task.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"context"
+	"time"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 )
@@ -110,8 +111,7 @@ func (s *SleeperTask) WorkDone() <-chan struct{} {
 func (s *SleeperTask) workerLoop() {
 	defer close(s.chDone)
 
-	ctx, cancel := s.chStop.NewCtx()
-	defer cancel()
+	ctx := &stopChanCtx{ch: s.chStop}
 
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


pkg/utils/sleeper_task.go
jmank88, line 1: A custom `context.Context` make me nervous. I do see some in the wild now, though most contain an embedded `context.Context` that they are extending and sometimes don't even override any methods 🤔 

Let's wait and see how the other fixes help before trying this.

Fletch153: Closing for now

PR #1988, opened 2026-04-20
pkg/services: wrap context to prevent mockery race

jmank88:
This is a reinterpretation of the fix proposed in https://github.com/smartcontractkit/chainlink-common/pull/1984, in lieu of https://github.com/stretchr/testify/pull/1689 or similar.

When the `-race` detector is active, we wrap contexts derived from `services.StopChan` so they don't flag the detector when printed by mockery via `%#v`.


mchain0 (approved):

Fletch153 (approved):

pavel-raykov (approved):

jmank88:
No changes - just trying to trigger CodeQL...

PR #1990, opened 2026-04-20
Add Stellar Service plumbing and GetLedgerEntries,GetLatestLedger

ilija42:


pkg/types/chains/stellar/stellar.go
```diff
@@ -0,0 +1,62 @@
+package stellar
+
+import "context"
+
+// XDR is a base64-encoded XDR-serialized value (envelope, data, result, key, etc.).
+type XDR string
+
+// LedgerHash is the hex-encoded SHA-256 hash of a Stellar ledger.
+type LedgerHash string
```
jmank88: Is it important for these to be independent types? I think the extra conversions could be avoided, even just with aliases:

```suggestion
// XDR is a base64-encoded XDR-serialized value (envelope, data, result, key, etc.).
type XDR = string

// LedgerHash is the hex-encoded SHA-256 hash of a Stellar ledger.
type LedgerHash = string
```

pkg/chains/stellar/proto_helpers.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid base64 XDR %q: %w", x, err)
+	}
+	return b, nil
+}
+
+// bytesToXDR base64-encodes raw binary XDR to the domain type.
+func bytesToXDR(b []byte) stellar.XDR {
+	return stellar.XDR(base64.StdEncoding.EncodeToString(b))
+}
```
jmank88: Would it be cleaner to define these in `package stellar`? e.g. `XDR) ToBytes() ([]byte, error)` and `stellar.ParseXDR([]byte)` 🤷

yashnevatia (dismissed):

dhaidashenko (approved):

Unheilbar (approved):

jmank88 (approved):

PR #1995, opened 2026-04-22
pkg/contexts: remove org ID normalization

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-3673

This PR updates `CRE.Normalized()` to leave the org id unmodified.

### DIFF
--- a/pkg/contexts/contexts.go
+++ b/pkg/contexts/contexts.go
@@ -38,10 +38,8 @@ type CRE struct {
 
 // Normalized returns a possibly modified CRE with normalized values.
 func (c CRE) Normalized() CRE {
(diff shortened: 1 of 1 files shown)


pkg/contexts/contexts.go
jmank88, line 47: Are these problematic too? Or OK since they are a stricter hex format?

patrickhuie19 (approved):

pkg/contexts/contexts.go
cedric-cordenier: I think those are OK 👍 We already do some normalization in various places so they're less problematic

pavel-raykov (approved):

PR #1998, opened 2026-04-27
Fix LOOP relaunch state reuse that kept clients bound to dead plugin sockets

krebernisak:
## Summary

This fixes a LOOP recovery bug where a plugin process restart did not rebuild the logical service that had been created from the old plugin generation.

Before this change, `PluginService` could relaunch the plugin process and replace the underlying plugin client/protocol, but still keep serving through the original logical service object. In relayer flows, that stale service could keep child clients and broker state tied to dead plugin generations. Under repeated crashes, that amplified into repeated dials against dead unix sockets and broader node health degradation.

The fix makes plugin relaunch semantics generation-safe:
- each relaunch builds a fresh logical service from the newly dispensed plugin instance
- the new service is swapped in atomically before the old generation is retired
- the superseded service and old plugin client are then closed

This PR also includes the minimal shared client recovery fix needed to keep the recovery path healthy:
- only one goroutine performs `clientConn` refresh work
- other callers wait without blocking behind a long-held write lock
- `Close` on an already-dead plugin client is treated as a no-op instead of reconnecting just to close again

## Root Cause

The old implementation mixed generations:
- fresh plugin process
- fresh plugin client/protocol
- stale logical service

That was safe only if services were completely generation-agnostic. They are not. The relayer service and its child clients hold state that is effectively bound to the plugin generation they were created from. Once the plugin crashed and came back on a new socket path, the old logical service could continue operating on stale clients and keep attempting to recover against dead sockets.

This explains the production symptom pattern:
- one plugin crashes
- keepalive relaunches the plugin process
- the node continues using stale logical/client state from the old generation
- stale clients keep dialing dead `/tmp/plugin...` sockets
- repeated retries spread degradation across the node and build up socket churn

### 1. Recreate logical services on relaunch

`PluginService` now:
- builds a fresh service and health reporter on every successful launch
- swaps the new generation into shared state under lock
- closes the old service only after the new generation is visible
- closes the old plugin client/protocol after swap

This keeps service lifetime aligned with plugin-generation lifetime.

### 2. Fix refresh-path lock convoying in `clientConn`

`clientConn.refresh()` previously held the write lock across the full reconnect/backoff loop. During failures, that could block unrelated readers and cleanup paths behind the lock and make recovery look like a broader stall.

It now:
- lets only one goroutine perform reconnect work
- lets other callers wait on a completion channel without holding the state lock
- publishes the new connection only once both broker dependencies and the gRPC client are ready

### 3. Avoid reconnecting dead child clients just to close them

If the underlying plugin is already gone, a `Close` RPC should not recreate the client just to send one last close call. This now returns success instead of triggering more churn during teardown.

## Tests

Added:
- `TestPluginService_Reset_RecreatesLogicalService`
  - proves a plugin restart creates a new logical service
  - proves the superseded service is closed

- `TestRelayerService_RepeatedRecoveryKeepsCurrentPluginSocketsBounded`
  - forces repeated plugin crashes with a helper plugin limited to one request
  - verifies relaunches happen
  - verifies that once the crash loop stops, current plugin sockets settle back to a small bounded set instead of growing with each crash

Updated:
- `TestRelayerService_recovery`
  - adjusts the degraded-health expectation to the valid in-flight recovery shape where only the top-level service and current logical relayer client may be ready enough to report health

### Requires

- N/A

### Supports

- N/A


pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -45,14 +46,16 @@ type PluginService[P grpcPlugin, S services.Service] struct {
 
 	newService NewService[S]
 
-	serviceCh      chan struct{} // closed when service is available
+	serviceCh chan struct{} // closed when service is available
+
+	// Deprecated: use CurrentService instead. This is not protected by the mutex and may be stale as Service gets replaced across plugin relaunches.
 	Service        S
```
jmank88, line 52: IIRC it was important for this field to be set only once, and for any references stored by callers to never become dead and require replacing. Does that property still hold? Or is it no longer necessary?

pkg/loop/internal/goplugin/plugin_service.go
```diff
@@ -179,14 +228,27 @@ func (s *PluginService[P, S]) Start(context.Context) error {
...
 	default:
 		return ErrPluginUnavailable
 	}
 }
 
 func (s *PluginService[P, S]) Name() string { return s.lggr.Name() }
 
+// CurrentService returns the currently published logical service for the active plugin generation.
+//
+// Callers should wait for service availability first via Wait or WaitCtx. The returned snapshot is
```
jmank88, line 242: > Callers should wait for service availability first via Wait or WaitCtx

Why don't we just do that for them?

krebernisak:
Closing for https://github.com/smartcontractkit/chainlink-common/pull/2027

PR #1999, opened 2026-04-27
Running a module respects the timeout on the provided context

nolag:


mchain0 (dismissed):

pkg/workflows/wasm/host/module.go
```diff
@@ -731,7 +740,7 @@ func runWasm[I, O proto.Message](
 	// Note - there is no other reliable signal on the error that can be used to infer it is due to epoch deadline
 	// being reached, so if an error is returned after the deadline it is assumed it is due to that and return
 	// context.DeadlineExceeded.
-	if err != nil && executionDuration >= *m.cfg.Timeout-m.cfg.TickInterval { // As start could be called just before epoch update 1 tick interval is deducted to account for this
+	if err != nil && ((executionDuration >= *m.cfg.Timeout-m.cfg.TickInterval) || ctxDeadline.Before(time.Now())) { // As start could be called just before epoch update 1 tick interval is deducted to account for this
```
jmank88: What if `hasDeadline == false`? Isn't a zero value deadline always before now?

jmank88: How about just going to the context directly, rather than computing deadlines manually? Then we don't have to worry about fixed deadlines vs. dynamic cancellation.
```suggestion
	if err != nil && ((executionDuration >= *m.cfg.Timeout-m.cfg.TickInterval) || ctx.Err() != nil)) { // As start could be called just before epoch update 1 tick interval is deducted to account for this
```

nolag: Thanks

jmank88 (approved):

DylanTinianov (approved):

PR #2002, opened 2026-04-27
Add L1Fee to evm service Receipt

ilija42:


pkg/chains/evm/proto_helpers.go
```diff
@@ -147,6 +147,7 @@ func ConvertReceiptToProto(receipt *evmtypes.Receipt) (*Receipt, error) {
 		BlockNumber:       valuespb.NewBigIntFromInt(receipt.BlockNumber),
 		TxIndex:           receipt.TransactionIndex,
 		EffectiveGasPrice: valuespb.NewBigIntFromInt(receipt.EffectiveGasPrice),
+		L1_Fee:            valuespb.NewBigIntFromInt(receipt.L1Fee),
```
dhaidashenko, line 150: That's a problematic change to read capability. During the update, part of the nodes will report receipts with the L1 fee, and others will report it without, so it's possible to see an increase in reads failure/

dhaidashenko (changes requested):

dhaidashenko (dismissed):

pkg/chains/evm/proto_helpers.go
dhaidashenko: Not a problem, since we chain-read uses a different structure.

dhaidashenko (approved):

vyzaldysanchez (approved):

PR #2003, opened 2026-04-28
pkg/services: add Engine.GoTickUntil

jmank88:


pavel-raykov (dismissed):

jmank88:
I removed `WaitFor`. Though it captured a pattern, it actually didn't integrate with `GoTickUtil` well, because of the indirection. It also didn't have support for an overridden context, which would have made things messier. Better to inline select statements as needed.

pavel-raykov (approved):

PR #2004, opened 2026-04-28
Enable workflows to request a TEE

nolag:


hendoxc (dismissed):

russell-stern (dismissed):

pkg/workflows/wasm/host/internal/rawsdk/helpers_wasip1.go
```diff
@@ -58,6 +58,7 @@ func SendError(err error) {
 func SendSubscription(subscriptions *sdk.TriggerSubscriptionRequest) {
 	execResult := &sdk.ExecutionResult{Result: &sdk.ExecutionResult_TriggerSubscriptions{TriggerSubscriptions: subscriptions}}
 	sendResponse(BufferToPointerLen(Must(proto.Marshal(execResult))))
+	os.Exit(0)
```
jmank88, line 61: Was this for testing?

nolag: no, it should have been there the whole time. Other tests just didn't do anything after calling it.

pkg/workflows/host/tee_selection_provider.go
```diff
@@ -0,0 +1,53 @@
...
+			return true
+		}
+
+		switch teet := tee.Item.(type) {
+		case *sdkpb.Tee_AnyRegions:
+			for _, provider := range providers {
+				if provider(tee) {
+					return true
+				}
+			}
```
jmank88, line 33: There's a shorthand func in package slices:
```suggestion
			if slices.ContainsFunc(providers, provider) {
				return true
			}
```

nolag: Not quite

func ContainsFunc[S ~[]E, E any](s S, f func(E) bool) bool

but this is []func(E)bool. Provider can't be an argument if you remove the loop, it's defined in it because it's the element.

pkg/workflows/artifacts/utils.go
```diff
@@ -78,10 +80,77 @@ func GetBuildCmd(inputFile string, outputFile string, rootFolder string) *exec.C
...
 
 	return buildCmd
 }
+
+// goToolchainFromMod returns a GOTOOLCHAIN value (e.g. "go1.26.2") derived
+// from the nearest go.mod walking up from startDir. Prefers the toolchain
+// directive when present, falling back to the go directive. Returns "" when
+// no go.mod or version can be determined.
+func goToolchainFromMod(startDir string) string {
+	goModPath := findNearestGoMod(startDir)
```
jmank88: FWIW you could run `go env GOMOD`

pkg/workflows/host/tee_selection_provider.go
nolag: Also, providers is a map, not a slice

pkg/workflows/artifacts/utils.go
```diff
@@ -78,10 +80,77 @@ func GetBuildCmd(inputFile string, outputFile string, rootFolder string) *exec.C
...
+			continue
+		}
+		fields := strings.Fields(line)
+		if len(fields) < 2 {
+			continue
+		}
+		switch fields[0] {
+		case "go":
+			goVersion = fields[1]
+		case "toolchain":
```
jmank88: IIUC `go env GOTOOLCHAIN` will give you this one directly

jmank88: And `go env GOVERSION` should give you the `go` line

pkg/workflows/artifacts/utils.go
nolag: will remove in favour of using go env TOOLCHAIN and go env GOVERSION if it's unset. I know I can do one call with go env, but I think it's easier to read the result this way and the call is fast.

nolag: nope, I'll use the GOMOD, GOVERSION is my version, not the mod file :P.

pkg/workflows/wasm/host/internal/rawsdk/helpers_wasip1.go
jmank88: Can you leave a comment explaining? This doesn't make any sense to me without context

pkg/workflows/host/tee_selection_provider.go
jmank88: Ah my bad

pkg/workflows/wasm/host/internal/rawsdk/helpers_wasip1.go
nolag: The SendError function above it does the same thing.

It's a very dumbed down version of [this code](https://github.com/smartcontractkit/cre-sdk-go/blob/main/cre/wasm/runner.go#L74) in the real SDK.

The internal one here is for bootstrapping, we can test the standard tests work, and that we can do things correctly with modules without needing the SDK. 

The SDK needs the standard tests to verify correct behaviour, so we can't create a circular dependency. Also, the goal is to test the host/guest interactions at a lower level to ensure it works before we ask the SDKs to implement it for their language

jmank88:
@erikburt Is API Diff failing here? These aliases are meant to be 100% compatible.

Edit: Wait, did it correct itself? maybe I just saw a stale comment

pkg/capabilities/v2/protoc/pkg/template_generator.go
```diff
@@ -259,6 +259,18 @@ func (t *TemplateGenerator) runTemplate(name, tmplText string, args any, partial
...
+			md, err := getCapabilityMetadata(s)
+			if err != nil {
+				return false, err
+			}
+			for _, env := range md.AdditionalEnvironments {
+				if env == generator.AdditionalEnvironments_ADDITIONAL_ENVIRONMENTS_TEE {
+					return true, nil
+				}
+			}
+			return false, nil
```
jmank88: How about this one?
```suggestion
			return slices.Contains(md.AdditionalEnvironments, generator.AdditionalEnvironments_ADDITIONAL_ENVIRONMENTS_TEE), nil
```

jmank88 (dismissed):

jmank88:
> @erikburt Is API Diff failing here? These aliases are meant to be 100% compatible.
> 
> Edit: Wait, did it correct itself? maybe I just saw a stale comment

Oh actually it is updating the two comments separately I guess? That may be confusing. But the main problem is that aliases seem to show as breaking.

jmank88 (approved):

DylanTinianov (approved):

PR #2008, opened 2026-04-28
loop/net: defer old relayer resource teardown until refresh succeeds

krebernisak:
## Summary

Fix relayer client refresh so a failed replacement attempt does not tear down the currently published broker-backed resources too early.

The issue was in `client.go:124`: `clientConn.refresh()` closed the active connection and its tracked deps before a replacement client had been successfully created and dialed. For relayer construction, those deps can include live broker-served resources such as keystore-related listeners. If the replacement attempt failed, the old generation was already torn down, which could leave callers with dead /tmp/plugin... sockets.

## Changes

- Updated `client.go:124` so refresh now:
    - keeps the current `cc` and `deps` alive while attempting the replacement
    - cleans up only failed-attempt deps immediately
    - publishes the new connection first
    - closes the previous generation only after successful cutover
- Added a focused regression in `client_test.go:58` that proves:
    - old published deps remain live across a failed refresh attempt
    - failed-attempt deps are cleaned up
    - old deps are closed only after the replacement is published


krebernisak:
Closing for https://github.com/smartcontractkit/chainlink-common/pull/2027

PR #2011, opened 2026-04-29
Split oversized chip ingress PublishBatch requests

pkcll:
## Summary
- split chip-ingress PublishBatch sends by serialized request size
- fail a single oversized event before calling PublishBatch
- add regression tests for request splitting and oversized single-event handling

## Tests
- go test ./batch
- go test ./...
- go build ./...


pkcll:
Its already in the main branch

PR #2017, opened 2026-04-30
fix(aptos): reject nil tx entries in AccountTransactionsReply [PLEX-2889]

Fletch153:
## Summary
- `ConvertTransactionFromProto(nil)` returns `(nil, nil)`, so `ConvertAccountTransactionsReplyFromProto` silently produced a slice containing nil transactions when a relayer reply contained nil entries.
- Downstream `TxInfoRetriever.scanTransactions` (in `capabilities`) dereferences `tx.Data` with no nil check — a malicious or buggy relayer could panic the capability.
- Reject any nil converted transaction at the converter boundary with an indexed error.

[PLEX-2889](https://smartcontract-it.atlassian.net/browse/PLEX-2889)

## Test plan
- [x] `go test ./pkg/chains/aptos/...` passes
- [x] New test asserts indexed error on nil entry

[PLEX-2889]: https://smartcontract-it.atlassian.net/browse/PLEX-2889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yashnevatia (approved):

pkg/chains/aptos/proto_helpers.go
```diff
@@ -406,6 +406,9 @@ func ConvertAccountTransactionsReplyFromProto(proto *AccountTransactionsReply) (
 		if err != nil {
 			return nil, fmt.Errorf("failed to convert transaction %d: %w", i, err)
 		}
+		if tx == nil {
+			return nil, fmt.Errorf("transaction %d is nil", i)
```
ilija42, line 414: better to catch all errors and then return, but won't block if this is urgent

Fletch153: Will think about this some more and create a ticket if it's worth doing. My concern with that approach is that there could be 100s of txs, that all fail with a similar error which leads to flooding

ilija42 (approved):

PR #2019, opened 2026-04-30
rm borsh encoder

Unheilbar:
Supports: https://github.com/smartcontractkit/chainlink/pull/22269


pkg/capabilities/consensus/ocr3/datafeeds/securemint_aggregator.go
jmank88, line 1: I'm concerned about making breaking changes here. Even if core is the only user and updates quickly, it creates a point in time that we cannot easily move past for things like backports etc. and often breaks integration-tests in the chain repos. Could we leave it deprecated for a while? Or at least avoid break compilation?

PR #2027, opened 2026-05-04
pkg/loop/internal: cleanup resources; fail TestRelayerService if goroutines leak

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-3973

Before this PR, `TestRelayerService` was leaking ~40 goroutines, mostly client side `CallbackSerializer`s. Now they are all cleaned up, and the test will continue to check for leaks going forward.
 
- fail `TestRelayerService` if goroutines leak
- return more dependency resources
- expose `Close` through `ClientConnInterface`
- close final client, not just refreshed ones
- bump github.com/hashicorp/go-plugin to v1.8.0, including fd leak fix

### DIFF
--- a/pkg/loop/internal/core/services/reportingplugin/ocr2/reporting_plugin_service.go
+++ b/pkg/loop/internal/core/services/reportingplugin/ocr2/reporting_plugin_service.go
@@ -60,7 +60,7 @@ func (m *ReportingPluginServiceClient) NewReportingPluginFactory(
 			pb.RegisterPipelineRunnerServiceServer(s, pipeline.NewRunnerServer(pipelineRunner))
 		})
 		if err != nil {
-			return 0, nil, err
+			return 0, deps, err
 		}
 		deps.Add(pipelineRunnerRes)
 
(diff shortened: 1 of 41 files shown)


pkg/loop/internal/core/services/reportingplugin/ocr2/reporting_plugin_service.go
```diff
@@ -60,23 +60,23 @@ func (m *ReportingPluginServiceClient) NewReportingPluginFactory(
 			pb.RegisterPipelineRunnerServiceServer(s, pipeline.NewRunnerServer(pipelineRunner))
 		})
 		if err != nil {
-			return 0, nil, err
+			return 0, deps, err
```
jmank88, line 63: I suspect that we had a bare return pattern and the nuance of returning `deps` in the case of failure was overlooked and lost along the way. Better to just be explicit

pavel-raykov (dismissed):

pavel-raykov (approved):

MStreet3 (approved):

krebernisak (approved):

ogtownsend (approved):

PR #2040, opened 2026-05-08
BaseTrigger: Remove backoff + jitter

DylanTinianov:
These were added for soaking on staging when there were underlaying traffic issues. We don't actually want these anymore.


pkg/capabilities/base_trigger.go
```diff
@@ -687,8 +650,7 @@ func (b *BaseTriggerCapability[T]) collectResendCandidate(
 	interval time.Duration,
 	toResend *[]PendingEvent,
 ) {
-	backoff := addJitter(retryBackoff(interval, rec.Attempts))
```
fernandezlautaro, line 690: aren't you missing removing addJitter and retryBackoff functions?

fernandezlautaro (changes requested):

pkg/capabilities/base_trigger.go
DylanTinianov: I did remove them above ^^

fernandezlautaro (dismissed):

fernandezlautaro (approved):

vyzaldysanchez (approved):

PR #2048, opened 2026-05-12
mark errors as limit error

ettec:


pkg/capabilities/errors/error.go
```diff
@@ -110,6 +114,12 @@ func NewPublicUserError(err error, errorCode ErrorCode) Error {
 	return NewError(err, VisibilityPublic, OriginUser, errorCode)
 }
 
+// NewLimitBreachedError creates a public user error that embeds the provided limit error, indicating that a CRE limit
+// breach has occurred. The error message will include the provided errorMsg as a prefix for additional context.
+func NewLimitBreachedError(errorMsg string, err limits.LimitError) Error {
```
jmank88: Is it important to restrict the kind of errors passed in?
```suggestion
func NewLimitBreachedError(errorMsg string, err error) Error {
```
It would be too restrictive if the caller was dealing with an already wrapped error.

jmank88: Separately, should we align the terminology?
```suggestion
func NewLimitExceededError(errorMsg string, err limits.LimitError) Error {
```

ettec: true, updated

pkg/settings/limits/errors.go
jmank88, line 1: I think we can leave this out now

pkg/capabilities/errors/error.go
```diff
@@ -110,6 +112,12 @@ func NewPublicUserError(err error, errorCode ErrorCode) Error {
 	return NewError(err, VisibilityPublic, OriginUser, errorCode)
 }
 
+// NewLimitBreachedError creates a public user error that embeds the provided limit error, indicating that a CRE limit
+// breach has occurred. The error message will include the provided errorMsg as a prefix for additional context.
+func NewLimitBreachedError(errorMsg string, err error) Error {
+	return NewPublicUserError(fmt.Errorf("%s: %w", errorMsg, err), LimitExceeded)
```
jmank88: Align on `Exceeded`?
```suggestion
func NewLimitExceededError(errorMsg string, err error) Error {
	return NewPublicUserError(fmt.Errorf("%s: %w", errorMsg, err), LimitExceeded)
```

agparadiso (approved):

cedric-cordenier (approved):

PR #2050, opened 2026-05-12
auto fix linter issues

jmank88:
- Upgraded linter to 2.12.2
- Ran with `--fix`
- Defensive `//nolint` for false positives
- Updated generated code header to prevent linting


pkg/codec/encodings/binary/gen/main.go
```diff
@@ -41,7 +41,7 @@ func main() {
 
 		res =
 			fmt.Appendf(nil,
-				"// DO NOT MODIFY: automatically generated from %s/main.go using the template %s\n\n%s",
+				"// Code generated by %s/main.go using the template %s. DO NOT EDIT.\n\n%s",
```
jmank88, line 44: Aligning with the generated grpc/pb header so the linter ignores them.

pavel-raykov (approved):

kirqz23 (approved):

PR #2052, opened 2026-05-13
PLEX-2683: wrap executableClient.Execute transport errors

fernandezlautaro:
Jira: https://smartcontract-it.atlassian.net/browse/PLEX-2683

Currently we are getting "error executing request: INTERNAL_ERROR : failed to execute capability" if the executableClient.Execute grpc transport fails due to 2 reasons: 1 is not adding the origina source of the problem and it is also not properly discriminating it as a caperror for further classification down the pipe.


pkg/loop/internal/core/services/capability/capabilities_test.go
fernandezlautaro, line 565: Removed duplicated `fetching an action capability, and executing it with private system error` test, word by word is the same test.

ettec (approved):

ilija42 (approved):

PR #2053, opened 2026-05-13
pkg/beholder: bridge prometheus metrics through otel

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-4426
Supports:
- https://github.com/smartcontractkit/chainlink/pull/22572


pavel-raykov (dismissed):

ogtownsend (dismissed):

pkg/loop/server.go
```diff
@@ -183,6 +187,15 @@ func (s *Server) start(opts ...ServerOpt) error {
...
 		}
 
+		// TODO repeat in core
+		var bridgeOpts []prometheus.Option
+		if !s.EnvConfig.TelemetryMetricPromBridgeEnabled {
+			// go_ runtime metrics only
+			bridgeOpts = append(bridgeOpts, prometheus.WithGatherer(&goGatherer{prometheus2.DefaultGatherer}))
+		}
+		// bridge prometheus metrics through otel
+		beholderCfg.MetricOptions = append(beholderCfg.MetricOptions, sdkmetric.WithReader(sdkmetric.NewManualReader(sdkmetric.WithProducer(prometheus.NewMetricProducer(bridgeOpts...)))))
```
pkcll: A ManualReader requires an explicit Collect() call to produce data. Since nothing calls it, the bridged Prometheus metrics will never actually be exported. It should be a PeriodicReader with the same exporter, or the Producer should be attached to the existing PeriodicReader.

pkg/beholder/client.go
```diff
@@ -509,15 +509,15 @@ func newMeterProvider(cfg Config, resource *sdkresource.Resource, auth Auth, cre
...
+			exporter,
+			sdkmetric.WithInterval(cfg.MetricReaderInterval), // Default is 10s
+		)),
 		sdkmetric.WithResource(resource),
-		sdkmetric.WithView(cfg.MetricViews...),
-	)
+		sdkmetric.WithView(cfg.MetricViews...)}
+	metricOpts = append(metricOpts, cfg.MetricOptions...)
+	mp := sdkmetric.NewMeterProvider(metricOpts...)
 	return mp, nil
```
pkcll: Could be something like this to setup prom producer

```suggestion
	readerOpts := []sdkmetric.PeriodicReaderOption{
		sdkmetric.WithInterval(cfg.MetricReaderInterval),
	}
	for _, p := range cfg.MetricProducers {
		readerOpts = append(readerOpts, sdkmetric.WithProducer(p))
	}

	metricOpts := []sdkmetric.Option{
		sdkmetric.WithReader(sdkmetric.NewPeriodicReader(exporter, readerOpts...)),
		sdkmetric.WithResource(resource),
		sdkmetric.WithView(cfg.MetricViews...),
	}
	metricOpts = append(metricOpts, cfg.MetricOptions...)
	mp := sdkmetric.NewMeterProvider(metricOpts...)
	return mp, nil
```

pkg/loop/server.go
pkcll:
```suggestion
        beholderCfg.MetricProducers = append(beholderCfg.MetricProducers, prometheus.NewMetricProducer(bridgeOpts...))

```

pkg/beholder/config.go
```diff
@@ -38,6 +38,7 @@ type Config struct {
 	MetricViews          []metric.View
 	// MetricCompressor sets the gRPC compressor for metrics. Valid values: "gzip" (default), "none".
 	MetricCompressor string
+	MetricOptions    []metric.Option
```
pkcll:
```suggestion
	MetricProducers  []metric.Producer 
```

pkg/beholder/client.go
jmank88: Do we need to have a single reader though? Or is this entangling unrelated things?

jmank88: Ah I guess we need that local exporter to construct one so we're stuck

pkcll (dismissed):

pavel-raykov (dismissed):

jmank88:
Removed comment and fixed condition - please re-ack

pavel-raykov (dismissed):

pkg/sqlutil/promutil/promutil.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	for _, m := range all {
+		if m.Name == nil {
+			continue
+		}
+		for _, prefix := range g.Prefixes {
+			if strings.HasPrefix(*m.Name, prefix) {
+				ret = append(ret, m)
+				break
+			}
```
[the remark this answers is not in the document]
jmank88: Updated

pavel-raykov (approved):

pkcll (approved):

PR #2057, opened 2026-05-14
BaseTrigger: Per org retransmit setting

DylanTinianov:
Scopes base trigger retransmit to Org

### DIFF
--- a/pkg/capabilities/base_trigger_test.go
+++ b/pkg/capabilities/base_trigger_test.go
@@ -121,6 +121,68 @@ func TestBaseTrigger_CRE_DynamicDisableStopsResend(t *testing.T) {
 	}
 }
 
+func TestBaseTrigger_CRE_PerOrgRetransmit(t *testing.T) {
+	lggr, err := logger.New()
+	require.NoError(t, err)
+	baseCtx := t.Context()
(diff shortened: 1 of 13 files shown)


pkg/capabilities/base_trigger_test.go
```diff
@@ -119,6 +121,68 @@ func TestBaseTrigger_CRE_DynamicDisableStopsResend(t *testing.T) {
...
+		}
+	}, 3*time.Second, 10*time.Millisecond, "orgA should receive at least one retransmit")
+
+	select {
+	case extra := <-chB:
+		t.Fatalf("orgB (retransmit disabled via global default) should not retransmit, got %+v", extra)
+	default:
+	}
+
+	recs, err := store.List(baseCtx)
```
fernandezlautaro: shouldn't you be using `ctxOrgA`? and also add another check to guarantee `ctxOrgB` has no rows inserted?

DylanTinianov: ctxOrgA doesn't matter for store.List in terms of how it fetches, but sure updated :)

Added check to insure no orgB rows were inserted

fernandezlautaro (dismissed):

vyzaldysanchez (dismissed):

fernandezlautaro (approved):

vyzaldysanchez (approved):

PR #2063, opened 2026-05-15
Add EnclaveConfig to confidentialrelay request params (PRIV-458)

nadahalli:
SecretsRequestParams and CapabilityRequestParams now carry an
EnclaveConfig field describing the current enclave Signers,
MasterPublicKey, T, and F. The relay-DON handler in chainlink/core
will use this to compare against onchain DON state before treating
an attested request as trusted.

Without this field, the Nitro attestation binds the request hash
but does not let the relay see the enclave config. A malicious host
can produce a genuinely-attested request over a forged enclave
config and have it accepted. This closes the gap Sigma Prime
CL112-01 identified, mirroring the pool.go-side hardening in
smartcontractkit/confidential-compute#329.

EnclaveConfig is a parallel struct in this package, not imported from
confidential-compute, to keep the dependency direction one-way. Fields
match types.EnclaveConfig there; the enclave will fill this struct
from its local instance before sending. Validate enforces non-empty
Signers, F > 0, non-empty MasterPublicKey. The relay-response hash
binds every field with Signers sorted so logically-equivalent configs
hash the same regardless of Signer ordering.

Rollout (Stage 2 of PRIV-458):
1. (this PR) chainlink-common adds the field.
2. confidential-compute bumps chainlink-common and fills EnclaveConfig
   on every outgoing relay request.
3. chainlink/core bumps both deps and verifies EnclaveConfig against
   onchain DON state in handleSecretsGet and handleCapabilityExecute.

Existing callers that do not set EnclaveConfig will now fail
Validate. This is intentional: the field is required for the relay
to perform the audit-recommended check.

Jira: https://smartcontract-it.atlassian.net/browse/PRIV-458


mchain0 (approved):

pkg/capabilities/v2/actions/confidentialrelay/types.go
```diff
@@ -25,6 +26,22 @@ const (
 	RelayResponseSignaturePrefix = "CONFIDENTIAL_RELAY_PAYLOAD_"
 )
 
+// EnclaveConfig mirrors the confidential-compute EnclaveConfig fields the
+// relay needs to verify against onchain DON state. The enclave fills this
+// from its local types.EnclaveConfig before sending each relay request.
+//
+// PRIV-458 / Sigma Prime CL112-01: without this field the request's Nitro
```
cedric-cordenier: Note I'm pretty sure we shouldn't be putting in audit tags in our code; I'm not entirely sure why that is, but the recommendation is to link to JIRA and add the audit ref there

cedric-cordenier (dismissed):

pkg/capabilities/v2/actions/confidentialrelay/types.go
nadahalli: Done. Dropped the audit tags from the comments and kept the PRIV-458 link; the CL112-01 ref lives in the ticket.

jmank88 (approved):

pavel-raykov (approved):

PR #2068, opened 2026-05-18
Vault refactor to support workflowOwner as owner of a secret

prashantkumar1982:
Supports **smartcontractkit/chainlink#22525**.

**Design:** [Vault DON design doc (new)](https://github.com/smartcontractkit/cre-docs/blob/main/vault_don_design_doc_new.md) — JWT Vault tenants and related behavior.

Tenant for the Vault JWT path is configured in **Chainlink job specs** (**`auth0.tenantID`**, alongside issuer and audience; see **chainlink#22525**). This **`chainlink-common`** change keeps **cresettings** aligned with that model.

## Coordination

- Land alongside **Go module bumps** on **chainlink#22525** pinned to the **`chainlink-common`** revision that contains this change.


pkg/capabilities/actions/vault/messages.proto
```diff
@@ -38,8 +38,8 @@ message SecretResponse {
 
 message GetSecretsRequest {
   repeated SecretRequest requests = 1;
-  string org_id = 2;
-  string workflow_owner = 3;
+  reserved 2, 3;
+  reserved "org_id", "workflow_owner";
```
jmank88: Did you consider this method of deprecation?
```suggestion
  string org_id = 2 [deprecated = true];
  string workflow_owner = 3 [deprecated = true];
```
The advantage being that the generated code will remain compatible.

pkg/settings/cresettings/settings.go
```diff
@@ -251,7 +254,7 @@ type Schema struct {
 	GatewayIncomingPayloadSizeLimit        Setting[config.Size]
 	GatewayVaultManagementEnabled          Setting[bool]
 	VaultJWTAuthEnabled                    Setting[bool]
-	VaultOrgIdAsSecretOwnerEnabled         Setting[bool]
+	VaultTenantID                          Setting[string]
```
jmank88: Please do not remove settings. Just mark them deprecated:
```suggestion
	VaultOrgIdAsSecretOwnerEnabled         Setting[bool] // Deprecated
	VaultTenantID                          Setting[string]
```

jmank88 (changes requested):

pkg/settings/cresettings/settings.go
jmank88: Where is `VaultTenantID` used?

pkg/capabilities/capabilities.go
jmank88, line 1: Why is this type affected by a vault change?

prashantkumar1982: This type field was added for Vault use, so now I was reverting it as Vault no longer needs it.
However, after offline discussions, this can be still be useful for other use-cases, so keeping it.

pkg/settings/cresettings/settings.go
prashantkumar1982: I have a follow-up PR in chainlink to use the TenantID. It is used to construct a derived workflowOwner from OrgID.

jmank88: Can you post a link? I don't see it used in the core PR.

pkg/capabilities/actions/vault/messages.proto
prashantkumar1982: I actually revisited this, and decided to delete entirely.
These fields were never used on prod or even staging. They are not not needed as per design changes. So best is to just delete them, as we know it is safe

jmank88: These fields are currently referenced from chainlink/v2 though. Please do not remove them in a breaking manner.

jmank88 (changes requested):

pkg/capabilities/capabilities.go
```diff
@@ -170,8 +170,9 @@ type SpendLimit struct {
 }
 
 type RequestMetadata struct {
-	WorkflowID               string
-	WorkflowOwner            string
+	WorkflowID    string
+	WorkflowOwner string
+	// OrgID is CRE organization context propagated with capability requests.
```
jmank88: Is this comment adding value? I think it could be worth noting that it is not guaranteed to be set.

pkg/settings/cresettings/settings.go
prashantkumar1982: https://github.com/smartcontractkit/chainlink/pull/22525

pkg/capabilities/actions/vault/messages.proto
prashantkumar1982: Removed them all here: https://github.com/smartcontractkit/chainlink/pull/22525

pkg/capabilities/capabilities.go
prashantkumar1982: I can remove the comment, rather than just say this maybe set or not.

pkg/capabilities/actions/vault/messages.proto
jmank88: That's great but the historical references are still there, so the incompatibility may still come up, e.g. when backporting to a release patch, or when working in chainlink-deployments or a chain integrations-tests/ module.

pkg/settings/cresettings/settings.go
jmank88: This field is not used in that PR as of a80dbb94fe584810c85139714200790b47a8b4c2

pkg/capabilities/actions/vault/messages.proto
prashantkumar1982: Ok, these fields were just added recently and never used/shipped so the only risk is when someone is backporting to an older release branch.
So bringing the fields back with deprecated tag.

pkg/settings/cresettings/settings.go
prashantkumar1982: Ohh, sorry I had forgotton to use it.
Added now. The latest PR should have it.

russell-stern (dismissed):

pkg/settings/cresettings/settings.go
```diff
@@ -251,7 +254,9 @@ type Schema struct {
 	GatewayIncomingPayloadSizeLimit        Setting[config.Size]
 	GatewayVaultManagementEnabled          Setting[bool]
 	VaultJWTAuthEnabled                    Setting[bool]
-	VaultOrgIdAsSecretOwnerEnabled         Setting[bool]
+	VaultOrgIdAsSecretOwnerEnabled         Setting[bool] // Deprecated
+	PropagateOrgIDInRequestMetadata        Setting[bool]
+	TenantID                               Setting[uint64]
```
jmank88: Is there any more we could add to this name? Because it may be confused with the settings package notion of tenant. I notice you already removed the Vault prefix.

prashantkumar1982: removed this, and added to job-specs

russell-stern (approved):

jmank88 (approved):

pkg/settings/cresettings/settings.go
```diff
@@ -52,12 +52,14 @@ var DefaultGetter Getter
...
-	WorkflowExecutionConcurrencyLimit:      Int(1000),
-	GatewayIncomingPayloadSizeLimit:        Size(1 * config.MByte),
-	GatewayVaultManagementEnabled:          Bool(true),
-	VaultJWTAuthEnabled:                    Bool(false),
+	WorkflowLimit:                     Int(1000),
+	WorkflowExecutionConcurrencyLimit: Int(1000),
+	GatewayIncomingPayloadSizeLimit:   Size(1 * config.MByte),
+	GatewayVaultManagementEnabled:     Bool(true),
+	VaultJWTAuthEnabled:               Bool(false),
+	// Deprecated: retained for backwards compatibility; workflow owner identifies secret ownership.
```
timothyF95, line 60: One thing to note here is that we never shipped a CLI version that used org owned secrets so we do not need to maintain backwards compatibility. This can be safely deprecated

prashantkumar1982: Thanks, yes agreed.
I initially deleted it, but @jmank88 asked to keep it deprecated.

I think it is just an artifact of this field existing in cld repo on some env, or if an older release branch needs to bump chainlink-common.

PR #2069, opened 2026-05-19
pkg/beholder: log chip ingress batch emitter drops on backpressure and send failure

pkcll:
Adds Info-level logging when chip ingress events are dropped due to backpressure (buffer full) or batch send failure, matching the style of the legacy synchronous emitter.

### DIFF
--- a/pkg/beholder/batch_emitter_service.go
+++ b/pkg/beholder/batch_emitter_service.go
@@ -158,6 +158,7 @@ func (e *ChipIngressBatchEmitterService) emitInternal(ctx context.Context, body 
 			// cancelled contexts.
 			if sendErr != nil {
 				e.metrics.eventsDropped.Add(ctx, 1, metricAttrs)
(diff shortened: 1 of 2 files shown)


pkg/beholder/batch_emitter_service.go
```diff
@@ -167,6 +168,7 @@ func (e *ChipIngressBatchEmitterService) emitInternal(ctx context.Context, body
 		})
 		if queueErr != nil {
 			e.metrics.eventsDropped.Add(ctx, 1, metricAttrs)
+			e.eng.Infow("failed to emit to chip ingress: backpressure", "error", queueErr, "domain", domain, "entity", entity)
```
jmank88: Why info instead of error?

pkcll:
Reopened here with more explicit branch name https://github.com/smartcontractkit/chainlink-common/pull/2070

PR #2071, opened 2026-05-19
[CRE-491] Move mercury protocol type from core to common

pavel-raykov:
Supports https://github.com/smartcontractkit/chainlink/pull/22531

### DIFF
--- a/pkg/types/mercury/types.go
+++ b/pkg/types/mercury/types.go
@@ -2,6 +2,7 @@ package mercury
 
 import (
 	"context"
+	"fmt"
 	"math/big"
 
 	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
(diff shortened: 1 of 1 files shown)


pkg/types/mercury/types.go
jmank88, line 1: Where is this used?

pavel-raykov: it is used to instantiate mercury transmitter in core - https://github.com/smartcontractkit/chainlink/pull/22531

jmank88: What is the reason for moving it here?

pavel-raykov: We are moving core/services/llo to chainlink-data-streams, so we need a place to store this config type which will be called both from chainlink-data-streams and chainlink-evm (once llo_provider is moved there)

jmank88: Is there a relayer API service that should reference it? (or some other type that includes it as a field?)

pavel-raykov: it is part of MercuryConfig which is a part of RelayerOpts which is used to instantiate a new EVM Relayer. 

It is not a part of the relayer API service.

pavel-raykov: Ok, will move MercuryTransmitterProtocol to chainlink-data-streams since it is not an essential  API part.

PR #2074, opened 2026-05-20
Bound capability call's max concurrency

justinkaseman:


pkg/workflows/wasm/host/module.go
```diff
@@ -192,6 +196,10 @@ func NewModule(ctx context.Context, modCfg *ModuleConfig, binary []byte, opts ..
 		modCfg.MaxFetchRequests = defaultMaxFetchRequests
 	}
 
+	if modCfg.MaxPendingCalls == 0 {
+		modCfg.MaxPendingCalls = defaultMaxPendingCalls
```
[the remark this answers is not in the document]
cedric-cordenier: Worth being a uint maybe?

pkg/workflows/wasm/host/execution.go
```diff
@@ -38,6 +39,13 @@ type execution[T any] struct {
 // channel and storing each channel with a unique identifier for future
 // retrieval on await.
 func (e *execution[T]) callCapAsync(ctx context.Context, req *sdkpb.CapabilityRequest) error {
+	// Acquire semaphore slot before spawning goroutine to bound concurrency.
+	select {
+	case e.pendingCallsSem <- struct{}{}:
```
cedric-cordenier: We should make this a ResourcePoolLimiter so we can control it via cresettings + have metrics on it

I'm worried that this will start backing up and we won't be able to know what's going on internally

justinkaseman: This is the way. Added.

pkg/workflows/wasm/host/module.go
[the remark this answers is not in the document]
justinkaseman: This doesn't add cleanly with the ResourcePoolLimiter. We would have to cast it everywhere.

I'll just add the guard.

cedric-cordenier (dismissed):

cedric-cordenier (dismissed):

pkg/workflows/wasm/host/module.go
```diff
@@ -192,6 +200,16 @@ func NewModule(ctx context.Context, modCfg *ModuleConfig, binary []byte, opts ..
...
 	}
 
+	if modCfg.PendingCallsLimiter == nil {
+		if modCfg.MaxPendingCalls == 0 {
+			modCfg.MaxPendingCalls = defaultMaxPendingCalls
+		}
+		if modCfg.MaxPendingCalls < 0 {
+			return nil, fmt.Errorf("MaxPendingCalls must be positive, got %d", modCfg.MaxPendingCalls)
+		}
+		modCfg.PendingCallsLimiter = limits.GlobalResourcePoolLimiter(modCfg.MaxPendingCalls)
```
jmank88: Why not connect this to cresettings?

justinkaseman: On the production path it is.

What you are seeing there is just the fallback for when `PendingCallsLimiter: h.engineLimiters.CapabilityConcurrency,` is not provided to `host.ModuleConfig` in the engine handler (in chainlink repo)

engineLimiters.CapabilityConcurrency (which is hooked up to cre-settings) gets shared between the engine and the WASM host. Since this is a scoped limiter (keyed by workflow ID from context) the engine's Available() read check and the WASM host's Wait() operate on the same per-workflow pool.

jmank88 (dismissed):

jmank88 (dismissed):

bolekk (dismissed):

bolekk (dismissed):

pkg/workflows/wasm/host/execution_semaphore_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+	ctx := t.Context()
+	var wg sync.WaitGroup
+
+	// Simulate sequential call-then-await pattern from a single WASM thread
+	// (the real case). We run it in parallel workers to stress-test the lock.
+	for w := 0; w < workers; w++ {
+		wg.Add(1)
+		go func(workerID int) {
+			defer wg.Done()
```
jmank88:
```suggestion
		wg.Gofunc(workerID int) {
```

jmank88 (dismissed):

bolekk (approved):

jmank88 (approved):

PR #2081, opened 2026-05-21
Create DurableEmitter pkg and Implement Services.Service

DylanTinianov:

### DIFF
--- a/pkg/beholder/durable_emitter.go
+++ b/pkg/beholder/durable_emitter.go
@@ -133,13 +133,15 @@ type insertResult struct {
 }
 
 type DurableEmitter struct {
+	services.Service
+	eng *services.Engine
+
 	store  DurableEventStore
 	client chipingress.Client
 	// isHostProcess determines if the emitter runs retransmit and cleanup loops.
 	// Should be set to false when initialized inside LOOP plugins.
 	isHostProcess bool
 	cfg           DurableEmitterConfig
-	log           logger.Logger
 
 	metrics *durableEmitterMetrics
 
@@ -154,7 +156,7 @@ type DurableEmitter struct {
 	// inserting is disabled.
 	insertCh chan *insertRequest
 	// insertShutdown stops new coalesced inserts; insertInFlight counts Emit
-	// callers inside the coalesced path so Close can close(insertCh) after wait
+	// callers inside the coalesced path so stop can close(insertCh) after wait
 	insertShutdown atomic.Bool
 	insertInFlight atomic.Int32
 
@@ -167,6 +169,9 @@ type DurableEmitter struct {
 	// publishCh buffers events for the batch publish loop.
 	publishCh chan publishWork
 
+	// stopCh signals background loops to exit. It is owned by DurableEmitter
+	// (not the service engine) because shutdown must drain coalesced inserts
+	// and close publishCh in a specific order after all workers stop.
 	stopCh services.StopChan
 	wg     sync.WaitGroup
 }
@@ -227,16 +232,16 @@ func NewDurableEmitter(
 	client chipingress.Client,
 	isHostProcess bool,
 	cfg DurableEmitterConfig,
-	log logger.Logger,
+	lggr logger.Logger,
 ) (*DurableEmitter, error) {
 	if store == nil {
-		return nil, errors.New("durable event store is nil")
+		return nil, fmt.Errorf("durable event store is nil")
 	}
 	if client == nil {
-		return nil, errors.New("chipingress client is nil")
+		return nil, fmt.Errorf("chipingress client is nil")
 	}
-	if log == nil {
-		return nil, errors.New("logger is nil")
+	if lggr == nil {
+		return nil, fmt.Errorf("logger is nil")
 	}
 	if cfg.PublishBatchSize < 1 {
 		cfg.PublishBatchSize = 1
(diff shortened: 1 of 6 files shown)


pkg/beholder/durable_emitter.go
```diff
@@ -227,16 +232,16 @@ func NewDurableEmitter(
 	client chipingress.Client,
 	isHostProcess bool,
 	cfg DurableEmitterConfig,
-	log logger.Logger,
+	lggr logger.Logger,
 ) (*DurableEmitter, error) {
 	if store == nil {
-		return nil, errors.New("durable event store is nil")
+		return nil, fmt.Errorf("durable event store is nil")
```
jmank88: What's the thinking here? The linter (and maybe even go fix) will recommend `errors.New` as it was.

pkg/beholder/durable_emitter_test.go
```diff
@@ -498,6 +498,16 @@ func TestNewDurableEmitter_ValidationErrors(t *testing.T) {
 	assert.ErrorContains(t, err, "logger")
 }
 
+func TestDurableEmitter_HealthReport(t *testing.T) {
+	em := newTestDurableEmitter(t, NewMemDurableEventStore(), &testChipClient{}, nil)
+	require.NoError(t, em.Start(t.Context()))
+	defer func() { _ = em.Close() }()
```
jmank88: You can use `servicetest.Run` for these (and it will assert no error on close too)
```suggestion
	servicetest.Run(t, em)
```

DylanTinianov: Sounds nice but...

```
package github.com/smartcontractkit/chainlink-common/pkg/beholder
	imports github.com/smartcontractkit/chainlink-common/pkg/services/servicetest from durable_emitter_test.go
	imports github.com/smartcontractkit/chainlink-common/pkg/utils/tests from run.go
	imports github.com/smartcontractkit/chainlink-common/pkg/beholder/beholdertest from beholder.go
	imports github.com/smartcontractkit/chainlink-common/pkg/beholder from beholder.go: import cycle not allowed in test
```

I'll add assertion for no error on close.

tarcisiozf (dismissed):

pkcll:
@DylanTinianov, consider wrapping [DurableEmitter.Emit](https://github.com/smartcontractkit/chainlink-common/blob/982704231e46c2e213c7d10e2ebed04f5f897dca/pkg/beholder/durable_emitter.go#L341) in `eng.IfStarted(...)`. See [here](https://github.com/smartcontractkit/chainlink-common/pull/2081#issuecomment-4510324661)

pkg/beholder/durable_emitter_test.go
DylanTinianov: After syncing with data team we decided to extract to our own pkg/durableemitter. Combining with this service change since it's small, and that allows us to use `servicetest.Run(t, em)` after all :)

pkg/durableemitter/durable_emitter_metrics.go
```diff
@@ -1,135 +1,137 @@
...
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
 )
 
 var (
-	durableEmitterMetricEmitSuccess = MetricInfo{
+	durableEmitterMetricEmitSuccess = beholder.MetricInfo{
```
patrickhuie19: @pkcll do we have to import beholder pkg for these metrics? is not great separation of concerns

jmank88 (dismissed):

pkg/durableemitter/durable_emitter_metrics.go
DylanTinianov: @patrickhuie19 What's the concern about using beholder metrics similar to most components? This is separate from Chip/ Beholder emitter related stuff.

pkcll: > do we have to import beholder pkg for these metrics? is not great separation of concerns

No, it’s better not to create coupling like that with the beholder package. Please try to use OpenTelemetry primitives from `go.opentelemetry.io/otel/attribute` and `go.opentelemetry.io/otel/metric` 

For example, pass otel Meter as a dependency to DurableEmitter when it’s initialized, then pass it to newDurableEmitterMetrics.

Also `MetricInfo` is not widely used. See [here](https://github.com/search?q=org%3Asmartcontractkit%20%22beholder.MetricInfo%22&type=code).

See how its setup here 
https://github.com/smartcontractkit/chainlink-common/blob/06757a49ab15cbb8a4426ccab02edc7c1fa516c1/pkg/beholder/batch_emitter_service.go#L194

jmank88 (approved):

pkcll (approved):

pkg/durableemitter/durable_emitter.go
```diff
@@ -99,7 +102,9 @@ func DefaultDurableEmitterConfig() DurableEmitterConfig {
 		PurgeInterval:       250 * time.Millisecond,
 		PurgeBatchSize:      500,
 		PublishBatchSize:    1,
-		Metrics:             &DurableEmitterMetricsConfig{},
+		// Metrics is opt-in: callers who want instrumentation must set this
+		// and pass a metric.Meter to NewDurableEmitter.
+		Metrics: nil,
```
patrickhuie19, line 107: I can't think of a case where we would benefit from not having metrics, whether for unit/integ/higher environment. Why not initialize with a reasonable default and avoid the foot gun?

patrickhuie19 (approved):

PR #2083, opened 2026-05-22
Add vault optimizations flag

russell-stern:


pkg/capabilities/actions/vault/messages.proto
```diff
@@ -20,6 +20,7 @@ message SecretRequest {
 message EncryptedShares {
     repeated string shares = 1;
     string encryption_key = 2;
+    repeated bytes binary_shares = 3;
```
prashantkumar1982, line 23: added an optional comment on chainlink repo changes reading this field.
add a util library in this namespace to read the binary share from the EncryptedShares proto. This way it can be reused across different places.

cedric-cordenier (dismissed):

prashantkumar1982 (approved):

jmank88 (approved):

PR #2090, opened 2026-05-26
DurableEmitter: Refactor with Chip BatchEmitter

DylanTinianov:
Refactors Durable Emitter to use Batch Emitter for publishing to Chip and includes the following changes:
- Added fallback client to retry failed events with a single event Publish.
- Extracts event store for reusability in LOOPs (tests are located in chainlink/core/services/durableemitter)
- Initializes DurableEmitter in LOOP server
- Extracts setup and adds a `globalEmitter` for accessability (similar to beholder client)

[Core Integration PR](https://github.com/smartcontractkit/chainlink/pull/22648)


pkg/durableemitter/auth.go
```diff
@@ -0,0 +1,195 @@
...
+	"maps"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/chipingress"
+)
+
+const (
+	authHeaderKey     = "X-Beholder-Node-Auth-Token"
```
DylanTinianov: Note: This file is just example code to show what we would need to do to achieve similar behaviour to beholder for injecting signer post initialization

patrickhuie19: @pkcll for some context - the rotating signer works today, but we can do better by upgrading to per message auth for ChIP as there is no passthrough. 

If, however, we believe that a streaming API is the best long term path for the durable emitter, than the rotating auth will be useful as generating a hash poses concurrency challenges.

pkcll: >upgrading to per message auth for ChIP
Signing every message adds CPU/latency on the publish hot path. Rotating signer is sufficient and gives us enough security IMHO.

pkcll:
@patrickhuie19 @DylanTinianov, I pushed this commit please review 
https://github.com/smartcontractkit/chainlink-common/pull/2090/changes/48205bc10812eae14f737b5b5c0e55ccdce1279d

pkg/durableemitter/auth.go
DylanTinianov: New commit seems simple enough for me.

patrickhuie19 (dismissed):

pkg/durableemitter/durable_emitter.go
```diff
@@ -307,58 +264,42 @@ func NewDurableEmitter(
...
-	batchWorkers := d.cfg.PublishBatchWorkers
-	if batchWorkers <= 0 {
-		d.eng.Warnw("configured batchWorkers <=0; defaulting to 1")
-		batchWorkers = 1
-	}
+// start launches the retransmit, expiry, purge, and insert-coalescing background
+// loops, then starts the batch emitter transport. It is invoked by the
+// services.Engine when the embedded Service is started.
+func (d *DurableEmitter) start(ctx context.Context) error {
+	d.batchEmitter.Start(ctx)
```
tarcisiozf, line 274: your previous comment said it should retain context beyond start, but here is being forwarded to batch emitter, is this correct?

pkg/durableemitter/durable_emitter.go
```diff
@@ -468,32 +409,113 @@ func (d *DurableEmitter) Emit(ctx context.Context, body []byte, attrKVs ...any)
...
+		publishElapsed := time.Since(t0Publish)
+
+		if h := d.cfg.Hooks; h != nil && h.OnBatchPublish != nil {
+			h.OnBatchPublish(publishElapsed, 1, sendErr)
+		}
+
+		cbCtx, cbCancel := context.WithTimeout(context.Background(), d.cfg.PublishTimeout)
+		defer cbCancel()
+
+		d.metrics.recordPublish(cbCtx, publishElapsed, "batch", sendErr)
```
tarcisiozf: check if metrics is nil

pkg/durableemitter/durable_emitter.go
DylanTinianov: ctx is not retained past start, we have a stopCh for ctx after start.

tarcisiozf (approved):

patrickhuie19 (approved):

PR #2096, opened 2026-05-27
refactor(beholder): extract newRotatingAuthFromConfig helper and add NewChipIngressHeaderProvider

pkcll:
[CRE-4422](https://smartcontract-it.atlassian.net/browse/CRE-4422)

## Summary

Introduces `pkg/chipingress.NewHeaderProvider`, a self-contained constructor for
the ChIP Ingress auth header provider. Lets non-beholder callers (e.g. the
DurableEmitter wiring in smartcontractkit/chainlink#22648) construct auth without
importing `pkg/beholder`.

## Motivation

`pkg/chipingress` is a separate Go module and cannot depend on `pkg/beholder`.
Today, `core/services/chainlink/application.go` (in smartcontractkit/chainlink#22648)
reaches into `beholder.NewStaticAuth` solely to obtain a
`chipingress.HeaderProvider`, and only the static-auth case is wired — rotating
auth is silently unsupported on the DurableEmitter path.

## Changes

- **New:** `pkg/chipingress/header_provider.go`
  - `Signer` interface (duck-typed; satisfied by `beholder.Signer`, `lazySigner`,
    `keystore.CSASigner`).
  - `HeaderProviderConfig` with fields: `AuthHeaders`, `AuthHeadersTTL`,
    `AuthPublicKeyHex`, `AuthKeySigner`, `InsecureConnection`.
  - `NewHeaderProvider` — uniformly returns nil / static / rotating provider
    based on config, mirroring `pkg/beholder/client.go:200-212` selection logic.
- `pkg/beholder/auth.go`: rename internal helper
  `newRotatingAuthFromConfig` -> `buildRotatingAuth` (single call site updated).
- **Removed:** `pkg/beholder/chipingress_auth.go` (+ its test) — replaced by the
  new package-local implementation.

## Parity Contract

The new chipingress provider preserves wire-protocol parity with beholder:

- Header name: `X-Beholder-Node-Auth-Token`
- V2 header format unchanged
- Identical validation error strings, e.g.
  `"auth: public key hex required for rotating auth (TTL > 0)"`,
  `"auth: headers TTL must be at least 10 minutes"`,
  `"auth: failed to decode public key hex"`

Documented as a doc comment on `NewHeaderProvider`. A cross-package parity test
is deferred to a follow-up PR — the chipingress module version must be bumped in
the chainlink-common root `go.mod` first so `pkg/beholder` can import the new
symbols.

## Public Surface

Exactly three exports from the new file: `Signer`, `HeaderProviderConfig`,
`NewHeaderProvider`. Static/rotating provider types are unexported.

## Testing

- `pkg/chipingress` — 18 test cases covering static, rotating, nil, transport
  security, and validation error paths.
- `pkg/beholder/...` — existing suite passes; helper rename is internal.

## Consumer

smartcontractkit/chainlink#22648 (DurableEmitter Core Integration) will adopt
this in `core/services/chainlink/application.go`, collapsing the inline auth
branch into a single `chipingress.NewHeaderProvider(...)` call.

## Follow-ups (separate PRs)

1. Bump `pkg/chipingress` version in chainlink-common root `go.mod`.
2. Add cross-package parity test:
   `pkg/beholder/chipingress_auth_parity_test.go` (table-driven, 8 cases)
   plus `pkg/beholder/export_test.go` exposing `BuildRotatingAuth`.
3. Integrate in smartcontractkit/chainlink#22648.

[CRE-4422]: https://smartcontract-it.atlassian.net/browse/CRE-4422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DylanTinianov (dismissed):

pkg/beholder/auth.go
```diff
@@ -219,3 +221,32 @@ func NewAuthHeaderV2(ctx context.Context, pubKey ed25519.PublicKey, signer Signe
 func authDialOpt(auth PerRPCCredentialsProvider) grpc.DialOption {
 	return grpc.WithPerRPCCredentials(auth.Credentials())
 }
+
+// newRotatingAuthFromConfig validates config and returns a rotating Auth and its lazySigner.
+// Returns (nil, nil, nil) if AuthHeadersTTL == 0 (no rotating auth configured).
+func newRotatingAuthFromConfig(cfg Config) (Auth, *lazySigner, error) {
```
patrickhuie19: is there a better name that RotatingAuth? if not, can we have a comment about what type of downstream consumers need this?

pkg/beholder/chipingress_auth.go
```diff
@@ -0,0 +1,24 @@
+package beholder
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/chipingress"
+)
+
+// NewChipIngressHeaderProvider creates a new chipingress.HeaderProvider
+// using the same auth logic as NewGRPCClient.
+// Returns nil provider (no error) if no auth is configured.
+func NewChipIngressHeaderProvider(cfg Config) (chipingress.HeaderProvider, error) {
```
DylanTinianov: Should this live in `pkg/chipingress`? Ideally durable emitter shouldn't need `pkg/beholder`

patrickhuie19: can this be extracted out of pkg/beholder into a chip specific pkg?

pkcll: It could, but that would require copying code (because otherwise it will create a circular dependency) and there is a chance it can diverge. 
Let me investigate how it will look like if we copy it

pkcll: Yes, I was thinking that’d be the next refactoring step, but we could probably just do it in one shot.

DylanTinianov (approved):

DylanTinianov (dismissed):

DylanTinianov (approved):

jmank88 (approved):

pkg/chipingress/header_provider.go
```diff
@@ -0,0 +1,226 @@
...
+type staticHeaderProvider struct {
+	headers    map[string]string
+	requireTLS bool
+}
+
+func (s *staticHeaderProvider) Headers(_ context.Context) (map[string]string, error) {
+	return s.headers, nil
+}
+
+func (s *staticHeaderProvider) RequireTransportSecurity() bool {
```
patrickhuie19, line 150: /nit for methods that return a bool, prefer to prefix with `Is...`, i.e. `IsTLSRequired`

pkg/chipingress/header_provider.go
```diff
@@ -0,0 +1,226 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// authHeaderKey is the header name used to carry the auth token. The value is
+// preserved verbatim from the original beholder-side implementation to keep
+// the wire protocol unchanged.
+const (
+	authHeaderKey = "X-Beholder-Node-Auth-Token"
```
patrickhuie19, line 20: /nit should this be named to something with `beholder` in it? the default for grpc authorization is `authorization`, so perhaps the difference in default behavior protects the naming as is.

pkg/chipingress/header_provider.go
```diff
@@ -0,0 +1,226 @@
...
+	returnHeader := make(map[string]string)
+	lastUpdated := time.Unix(0, r.lastUpdatedNanos.Load())
+
+	if time.Since(lastUpdated) > r.ttl {
+		r.mu.Lock()
+		defer r.mu.Unlock()
+
+		// Double-check after acquiring the lock in case another goroutine
+		// already refreshed.
+		lastUpdated = time.Unix(0, r.lastUpdatedNanos.Load())
```
patrickhuie19, line 177: /nit can leave for a follow up, but technically we could just acquire the lock and only make the top level `if time.Since(lastUpdated) > r.ttl {` check as its duplicated on line 178 to check under the lock

patrickhuie19 (approved):

hendoxc (approved):

pkg/chipingress/header_provider.go
pkcll: The intention is to keep 1:1 parity with existing implementation in pkg/beholder

pkg/chipingress/header_provider.go
pkcll: >grpc authorization is authorization

Yes, the `Authorization` header is the standard header for auth tokens in HTTP/gRPC requests.
We have tickets in place to support this. Most of the work has been completed, and only a few items remain to fully wire it up.

pkcll: > should this be named to something with beholder in it
we were going to replace it with `authorization` header at some point

pkg/chipingress/header_provider.go
pkcll: The intention of this PR is to extract auth, keep the surface area minimal, and keep it as close as possible to pkg/beholder.

PR #2097, opened 2026-05-28
[CRE-3889] Retry org resolution

agparadiso:
### Description
This [pr](https://github.com/smartcontractkit/chainlink-common/pull/1941) was already adding a single retry + cache.

What we are adding here is an increase in the scope of errors in which we retry (before it was only on `unavailable`) + adding exponential backoff with a max of 3 retries


pkg/services/orgresolver/fallback.go
```diff
@@ -25,47 +29,82 @@ type OrgResolverFallback struct {
...
+	return after(backoff)
+}
+
+func (c *OrgResolverFallback) Get(ctx context.Context, owner string) (string, error) {
+	var (
+		lastErr error
+		backoff = initialRetryBackoff
+	)
+
+	for attempt := 1; attempt <= maxGetAttempts; attempt++ {
```
pavel-raykov: Could you use "github.com/smartcontractkit/chainlink-common/pkg/utils/retry" here?

pkg/services/orgresolver/fallback.go
```diff
@@ -35,37 +62,46 @@ func NewOrgResolverWithFallback(inner OrgResolver, logger log.Logger) *OrgResolv
...
+}
 
-	return "", err
+func isRetriableGRPCCode(code codes.Code) bool {
+	switch code {
+	case codes.Unavailable, codes.DeadlineExceeded, codes.ResourceExhausted, codes.Aborted, codes.Unknown:
+		return true
+	default:
+		return false
+	}
```
pavel-raykov:
```suggestion
	return slices.Contains([]codes.Code{codes.Unavailable, codes.DeadlineExceeded, codes.ResourceExhausted, codes.Aborted, codes.Unknown},
		code)
```

pavel-raykov (dismissed):

pkg/services/orgresolver/fallback.go
pavel-raykov: feel free to also use a one-liner `Contains` here.

pavel-raykov (approved):

george-dorin (approved):

PR #2099, opened 2026-05-28
pkg/services/orgresolver: track pass/fail metrics

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-3887
Support passing a `metric.Meter` to the org resolver to report pass/fail metrics.

### DIFF
--- a/pkg/services/orgresolver/linking.go
+++ b/pkg/services/orgresolver/linking.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
@@ -33,6 +34,9 @@ type Config struct {
 	WorkflowRegistryAddress       string
 	WorkflowRegistryChainSelector uint64
 	JWTGenerator                  JWTGenerator
+
+	Client linkingclient.LinkingServiceClient // optional
+	Meter  metric.Meter                       // optional
 }
 
 // orgResolver makes direct calls to the linking service to resolve organization IDs from workflow owners.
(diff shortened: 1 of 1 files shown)


pkg/services/orgresolver/linking.go
```diff
@@ -119,9 +146,15 @@ func (o *orgResolver) Get(ctx context.Context, owner string) (string, error) {
 
 	resp, err := o.client.GetOrganizationFromWorkflowOwner(ctx, req)
 	if err != nil {
+		if o.failCount != nil {
+			o.failCount.Add(ctx, 1) //TODO owner attribute?
```
jmank88: Are owner attributes helpful? Or just noise with extra cardinality?

pkg/services/orgresolver/linking.go
```diff
@@ -45,24 +49,34 @@ type orgResolver struct {
...
 }
 
 // NewOrgResolverWithClient creates a new org resolver with an optional injected client (for testing)
+// Deprecated: Use Config.New
 func NewOrgResolverWithClient(cfg Config, client linkingclient.LinkingServiceClient, logger log.Logger) (*orgResolver, error) {
+	cfg.Client = client
+	return cfg.New(logger)
+}
+
+func (cfg *Config) New(logger log.Logger) (*orgResolver, error) {
```
patrickhuie19, line 70: Why the indirection with exporting `New` over `*Config` vs just calling into the exported `NewOrgResolverWithClient`? The old pattern better supported testing

pkg/services/orgresolver/linking.go
```diff
@@ -84,6 +98,19 @@ func NewOrgResolverWithClient(cfg Config, client linkingclient.LinkingServiceCli
 		resolver.client = linkingclient.NewLinkingServiceClient(conn)
 	}
 
+	if cfg.Meter != nil {
+		var err error
+		//TODO names
+		resolver.passCount, err = cfg.Meter.Int64Counter("foo.bar.TODO.pass")
```
patrickhuie19: i think just the straight name here is fine, even without dot notations to give a path. i.e. "org_resolver_success" and "org_resolver_fail"

pkg/services/orgresolver/linking.go
patrickhuie19: good q, they are helpful. we generally pass a labeler around in the engine to have KV attributes we want. The potential problem is that we can't have high cardinality values, for instance `WorkflowExecutionID`. So to give clients the option, we could define a map on Config that is optional for clients creating an org resolver.

pkg/services/orgresolver/linking.go
jmank88: I added a new field `Meter metric.Meter`. The `XWithY` suffix pattern doesn't scale. Why is testing more difficult?

pkg/services/orgresolver/linking.go
jmank88: But do we actually expect the information to be relevant? I wouldn't think problems at this level would be correlated with particular owners. (and even if they were, that would imply the requests are being received by the server, who should be reporting better information than we can as clients?)

pkg/services/orgresolver/linking.go
patrickhuie19: I don't seen a need for it to scale. For instance, why would we ever have metrics nil or disabled?

patrickhuie19 (approved):

pkg/services/orgresolver/linking.go
jmank88: nil/disabled metrics are the implicit default that has to be supported to remain compatible though.

pavel-raykov (approved):

PR #2100, opened 2026-05-29
limit error marker interface

ettec:
To make it easy to identify when the underlying cause of an error is due to a limit breach.

### DIFF
--- a/pkg/settings/limits/errors.go
+++ b/pkg/settings/limits/errors.go
@@ -11,6 +11,16 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/settings"
 )
 
+// LimitError is implemented by errors returned when a limit is exceeded.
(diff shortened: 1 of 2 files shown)


george-dorin (approved):

pkg/settings/limits/errors.go
```diff
@@ -11,6 +11,16 @@ import (
...
 )
 
+// LimitError is implemented by errors returned when a limit is exceeded.
+// Use [errors.As] to identify limit errors, for example:
+//
+//	var limitErr LimitError
+//	if errors.As(err, &limitErr) { ... }
+type LimitError interface {
+	error
+	limitError()
```
pavel-raykov, line 21: why do you need this method?

pavel-raykov: oh, I see it is a marker one

pavel-raykov (approved):

PR #2105, opened 2026-05-29
Add configurable min identical responses to aggregate

amit-momin:


pkg/capabilities/pb/registry.proto
```diff
@@ -52,6 +52,12 @@ message RemoteExecutableConfig {
   google.protobuf.Duration request_timeout = 6;
   uint32 server_max_parallel_requests = 7;
   RequestHasherType request_hasher_type = 8;
+
+  // MinResponsesToAggregate is the minimum number of chain-capability DON nodes that must
```
bolekk: nit: not only chain capability DON. Any capability DON

pkg/capabilities/capabilities.go
```diff
@@ -664,6 +664,12 @@ type RemoteExecutableConfig struct {
 	RequestTimeout            time.Duration
 	ServerMaxParallelRequests uint32
 	RequestHasherType         RequestHasherType
+
+	// MinResponsesToAggregate is the minimum number of chain-capability DON nodes that must
```
bolekk: same as above

pkg/capabilities/v2/chain-capabilities/consensus/ocr3/types/ocr3_chain_capabilities_config_types.proto
```diff
@@ -14,4 +14,9 @@ message ReportingPluginConfig {
   uint32 maxReportCount = 5;
 
   uint32 maxBatchSize = 6;
+
+  // MinResponsesToAggregate is the minimum number of nodes that must report
+  // identical observations for a value to be accepted as the read result.
+  // 0 defaults to F+1. Must satisfy F+1 <= MinResponsesToAggregate <= N when non-zero.
+  uint32 minResponsesToAggregate = 7;
```
bolekk, line 21: Why do we need it for OCR? Is this config for the ReadChain plugin?

amit-momin: Think the capabilities [PR](https://github.com/smartcontractkit/capabilities/pull/626/changes#diff-47c4c0be609c1c745c68c58d46f85015f10e29e907da6e550fc7e098960dbbe7) may help add some color here. It's used in the Reporting Plugin to overwrite the hardcoded F+1 with a custom value so we can increase this for specific DONs

pkg/capabilities/pb/registry.proto
amit-momin: Fixed in the latest commit

Unheilbar (approved):

Krish-vemula (approved):

silaslenihan (approved):

jmank88 (approved):

ilija42 (approved):

PR #2121, opened 2026-06-03
(feat): Enable per-owner overrides for the max ciphertext size

justinkaseman:
Before:
- `global.VaultCiphertextSizeLimit`

After:
- `owner.<addr>.PerOwner.VaultCiphertextSizeLimit` = per-owner override
- `global.PerOwner.VaultCiphertextSizeLimit` = global override (applies to all owners)

**Backwards compatibility:**
1. All nodes use the hardcoded 2KB default.
2. If any nodes have the old key path via `CL_CRE_SETTINGS`, it stops being found so falls back to 2KB default

For OCR safety: only CL_CRE_SETTINGS needs to be checked for the old key, and configuring the new override should only be done after all nodes of the DON have upgraded.


pkg/settings/cresettings/settings.go
```diff
@@ -78,7 +78,6 @@ var Default = Schema{
 	// used by the Vault OCR plugin -- changing these values could cause issues with the plugin during an image
 	// upgrade as nodes apply the old and new values inconsistently. A safe upgrade path
 	// must ensure that we are overriding the default in the onchain configuration for the contract.
-	VaultCiphertextSizeLimit:          Size(2 * config.KByte),
```
cedric-cordenier, line 81: I would avoid a breaking change here; add the new PerOwner field, mark this as deprecated and then go and update the places where it's used

I know it's a bit more work (3 PRs rather than 2, if you choose to cleanup the field) but I think it'll make things less painful for others as we're working the change through core

cedric-cordenier (dismissed):

cedric-cordenier (dismissed):

pkg/settings/cresettings/README.md
```diff
@@ -222,7 +222,6 @@ flowchart
         end
     end
     subgraph vault
-        VaultCiphertextSizeLimit{{VaultCiphertextSizeLimit}}:::bound
```
cedric-cordenier, line 225: @justinkaseman You're missing this I think

cedric-cordenier (approved):

russell-stern (approved):

PR #2126, opened 2026-06-04
Plex 2406/pipe get program accounts

Unheilbar:
implementation:https://github.com/smartcontractkit/chainlink-solana/pull/1564


pkg/types/chains/solana/solana.go
```diff
@@ -110,6 +110,11 @@ type Client interface {
 	// In: ctx, {Receiver, EncodedTransaction, Opts(SigVerify, Commitment, ReplaceRecentBlockhash, Accounts)}.
 	// Out: {Err, Logs, Accounts, UnitsConsumed}, error.
 	SimulateTX(ctx context.Context, req SimulateTXRequest) (*SimulateTXReply, error)
+
+	// GetProgramAccounts: all accounts owned by program.
+	// In: ctx, {Program, Opts(Encoding, Commitment, DataSlice, Filters), IsExternal}.
+	// Out: {Value([]*KeyedAccount)}, error.
+	GetProgramAccounts(ctx context.Context, req GetProgramAccountsRequest) (*GetProgramAccountsReply, error)
```
jmank88, line 117: Do we have an `UnimplementedClient`? If we do/did and we added an unexported method, then adding methods here in the future won't be a breaking change. For example, the GRPC style is like:
```suggestion
	GetProgramAccounts(ctx context.Context, req GetProgramAccountsRequest) (*GetProgramAccountsReply, error)
	
	mustEmbedUnimplementedClient()
```

Unheilbar: We do, its added:
```go
func (uss *UnimplementedSolanaService) GetProgramAccounts(ctx context.Context, req solana.GetProgramAccountsRequest) (*solana.GetProgramAccountsReply, error) {
	return nil, status.Errorf(codes.Unimplemented, "method GetProgramAccounts not implemented")
}
```
so it will still compile, just method consumers will get gRPC-style unimplemented error
or you mean something else?

jmank88: Without the unexported method, every change to this interface is still breaking, as indicated by API Diff: https://github.com/smartcontractkit/chainlink-common/pull/2126#issuecomment-4622626216

pkg/chains/solana/proto_helpers.go
```diff
@@ -1067,6 +1067,171 @@ func ConvertGetMultipleAccountsReplyToProto(r *solana.GetMultipleAccountsReply)
...
+	return out
+}
+
+func ConvertRPCFiltersToProto(filters []solana.RPCFilter) []*RPCFilter {
+	if len(filters) == 0 {
+		return nil
+	}
+	out := make([]*RPCFilter, 0, len(filters))
+	for _, f := range filters {
+		if pf := ConvertRPCFilterToProto(f); pf != nil {
```
dhaidashenko: I'm worried about skipping nil pf. Should we return an error if the user passed an invalid RPCFilter?
This could help identify odd bugs caused by user typo/invalid args

dhaidashenko (approved):

jmank88 (approved):

PR #2129, opened 2026-06-05
Remind user that cre_settings changes might involve re-proposing job-specs

yashnevatia:

### DIFF
--- a/.github/workflows/cre-settings-schema-reminder.yml
+++ b/.github/workflows/cre-settings-schema-reminder.yml
@@ -39,9 +39,7 @@ jobs:
 
             This PR adds `// Deprecated` in [`pkg/settings/cresettings/settings.go`](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/settings/cresettings/settings.go).
 
-            When making the corresponding **capabilities** change (where this settings field is consumed), keep the change **backward compatible** — the new code should still respect the old field.
+			If the deprecated field is being replaced by a new one, then a smooth deployment migration requires:
 
-            **Rollout:**
-            1. Ship the capability change that reads both old and new fields.
-            2. After the release is live, propose job specs in [chainlink-deployments](https://github.com/smartcontractkit/chainlink-deployments) using the new schema (check `domains/cre/*/settings/*.toml` for existing overrides on the deprecated field).
-            3. In a **subsequent** release, remove the old field from the capability code.
+			1. Before deploying the code, any overridden values on the old field must be duplicated to the new field.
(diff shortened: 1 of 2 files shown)


.github/workflows/cre-settings-schema-reminder.yml
jmank88, line 1: Updating the default values here in common is not related to the deployed overrides set anywhere else. Defaults take effect at deploy time if no override is present, and if there is an override then it continues to be in effect no matter how the default changes with the common version.

jmank88 (changes requested):

.github/workflows/cre-settings-schema-reminder.yml
```diff
@@ -0,0 +1,47 @@
...
+            ### CRE settings field deprecated
+
+            This PR adds `// Deprecated` in [`pkg/settings/cresettings/settings.go`](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/settings/cresettings/settings.go).
+
+            When making the corresponding **capabilities** change (where this settings field is consumed), keep the change **backward compatible** — the new code should still respect the old field.
+
+            **Rollout:**
+            1. Ship the capability change that reads both old and new fields.
+            2. After the release is live, propose job specs in [chainlink-deployments](https://github.com/smartcontractkit/chainlink-deployments) using the new schema (check `domains/cre/*/settings/*.toml` for existing overrides on the deprecated field).
+            3. In a **subsequent** release, remove the old field from the capability code.
```
jmank88: Let's keep this as general as possible, and it is still only relevant if migrating to a new field. How about something like:
```suggestion
			If the deprecated field is being replaced by a new one, then a smooth deployment migration requires:

			1. Before deploying the code, any overridden values on the old field must be duplicated to the new field.
			2. Old field overrides must remain in-place and kept in-sync until all nodes are migrated.
```

pkg/settings/cresettings/settings.go
```diff
@@ -212,7 +212,7 @@ var Default = Schema{
 				GasLimit:        PerChainSelector(Uint32(300_000), map[string]uint32{}),
 			},
 			Aptos: aptosChainWrite{
-				ReportSizeLimit: Size(5 * config.KByte),
+				ReportSizeLimit: Size(5 * config.KByte), // Deprecated
```
jmank88: We document the field definitions instead, but this is just a test, right?

yashnevatia: yes was just a test

jmank88 (approved):

ilija42 (approved):

PR #2131, opened 2026-06-05
feat(cre): extends gateway connector to support multiple DONs

MStreet3:
* extends GatewayConnector interface via the MultiGatewayConnector
* updates proto definitions for loop use
* design [doc](https://docs.google.com/document/d/124QL5bCWB9Saq9y3L1mWjuTQqUun1nVoV7pZaRPjhxk/edit?tab=t.0#heading=h.t0ctqojd7qkq)


jmank88 (dismissed):

justinkaseman (dismissed):

justinkaseman (dismissed):

jmank88 (dismissed):

jmank88 (approved):

justinkaseman (approved):

pkg/loop/internal/core/services/gateway/gateway_connector.go
```diff
@@ -14,7 +14,10 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/types/core"
 )
 
-var _ core.GatewayConnector = (*GatewayConnectorClient)(nil)
+var (
+	_ core.GatewayConnector         = (*GatewayConnectorClient)(nil)
+	_ core.MultiGatewayConnector = (*GatewayConnectorClient)(nil)
```
cedric-cordenier, line 19: FWIW it's possible API diff was wrong here, it probably just doesn't understand the Unimplemented pattern we added

If the build core job passed then it would have compiled correctly

PR #2134, opened 2026-06-08
DurableEmitter: Instrument Callsites

DylanTinianov:
Add emit callsites

### DIFF
--- a/pkg/durableemitter/setup.go
+++ b/pkg/durableemitter/setup.go
@@ -17,6 +17,11 @@ import (
 // globalEmitter holds the process-wide DurableEmitter instance, set by Setup.
 var globalEmitter atomic.Pointer[DurableEmitter]
 
+var (
+	ErrNotInitialized = errors.New("durable emitter not initialized")
+	ErrEmitFailed     = errors.New("durable emitter emit failed")
+)
+
 // SetGlobalEmitter sets the global DurableEmitter.
 func SetGlobalEmitter(d *DurableEmitter) {
 	globalEmitter.Store(d)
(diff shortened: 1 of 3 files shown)


tarcisiozf (dismissed):

pkg/durableemitter/setup.go
```diff
@@ -28,11 +28,12 @@ func GetGlobalEmitter() *DurableEmitter {
...
 
 // GlobalEmit emits an event via the global DurableEmitter.
-// Returns a non-nil error when the global emitter has not been initialized.
+// Returns nil if the global emitter has not been initialized, so callers can
+// treat a non-nil error as a real emission failure rather than "disabled".
 func GlobalEmit(ctx context.Context, body []byte, attrKVs ...any) error {
 	d := globalEmitter.Load()
 	if d == nil {
-		return errors.New("global DurableEmitter not initialized; call durableemitter.Setup first")
+		return nil
```
patrickhuie19: this coalesces a negative path with a happy path which is not ideal. IMHO a better path forward is to either keep the error path as is, or define explicit error types that a consumer can switch case on.

DylanTinianov: Lets go with the explicit error path

pkg/loop/plugin_relayer_emitter.go
```diff
@@ -111,6 +113,18 @@ func (e *pluginRelayerConfigEmitter) emit(ctx context.Context) {
 		return
 	}
 
+	err = durableemitter.GlobalEmit(ctx, payloadBytes, "source", beholderDomain, "type", beholderEntity)
```
patrickhuie19, line 116: we can instrument here, but these aren't events that we necessarily need to be durable

pkg/workflows/events/trigger_events.go
```diff
@@ -88,8 +90,17 @@ func EmitTriggerExecutionStarted(ctx context.Context, labeler custmsg.MessageEmi
...
+	const entity = "workflows.v2.TriggerExecutionStarted"
+	if err := beholder.GetEmitter().Emit(ctx, b,
 		"beholder_data_schema", "workflows.v2.trigger_execution_started", // required
 		"beholder_domain", "platform", // required
-		"beholder_entity", "workflows.v2.TriggerExecutionStarted") // required
+		"beholder_entity", entity); err != nil { // required
+		return err
+	}
+
+	err = durableemitter.GlobalEmit(ctx, b, "source", "platform", "type", entity)
```
patrickhuie19, line 101: same with here, the actual TriggerExecutionStarted emits happen in the capabilities repo, so doing this won't help us view any events on staging. IFIUC this is a legacy code path (maybe invoked in testing)

patrickhuie19 (approved):

tarcisiozf (approved):

PR #2137, opened 2026-06-09
[CRE-4317] Implement genericless adapter for ocr3types.OnchainKeyring2

pavel-raykov:
Supports https://github.com/smartcontractkit/chainlink-evm/pull/469


keystore/ocr3/utils.go
```diff
@@ -0,0 +1,44 @@
...
+	Has(publicKey ocrtypes.OnchainPublicKey) bool
+	MaxSignatureLength() int
+	DebugIdentifier() string
+}
+
+var _ ocr3types.OnchainKeyring2[struct{}] = &OnchainKeyring2ToGenericAdapter[struct{}]{}
+
+type OnchainKeyring2ToGenericAdapter[RI any] struct {
+	K OnchainKeyring2Genericless
+}
```
jmank88: How about adding a constructor to avoid the name collision so we can drop the suffixes?
```suggestion
type OnchainKeyring2 interface {
	Sign(configDigest ocrtypes.ConfigDigest, seqNr uint64, report types.Report) (signature []byte, err error)
	Verify(publicKey ocrtypes.OnchainPublicKey, configDigest ocrtypes.ConfigDigest, seqNr uint64, report types.Report, signature []byte) bool
	Has(publicKey ocrtypes.OnchainPublicKey) bool
	MaxSignatureLength() int
	DebugIdentifier() string
}

var _ ocr3types.OnchainKeyring2[struct{}] = &onchainKeyring2[struct{}]{}

type onchainKeyring2[RI any] struct {
	k OnchainKeyring2
}

func AsOCR3OnchainKeyring2[RI any](k OnchainKeyring2) ocr3types.OnchainKeyring2[RI]
```

pavel-raykov: Done.

keystore/ocr3/utils.go
```diff
@@ -0,0 +1,48 @@
+package ocr3
```
jmank88: Should this be named more specifically to avoid collisions?
Is this a general ocr3 package?
```suggestion
package ocr3util
```
Or a keys only package?
```suggestion
package ocr3keys
```

keystore/ocr3/utils.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"github.com/smartcontractkit/libocr/offchainreporting2/types"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+	ocrtypes "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+)
+
+// OnchainKeyring2Genericless is a genericless counterpart of ocr3types.OnchainKeyring2. If generic RI does not matter,
+// then it is more convenient to implement this interface and then use OnchainKeyring2ToGenericAdapter to use it as
```
jmank88:
```suggestion
// OnchainKeyring2 is a genericless counterpart of ocr3types.OnchainKeyring2. If generic RI does not matter,
// then it is more convenient to implement this interface and then use AsOCR3OnchainKeyring2 to use it as
```

keystore/ocr3/utils.go
pavel-raykov: I think util seems like a better match here.

keystore/ocr3util/utils.go
```diff
@@ -0,0 +1,48 @@
+package ocr3util
```
jmank88, line 1: Let's align the directory name too?

pavel-raykov: Done.

jmank88 (dismissed):

keystore/ocr3util/utils.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (o *onchainKeyring2[RI]) MaxSignatureLength() int {
+	return o.k.MaxSignatureLength()
+}
+
+func (o *onchainKeyring2[RI]) DebugIdentifier() string {
+	return o.k.DebugIdentifier()
+}
+
+func AsOCR3OnchainKeyring2[RI any](k OnchainKeyring2) ocr3types.OnchainKeyring2[RI] {
```
kaleofduty: As per discussion with @pavel-raykov , we want to make slight adjustments to make it more misuse reistant:
```
// Be careful! If you use this shim, any information in RI is dropped and *not* signed over. If you have security-critical information in RI, you should make your own ocr3types.OnchainKeyring2 that does use the information in RI for signing.
```
Rename to 
`AsOCR3OnchainKeyring2IgnoringRI`

pavel-raykov: Done.

kaleofduty (approved):

george-dorin (approved):

PR #2138, opened 2026-06-09
SDK Standard Test for capability errors

ettec:
The standard test defined here is used to check capability error handling, this test is also used in the SDK to ensure that new error codes stay in sync.

### DIFF
--- a/pkg/capabilities/errors/error_codes.go
+++ b/pkg/capabilities/errors/error_codes.go
@@ -135,6 +135,29 @@ const (
 	InsufficientObservations ErrorCode = 102
 )
 
+// AllErrorCodes lists every defined capability error code in ascending order.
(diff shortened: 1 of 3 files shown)


pkg/capabilities/errors/error_codes.go
```diff
@@ -135,6 +135,29 @@ const (
 	InsufficientObservations ErrorCode = 102
 )
 
+// AllErrorCodes lists every defined capability error code in ascending order.
+var AllErrorCodes = []ErrorCode{
+	Canceled,
```
ettec, line 140: Todo in a different PR: Should look at generating the error_codes class to ensure all the parts stay in sync

russell-stern (approved):

george-dorin (approved):

agparadiso (approved):

PR #2147, opened 2026-06-11
Add Error field to Stellar SubmitTransactionResponse

Krish-vemula:
## Summary

- Add `Error string` to `SubmitTransactionResponse` in the Stellar domain types, mirroring `ReadContractResponse.Error`
- Extend the `SubmitTransactionResponse` gRPC message with an `error` field
- Wire the field through `ConvertSubmitTransactionResponseToProto` / `ConvertSubmitTransactionResponseFromProto`
- Add a proto round-trip test for failed on-chain submissions with a non-empty error

## Motivation

`SubmitTransaction` needs to distinguish a successful API call with an on-chain failure (`TxFailed`) from an infrastructure error (Go `error`). After that change, callers inspect `TxStatus` and need a human-readable revert reason without decoding `ResultXDR`.

`ReadContract` already returns simulation failures via `response.Error` with a nil Go error. This PR brings `SubmitTransaction` to the same pattern for on-chain reverts.

## Changes

- `pkg/types/chains/stellar/stellar.go` — add `Error` to `SubmitTransactionResponse`
- `pkg/chains/stellar/stellar.proto` — add `error = 6` to `SubmitTransactionResponse`
- `pkg/chains/stellar/stellar.pb.go` — regenerated
- `pkg/chains/stellar/proto_helpers.go` — map `Error` in both conversion directions
- `pkg/chains/stellar/proto_helpers_test.go` — add `TestConvertSubmitTransactionResponse_RoundTrip_WithError`


silaslenihan (dismissed):

pkg/chains/stellar/proto_helpers_test.go
```diff
@@ -395,6 +395,7 @@ func TestConvertSubmitTransactionResponse_RoundTrip(t *testing.T) {
 		TxIdempotencyKey: "idem-456",
 		ResultXDR:        base64.StdEncoding.EncodeToString([]byte("result")),
 		ResultMetaXDR:    base64.StdEncoding.EncodeToString([]byte("meta")),
+		Error:            "",
```
ilija42, line 398: shouldn't this test have all fields filled out to test the converters properly

Krish-vemula: For `TxSuccess`, `Error` is intentionally empty the type documents it as only set when the transaction was accepted but failed on-chain. The success round-trip already exercises the other fields (`ResultXDR`, `ResultMetaXDR`, etc.), and I added an explicit `require.Empty(t, proto.GetError())` to make that clear now. Non-empty `Error` is covered in `TestConvertSubmitTransactionResponse_RoundTrip_WithError`.

ilija42 (approved):

Unheilbar (approved):

PR #2151, opened 2026-06-12
DurableEmitter: Seed pending count on start

DylanTinianov:
Also reduce event TTL default to 1hr.

### DIFF
--- a/pkg/durableemitter/durable_emitter.go
+++ b/pkg/durableemitter/durable_emitter.go
@@ -279,6 +279,10 @@ func NewDurableEmitter(
 func (d *DurableEmitter) start(ctx context.Context) error {
 	d.batchEmitter.Start(ctx)
 
+	// Seed pendingCount from the on-disk queue depth so the gauge survives
+	// process restarts.
+	d.seedPendingFromStore(ctx)
+
 	insertWorkers := d.cfg.InsertBatchWorkers
 	if insertWorkers <= 0 {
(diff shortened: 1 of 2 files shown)


pkg/durableemitter/durable_emitter.go
```diff
@@ -302,6 +306,32 @@ func (d *DurableEmitter) start(ctx context.Context) error {
...
+		return
+	}
+	if st.Depth <= 0 {
+		return
+	}
+	d.pendingCount.Store(st.Depth)
+	d.pendingMax.Store(st.Depth)
+	if d.metrics != nil {
+		d.metrics.queueDepth.Record(ctx, st.Depth)
+		d.metrics.queueDepthMax.Record(ctx, st.Depth)
```
patrickhuie19, line 330: Can you help remind me - why did we want to record queue depth max as an additional metric? Is there a reason we shouldn't be calculating it with sql or promql expressions on our dashboards?

patrickhuie19 (dismissed):

patrickhuie19 (approved):

pkg/durableemitter/durable_emitter.go
DylanTinianov: Agreed this should probably just be done on Grafana. It's just leftover from original testing. I can make a follow-up PR to remove it.

tarcisiozf (approved):

PR #2158, opened 2026-06-15
Resource Managers: durable metering for the CRE and other products

patrickhuie19:
Resource Managers record and measure utilization of long lived resources that drawdown from unified account balances. i.e. trigger subscriptions drawing down from a CRE account balance.

### Requires
https://github.com/smartcontractkit/chainlink-protos/pull/400/commits + https://github.com/smartcontractkit/chainlink-protos/pull/419 (merged)

### Supports
https://github.com/smartcontractkit/chainlink/pull/22845


pkg/loop/config.go
```diff
@@ -85,6 +85,12 @@ const (
 	envTelemetryPrometheusBridgeEnabled   = "CL_TELEMETRY_PROMETHEUS_BRIDGE_ENABLED"
 	envTelemetryPrometheusBridgePrefixes  = "CL_TELEMETRY_PROMETHEUS_BRIDGE_PREFIXES"
 	envTelemetryLogCompressor             = "CL_TELEMETRY_LOG_COMPRESSOR"
+	envMeterRecordsEnabled                = "CL_METER_RECORDS_ENABLED"
+	envMeterSnapshotsEnabled              = "CL_METER_SNAPSHOTS_ENABLED"
+	envMeteringProduct                    = "CL_METERING_PRODUCT"
+	envMeteringEnvironment                = "CL_METERING_ENVIRONMENT"
+	envMeteringZone                       = "CL_METERING_ZONE"
+	envMeteringNodeID                     = "CL_METERING_NODE_ID"
```
patrickhuie19, line 96: note to reader, DonID isn't here b/c a node can belong to many DONs; it is the responsibility of the loop to properly receive its DonID from `Initialise` in the standard cap interface (or for non-CRE use cases, elsewhere)

pkg/durableemitter/durable_emitter.go
```diff
@@ -19,6 +19,12 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
```
patrickhuie19, line 22: note to reader: changes made here instead of pkg/beholder as we are trying to leave beholder as is without accruing much more complexity, and durableemitter pkg is completely owned by foundations, so is a cleaner division of responsibility

cc @pkcll

pkg/loop/config.go
```diff
@@ -85,6 +85,13 @@ const (
 	envTelemetryPrometheusBridgeEnabled   = "CL_TELEMETRY_PROMETHEUS_BRIDGE_ENABLED"
 	envTelemetryPrometheusBridgePrefixes  = "CL_TELEMETRY_PROMETHEUS_BRIDGE_PREFIXES"
 	envTelemetryLogCompressor             = "CL_TELEMETRY_LOG_COMPRESSOR"
+	envMeterRecordsEnabled                = "CL_METER_RECORDS_ENABLED"
+	envMeterSnapshotsEnabled              = "CL_METER_SNAPSHOTS_ENABLED"
+	envMeteringProduct                    = "CL_METERING_PRODUCT"
```
jmank88: Is the difference of `meter` vs `metering` signifiant?

pkg/resourcemanager/utilization.go
```diff
@@ -0,0 +1,47 @@
...
+
+// NewUtilization builds a Utilization with int64 quantity encoded as a decimal
+// string value.
+func NewUtilization(value int64, fields UtilizationFields) *meteringpb.Utilization {
+	return NewUtilizationString(strconv.FormatInt(value, 10), fields)
+}
+
+// NewUtilizationString builds a Utilization from a pre-formatted numeric string
+// value.
+func NewUtilizationString(value string, fields UtilizationFields) *meteringpb.Utilization {
```
jmank88: Does this one deserve the plain constructor name, since it is accepting the raw string?

pkg/loop/config.go
patrickhuie19: no, its a smell. i'll cleanup

pkg/resourcemanager/utilization.go
patrickhuie19: i.e. you are saying it should be called `NewUtilization`?

jmank88: Yeah with NewUtilizationInt or similar for the other one

pkg/resourcemanager/meterable.go
```diff
@@ -0,0 +1,145 @@
...
+	Utilizations []*meteringpb.Utilization
+}
+
+// DeploymentIdentity carries the static deployment + node identity dimensions
+// that are fixed for a LOOP plugin process. They are resolved once from node
+// config by the host and delivered to every LOOP plugin over the environment
+// (loop.EnvConfig), not the standard-capabilities boundary, so any LOOP plugin
+// can populate the coarse metering rollup dimensions. Any field may be empty if
+// the host did not provide it.
+type DeploymentIdentity struct {
```
justinkaseman, line 51: Where does this get used? Are the env vars meant to plug into this struct?

patrickhuie19: yes, the env vars plug into, passed over to the loop via env. They should then be set on the ResourceManagers config so that they can be sourced into the emission of each event.

pkg/resourcemanager/utilization.go
```diff
@@ -0,0 +1,47 @@
...
+func NewUtilizationBig(value *big.Int, fields UtilizationFields) *meteringpb.Utilization {
+	if value == nil {
+		return NewUtilizationString("0", fields)
+	}
+	return NewUtilizationString(value.String(), fields)
+}
+
+// NewUtilizationFloat builds a Utilization from a floating-point value.
+func NewUtilizationFloat(value float64, fields UtilizationFields) *meteringpb.Utilization {
+	return NewUtilizationString(strconv.FormatFloat(value, 'g', -1, 64), fields)
```
justinkaseman: This returns floats formatted into scientific notation strings. Is that how we kept them before?

May be worth documenting for consumers.

patrickhuie19: good catch, it should be standard decimal notation.

pkg/resourcemanager/meterable.go
```diff
@@ -0,0 +1,149 @@
...
+	Zone string
+	// NodeID is the node's logical name, e.g. "clp-cre-wf-zone-a-1". It is NOT
+	// the CSA public key; it is a stable name the billing service can use to
+	// look up the node's CSA key in the workflow registry. The CSA key itself is
+	// carried separately as the node_csa_key event attribute.
+	NodeID string
+}
+
+// DonIdentifier captures DON-specific identity dimensions as one unit.
+type DonIdentifier struct {
```
jmank88: What is the difference between `Identifier`, `Identity`, and `ID`?

patrickhuie19: Aligned offline: ...ID = one scalar string
...Identity = a struct grouping multiple dimensions

fix starting in protos: https://github.com/smartcontractkit/chainlink-protos/pull/420

patrickhuie19: changed: https://github.com/smartcontractkit/chainlink-common/pull/2158/commits/6494cc6a8142247de97df16ec16f64b1615acb8c

jmank88 (dismissed):

jmank88 (approved):

DylanTinianov (approved):

PR #2161, opened 2026-06-16
DurableEmitter: Drop non retryable events

DylanTinianov:

### DIFF
--- a/pkg/durableemitter/non_retryable.go
+++ b/pkg/durableemitter/non_retryable.go
@@ -1,0 +1,64 @@
+package durableemitter
+
+import (
+	"strings"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/chipingress"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/durableemitter/non_retryable.go
```diff
@@ -0,0 +1,64 @@
...
+		"eventID", eventPb.GetId(),
+		"source", eventPb.GetSource(),
+		"type", eventPb.GetType(),
+		"reason", reason,
+	)
+
+	ctx, cancel := d.stopCh.CtxWithTimeout(d.cfg.PublishTimeout)
+	defer cancel()
+
+	if err := d.store.Delete(ctx, id); err != nil {
```
patrickhuie19, line 55: can do in a followup, but I think as is, this will be inefficient as we are going to collect many many non retryable events at a time

patrickhuie19 (approved):

patrickhuie19:
@DylanTinianov can we close this?

PR #2165, opened 2026-06-17
pkg/settings/cresettings: add ConfidentialWorkflows.Enabled feature gate

nadahalli:
Adds `PerWorkflow.ConfidentialWorkflows.Enabled` (Setting[bool], default true) to the CRE settings schema.

This gives the confidential-workflows capability a feature flag that can be flipped in production through the scoped settings registry, with overrides resolvable at workflow, owner, org, or global level. The consuming GateLimiter lands separately in confidential-compute.

Regenerated defaults.toml / defaults.json and added the key to the README flowchart (both required by the package tests).

### DIFF
--- a/pkg/settings/cresettings/defaults.json
+++ b/pkg/settings/cresettings/defaults.json
@@ -151,6 +151,9 @@
 			"RequestSizeLimit": "10kb",
 			"ResponseSizeLimit": "100kb"
 		},
+		"ConfidentialWorkflows": {
+			"Enabled": "true"
+		},
 		"Secrets": {
 			"CallLimit": "5"
 		},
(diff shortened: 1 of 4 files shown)


pkg/settings/cresettings/defaults.json
```diff
@@ -151,6 +151,9 @@
 			"RequestSizeLimit": "10kb",
 			"ResponseSizeLimit": "100kb"
 		},
+		"ConfidentialWorkflows": {
+			"Enabled": "true"
```
prashantkumar1982: Shouldn't the default be set to false?

nadahalli: Yes. https://github.com/smartcontractkit/chainlink-common/pull/2165/changes/743b7da6320baf7083d2ff346b65f96abfe64f87

prashantkumar1982 (dismissed):

vreff (dismissed):

prashantkumar1982 (approved):

bolekk (approved):

PR #2168, opened 2026-06-17
DurableEmitter: Fix queue depth bug

DylanTinianov:
Leave delivered events to be purged by purge loop.

### DIFF
--- a/pkg/durableemitter/store.go
+++ b/pkg/durableemitter/store.go
@@ -149,7 +149,8 @@ func (s *PgDurableEventStore) DeleteExpired(ctx context.Context, ttl time.Durati
 	const q = `
 WITH deleted AS (
     DELETE FROM ` + chipDurableEventsTable + `
-    WHERE created_at <= now() - $1::interval
+    WHERE delivered_at IS NULL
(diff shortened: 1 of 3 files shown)


pkg/durableemitter/store.go
```diff
@@ -149,7 +149,8 @@ func (s *PgDurableEventStore) DeleteExpired(ctx context.Context, ttl time.Durati
 	const q = `
 WITH deleted AS (
     DELETE FROM ` + chipDurableEventsTable + `
-    WHERE created_at <= now() - $1::interval
+    WHERE delivered_at IS NULL
```
patrickhuie19, line 152: For future self:

msg delivered_at created_at
"bob" 14:00 13:59
"uncle" 12:05 12:04
"barb" nil 14:01

TTL is h1r, and its 16:00
loop runs, deletes all 3.
Queue depth should go down by 1. in old version went down by 3

patrickhuie19 (dismissed):

patrickhuie19 (dismissed):

tarcisiozf (approved):

patrickhuie19 (approved):

PR #2170, opened 2026-06-18
fix(teeattestation/nitro): enforce max age on attestation timestamp [CL112-10]

nadahalli:
`verifyAttestationDocument` only checked the attestation timestamp was non-zero; `currentTime` was used solely for cert validity (a 1-2 year window). A valid attestation therefore stayed accepted until the leaf cert expired, even though our protocol produces attestations immediately before consumption.

Adds `MaxAttestationAge` (5 min) and rejects attestations older than that. Test covers a fresh attestation verifying and a stale one being rejected while the cert is still valid.

Audit finding CL112-10 / PRIV-438.


prashantkumar1982 (approved):

vreff (approved):

marshalokos:
Make it Ready

PR #2173, opened 2026-06-18
DurableEmitter: Poll DB queue depth + Update defaults

DylanTinianov:
Removes in-memory queue depth tracking and instead uses metrics loop to poll DB for queue depth.
Note: queue depth metric reflects the DB table size.

Also sets defaults for batching
```
InsertBatchFlushInterval: 100 * time.Millisecond,
InsertBatchSize:          100,
```


patrickhuie19 (dismissed):

tarcisiozf (dismissed):

patrickhuie19 (dismissed):

tarcisiozf (dismissed):

patrickhuie19 (approved):

tarcisiozf (approved):

pkg/durableemitter/durable_emitter.go
```diff
@@ -177,12 +179,6 @@ type DurableEmitter struct {
 	insertShutdown atomic.Bool
 	insertInFlight atomic.Int32
 
-	// pendingCount is an exact, atomic count of rows inserted but not yet
```
patrickhuie19, line 180: noting for historical purposes; the other approach would be to:

"Have only host process (RetransmitEnabled=true) emitters report queue depth, and they can track atomic queue size since it runs purge + expiry loops." - @DylanTinianov :)

this will ultimately be more long term efficient since you can have multiple emitter created in LOOPs and they don't all need to be querying the DB for queue depth size. this is good enough for now

PR #2180, opened 2026-06-22
[CRE][DonTime] Move request expiry outside the OCR plugin

bolekk:
[CRE-5067] Handling expiry inside the plugin comes with a risk of accumulating pending requests indefinitely when OCR is not progressing.

[CRE-5067]: https://smartcontract-it.atlassian.net/browse/CRE-5067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

1. Make Request SendResponse/SendTimeout thread safe and at-most-once.
2. Add timer to each request.
3. Remove expiry from the OCR plugin.
4. Remove unused Copy() method.
5. Remove unused context params.

### DIFF
--- a/pkg/workflows/dontime/plugin.go
+++ b/pkg/workflows/dontime/plugin.go
@@ -113,15 +113,7 @@ func (p *Plugin) Observation(_ context.Context, outctx ocr3types.OutcomeContext,
 	}
 
 	requests := map[string]int64{} // Maps executionID --> seqNum
-	timeoutCheck := time.Now()
 	for _, req := range p.store.GetRequests() {
-		if req.ExpiryTime().Before(timeoutCheck) {
-			// Request has been sitting in queue too long
-			p.store.RemoveRequest(req.WorkflowExecutionID)
-			req.SendTimeout(nil)
-			continue
-		}
-
 		// Validate request sequence number
 		numObservedDonTimes := 0
 		times, ok := previousOutcome.ObservedDonTimes[req.WorkflowExecutionID]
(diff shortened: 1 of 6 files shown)


DylanTinianov (approved):

pkg/workflows/dontime/plugin.go
```diff
@@ -132,14 +124,13 @@ func (p *Plugin) Observation(_ context.Context, outctx ocr3types.OutcomeContext,
...
 			p.store.RemoveRequest(req.WorkflowExecutionID)
-			req.SendResponse(nil,
-				Response{
-					WorkflowExecutionID: req.WorkflowExecutionID,
-					SeqNum:              req.SeqNum,
-					Timestamp:           0,
-					Err: fmt.Errorf("requested seqNum %d for executionID %s is greater than the number of observed don times %d",
-						req.SeqNum, req.WorkflowExecutionID, numObservedDonTimes),
-				})
+			req.SendResponse(Response{
```
cedric-cordenier, line 127: Can we do anything to avoid a side effect here? This still suffers from the same problem you mentioned of requests hanging when the plugin is dead

pkg/workflows/dontime/request.go
```diff
@@ -15,42 +15,39 @@ type Request struct {
...
 
-func (r *Request) SendResponse(_ context.Context, resp Response) {
-	select {
-	case r.CallbackCh <- resp:
+func (r *Request) SendResponse(resp Response) {
+	r.respondOnce.Do(func() {
+		r.stopExpiry()
+		select {
+		case r.CallbackCh <- resp:
+		default: // Don't block trying to send
```
cedric-cordenier, line 38: This means that if CallbackCh is blocked for whatever reason, we'll just abort the request and then due to the respondOnce we'll never send a response. Is that desirable?

pkg/workflows/dontime/request.go
cedric-cordenier, line 9: We already have a library for this that we use for Vault + Consensus caps, should we consider using that here too? Looks like we are introducing identical semantics

cedric-cordenier:
@bolekk Is it worth refactoring this so it uses this lib instead: https://github.com/smartcontractkit/chainlink-common/blob/fde1bc6ad94dba5b18b2f068620c56edefb52513/pkg/capabilities/consensus/requests/handler.go?

pkg/workflows/dontime/request.go
bolekk: Technically yes but callbackCh is always created with size=1 so this should never happen. 
Still, if it happens I'd rather fail than block.

pkg/workflows/dontime/plugin.go
bolekk: WDYM? Now they will time out if the plugin is down.

pkg/workflows/dontime/request.go
bolekk: Good point, let me take a look. Might be cleaner to do it seperately.

cedric-cordenier (approved):

pkg/workflows/dontime/plugin.go
bolekk: Discussed offline that we might consider moving SendResponse outside of the plugin but that's not part of this PR.

pkg/workflows/dontime/request.go
bolekk: Will address separately.

PR #2182, opened 2026-06-22
Allow a prehook for restricting workflow execution behaviour

nolag:


vreff (dismissed):

vreff (dismissed):

pkg/workflows/host/execution_restrictions.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if len(allowed) == 0 {
+		return responses, nil
+	}
+
+	inner, err := e.ExecutionHelper.GetSecrets(ctx, &sdk.GetSecretsRequest{Requests: allowed})
+	if err != nil {
+		return nil, err
+	}
+	return append(responses, inner...), nil
```
russell-stern: This could change the return order to be different than what we pass in since we're always putting the errors first

pkg/workflows/wasm/host/module.go
```diff
@@ -41,6 +41,7 @@ const v2ImportPrefix = "version_v2"
 var (
 	defaultTickInterval              = 100 * time.Millisecond
 	defaultTimeout                   = 10 * time.Minute
+	defaultPreeHookTimeout           = 10 * time.Second
```
russell-stern: typo: defaultPreeHookTimeout -> defaultPrehookTimeout

pkg/workflows/host/requirement_selecting_module.go
```diff
@@ -119,13 +120,35 @@ func (r *requirementSelectingModule) trigger(ctx context.Context, request *sdk.E
...
+			}
+
+			switch preHookResult.Result.(type) {
+			case *sdk.ExecutionResult_Error:
+				return preHookResult, nil
+			}
+
+			restrictions := preHookResult.GetRestrictions()
+
+			handler = NewRestrictedExecutionHelper(handler, restrictions)
```
russell-stern, line 139: If restrictions is nil we fail open in `NewRestrictedExecutionHelper`. Not necessarily a bug but should we fail close to let the workflow owner know they're using prehook wrong?

russell-stern (changes requested):

pkg/workflows/host/requirement_selecting_module.go
nolag: Let's leave it on the SDK to do that if we want to. You need a way to say this is unrestricted. In Go, or TS, if you return nil or an empty object, it seems intentional enough to me.

vreff (approved):

russell-stern (approved):

PR #2188, opened 2026-06-23
ChipIngress: Introduce per-event idempotency keys

tarcisiozf:
This pull request adds support for per-event idempotency keys in batch publishing for the ChipIngress client, allowing each event in a batch to carry its own idempotency identifier. This is achieved by introducing a new `Idempotency` message in the protobuf schema and updating the client code to populate and transmit these keys in a positionally-aligned manner with the events. Comprehensive unit tests are included to ensure correctness and backward compatibility.

**Batch publishing and idempotency support:**

* Added a new `Idempotency` message to the protobuf schema (`chip_ingress.pb.go`) and extended the `CloudEventBatch` message with a repeated `idempotency` field, ensuring each entry aligns positionally with its corresponding event. This allows future extensibility for idempotency metadata. 
* Updated the `newBatchRequest` function in `client.go` to extract idempotency keys from event attributes and populate the new `Idempotency` field in batch requests, maintaining positional alignment with the events. 

**Event creation enhancements:**

* Modified `NewEvent` to set the idempotency key as an extension attribute only if it is present and non-empty, ensuring correct propagation of idempotency information at event creation.


patrickhuie19 (approved):

pkg/chipingress/batch/client.go
```diff
@@ -428,17 +428,24 @@ func splitMessagesByRequestSize(messages []*messageWithCallback, maxRequestSize
...
 	events := make([]*chipingress.CloudEventPb, len(messages))
+	idem := make([]*chipingress.Idempotency, len(messages))
 	for i, msg := range messages {
 		events[i] = msg.event
+		key := ""
+		if av := msg.event.Attributes[chipingress.IdempotencyKeyAttr]; av != nil {
+			key = av.GetCeString()
+		}
+		idem[i] = &chipingress.Idempotency{Key: key}
 	}
```
hendoxc, line 439: can't we derive this on the server side if they are part of the attributes on each event anyways? whats the thinking here?

pkcll: +1

patrickhuie19: I see the argument. On-balance though if we push the compute to the clients doesn't this help the server scale?

DylanTinianov (approved):

pkg/chipingress/pb/chip_ingress.proto
```diff
@@ -72,10 +72,20 @@ message PublishError {
...
+}
+
 // CloudEventBatch is used to send many ChipIngress
 message CloudEventBatch {
     repeated io.cloudevents.v1.CloudEvent events = 1;
     PublishOptions options = 2;
+    // idempotency[i] corresponds to events[i] (positional, like
+    // PublishResponse.results). Optional/additive: may be empty or shorter than
+    // events; servers that don't understand field 3 ignore it.
+    repeated Idempotency idempotency = 3;
```
pkcll, line 88: @patrickhuie19 @tarcisiozf I need a better understanding of the full picture here, specifically which part of the pipeline will own this, but if the idempotency key is being passed from the client, it could be part of PublishOptions.  PublishOptions was introduced to encapsulate flags like that.

pkg/chipingress/batch/client.go
pkcll: We already perform this type of processing on the server side, so handling one additional attribute should not make any noticeable difference from a computational standpoint.

PR #2196, opened 2026-06-25
teeattestation: add Document.VerifyPCR for per-instance PCRs

nadahalli:
## What

Adds `Document.VerifyPCR(index, expected)` and a `VerifyExpectedPCRs(map)` convenience on the parsed attestation `Document` (from `ValidateAndParse`).

## Why

`ValidateAndParse` checks PCR0/1/2 against the shared `trustedMeasurements` (image measurements, identical across the pool). Per-instance PCRs, notably **PCR4** (the SHA384 of the parent instance ID), differ per enclave, so they can't live in those shared measurements and need a separate per-enclave comparison. This is the building block for binding an attestation to a specific host instance.

`VerifyPCR` rejects:
- an empty expected value,
- an absent or length-mismatched PCR,
- an **all-zero** PCR, which is what debug-mode enclaves emit and must never be accepted as a real measurement.

## Scope

Pure addition, no behavior change to existing validation. Consumers (confidential-compute pool.go, and potentially the relay DON) will call this against a per-enclave expected PCR4 sourced from config.

### DIFF
--- a/pkg/teeattestation/nitro/validate.go
+++ b/pkg/teeattestation/nitro/validate.go
@@ -76,6 +76,56 @@ type Document struct {
 	ModuleID string
 }
 
+// VerifyPCR checks that the PCR at index equals expected. Use it for
+// per-instance PCRs such as PCR4 (the SHA384 of the parent instance ID), which
+// differ per enclave and therefore cannot be part of the shared trusted
+// measurements checked by ValidateAndParse.
+//
+// It returns an error if expected is empty, the PCR is absent or
+// length-mismatched, the PCR is all zero (debug-mode enclaves emit all-zero
+// PCRs, which must never be accepted as a real measurement), or the values
+// differ.
+func (d *Document) VerifyPCR(index uint, expected []byte) error {
+	if len(expected) == 0 {
+		return fmt.Errorf("expected PCR%d value is empty", index)
+	}
+	actual, ok := d.PCRs[index]
+	if !ok {
+		return fmt.Errorf("attestation has no PCR%d", index)
+	}
+	if len(actual) != len(expected) {
+		return fmt.Errorf("PCR%d length mismatch: expected %d bytes, got %d", index, len(expected), len(actual))
+	}
+	if allZero(actual) {
+		return fmt.Errorf("PCR%d is all zero (debug-mode enclave), refusing", index)
+	}
+	if !bytes.Equal(actual, expected) {
+		return fmt.Errorf("PCR%d mismatch", index)
+	}
+	return nil
+}
+
+// VerifyExpectedPCRs checks each index/value in expected against the document
+// via VerifyPCR. It is a convenience for callers asserting several per-instance
+// PCRs at once and returns the first failure.
+func (d *Document) VerifyExpectedPCRs(expected map[uint][]byte) error {
+	for index, value := range expected {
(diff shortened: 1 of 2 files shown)


pkg/teeattestation/nitro/validate.go
```diff
@@ -76,6 +76,56 @@ type Document struct {
...
+// via VerifyPCR. It is a convenience for callers asserting several per-instance
+// PCRs at once and returns the first failure.
+func (d *Document) VerifyExpectedPCRs(expected map[uint][]byte) error {
+	for index, value := range expected {
+		if err := d.VerifyPCR(index, value); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
nadahalli: Done in fc06c95f

pkg/teeattestation/nitro/validate.go
```diff
@@ -76,6 +76,56 @@ type Document struct {
...
+	}
+	if len(actual) != len(expected) {
+		return fmt.Errorf("PCR%d length mismatch: expected %d bytes, got %d", index, len(expected), len(actual))
+	}
+	if allZero(actual) {
+		return fmt.Errorf("PCR%d is all zero (debug-mode enclave), refusing", index)
+	}
+	if !bytes.Equal(actual, expected) {
+		return fmt.Errorf("PCR%d mismatch", index)
+	}
```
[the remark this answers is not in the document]
nadahalli: Done in fc06c95f

pkg/teeattestation/nitro/validate.go
```diff
@@ -76,6 +77,62 @@ type Document struct {
...
+// VerifyPCR checks that the PCR at index equals expected. Use it for
+// per-instance PCRs such as PCR4 (the SHA384 of the parent instance ID), which
+// differ per enclave and therefore cannot be part of the shared trusted
+// measurements checked by ValidateAndParse.
+//
+// It returns an error if expected is empty, the PCR is absent or
+// length-mismatched, the PCR is all zero (debug-mode enclaves emit all-zero
+// PCRs, which must never be accepted as a real measurement), or the values
+// differ.
+func (d *Document) VerifyPCR(index uint, expected []byte) error {
```
vreff, line 89: Why aren't we just returning a generic []bytes ID field from the verifyMeasurements function and just pass the PCR4 through there? We are really gating ourselves into Nitro here. Although that probably doesn't matter, it's still quite specific.

nadahalli: This is already inside teeattestation/nitro with cose parsing, etc. I think it should be ok at this level. If we ever move to proper TEEAL, we will need to rework a lot of other code. I'd rather do that then.

vreff: How about when we move this code back into `confidential-compute` once it is open sourced?

vreff: This code was previously very simple. It would be nice to revert to that.

vreff: See the original PR for reference: https://github.com/smartcontractkit/confidential-compute/pull/293/changes#diff-822b5dfcbad5be1fed590586f99dd1ba4d0cba29c1b45761d03f659c2007cc94R13.

By simply extending the interface to include a []bytes field our consumers didn't have to dig into Nitro details.

PR #2202, opened 2026-06-26
Add GetTransaction and GetSigningAccount to Stellar relayer API

Krish-vemula:


ilija42 (commented):
Why set the transmitter in the capability code instead of just letting the chain service handle it?

ilija42 (approved):
This looks fine, but lets spend a minute to think about what are the odds of the signing account retrieved for the args being different from the actual one

Unheilbar (approved):

PR #2210, opened 2026-07-01
[1/2] chipingress: expose drop failure classification for partial delivery metrics

pkcll:
Part 1 of 2 — adds the chipingress batch client helpers needed for the beholder partial-delivery metric change.

- Export `ErrMessageBufferFull` and `ErrClientShutdown` sentinels from `pkg/chipingress/batch`.
- Add `batch.ClassifyDropFailure` to map send/queue errors into shared metric dimensions (`error_type`, `error_code`, `error_reason`).

Part 2: #2266
Jira: INFOPLAT-13901

### DIFF
--- a/pkg/chipingress/batch/drop_failure.go
+++ b/pkg/chipingress/batch/drop_failure.go
@@ -1,0 +1,53 @@
+package batch
+
+import (
+	"context"
+	"errors"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	ErrorTypePartialDelivery = "partial_delivery"
+	ErrorTypeRPCError        = "rpc_error"
+	ErrorTypeBufferFull      = "buffer_full"
+	ErrorTypeClientError     = "client_error"
+)
+
+// ErrorCodeFor returns a bounded error code string for a batch send/queue error.
+// The returned code is intended for metric dimensions; callers can infer the
+// error category from the code:
+//   - PUBLISH_ERROR_CODE_*  -> partial_delivery
+//   - "buffer_full"         -> buffer_full
+//   - "client_shutdown"     -> client_error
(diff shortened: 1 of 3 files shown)


pkg/chipingress/batch/drop_failure.go
```diff
@@ -0,0 +1,79 @@
...
+	ErrorTypeBufferFull      = "buffer_full"
+	ErrorTypeClientError     = "client_error"
+)
+
+// DropFailure describes a dropped event for metrics and logging.
+type DropFailure struct {
+	ErrorType   string // partial_delivery | rpc_error | buffer_full | client_error
+	ErrorCode   string // PUBLISH_ERROR_CODE_* or gRPC code name (DeadlineExceeded, …)
+	ErrorReason string // server reason, status message, or client error text
+}
```
pavel-raykov: It seems that you don't really need a separate type `DropFailure` here.

In https://github.com/smartcontractkit/chainlink-common/pull/2266, you can:
 * use `sendErr.Error()` instead of `drop.ErrorReason,`
 * use only ErrorCode instead of ErrorType (one can infer one from another)
 * In this PR (or you can drop it and update https://github.com/smartcontractkit/chainlink-common/pull/2266 directly)have one method generating `ErrorCode` with separate if's based on `errors.AsType`.

pavel-raykov (changes requested):

pkg/chipingress/batch/drop_failure.go
pkcll: Addressed in https://github.com/smartcontractkit/chainlink-common/pull/2266

pkcll:
Folding this into #2266 — moved drop_failure.go there and simplified ErrorCodeFor to return (errorType, errorCode) directly, so there's no cross-PR dependency to coordinate.

PR #2207, opened 2026-06-29
feat: add chip-ingress batch config keys and tune defaults

pkcll:
## Summary

Add 7 chip-ingress batch configuration keys to `beholder.Config` and wire them through the batch emitter and LOOP plugin env config.

This PR also updates the batch emitter defaults based on staging/prod capacity analysis and fixes the histogram bucket schema for `chip_ingress.batch.request_size_*`.

## Default changes

| Key | Before | After | Rationale |
|---|---|---|---|
| ChipIngressBufferSize | 1000 | **10000** | Buffer headroom for transient send stalls |
| ChipIngressMaxBatchSize | 500 | **1000** | Burst-safety ceiling (batches are interval-bound, not size-bound) |
| ChipIngressSendInterval | 100ms | **500ms** | Reduce RPC volume on interval-bound batches |
| ChipIngressSendTimeout | 3s | **10s** | Match observed prod latency tail; preserve flush margin |
| ChipIngressDrainTimeout | 10s | **30s** | Preserve shutdown-flush margin over the larger SendTimeout |
| ChipIngressMaxConcurrentSends | 10 | 10 | Unchanged |
| ChipIngressMaxGRPCRequestSize | 10 MiB | 10 MiB | Unchanged |

Batches are interval-bound at steady state (p99 ~10 msgs/batch), so `MaxBatchSize` is a burst-safety ceiling rather than the active flush lever.

## Changes

- `pkg/beholder/config.go`: added `ChipIngressMaxGRPCRequestSize` + updated defaults
- `pkg/beholder/batch_emitter_service.go`: wired `WithMaxGRPCRequestSize`
- `pkg/chipingress/batch/client.go`: fixed histogram bucket boundaries for `request_size_messages` and `request_size_bytes`
- `pkg/loop/config.go`: added `CL_CHIP_INGRESS_*` env vars, `EnvConfig` fields, serialization/parsing
- `pkg/loop/server.go`: wired env fields into `beholder.Config`

## Related PRs

- **Downstream (chainlink TOML wiring):** https://github.com/smartcontractkit/chainlink/pull/22987

## Testing

- `go test ./pkg/loop/... ./pkg/beholder/... ./pkg/chipingress/...`

## Notes

- DurableEmitter batch client intentionally unchanged.
- Histogram bucket schema change causes transient mixed-schema skew until the fleet upgrades; this is expected and self-healing.
- The previous `Co-authored-by: Cursor` attribution has been removed from the commit history.

PR #2216, opened 2026-07-02
beholder: send resource attributes to chip ingress as gRPC metadata

pkcll:
## Summary

Resource attributes set in node TOML (`[Telemetry.ResourceAttributes]`) reach the OTel collector
path but never the ChipIngress path, so they appear as headers on `beholder__platform__messages`
and not on `cre`. This wires them into the ChipIngress client.

They travel **once per request as gRPC metadata**, not stamped on every CloudEvent. Resource
attributes describe the producer rather than any individual event — the same reason OTLP factors
`resource` out of the payload instead of repeating it on every log record. Per-event stamping would
also repeat identical bytes for every event in a batch: ~10 attributes across a 1,000-event batch is
roughly 300 KB of duplication, which counts against `maxGRPCRequestSize` and therefore reduces how
many events fit per batch.

This also matches an existing chip-ingress pattern. `nopInfoHeadersFromContext` already reads a
connection-scoped value once and `baseKafkaHeaders` fans it onto every Kafka record, cached per
`(domain, entity, specVersion)`.

## Changes

- `pkg/beholder/client.go` — pass resource attributes to the ChipIngress client as gRPC metadata via
  `chipingress.WithResourceAttributeHeaders`.
- `pkg/beholder/resource_attributes.go` — add `resourceAttributesToStringMap`, the single conversion
  point, using `attribute.Value.Emit` for canonical stringification of any value type. (Its doc
  comment is worded to make explicit that this feeds gRPC metadata only — resource attributes are
  never stamped as CloudEvent extensions.)
- `pkg/beholder/resource_attributes_test.go` — unit coverage for `resourceAttributesToStringMap`,
  including the empty-input case.
- `pkg/beholder/chip_ingress_emitter.go` — doc comment on `ChipIngressEmitter` making the same point:
  resource attributes are not stamped on individual events here, they travel once per request as
  client-level gRPC metadata. Plus an incidental `gofmt` alignment fix.
- `pkg/beholder/batch_emitter_service_test.go` — incidental stray-blank-line `gofmt` fix; plus a
  mock fix required by the pin bump (see "Depends on #2288"): `PublishBatch` success mocks now
  return a response with successful results instead of `(nil, nil)`.
- `pkg/beholder/client_test.go` — regression test asserting on a real gRPC connection that
  configuring `AuthHeaders` together with `ResourceAttributes` leaves the CSA node auth token intact
  and delivers the attributes alongside it.
- Root `go.mod`/`go.sum` — bump `pkg/chipingress` to the reworked #2288 head (whitelisted
  `chainlink-*` headers).

No exported API is added or changed in `pkg/beholder`.

## Why the auth test

The CSA node auth token (`X-Beholder-Node-Auth-Token`) travels as `credentials.PerRPCCredentials`,
while resource attributes travel through a unary interceptor that calls
`metadata.AppendToOutgoingContext`. Nothing previously covered the two coexisting. Two properties
matter and both are asserted:

- the token arrives **exactly once**, unmodified — the interceptor appends rather than replaces, so a
  collision would send two values under one key rather than overwriting
- grpc-go validates the whole outgoing metadata map in `newClientStream`, so a single non-printable
  attribute value would fail the entire RPC with `codes.Internal`, auth included. `pkg/chipingress`
  `SanitizeMetadataHeaders` prevents this upstream by omitting an attribute with an invalid key or a
  non-printable value rather than sending it malformed — see #2288.

## Not yet visible to consumers

Chip-ingress currently reads incoming gRPC metadata **only to authenticate**
(`metadata.FromIncomingContext` appears solely in `internal/auth/csa_auth.go`). Kafka headers come
from CloudEvent extensions and core attributes plus a fixed set the server derives itself.

So this PR makes the attributes visible to the service but **not to a Kafka consumer**. Forwarding
them onto records is smartcontractkit/atlas#13201 (open, draft, implemented and passing its own
suite), which reads a closed whitelist of ten fixed `chainlink-*` metadata headers and translates
each to its mapped `resource_<attribute key>` Kafka header — keeping the namespace closed so a
resource attribute cannot shadow `ce_*`, `table_route_name`, or an identity header the server
derives from the verified auth token.

On its own, therefore, this PR does not close the header gap between `cre` and
`beholder__platform__messages`. It is one of three parts: this, #2288, and atlas#13201.

## Depends on #2288

The publisher-side contract lives in #2288, reworked 2026-09-14 to a **closed whitelist**: ten fixed,
hand-named `chainlink-*` gRPC metadata headers (`chainlink-resource-csa-public-key`, `chainlink-resource-service-name`,
... — see `chipingress.ResourceAttributeHeaders`), matched case-insensitively against the configured
resource attributes. Attributes outside the whitelist are ignored; non-printable values are omitted,
never rewritten. Because no operator-defined key can ever become a header name, the earlier revision
charset validation, caps, drop-reporting and deny-lists were all deleted. This PR needs **no
functional code change** for any of that: it calls `WithResourceAttributeHeaders`, so all of it
happens inside `pkg/chipingress`.

The pin is now bumped to the reworked #2288 branch head
(`v0.0.11-0.20260915043145-8e5ac5170bc5`). Until #2288 merges, the
`Validate go.mod dependencies` job is expected to fail, as it has throughout this stack.

**One test-only fix came along with the pin bump.** The new pin also pulls in the batch client
partial-delivery callback path, which dispatches per-event callbacks from `PublishResponse.Results`.
The batch-emitter test mocks returned `(nil, nil)` for `PublishBatch` — realistic enough for the old
client, a nil-deref / synthetic `RESULTS_MISMATCH` for the new one — so
`pkg/beholder/batch_emitter_service_test.go` now returns a response with a generous tail of
successful results, matching the real server contract of one result per event. No production code
changed.

Consumers reading `platformEnv` on the legacy topic will find it as `resource_platformEnv` on `cre`.
They already need changes for two other headers on these topics regardless: `beholder_entity` is
`ce_type`, and `csa_public_key` is `ce_csapublickey` — auth-derived, so more trustworthy than the
self-reported `resource_csa_public_key`.

## Related

- Upstream (chipingress primitives): #2267 (merged), #2288 (prefixing + validation/limits, required)
- Server: smartcontractkit/atlas#13201
- Downstream (chainlink): smartcontractkit/chainlink#23014

### Related PRs

- https://github.com/smartcontractkit/chainlink-common/pull/2288 (upstream — chipingress client
  primitives this PR depends on)
  - https://github.com/smartcontractkit/chainlink-common/pull/2216 (this PR — beholder wiring)
  - https://github.com/smartcontractkit/atlas/pull/13201 (sibling downstream, different repo —
    chip-ingress server; forwards the metadata this PR sends onto Kafka headers; does not depend on
    this PR, only on the #2288 wire contract)

## Test plan

```
go test ./pkg/beholder/...
go vet ./pkg/beholder/...
```


go.mod
```diff
@@ -158,3 +158,5 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20260401024825-9d38bb4040a9 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/smartcontractkit/chainlink-common/pkg/chipingress => ./pkg/chipingress
```
patrickhuie19: any way to avoid the local replace?

patrickhuie19 (dismissed):

go.mod
pkcll: For sure, will clean this up.

engnke:
overall approach lgtm. Adding the CloudEvent extensions as the guaranteed propagation path makes sense, and we're keeping the gRPC header route for observability is reasonable

pkcll:
Split into two PRs:
- pkg/chipingress resource attributes: #2267
- pkg/beholder wiring: #2207
- downstream chainlink bump: #23014

thomaska (dismissed):

thomaska (dismissed):

4of9 (dismissed):

pkg/beholder/resource_attributes.go
```diff
@@ -0,0 +1,15 @@
...
+// using attribute.Value.Emit for canonical stringification of any value type. This is the
+// single source of truth used to derive both the gRPC metadata headers and the CloudEvent
+// extension keys/values sent to ChipIngress, so both mechanisms stay consistent.
+func resourceAttributesToStringMap(attrs []attribute.KeyValue) map[string]string {
+	m := make(map[string]string, len(attrs))
+	for _, kv := range attrs {
+		m[string(kv.Key)] = kv.Value.Emit()
+	}
+	return m
+}
```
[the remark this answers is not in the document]
pkcll: Done in a4260cda0 — resourceAttributesToStringMap now returns nil for empty input, so the no-attributes case (the common one) allocates nothing. The caller gates on len(resourceAttrs) > 0, which holds for a nil map. Test updated to assert nil for both nil and empty-slice input.

pkcll:
## Data flow (cross-posted from the server-side design doc in smartcontractkit/atlas#13201)

For easier review — what this PR wires up, end to end:

```mermaid
flowchart LR
    subgraph node["Chainlink node"]
        toml["TOML config<br/>Telemetry.ResourceAttributes"]
        beholder["beholder.Config.ResourceAttributes<br/>pkg/beholder"]
        client["chipingress client<br/>SanitizeMetadataHeaders:<br/>whitelist projection, once per client<br/>pkg/chipingress"]
        toml --> beholder --> client
    end

    subgraph ingress["chip-ingress"]
        auth["auth interceptor<br/>verified CSA token"]
        fwd["resourceHeadersFromContext<br/>reads the 10 fixed keys only,<br/>first value wins, sorted"]
        base["baseKafkaHeaders<br/>built once per request,<br/>reused by every record"]
        auth -->|"server-derived:<br/>ce_csapublickey<br/>ce_nodeoperatorname"| base
        fwd -->|"client-asserted:<br/>resource_*"| base
    end

    kafka[("Kafka records<br/>cre topic")]

    client -->|"gRPC metadata, once per request:<br/>chainlink-resource-service-name: chainlink<br/>chainlink-resource-csa-public-key: abc123<br/>(non-whitelisted attributes dropped client-side)"| ingress
    base -->|"fanned out onto every record:<br/>resource_service.name: chainlink<br/>resource_csa_public_key: abc123"| kafka
```

This PR is the `pkg/beholder` box: it converts `beholder.Config.ResourceAttributes` to a string map and hands it to `chipingress.WithResourceAttributeHeaders`; the whitelist projection, header naming, and server fan-out all live in #2288 (merged) and smartcontractkit/atlas#13201.

skippaDaBitFlippa (approved):

jmank88 (approved):

PR #2217, opened 2026-07-02
Handle nil pointer error

yashnevatia:
- Handle nil pointer error when fetching dashboards.
- Add option to deploy using folder UID. Generally the folder UID resolves properly but I hit a case when it did not and as a result I had missing dashboards. This is an additive change which does not hurt in case we run into the same issue again.

### DIFF
--- a/observability-lib/grafana/dashboard.go
+++ b/observability-lib/grafana/dashboard.go
@@ -3,6 +3,7 @@ package grafana
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 
 	"github.com/grafana/grafana-foundation-sdk/go/alerting"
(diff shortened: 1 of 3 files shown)


observability-lib/grafana/dashboard.go
```diff
@@ -39,11 +40,29 @@ type DeployOptions struct {
...
+		if err != nil {
+			return nil, err
+		}
+		if folder == nil {
+			return nil, fmt.Errorf("folder with UID %q not found", options.FolderUID)
+		}
+		return folder, nil
+	}
+	if options.FolderName != "" {
+		return client.FindOrCreateFolder(options.FolderName)
```
dhaidashenko, line 61: We also want to return error from FindOrCreateFolder if we see two folders with the same name

dhaidashenko: Never mind, seems like it's not possible

dhaidashenko (dismissed):

dhaidashenko (approved):

ilija42 (approved):

PR #2224, opened 2026-07-06
DurableEmitter Optimizations

DylanTinianov:
- Coalesce deletes instead of marking delivered + purge loop.
- Removes Fallback client (doesn't provide any value and adds complexity during outages).
- Added page cursor to cycle through DB start point in retransmit loop to be proactive against events that never deliver from blocking other events from being resent.
- Cleans up unnecessary DB calls and logs.

Tested on Reliability Env


pkg/custmsg/custom_message.go
```diff
@@ -119,5 +120,10 @@ func sendLogAsCustomMessageW(ctx context.Context, msg string, labels map[string]
 		return fmt.Errorf("sending custom message failed on emit: %w", err)
 	}
 
+	err = durableemitter.GlobalEmit(ctx, payloadBytes, "source", "platform", "type", "BaseMessage")
+	if err != nil {
+		return fmt.Errorf("DurableEmitter: sending custom message failed on emit: %w", err)
```
tarcisiozf: should we log instead of return error here? GlobalEmit may return ErrNotInitialized if none is registered

pkg/custmsg/custom_message.go
```diff
@@ -119,5 +120,10 @@ func sendLogAsCustomMessageW(ctx context.Context, msg string, labels map[string]
 		return fmt.Errorf("sending custom message failed on emit: %w", err)
 	}
 
+	err = durableemitter.GlobalEmit(ctx, payloadBytes, "source", "platform", "type", "BaseMessage")
```
tarcisiozf: we need `beholder_data_schema` attribute, right?

pkg/durableemitter/store.go
```diff
@@ -69,51 +69,19 @@ func (s *PgDurableEventStore) Delete(ctx context.Context, id int64) error {
...
-func (s *PgDurableEventStore) MarkDelivered(ctx context.Context, id int64) error {
-	const q = `UPDATE ` + chipDurableEventsTable + ` SET delivered_at = now() WHERE id = $1 AND delivered_at IS NULL`
-	if _, err := s.ds.ExecContext(ctx, q, id); err != nil {
-		return fmt.Errorf("failed to mark chip durable event delivered id=%d: %w", id, err)
-	}
-	return nil
-}
-
-func (s *PgDurableEventStore) MarkDeliveredBatch(ctx context.Context, ids []int64) (int64, error) {
+func (s *PgDurableEventStore) BatchDelete(ctx context.Context, ids []int64) (int64, error) {
```
tarcisiozf, line 72: we are not enforcing a limit for number of ids, and during our POC I remember break this into smaller chunks helped me with performance locally because you had smaller WAL entries + less roundtrips than individual deletes

DylanTinianov: The limit is enforced by the configured batch size. The ORM implementation doesn't need to care IMO :P

pkg/custmsg/custom_message.go
DylanTinianov: I don't think we need that

pkg/durableemitter/store.go
tarcisiozf: let me rephrase this, a batch limit of 2k might be good for publish throughput but terrible for db clean up, I rather have two separate limits, e.g 2k batch size and 100~500 delete chunk size

DylanTinianov: Good thing we have that already then ;) 

[DeleteBatchSize](https://github.com/smartcontractkit/chainlink-common/pull/2224/changes#diff-bf5153785824bcce4dd2d71df28ee6391f3def652716abe5a72a4c365ffb4b74R92)

pkg/custmsg/custom_message.go
tarcisiozf: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/beholder/message.go#L26

DylanTinianov: It resolves/validates schema by source + type (domain + entity)

tarcisiozf: this only fails in direct beholder path, we should be good

patrickhuie19 (approved):

tarcisiozf (approved):

PR #2225, opened 2026-07-07
Add Beholder OTel SDK metric cardinality limit

pkcll:
This PR adds a configurable cardinality limit to the Beholder OTel SDK metric provider. It prevents unbounded memory growth when high-cardinality labels (e.g., `event_id`, `workflow_execution_id`) are accidentally attached to metrics.

### How `sdkmetric.WithCardinalityLimit` works

`WithCardinalityLimit(n)` sets a **hard limit on distinct attribute sets (time series) per metric instrument**. Each unique combination of attributes creates a separate aggregation state in the SDK.

**Example:**

```go
requestsTotal.Add(ctx, 1,
    attribute.String("method", method),
    attribute.String("status", status),
)
```

Creates series:
- `(method=GET, status=200)`
- `(method=GET, status=500)`
- `(method=POST, status=200)`

If someone adds `user_id` (unbounded), series explode. The limit prevents this.

**What happens when the limit is reached?**

When the (n+1)th unique attribute set arrives, it does **not** create a new aggregation. Instead, it is recorded into an **overflow bucket** with attribute `otel.metric.overflow=true`:

```
http_requests_total{method=GET,status=200}     15231
http_requests_total{method=POST,status=500}    183
...
http_requests_total{otel.metric.overflow=true} 428912
```

The total count remains correct, but per-identity of overflowed combinations is lost.

**Important:** This does **not** drop the metric or fail exports. It is a safety net, not a substitute for good metric design.

### Scope

- `pkg/beholder/config.go`: Add `MetricCardinalityLimit` field
- `pkg/beholder/meter_provider.go`: Wire `sdkmetric.WithCardinalityLimit`
- `pkg/loop/config.go`: Propagate limit via env var `CL_TELEMETRY_METRIC_CARDINALITY_LIMIT`

### Limitations

This limit is **per instrument**, not global across the application. Unbounded labels should still be avoided or filtered via views; the cardinality limit exists as a last-resort memory guard.

### DIFF
--- a/pkg/beholder/config.go
+++ b/pkg/beholder/config.go
@@ -41,7 +41,7 @@ type Config struct {
 	MetricRetryConfig    *RetryConfig
 	MetricViews []metric.View
 	// MetricCardinalityLimit sets the SDK per-instrument attribute-set limit (0 = disabled).
-	// DefaultConfig uses 100000 as a production safety valve for high-cardinality workloads.
+	// DefaultConfig uses DefaultMetricCardinalityLimit as a production safety valve for high-cardinality workloads.
 	MetricCardinalityLimit int
 	// MetricCompressor sets the gRPC compressor for metrics. Valid values: "gzip" (default), "none".
 	MetricCompressor string
@@ -111,6 +111,9 @@ var defaultRetryConfig = RetryConfig{
 const (
 	defaultPackageName        = "beholder"
 	defaultMaxConcurrentSends = 10
+
+	// DefaultMetricCardinalityLimit is the production safety valve for high-cardinality workloads.
+	DefaultMetricCardinalityLimit = 100000
 )
 
 var defaultOtelAttributes = []attribute.KeyValue{
@@ -142,7 +145,7 @@ func DefaultConfig() Config {
 		// Metric
 		MetricReaderInterval:   1 * time.Second,
 		MetricCompressor:       "gzip",
-		MetricCardinalityLimit: 100000,
+		MetricCardinalityLimit: DefaultMetricCardinalityLimit,
 		// OTel metric exporter retry config
 		MetricRetryConfig: defaultRetryConfig.Copy(),
 		// Log
--- a/pkg/beholder/config.go
+++ b/pkg/beholder/config.go
@@ -26,7 +26,7 @@ type Config struct {
 	EmitterMaxQueueSize       int
 	// EmitterBatchProcessor controls custom-message export mode:
 	// true = batched async export; false = immediate per-record export.
-	EmitterBatchProcessor     bool
+	EmitterBatchProcessor bool
 
 	// OTel Trace
 	TraceSampleRatio  float64
@@ -39,7 +39,10 @@ type Config struct {
 	// OTel Metric
 	MetricReaderInterval time.Duration
 	MetricRetryConfig    *RetryConfig
-	MetricViews          []metric.View
+	MetricViews []metric.View
+	// MetricCardinalityLimit sets the SDK per-instrument attribute-set limit (0 = disabled).
+	// DefaultConfig uses 100000 as a production safety valve for high-cardinality workloads.
+	MetricCardinalityLimit int
 	// MetricCompressor sets the gRPC compressor for metrics. Valid values: "gzip" (default), "none".
 	MetricCompressor string
 	MetricProducers  []metric.Producer // For example, a prometheus bridge
@@ -127,7 +130,7 @@ func DefaultConfig() Config {
 		EmitterExportInterval:     1 * time.Second,
 		EmitterMaxQueueSize:       2048,
 		// Keep batched export enabled by default for throughput.
-		EmitterBatchProcessor:     true,
+		EmitterBatchProcessor: true,
 		// OTel message log exporter retry config
 		LogRetryConfig: defaultRetryConfig.Copy(),
 		// Trace
@@ -137,8 +140,9 @@ func DefaultConfig() Config {
 		// OTel trace exporter retry config
 		TraceRetryConfig: defaultRetryConfig.Copy(),
 		// Metric
-		MetricReaderInterval: 1 * time.Second,
-		MetricCompressor:     "gzip",
+		MetricReaderInterval:   1 * time.Second,
+		MetricCompressor:       "gzip",
+		MetricCardinalityLimit: 100000,
 		// OTel metric exporter retry config
 		MetricRetryConfig: defaultRetryConfig.Copy(),
 		// Log
@@ -148,8 +152,8 @@ func DefaultConfig() Config {
 		LogMaxQueueSize:       2048,
 		LogBatchProcessor:     true,
 		LogStreamingEnabled:   true, // Enable logs streaming by default
-		LogLevel:      zapcore.InfoLevel,
-		LogCompressor: "gzip",
+		LogLevel:              zapcore.InfoLevel,
+		LogCompressor:         "gzip",
 		// Chip Ingress Batch Emitter
 		ChipIngressBatchEmitterEnabled: false,
 		ChipIngressBufferSize:          1000,
@@ -173,6 +177,7 @@ func TestDefaultConfig() Config {
 	config.LogRetryConfig.MaxElapsedTime = 0    // Retry is disabled
 	config.TraceRetryConfig.MaxElapsedTime = 0  // Retry is disabled
 	config.MetricRetryConfig.MaxElapsedTime = 0 // Retry is disabled
+	config.MetricCardinalityLimit = 0           // Disable overflow aggregation in unit tests
 	// Auth disabled for testing (TTL=0 means static auth mode)
 	config.AuthHeadersTTL = 0
 	return config
@@ -186,6 +191,7 @@ func TestDefaultConfigHTTPClient() Config {
 	config.LogBatchProcessor = false
 	config.OtelExporterGRPCEndpoint = ""
 	config.OtelExporterHTTPEndpoint = "localhost:4318"
+	config.MetricCardinalityLimit = 0
 	// Auth disabled for testing (TTL=0 means static auth mode)
 	config.AuthHeadersTTL = 0
 	return config
--- a/pkg/beholder/meter_provider.go
+++ b/pkg/beholder/meter_provider.go
@@ -4,9 +4,11 @@ import (
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 )
 
-// metricOptions returns sdkmetric.Option values for caller MetricViews and the
-// SDK per-instrument cardinality limit.
-func (cfg Config) metricOptions(opts ...sdkmetric.Option) []sdkmetric.Option {
+// metricOptions returns cfg-derived sdkmetric.Option values (caller MetricViews
+// and the SDK per-instrument cardinality limit). Callers append reader/resource
+// options at the call site.
+func (cfg Config) metricOptions() []sdkmetric.Option {
+	var opts []sdkmetric.Option
 	if len(cfg.MetricViews) > 0 {
 		opts = append(opts, sdkmetric.WithView(cfg.MetricViews...))
 	}
--- a/pkg/beholder/meter_provider.go
+++ b/pkg/beholder/meter_provider.go
@@ -1,0 +1,17 @@
+package beholder
+
+import (
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+)
+
+// metricOptions returns sdkmetric.Option values for caller MetricViews and the
+// SDK per-instrument cardinality limit.
+func (cfg Config) metricOptions(opts ...sdkmetric.Option) []sdkmetric.Option {
+	if len(cfg.MetricViews) > 0 {
+		opts = append(opts, sdkmetric.WithView(cfg.MetricViews...))
+	}
+	if cfg.MetricCardinalityLimit > 0 {
+		opts = append(opts, sdkmetric.WithCardinalityLimit(cfg.MetricCardinalityLimit))
+	}
+	return opts
+}
--- a/pkg/beholder/meter_provider_test.go
+++ b/pkg/beholder/meter_provider_test.go
@@ -1,0 +1,50 @@
+package beholder
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+)
+
+func TestConfig_metricOptions_cardinalityLimit(t *testing.T) {
+	t.Parallel()
+
+	const (
+		uniqueAttributes = 10
+		limit            = 5
+	)
+
+	reader := sdkmetric.NewManualReader()
+	cfg := DefaultConfig()
+	cfg.MetricCardinalityLimit = limit
+
+	mpOpts := cfg.metricOptions(sdkmetric.WithReader(reader))
+	mp := sdkmetric.NewMeterProvider(mpOpts...)
+	t.Cleanup(func() { _ = mp.Shutdown(context.Background()) })
+
+	meter := mp.Meter("test")
+	counter, err := meter.Int64Counter("overflow_test_total")
+	require.NoError(t, err)
+
+	for i := range uniqueAttributes {
+		counter.Add(context.Background(), 1, metric.WithAttributes(attribute.Int("key", i)))
+	}
+
+	var rm metricdata.ResourceMetrics
+	require.NoError(t, reader.Collect(context.Background(), &rm))
+
+	sum := rm.ScopeMetrics[0].Metrics[0].Data.(metricdata.Sum[int64])
+	assert.Len(t, sum.DataPoints, limit)
+
+	var total int64
+	for _, dp := range sum.DataPoints {
+		total += dp.Value
+	}
+	assert.Equal(t, int64(uniqueAttributes), total)
+}
--- a/pkg/beholder/meter_provider_test.go
+++ b/pkg/beholder/meter_provider_test.go
@@ -24,7 +24,7 @@ func TestConfig_metricOptions_cardinalityLimit(t *testing.T) {
 	cfg := DefaultConfig()
 	cfg.MetricCardinalityLimit = limit
 
-	mpOpts := cfg.metricOptions(sdkmetric.WithReader(reader))
+	mpOpts := append(cfg.metricOptions(), sdkmetric.WithReader(reader))
 	mp := sdkmetric.NewMeterProvider(mpOpts...)
 	t.Cleanup(func() { _ = mp.Shutdown(context.Background()) })
 
--- a/pkg/loop/config.go
+++ b/pkg/loop/config.go
@@ -82,6 +82,7 @@ const (
 	envTelemetryLogMaxQueueSize           = "CL_TELEMETRY_LOG_MAX_QUEUE_SIZE"
 	envTelemetryTraceCompressor           = "CL_TELEMETRY_TRACE_COMPRESSOR"
 	envTelemetryMetricCompressor          = "CL_TELEMETRY_METRIC_COMPRESSOR"
+	envTelemetryMetricCardinalityLimit    = "CL_TELEMETRY_METRIC_CARDINALITY_LIMIT"
 	envTelemetryPrometheusBridgeEnabled   = "CL_TELEMETRY_PROMETHEUS_BRIDGE_ENABLED"
 	envTelemetryPrometheusBridgePrefixes  = "CL_TELEMETRY_PROMETHEUS_BRIDGE_PREFIXES"
 	envTelemetryLogCompressor             = "CL_TELEMETRY_LOG_COMPRESSOR"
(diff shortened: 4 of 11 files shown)


pkcll:
Response to Copilot review comments:

| # | Finding | Status |
|---|---------|--------|
| 1 | AsCmdEnv drops 0 → child re-defaults to 10000 | Already fixed by an earlier commit (switched the field to `*int` so unset vs. explicit-0 are distinguishable) — Copilot was reviewing a stale diff |
| 2 | Negative limit values silently accepted | Fixed — `parse()` now rejects negative values with an explicit error; added a test case |
| 3 | Test name says "prepends", code appends | Fixed — renamed to `TestMergeMetricViews_appendsDefaultsAfterCallerViews` |
| 4 | Base-trigger allowlist test doesn't prove the specific view matched | Fixed — added an attribute (`attempts`) that's absent from both the allow- and deny-lists, so the test only passes if the base-trigger view itself applied, not just the global catch-all |

pkg/loop/config.go
```diff
@@ -290,6 +292,9 @@ func (e *EnvConfig) AsCmdEnv() (env []string) {
 	add(envTelemetryLogMaxQueueSize, strconv.Itoa(e.TelemetryLogMaxQueueSize))
 	add(envTelemetryTraceCompressor, e.TelemetryTraceCompressor)
 	add(envTelemetryMetricCompressor, e.TelemetryMetricCompressor)
+	if e.TelemetryMetricCardinalityLimit != 0 {
+		add(envTelemetryMetricCardinalityLimit, strconv.Itoa(e.TelemetryMetricCardinalityLimit))
+	}
```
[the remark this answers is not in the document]
pkcll: Already fixed by an earlier commit (switched the field to `*int` so unset vs. explicit-0 are distinguishable) — Copilot was reviewing a stale diff.

pkg/loop/config.go
```diff
@@ -530,6 +535,15 @@ func (e *EnvConfig) parse() error {
 		}
 		e.TelemetryTraceCompressor = os.Getenv(envTelemetryTraceCompressor)
 		e.TelemetryMetricCompressor = os.Getenv(envTelemetryMetricCompressor)
+		if v, ok := os.LookupEnv(envTelemetryMetricCardinalityLimit); ok {
+			limit, err := strconv.Atoi(v)
+			if err != nil {
+				return fmt.Errorf("failed to parse %s: %w", envTelemetryMetricCardinalityLimit, err)
+			}
+			e.TelemetryMetricCardinalityLimit = limit
+		} else {
```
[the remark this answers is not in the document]
pkcll: Fixed — `parse()` now rejects negative values with an explicit error; added a test case.

pkg/beholder/meter_provider_test.go
```diff
@@ -0,0 +1,67 @@
...
+	assert.Len(t, sum.DataPoints, limit)
+
+	var total int64
+	for _, dp := range sum.DataPoints {
+		total += dp.Value
+	}
+	assert.Equal(t, int64(uniqueAttributes), total)
+}
+
+func TestMergeMetricViews_prependsDefaults(t *testing.T) {
```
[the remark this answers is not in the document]
pkcll: Fixed — renamed to `TestMergeMetricViews_appendsDefaultsAfterCallerViews`.

pkg/beholder/metricviews/views_test.go
```diff
@@ -0,0 +1,145 @@
...
+	)
+
+	var rm metricdata.ResourceMetrics
+	require.NoError(t, reader.Collect(context.Background(), &rm))
+
+	keys := attributeKeysFromSum(t, rm)
+	assert.Contains(t, keys, attribute.Key("capability_id"))
+	assert.NotContains(t, keys, attribute.Key("trigger_id"))
+	assert.NotContains(t, keys, attribute.Key("event_id"))
+}
```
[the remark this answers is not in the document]
pkcll: Fixed — added an attribute (`attempts`) that's absent from both the allow- and deny-lists, so the test only passes if the base-trigger view itself applied, not just the global catch-all.

pkg/beholder/metricviews/views.go
```diff
@@ -0,0 +1,72 @@
...
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+)
+
+const (
+	baseTriggerInstrumentGlob  = "capabilities_base_trigger_*"
+	stoppedResendingInstrument = "capabilities_base_trigger_stopped_resending_timestamp"
+)
+
+var (
+	globalHighCardinalityDeny = attribute.NewDenyKeysFilter(
```
pavel-raykov: this file reads as if it is applied globally to all metrics, instead these are specific limiters for capabilities - is this intended?

pkg/beholder/metricviews/views.go
```diff
@@ -0,0 +1,72 @@
...
+		attribute.Key("event_id"),
+		attribute.Key("workflow_execution_id"),
+	)
+
+	baseTriggerAllow = attribute.NewAllowKeysFilter(
+		attribute.Key("capability_id"),
```
pavel-raykov: is there some beholder data showing that filtering based on these attributes is needed?

pkg/beholder/config.go
```diff
@@ -40,6 +40,11 @@ type Config struct {
 	MetricReaderInterval time.Duration
 	MetricRetryConfig    *RetryConfig
 	MetricViews          []metric.View
+	// MetricViewsDisabled skips DefaultViews attribute filtering (for tests).
+	MetricViewsDisabled bool
```
pavel-raykov: this is a strange approach - can't you default filters only in the place where they are created in prod? For example, do you also need these filters in staging?

pkg/beholder/meter_provider.go
```diff
@@ -0,0 +1,37 @@
...
+// Final order:
+//  1. cfg.MetricViews — caller overrides (e.g. chainlink metricViews() histogram buckets)
+//  2. metricviews.DefaultViews() — attribute filters for high-cardinality labels
+//
+// Caller views must come first: when multiple views match an instrument and
+// resolve to the same output stream (same name/description/unit/kind), the SDK
+// keeps only the first in registration order and drops the rest. If the default
+// "*" denylist ran first, caller histogram-bucket views would be silently dropped.
+//
+// A consequence is that filters do not compose: once a caller view wins for a
```
pavel-raykov: this I don't understand - what is the difficulty of making composable filters? (which are the most natural)

pkg/beholder/metricviews/views.go
pkcll: Yes this is global filter for labels which were added. Historically workflow_execution_id was added multiple times for metrics in multiple places. This view/filter attempts to prevent this from happening in the future for the top offenders. This is basically a black list for labels which we know for sure are not bounded.

pkg/beholder/metricviews/views.go
pkcll: I dont have exact number for this label cardinality, ideally I would suggest to drop anything that has word ID in it. We should have only low cardinality labels to not put too much pressure on our Observability backends. But thats a long term goal. For now Im keeping this white list here for labels which do look suspicious.

pavel-raykov: is there a way to compute the traffic of _id* labels vs others now?

pkcll: Not directly from the OTel pipeline, you'd need to either instrument the otel exporter on the node side or tap into it somewhere in the otel collector upstream. That would require custom code. New VictoriaMetrics backend has a native cardinality API and UI that shows number of time series for metrics per label: https://vm-ui-prod.ops.prod.cldev.sh/select/0/vmui/?#/cardinality?date=2026-07-09&topN=10&focusLabel=event_id 

I think it supports wildcard too `*_id` 

https://vm-ui-prod.ops.prod.cldev.sh/select/0/vmui/?#/cardinality?date=2026-07-09&topN=10&focusLabel=*_id

Image

pkcll: Some reports generated from querying VM API 
https://vmselect-prod.ops.prod.cldev.sh:8481

```
Distinct label values

  ┌─────────────────────────────┬───────────────┬───────────────────────┐
  │ Scope                       │ capability_id │ trigger_id            │
  ├─────────────────────────────┼───────────────┼───────────────────────┤
  │ Global (all metrics)        │ 160           │ 117                   │
  ├─────────────────────────────┼───────────────┼───────────────────────┤
  │ capabilities_*              │ 59            │ 117                   │
  ├─────────────────────────────┼───────────────┼───────────────────────┤
  │ capabilities_base_trigger_* │ 59            │ 117                   │
  ├─────────────────────────────┼───────────────┼───────────────────────┤
  │ platform_launcher_*         │ 102           │ 0 (label not present) │
  └─────────────────────────────┴───────────────┴───────────────────────┘

  trigger_id only appears on capabilities_base_trigger_* metrics.
  capability_id also appears on platform_launcher_* (launcher metrics), which adds ~43 extra values beyond the capabilities family.

  ────────────────────────────────────────

  Series count (key metrics)

  ┌─────────────────────────────────────────────────┬──────────────┬────────────────────────┬─────────────────────┐
  │ Metric                                          │ Total series │ Distinct capability_id │ Distinct trigger_id │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ capabilities_base_trigger_ack_total             │ 215,546      │ 52                     │ 116                 │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ capabilities_base_trigger_retry_total           │ 3,358        │ 36                     │ 51                  │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ capabilities_base_trigger_active_registrations  │ 543          │ 57                     │ —                   │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ capabilities_base_trigger_pending_events        │ 457          │ 55                     │ —                   │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ capabilities_base_trigger_ack_attempts_bucket   │ 2,155,460    │ 52                     │ 116                 │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ capabilities_base_trigger_time_to_ack_ms_bucket │ 2,586,552    │ (timed out)            │ 116                 │
  ├─────────────────────────────────────────────────┼──────────────┼────────────────────────┼─────────────────────┤
  │ platform_launcher_* (all)                       │ 7,299        │ 102                    │ —                   │
  └─────────────────────────────────────────────────┴──────────────┴────────────────────────┴─────────────────────┘

  Histogram buckets (*_bucket) dominate series count because of the le label — each capability_id/trigger_id is multiplied across bucket boundaries and other labels (donID, host_name, zone, platformEnv, etc.).

  ────────────────────────────────────────

  Top cardinality drivers (ack_total)

  capability_id — up to ~10,912 series per value:
  • evm:ChainSelector:11344663589394136015@1.0.0 (56)
  • evm:ChainSelector:1523760397290643893@1.0.0 (5734951)

  trigger_id — up to ~9,960 series per value (hash-style trigger_reg_* IDs)

  ────────────────────────────────────────

  Notes

  • VM has ~164M total series; global count({capability_id!=""}) and count({trigger_id!=""}) timed out at the 30s query limit.
  • Scoped queries and the label-values API completed successfully and are the reliable approach here.
  • trigger_id cardinality (117) is moderate; series explosion comes from label combinations (especially on histogram metrics), not from hundreds of thousands of unique trigger IDs.
```

pkcll: event_id label cardinality — VictoriaMetrics Global
Source: https://vmselect-prod.ops.prod.cldev.sh:8481/select/0/prometheus

```
Distinct values

  ┌─────────────────────────────┬──────────────────────────┐
  │ Scope                       │ Distinct event_id values │
  ├─────────────────────────────┼──────────────────────────┤
  │ Global                      │ 88,529                   │
  ├─────────────────────────────┼──────────────────────────┤
  │ capabilities_*              │ 88,529                   │
  ├─────────────────────────────┼──────────────────────────┤
  │ capabilities_base_trigger_* │ 88,529                   │
  ├─────────────────────────────┼──────────────────────────┤
  │ platform_launcher_*         │ 0 (label not present)    │
  ├─────────────────────────────┼──────────────────────────┤
  │ VM tsdb status (top labels) │ 72,955                   │
  └─────────────────────────────┴──────────────────────────┘

  event_id is orders of magnitude higher than the other labels we checked:

  ┌───────────────┬────────────────────────┐
  │ Label         │ Global distinct values │
  ├───────────────┼────────────────────────┤
  │ event_id      │ 88,529                 │
  ├───────────────┼────────────────────────┤
  │ capability_id │ 160                    │
  ├───────────────┼────────────────────────┤
  │ trigger_id    │ 117                    │
  └───────────────┴────────────────────────┘

  ────────────────────────────────────────

  Where it appears

  • Only on capabilities_* metrics (same family as trigger_id)
  • Co-labeled with capability_id, trigger_id, donID, host_name, zone, platformEnv, etc. (42 labels total)
  • Not on platform_launcher_* or gauges like active_registrations / pending_events

  Values look like composite hashes, e.g.
  00003f2477...:a873ff00...:0

  ────────────────────────────────────────

  Series count by metric

  ┌─────────────────────────────────────────────────┬──────────────────────┬───────────────────┐
  │ Metric                                          │ Series with event_id │ Distinct event_id │
  ├─────────────────────────────────────────────────┼──────────────────────┼───────────────────┤
  │ capabilities_base_trigger_ack_total             │ 217,447              │ 82,391            │
  ├─────────────────────────────────────────────────┼──────────────────────┼───────────────────┤
  │ capabilities_base_trigger_retry_total           │ 2,792                │ 2,792 (1:1)       │
  ├─────────────────────────────────────────────────┼──────────────────────┼───────────────────┤
  │ capabilities_base_trigger_ack_attempts_bucket   │ 2,174,470            │ 82,391            │
  ├─────────────────────────────────────────────────┼──────────────────────┼───────────────────┤
  │ capabilities_base_trigger_time_to_ack_ms_bucket │ 2,609,520            │ 82,398            │
  ├─────────────────────────────────────────────────┼──────────────────────┼───────────────────┤
  │ capabilities_base_trigger_active_registrations  │ —                    │ —                 │
  ├─────────────────────────────────────────────────┼──────────────────────┼───────────────────┤
  │ capabilities_base_trigger_pending_events        │ —                    │ —                 │
  └─────────────────────────────────────────────────┴──────────────────────┴───────────────────┘

  On ack_total, ~2.6 series per event_id on average (other labels multiply). Histogram buckets push total series into the millions.

  ────────────────────────────────────────

  Takeaway

  event_id is the dominant high-cardinality label in this family — ~82k+ unique values vs ~117 trigger_id and ~59 capability_id on the same metrics. It's a strong candidate for aggregation, dropping at scrape/export, or scoping to debug-only metrics.

```

pkg/beholder/config.go
pkcll: Sorry, `// MetricViewsDisabled skips DefaultViews attribute filtering (for tests).` -  this is confusing comment, its not just for tests (will fix that).

MetricViewsDisabled is a feature flag to disable the filters in case we need it for emergency situations. It will be added to core.toml config and passed from the core node, the same as other Beholder/ChIP config parameters.

pkg/beholder/meter_provider.go
pkcll: The difficulty is that OTel does not support chaining/composing filters out of the box. It picks the first match and applies it. So if you are adding a blacklist view to filter event_id and there is already a view passed from the core to Beholder for existing metrics (we have cases like that for configuring histogram buckets), the blacklist will not be applied.

pkg/beholder/config.go
pkcll: Fixed https://github.com/smartcontractkit/chainlink-common/pull/2225/changes/80de37796ab4501f6b2e275fd5dfa55b73bc654d

pkg/beholder/meter_provider.go
```diff
@@ -0,0 +1,37 @@
...
+// because caller views target metrics that do not emit the denied labels; capability
+// trigger metrics carry no caller view and fall through to the defaults below.
+func mergeMetricViews(cfg Config) []sdkmetric.View {
+	if cfg.MetricViewsDisabled {
+		return cfg.MetricViews
+	}
+	return append(append([]sdkmetric.View{}, cfg.MetricViews...), metricviews.DefaultViews()...)
+}
+
+func appendMeterProviderOptions(cfg Config, opts ...sdkmetric.Option) []sdkmetric.Option {
```
jmank88: I think it is unintuitive for an `append` func to reverse the order. What if we just had a simple method and caller did the append?
```suggestion
func (cfg Config) metricOptions() []sdkmetric.Option {
```

pkg/beholder/meter_provider.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// A consequence is that filters do not compose: once a caller view wins for a
+// stream, the default attribute filter no longer applies to it. That is acceptable
+// because caller views target metrics that do not emit the denied labels; capability
+// trigger metrics carry no caller view and fall through to the defaults below.
+func mergeMetricViews(cfg Config) []sdkmetric.View {
+	if cfg.MetricViewsDisabled {
+		return cfg.MetricViews
+	}
+	return append(append([]sdkmetric.View{}, cfg.MetricViews...), metricviews.DefaultViews()...)
```
jmank88: Appending to empty?
```suggestion
	return append(cfg.MetricViews, metricviews.DefaultViews()...)
```

pkg/beholder/meter_provider.go
```diff
@@ -0,0 +1,37 @@
...
+// Caller views must come first: when multiple views match an instrument and
+// resolve to the same output stream (same name/description/unit/kind), the SDK
+// keeps only the first in registration order and drops the rest. If the default
+// "*" denylist ran first, caller histogram-bucket views would be silently dropped.
+//
+// A consequence is that filters do not compose: once a caller view wins for a
+// stream, the default attribute filter no longer applies to it. That is acceptable
+// because caller views target metrics that do not emit the denied labels; capability
+// trigger metrics carry no caller view and fall through to the defaults below.
+func mergeMetricViews(cfg Config) []sdkmetric.View {
```
jmank88: This seems like it could be a method as well:
```suggestion
func (cfg Config) metricViews() []sdkmetric.View {
```

pkg/beholder/meter_provider.go
pkcll: Done — replaced `appendMeterProviderOptions` with zero-arg `(cfg Config) metricOptions()`; callers use `append(cfg.metricOptions(), ...)` for reader/resource options.

pkg/beholder/meter_provider.go
pkcll: Addressed in #2238 after the PR split (metric filter views). `metricViews()` uses `append(cfg.MetricViews, metricviews.DefaultViews()...)` as suggested.

pkg/beholder/meter_provider.go
pkcll: Addressed in #2238 after the PR split. `mergeMetricViews` became `(cfg Config) metricViews()`.

pkg/loop/config_test.go
```diff
@@ -147,6 +156,8 @@ func TestEnvConfig_parse(t *testing.T) {
 	}
 }
 
+func ptr[T any](v T) *T { return &v }
```
jmank88: These aren't necessary anymore as of go 1.26. You can take the address inline.

pkcll: Dropped the `ptr` helper and switched call sites to `new(100000)`, `new(0)`, and `new(500)`.

pkg/beholder/config.go
```diff
@@ -137,8 +140,9 @@ func DefaultConfig() Config {
 		// OTel trace exporter retry config
 		TraceRetryConfig: defaultRetryConfig.Copy(),
 		// Metric
-		MetricReaderInterval: 1 * time.Second,
-		MetricCompressor:     "gzip",
+		MetricReaderInterval:   1 * time.Second,
+		MetricCompressor:       "gzip",
+		MetricCardinalityLimit: 100000,
```
pavel-raykov: this 100000 is defined twice (also below). Can we avoid doing that?

pkg/loop/config_test.go
```diff
@@ -205,6 +214,7 @@ var envCfgFull = EnvConfig{
 	TelemetryEmitterExportMaxBatchSize: 100,
 	TelemetryEmitterMaxQueueSize:       1000,
 	TelemetryLogStreamingEnabled:       false,
+	TelemetryMetricCardinalityLimit:    new(100000),
```
pavel-raykov: can you skip specifying it here to use the default (which is the same)?

pkg/beholder/config.go
pkcll: Fixed in d4e5f859 -- extracted DefaultMetricCardinalityLimit = 100000 as a package-level constant in pkg/beholder/config.go, and updated both pkg/beholder/config.go (DefaultConfig) and pkg/loop/config.go (env fallback) to reference it instead of duplicating the literal.

pavel-raykov (approved):

jmank88 (approved):

4of9 (approved):

thomaska (approved):

pkg/loop/config_test.go
pkcll: done

PR #2227, opened 2026-07-07
Add TenantID cresettings limit with default 0

prashantkumar1982:
## Summary
- Add global `TenantID` cresettings field as `Setting[uint64]` with default `0`
- Regenerate `defaults.toml` and `defaults.json` golden files
- Document the new limit in the cresettings README flowchart


pavel-raykov (dismissed):

cedric-cordenier (dismissed):

pkg/settings/cresettings/README.md
```diff
@@ -61,6 +61,7 @@ flowchart
 
     subgraph Store.FetchWorkflowArtifacts
         CentralizedWorkflowOwnerVerificationEnabled[/CentralizedWorkflowOwnerVerificationEnabled\]:::gate
+        TenantID{{TenantID}}:::bound
```
jmank88: Can we label it differently? Not sure it belongs here either

pkg/settings/cresettings/defaults.json
```diff
@@ -5,6 +5,7 @@
 	"GatewayVaultManagementEnabled": "true",
 	"VaultJWTAuthEnabled": "false",
 	"CentralizedWorkflowOwnerVerificationEnabled": "false",
+	"TenantID": "1",
```
bolekk: How about we default to 0 (= "unassigned")? Otherwise we risk unrelated envs using 1 and polluting prod.

prashantkumar1982: done

jmank88 (approved):

pkg/settings/cresettings/README.md
```diff
@@ -11,20 +11,23 @@ flowchart
...
     time:::time
+    setting:::setting
 
     classDef bound stroke:#f00
     classDef gate stroke:#0f0
     classDef queue stroke:#00f
     classDef rate stroke:#ff0
     classDef resource stroke:#f0f
     classDef time stroke:#0ff
+    classDef setting stroke:#999,fill:#f5f5f5
```
jmank88, line 30: Probably should exclude fill since it's dynamic depending on the theme
```suggestion
    classDef setting stroke:#999
```

russell-stern (approved):

PR #2235, opened 2026-07-08
emit service health errors via Beholder

skudasov:


pkg/services/beholderhealth/beholderhealth.go
jmank88, line 1: What is sot? And should this package be under the `health` subtree?

pkg/services/health.go
```diff
@@ -88,11 +91,13 @@ func NewHealthChecker(ver, sha string) *HealthChecker {
 type HealthCheckerConfig struct {
 	// Optionally override debug.BuildInfo
 	Ver, Sha string
+	Logger   logger.Logger
 	// Optional hooks for reporting.
-	IncVersion func(ctx context.Context, ver string, sha string)
-	AddUptime  func(ctx context.Context, duration time.Duration)
-	SetStatus  func(ctx context.Context, name string, status int)
-	Delete     func(ctx context.Context, name string)
+	EmitSoTHealthData func(ctx context.Context, svcHealth map[string]error)
```
jmank88: Isn't `Emit` an implementation detail? This hook can be used by anything that wants to inspect the map. How about something like `UpdateHealthy` or `WithHealthy`? 🤷

pkg/services/health.go
```diff
@@ -204,6 +213,9 @@ func (c *HealthChecker) update(ctx context.Context) {
 	}
 	c.cfg.AddUptime(ctx, interval)
 
+	// emit beholder metrics for rane-sot-app service
```
jmank88: Let's avoid implementation details. This is just calling a hook that can be used in any way the caller wants.

pkg/services/beholderhealth/beholderhealth.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewChecker(ver, sha string) (*services.HealthChecker, error) {
+	cfg, err := ConfigureHooks(services.HealthCheckerConfig{Ver: ver, Sha: sha})
+	if err != nil {
+		return nil, err
+	}
+	return cfg.New(), nil
+}
+
+func ConfigureHooks(orig services.HealthCheckerConfig) (services.HealthCheckerConfig, error) {
```
jmank88: Why return an error?

pkg/services/beholderhealth/beholderhealth.go
```diff
@@ -0,0 +1,64 @@
...
+	cfg := orig // copy
+	cfg.EmitSoTHealthData = func(ctx context.Context, svcHealth map[string]error) {
+		emitter := beholder.GetEmitter()
```
jmank88: Please inject this instead of fetching it via the gobal func.

pkg/services/beholderhealth/beholderhealth.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func NewChecker(ver, sha string) *services.HealthChecker {
+	cfg := ConfigureHooks(services.HealthCheckerConfig{Ver: ver, Sha: sha})
+	return cfg.New()
+}
+
+func ConfigureHooks(orig services.HealthCheckerConfig) services.HealthCheckerConfig {
+	cfg := orig // copy
+	cfg.EmitHealthInfoToBeholder = func(ctx context.Context, emitter services.Emitter, svcHealth map[string]error) {
```
jmank88: I think that injecting from here will be cleaner, and do we still need a separate interface in that case?
```suggestion
func ConfigureHooks(orig services.HealthCheckerConfig, emitter services.Emitter) services.HealthCheckerConfig {
	cfg := orig // copy
	cfg.EmitHealthInfoToBeholder = func(ctx context.Context, svcHealth map[string]error) {
```

pkg/services/health.go
```diff
@@ -50,7 +52,8 @@ type HealthChecker struct {
 	chStop chan struct{}
 	chDone chan struct{}
 
-	cfg HealthCheckerConfig
+	cfg     HealthCheckerConfig
+	lggr    logger.SugaredLogger
```
jmank88: Is this unused now?

pkg/services/beholderhealth/beholderhealth.go
```diff
@@ -0,0 +1,60 @@
...
+	beholderProtobufDataSchema = "/node-platform/common/v1"
+)
+
+func NewChecker(ver, sha string, emitter beholder.Emitter) *services.HealthChecker {
+	cfg := ConfigureHooks(services.HealthCheckerConfig{Ver: ver, Sha: sha}, emitter)
+	return cfg.New()
+}
+
+func ConfigureHooks(cfg services.HealthCheckerConfig, emitter beholder.Emitter) services.HealthCheckerConfig {
+	cfg.EmitHealthInfoToBeholder = func(ctx context.Context, svcHealth map[string]error) {
```
jmank88: I think this should have the same logic as the others where a copy is made `cfg := orig // copy` to call any existing hooks instead of replacing them.

pkg/services/health.go
```diff
@@ -88,11 +91,13 @@ func NewHealthChecker(ver, sha string) *HealthChecker {
 type HealthCheckerConfig struct {
 	// Optionally override debug.BuildInfo
 	Ver, Sha string
+	Logger   logger.Logger
 	// Optional hooks for reporting.
-	IncVersion func(ctx context.Context, ver string, sha string)
-	AddUptime  func(ctx context.Context, duration time.Duration)
-	SetStatus  func(ctx context.Context, name string, status int)
-	Delete     func(ctx context.Context, name string)
+	EmitHealthInfoToBeholder func(ctx context.Context, svcHealth map[string]error)
```
jmank88: Let's name this something more generic.

pkg/services/beholderhealth/beholderhealth.go
```diff
@@ -0,0 +1,61 @@
...
+	beholderProtobufDataSchema = "/node-platform/common/v1"
+)
+
+func NewChecker(ver, sha string, emitter beholder.Emitter) *services.HealthChecker {
+	cfg := ConfigureHooks(services.HealthCheckerConfig{Ver: ver, Sha: sha}, emitter)
+	return cfg.New()
+}
+
+func ConfigureHooks(orig services.HealthCheckerConfig, emitter beholder.Emitter) services.HealthCheckerConfig {
+	cfg := orig // copy
```
jmank88, line 37: The original hook needs to be called from the new one to preserve it

pkg/services/health.go
```diff
@@ -88,11 +90,13 @@ func NewHealthChecker(ver, sha string) *HealthChecker {
 type HealthCheckerConfig struct {
 	// Optionally override debug.BuildInfo
 	Ver, Sha string
+	Logger   logger.Logger
```
jmank88:
```suggestion
```

skudasov: What about `EmitHealthInfo` and
```
			if cfg.Logger != nil {
				cfg.Logger.Errorw("failed to marshal node-platform health info", "err", err)
			}

```
then?

skudasov: Ok, can inject it too

pkg/services/health.go
```diff
@@ -89,10 +89,11 @@ type HealthCheckerConfig struct {
...
 	// Optional hooks for reporting.
-	IncVersion func(ctx context.Context, ver string, sha string)
-	AddUptime  func(ctx context.Context, duration time.Duration)
-	SetStatus  func(ctx context.Context, name string, status int)
-	Delete     func(ctx context.Context, name string)
+	EmitHealthInfo func(ctx context.Context, svcHealth map[string]error)
+	IncVersion     func(ctx context.Context, ver string, sha string)
+	AddUptime      func(ctx context.Context, duration time.Duration)
+	SetStatus      func(ctx context.Context, name string, status int)
+	Delete         func(ctx context.Context, name string)
```
jmank88: This is still describing an implementation detail. Let's align it with the others:
```suggestion
	IncVersion     func(ctx context.Context, ver string, sha string)
	AddUptime      func(ctx context.Context, duration time.Duration)
	SetStatus      func(ctx context.Context, name string, status int)
	Delete         func(ctx context.Context, name string)
	SetHealth func(ctx context.Context, health map[string]error)
```

jmank88 (approved):

gheorghestrimtu (approved):

PR #2236, opened 2026-07-09
Async commit for batch deletes

DylanTinianov:

### DIFF
--- a/pkg/durableemitter/store.go
+++ b/pkg/durableemitter/store.go
@@ -79,12 +79,27 @@ DELETE FROM ` + chipDurableEventsTable + ` WHERE id IN (
     WHERE id = ANY($1)
     FOR UPDATE SKIP LOCKED
 )`
-	res, err := s.ds.ExecContext(ctx, q, pq.Array(ids))
+	// Delete durability is not required under delete-on-delivery: a delete commit
+	// lost to a crash just leaves the row to be re-delivered (idempotent) or reclaimed
+	// by the expiry loop. We can relax synchronous_commit for the delete transaction,
+	// which removes the commit WAL-flush wait (IO:XactSync) which dominates at high
+	// delete rates, especially on Aurora where commit blocks on a storage-quorum ack.
+	var deleted int64
+	err := sqlutil.TransactDataSource(ctx, s.ds, nil, func(tx sqlutil.DataSource) error {
+		if _, err := tx.ExecContext(ctx, "SET LOCAL synchronous_commit = off"); err != nil {
+			return fmt.Errorf("failed to set synchronous_commit=off: %w", err)
+		}
+		res, err := tx.ExecContext(ctx, q, pq.Array(ids))
+		if err != nil {
+			return err
+		}
+		deleted, _ = res.RowsAffected()
+		return nil
+	})
 	if err != nil {
 		return 0, fmt.Errorf("failed to batch delete delivered chip durable events: %w", err)
 	}
-	n, _ := res.RowsAffected()
-	return n, nil
+	return deleted, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/durableemitter/store.go
```diff
@@ -73,18 +73,36 @@ func (s *PgDurableEventStore) BatchDelete(ctx context.Context, ids []int64) (int
 	if len(ids) == 0 {
 		return 0, nil
 	}
+	// FOR UPDATE SKIP LOCKED avoids deadlocks (40P01) when a backlog re-delivers
```
patrickhuie19: /nit what is 40P01? and is there a FOR UPDATE SKIP LOCKED in `BatchDelete` anymore?

pkg/durableemitter/store.go
```diff
@@ -73,18 +73,36 @@ func (s *PgDurableEventStore) BatchDelete(ctx context.Context, ids []int64) (int
...
     SELECT id FROM ` + chipDurableEventsTable + `
     WHERE id = ANY($1)
     FOR UPDATE SKIP LOCKED
 )`
-	res, err := s.ds.ExecContext(ctx, q, pq.Array(ids))
+	// Delete durability is not required under delete-on-delivery: a delete commit
+	// lost to a crash just leaves the row to be re-delivered (idempotent) or reclaimed
+	// by the expiry loop. We can relax synchronous_commit for the delete transaction,
+	// which removes the commit WAL-flush wait (IO:XactSync) which dominates at high
+	// delete rates, especially on Aurora where commit blocks on a storage-quorum ack.
```
patrickhuie19: /nit want to be a little careful with over-indexing on Aurora in comments. Most NOPs are running their DBs on baremetal with the default postgres implementation.

pkg/durableemitter/store.go
```diff
@@ -73,18 +73,36 @@ func (s *PgDurableEventStore) BatchDelete(ctx context.Context, ids []int64) (int
...
+	var deleted int64
+	err := sqlutil.TransactDataSource(ctx, s.ds, nil, func(tx sqlutil.DataSource) error {
+		if _, err := tx.ExecContext(ctx, "SET LOCAL synchronous_commit = off"); err != nil {
```
patrickhuie19, line 89: is this necessary to set per tx? is there a way we can have a particular db client always have this be true? wondering if we can split the delete client from the insert client. would this save on any latency?

pkg/durableemitter/store.go
DylanTinianov: It's an error that you're trying to delete the same row in multiple commits. And yes it's still there. Just added a comment so it's more clear why we need that.

It can happen when draining:
> retransmit re-queues them once per cursor sweep, and if a delivery is slower than the sweep period the same row gets re-queued before the prior attempt deleted it causing row-lock contention (40P01 error).

pkg/durableemitter/store.go
DylanTinianov: We could but it's probably not worth it right now for the overhead.

Would at most save microseconds per event. Crucially, it does not improve the big win; the IO:XactSync flush wait is already eliminated by the async commit itself. Splitting only trims the per-batch transaction bookkeeping, not the flush.

Claude's Take:
> Cost of splitting: a second connection pool = extra plumbing (thread a second ds into PgDurableEventStore + wire it in application.go) and it consumes from the node's Postgres connection budget.

> The saving is microseconds/event and the flush win is already captured by SET LOCAL; you'd be adding a whole connection pool + pgbouncer config care to shave a few round-trips per batch. I'd keep the per-tx SET LOCAL unless profiling shows the delete-transaction overhead (not the flush) is actually a bottleneck — then a dedicated async delete pool is the right move.

patrickhuie19: 👍 lets have a ticket in the backlog to remind us to check in on this and we can move on

patrickhuie19 (approved):

tarcisiozf (approved):

PR #2239, opened 2026-07-10
Durable Emitter - TTL budget metric

tarcisiozf:
This pull request refactors how the durable emitter tracks and reports queue expiry pressure metrics. Instead of counting the number of events near expiry, it now measures the "TTL budget"—the remaining time before the oldest pending event expires. This simplifies the code and provides a more direct signal for alerting on dead-letter queue (DLQ) pressure.

**Metrics and Observability Improvements:**

* Replaced the `NearTTLCount` metric (number of events near expiry) with a new `TTLBudget` metric, which tracks the remaining TTL headroom for the oldest pending event. This is now reported as `queue.ttl_budget_seconds`.
* Updated the `DurableQueueStats` struct and the `ObserveDurableQueue` interface to remove the `nearExpiryLead` parameter and field, and instead use the new `TTLBudget` field.
* Updated metric recording and gauge definitions to support the new TTL budget metric, including changes to metric names and types.
* Simplified the queue statistics observation logic in all stores (`PgDurableEventStore`, `MemDurableEventStore`, and metric wrappers) to remove all logic related to "near expiry" windows and focus only on TTL budget calculation.

### DIFF
--- a/pkg/durableemitter/durable_emitter_metrics.go
+++ b/pkg/durableemitter/durable_emitter_metrics.go
@@ -62,7 +62,7 @@ type durableEmitterMetrics struct {
 	queueDepth         metric.Int64Gauge
 	queuePayloadBytes  metric.Int64Gauge
 	queueOldestAgeSec  metric.Float64Gauge
-	queueTTLBudgetSec  metric.Float64Gauge
+	queueTTLBudgetSec  metric.Int64Gauge
 	queueCapacityRatio metric.Float64Gauge
 	procHeapInuse      metric.Int64Gauge
 	procHeapSys        metric.Int64Gauge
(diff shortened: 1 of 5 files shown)


pkg/durableemitter/durable_emitter_metrics.go
```diff
@@ -64,7 +62,7 @@ type durableEmitterMetrics struct {
 	queueDepth         metric.Int64Gauge
 	queuePayloadBytes  metric.Int64Gauge
 	queueOldestAgeSec  metric.Float64Gauge
-	queueNearTTL       metric.Int64Gauge
+	queueTTLBudgetSec  metric.Float64Gauge
```
patrickhuie19: Is a float necessary? From the PR description:

> Instead of counting the number of events near expiry, it now measures the "TTL budget"

A budget in integer seconds is precise enough for our needs

patrickhuie19:
```
func (d *DurableEmitter) metricsLoop() {
	mc := d.cfg.Metrics
	poll := mc.PollInterval
	if poll <= 0 {
		poll = 500 * time.Millisecond
	}
	ctx, cancel := d.stopCh.NewCtx()
	defer cancel()
```

the default of pollInteval is every 10 seconds, which is the preferred cadence over every 1/2 second. IFIUC this is where we set the default, so the implementation doesn't match the comment. Let's set to 10 seconds please

patrickhuie19 (approved):

patrickhuie19 (approved):

DylanTinianov (approved):

PR #2252, opened 2026-07-14
Helpers for Resource Managers

patrickhuie19:
ResourceManagers share needs around populating deployment + resource identities. There is also shared code like generating event_ids for snapshots or deltas.


pkg/resourcemanager/resourcemanager.go
```diff
@@ -212,15 +240,29 @@ func (rm *ResourceManager) start(_ context.Context) error {
...
+		timer := rm.clock.NewTimer(firstTick.Sub(now))
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return
+		case <-timer.Chan():
+		}
+		rm.emitSnapshots(ctx, firstTick)
+
 		ticker := rm.clock.NewTicker(rm.snapshotInterval)
```
jmank88, line 264: Could you use a `services.TickerConfig`? Or I guess there is a fake clock involved here too?

pkg/resourcemanager/resourcemanager.go
```diff
@@ -249,42 +291,84 @@ func (rm *ResourceManager) Register(m Meterable) (unregister func()) {
...
+			rctx, cancel := context.WithTimeout(ctx, timeout)
+			defer cancel()
+			rm.emitSnapshot(rctx, m, tickTime)
+			if rctx.Err() != nil && ctx.Err() == nil {
+				rm.lggr.Warnw("snapshot registrant exceeded its per-tick deadline; skipped",
+					"timeout", timeout,
+					"err", rctx.Err(),
+				)
+			}
+		}(m)
```
jmank88: If you add a context param then you can use normal naming within the func:
```suggestion
		func(ctx context.Context, m Meterable) {
			var cancel context.CancelFn
			ctx, cancel = context.WithTimeout(ctx, timeout)
			defer cancel()
			rm.emitSnapshot(ctx, m, tickTime)
			if ctx.Err() == nil {
				rm.lggr.Warnw("snapshot registrant exceeded its per-tick deadline; skipped",
					"timeout", timeout,
					"err", ctx.Err(),
				)
			}
		}(ctx, m)
```
No need to check both errors either. The parent will propogate.

pkg/resourcemanager/config_from_env.go
```diff
@@ -0,0 +1,59 @@
+package resourcemanager
+
+import (
+	"github.com/smartcontractkit/chainlink-common/pkg/durableemitter"
+	"github.com/smartcontractkit/chainlink-common/pkg/loop"
```
jmank88: It looks like this convenience func is the only reason for this import - can we try to avoid it? Or even just reversed might be better (e.g. move it to a method on loop.EnvConfig)

pkg/resourcemanager/config_from_env.go
```diff
@@ -0,0 +1,59 @@
...
+func ConfigFromEnv(envCfg *loop.EnvConfig) Config {
+	if envCfg == nil {
+		return Config{}
+	}
+	// Source the durable emitter. GetGlobalEmitter returns a typed nil pointer
+	// when Setup has not run; assign through a local Emitter so an unconfigured
+	// durable emitter yields a true nil interface (not a non-nil interface
+	// wrapping a nil pointer), which the ResourceManager treats as no-op.
+	var emitter Emitter
+	if de := durableemitter.GetGlobalEmitter(); de != nil {
```
jmank88: Can we inject this to avoid calling a global func from such a low level helper?

pkg/resourcemanager/resourcemanager.go
patrickhuie19: >Or I guess there is a fake clock involved here too?

yes, exactly. so the `services.TickerConfig` wouldn't propagate to tests appropriately ifiuc

pkg/resourcemanager/resourcemanager.go
patrickhuie19: agreed, this is awkward. resolved in next commit, and dropped the anonymous func in favor of a named one

pkg/resourcemanager/config_from_env.go
patrickhuie19: agreed, inverted the dependency

pkg/resourcemanager/config_from_env.go
patrickhuie19: yes, agreed

pkg/loop/metering_config.go
```diff
@@ -0,0 +1,38 @@
...
+// into every producer main. The CL_METER_* fields it reads are populated by
+// EnvConfig.AsCmdEnv from the core node's TOML [Metering] config; env is only
+// the LOOP transport.
+//
+// emitter is injected by the caller (can be process global durableemitter.GetGlobalEmitter())
+// rather than looked up here, so this mapping stays a pure function of its
+// inputs. A nil emitter is valid: the ResourceManager treats it as a no-op —
+// fail-open, so metering never blocks startup. A nil EnvConfig yields a zero
+// Config (metering disabled).
+func (e *EnvConfig) MeteringConfig(emitter resourcemanager.Emitter) resourcemanager.Config {
```
jmank88, line 18: Is this used yet? Now I'm questioning whether we need a helper at all - won't it just be called in one place, from the `loop.Server` during startup?

patrickhuie19: not yet merged, but here's a draft PR with its intended use in Capabilities: https://github.com/smartcontractkit/capabilities/pull/657/commits/07b548bd1855b99748f68aeabda9747c6f2ed00b#diff-403455efbeefabb9b70f2cae18606f1ccc9f752b35bac5139b4976480b4ca0bdR81

jmank88: Can we do it once for all loops in the shared code?

patrickhuie19: yes, 👍

russell-stern (approved):

jmank88 (approved):

PR #2253, opened 2026-07-15
cre-5476: pointer improvement

mchain0:
[cre-5476](https://smartcontract-it.atlassian.net/browse/CRE-5476)

### DIFF
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -1163,8 +1163,16 @@ func writeUInt32(memory []byte, ptr int32, val uint32) int64 {
 
 func truncateWasmWrite(caller *wasmtime.Caller, src []byte, ptr int32, size int32) int64 {
 	memory := wasmMemoryAccessor(caller)
-	if int32(len(memory)) < ptr+size {
+	if ptr < 0 || size < 0 || int64(ptr) > int64(len(memory)) {
+		return -1
+	}
+
+	// widen to int64 so a guest-supplied ptr/size near math.MaxInt32 cannot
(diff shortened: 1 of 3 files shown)


jmank88 (approved):

pkg/workflows/wasm/host/module.go
```diff
@@ -1175,15 +1183,17 @@ func truncateWasmWrite(caller *wasmtime.Caller, src []byte, ptr int32, size int3
...
 	}
 
 	if len(src) > int(maxSize) {
 		return -1
 	}
 
-	if int32(len(memory)) < ptr+maxSize {
+	// widen to int64 so a guest-supplied ptr near math.MaxInt32 cannot overflow
+	// the bounds check into a negative value and bypass it.
+	if int64(ptr)+int64(maxSize) > int64(len(memory)) {
```
pavel-raykov, line 1196: why this check? at this point `len(src) <= int(maxSize)` and below we access memory up to `ptr+int32(len(src))`, should not we check for `ptr+src > len(memory)` here?

mchain0: it's done this way to keep the security fix behavior-preserving

pavel-raykov: Sorry, but if this math does not make sense, then it is not clear what we are trying to fix

MStreet3: the math is worst case scenario and behavior preserving.  the fix is avoiding an overflow of adding two large int32s by casting int32 to int64 before adding.

MStreet3 (approved):

PR #2262, opened 2026-07-20
[CRE-5574] Reject duplicate IDs

cedric-cordenier:


pkg/workflows/wasm/host/execution.go
```diff
@@ -37,6 +37,8 @@ type execution[T any] struct {
 	nodeSeed             int64
 	donLogCount          uint32
 	nodeLogCount         uint32
+
+	usedCallbackIds map[string]bool
```
patrickhuie19: should this be a concurrent safe map?

pkg/workflows/wasm/host/execution.go
```diff
@@ -54,6 +56,14 @@ func (e *execution[T]) callCapAsync(ctx context.Context, req *sdkpb.CapabilityRe
 	defer e.lock.Unlock()
 	e.capabilityResponses[req.CallbackId] = ch
 
+	normalizedCallbackId := fmt.Sprintf("cap__%d", req.CallbackId)
```
patrickhuie19: IFIUC, the though with the `cap__` and `secret__` prefixes is that the CallbackIds should conflict?

nit: don't we generally try to name `IDs` over `Ids`?

pkg/workflows/wasm/host/execution.go
cedric-cordenier: It should be safely accessed yep, you can see we're doing everything under the lock we acquired to track capabilityResponses/secretsResponses

pkg/workflows/wasm/host/execution.go
cedric-cordenier: Yes 👍 I'll fix this

nolag (approved):

pkg/workflows/wasm/host/execution.go
```diff
@@ -49,9 +51,17 @@ func (e *execution[T]) callCapAsync(ctx context.Context, req *sdkpb.CapabilityRe
...
+
+	normalizedCallbackID := fmt.Sprintf("cap__%d", req.CallbackId)
+	if e.usedCallbackIDs[normalizedCallbackID] {
+		free()
+		return errors.New("duplicate callback id not allowed")
+	}
+	e.usedCallbackIDs[normalizedCallbackID] = true
+
+	ch := make(chan *sdkpb.CapabilityResponse, 1)
 	e.capabilityResponses[req.CallbackId] = ch
```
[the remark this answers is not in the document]
patrickhuie19: the map gets gc'd per execution, no-op imho

patrickhuie19 (approved):

PR #2263, opened 2026-07-20
pkg/settings/limits: raise missing org ID tenant log level from warn to error

jmank88:
Towards https://smartcontract-it.atlassian.net/browse/CRE-1707

Raise level, and add logs for when just doing a setting lookup, which was a silent failure before via omission of the key.

I also cleaned up some production loggers being used from tests (which just spams std out in real time, even when a test is passing)

### DIFF
--- a/pkg/settings/limits/bound.go
+++ b/pkg/settings/limits/bound.go
@@ -227,7 +227,7 @@ func (b *boundLimiter[N]) get(ctx context.Context) (tenant string, bound N, err 
 		if tenant == "" {
 			if !b.scope.IsTenantRequired() {
 				kvs := contexts.CREValue(ctx).LoggerKVs()
-				b.lggr.Warnw("Unable to get scoped bound limit due to missing tenant: failing open", append([]any{"scope", b.scope}, kvs...)...)
+				b.lggr.Errorw("Unable to get scoped bound limit due to missing tenant: failing open", append([]any{"scope", b.scope}, kvs...)...)
 				return
 			}
 			err = fmt.Errorf("unable to get scoped bound limit due to missing tenant for scope: %s", b.scope)
--- a/pkg/settings/toml.go
+++ b/pkg/settings/toml.go
@@ -3,12 +3,15 @@ package settings
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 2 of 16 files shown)


pkg/settings/limits/bound.go
```diff
@@ -227,7 +227,7 @@ func (b *boundLimiter[N]) get(ctx context.Context) (tenant string, bound N, err
 		if tenant == "" {
 			if !b.scope.IsTenantRequired() {
 				kvs := contexts.CREValue(ctx).LoggerKVs()
-				b.lggr.Warnw("Unable to get scoped bound limit due to missing tenant: failing open", append([]any{"scope", b.scope}, kvs...)...)
+				b.lggr.Errorw("Unable to get scoped bound limit due to missing tenant: failing open", append([]any{"scope", b.scope}, kvs...)...)
```
prashantkumar1982, line 230: As discussed offline, will this just spam the logs if the limit in question was designed to only work at a global scope?

jmank88: No, global limits won't enter this block b/c of the check at the top.

pkg/settings/toml.go
```diff
@@ -134,25 +137,40 @@ func (t *tomlSettings) getFirst(keys ...string) (string, error) {
...
 
 func (t *tomlGetter) GetScoped(ctx context.Context, scope Scope, key string) (value string, err error) {
 	keys, err := scope.rawKeys(ctx, key)
 	if err != nil {
-		return "", fmt.Errorf("failed to get raw keys: %w", err)
+		tme, ok := errors.AsType[tenantMissingError](err)
+		if ok && tme.Scope.IsTenantRequired() {
+			return "", fmt.Errorf("failed to get raw keys: %w", err)
+		} else {
+			t.lggr.Errorf("Settings lookup for "+key+" limited by missing tenant at scope "+tme.Scope.String(), "key", key, "err", err)
```
jmank88, line 172: @prashantkumar1982 These could be noisier than the other kind, since they are per lookup, but there are still not false positives.

pavel-raykov (approved):

prashantkumar1982 (approved):

pavel-raykov (approved):

kirqz23 (approved):

PR #2264, opened 2026-07-21
gomods -go fix ./...

jmank88:


observability-lib/api/notification-policy_test.go
```diff
@@ -7,10 +7,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Pointer[T any](d T) *T {
```
jmank88, line 10: I'm surprised that it removed this one, since it is exported. Is that OK?

pavel-raykov (approved):

thomaska (approved):

PR #2266, opened 2026-07-21
beholder: handle partial-delivery errors via metric and log suppression

pkcll:
With partial delivery enabled, per-event `PublishError` failures are now returned in the callback. This PR:

- Adds `batch.ErrorCodeFor` to `pkg/chipingress/batch/client.go`, classifying a send/queue error into a bounded `error_code` string via `errors.As` / `errors.Is` / `status.FromError`.
- Suppresses ERROR logging for per-event `PublishError` failures — detected directly via `errors.As(sendErr, &pubErr)` in beholder — since they're already captured by the `chip_ingress.events_dropped` metric.
- Continues to log other send/queue errors at ERROR, enriched with `error_code` and `error_reason` (`sendErr.Error()`).
- Adds an `error_code` attribute to the `chip_ingress.events_dropped` metric.

Self-contained: the chipingress-side classification helper (previously split out as its own PR, #2210, closed) now lives in this PR alongside its beholder consumer, so there's no cross-PR `go.mod` dependency to coordinate.

Jira: INFOPLAT-13901


pavel-raykov (dismissed):

pkg/chipingress/batch/client.go
```diff
@@ -51,6 +58,37 @@ const (
...
+// ErrorCodeFor returns a bounded error code string for a batch send/queue
+// error, suitable for use as a metric dimension.
+func ErrorCodeFor(err error) string {
+	if err == nil {
+		return ""
+	}
+
+	var pubErr *PublishError
+	if errors.As(err, &pubErr) {
+		return pubErr.Code.String()
```
thomaska, line 73: This can be problematic for `ErrCodeResultsMismatch` that is `chipingress.PublishErrorCode(-1)` ie. not in the 0-4 range of the proto enum, so this will return a cryptic "-1"
Maybe we could do explicit handling here:
```go
	var pubErr *PublishError
	if errors.As(err, &pubErr) {
		if pubErr.Code == ErrCodeResultsMismatch {
			return "results_mismatch"
		}
		return pubErr.Code.String()
	}
```

thomaska (dismissed):
👍 💯 Added a small comment, but overall sgtm!

thomaska (dismissed):

thomaska (approved):

jmank88 (approved):

PR #2267, opened 2026-07-21
Add resource attribute support to ChipIngress client

pkcll:
## Summary

Add resource-attribute propagation to the ChipIngress client so that node resource attributes from `[Telemetry.ResourceAttributes]` can reach the `cre` topic via two mechanisms:

1. **gRPC metadata headers** — raw keys forwarded as gRPC metadata (parity with `beholder__platform__messages` once the chip-ingress server forwards arbitrary metadata).
2. **CloudEvent extensions** — sanitized keys per the CloudEvents spec (`^[a-z0-9]+$`), surfaced as `ce_<name>` Kafka headers by the CloudEvents Kafka binding.

## Changes

- `pkg/chipingress/client.go`: add `NewEventWithOpts` / `WithResourceAttributeExtensions`, `SanitizeExtensionName`, and fix interceptor chaining so header providers compose.
- `pkg/chipingress/header_provider.go`: add `NewStaticHeaderProvider`, `SanitizeMetadataHeaders`, `SanitizeMetadataValue`.
- `pkg/chipingress/types.go`: add `reservedExtensionNames` and `reservedMetadataKeys`.
- Root `go.mod`: bump `pkg/chipingress` pseudo-version to this branch.

## Related PRs

- **Downstream (beholder wiring):** https://github.com/smartcontractkit/chainlink-common/pull/2216
- **Downstream (chainlink):** https://github.com/smartcontractkit/chainlink/pull/23014

## Test plan

- `go test ./pkg/chipingress/...`

## Notes

- gRPC metadata → Kafka header forwarding is unconfirmed on the chip-ingress server side; the CE-extension path guarantees `ce_<name>` headers immediately.
- The previous `Co-authored-by: Cursor` attribution has been removed from the commit history.


pkg/chipingress/client.go
```diff
@@ -267,9 +277,60 @@ func newHeaderInterceptor(provider HeaderProvider) grpc.UnaryClientInterceptor {
...
+	}
+	return b.String()
+}
+
+// WithResourceAttributeExtensions returns an EventOpt that sets a CloudEvent extension for each
+// entry in attrs, sanitizing keys via SanitizeExtensionName so they satisfy the CloudEvents
+// extension-name character set. Entries that sanitize to an empty string, or that collide with a
+// reserved extension name (see reservedExtensionNames), are skipped. Keys are applied in sorted
+// order so that if two distinct keys sanitize to the same name, the result is deterministic.
+func WithResourceAttributeExtensions(attrs map[string]string) EventOpt {
```
thomaska, line 306: Opus picked up this:
WithResourceAttributeExtensions (client.go) and SanitizeMetadataHeaders (header_provider.go) both implement the same "collect keys → sort → sanitize via SanitizeExtensionName → skip empty/reserved/dupe" loop, differing only in the extra reservedMetadataKeys check and what happens to a surviving key (event.SetExtension vs. sanitize-value-and-store). Worth extracting a shared helper that returns the deduplicated, sorted (sanitizedName, originalKey) pairs, with each caller supplying just its own "apply" step. Saves ~15 duplicated lines and means any future change to the collision/ordering rule (e.g. "last wins" instead of "first wins") only has to happen once.

thomaska (dismissed):
sgtm!

pkg/chipingress/client.go
```diff
@@ -267,9 +277,60 @@ func newHeaderInterceptor(provider HeaderProvider) grpc.UnaryClientInterceptor {
 	}
 }
 
+// EventOpt configures a CloudEvent after its well-known attributes have been set by NewEvent.
+type EventOpt func(*ce.Event)
+
+// SanitizeExtensionName lower-cases name and strips every rune outside [a-z0-9], the character
+// set the CloudEvents spec requires for extension attribute names.
+func SanitizeExtensionName(name string) string {
```
jmank88: Does this need to be exported?

pkcll: Addressed

pkg/chipingress/client.go
pkcll: Addressed

thomaska (dismissed):

pkg/chipingress/header_provider.go
```diff
@@ -108,6 +108,55 @@ func newStaticHeaderProvider(headers map[string]string, requireTLS bool) HeaderP
 	return &staticHeaderProvider{headers: headers, requireTLS: requireTLS}
 }
 
+// NewStaticHeaderProvider returns a HeaderProvider that always returns the given headers,
+// for use with WithHeaderProvider to attach fixed, non-auth gRPC metadata (e.g. resource
+// attributes) to every request.
+func NewStaticHeaderProvider(headers map[string]string) HeaderProvider {
```
thomaska, line 114: it would be useful to chain the creation of HeaderProvider with sanitization, considering how important it is. 
ie. Could we add a convenience Opt that composes sanitize + construct so the safe path is the only path, e.g.:
```
func WithResourceAttributeHeaders(attrs map[string]string) Opt {
	return WithHeaderProvider(NewStaticHeaderProvider(SanitizeMetadataHeaders(attrs)))
}
```

thomaska (dismissed):

4of9 (approved):

jmank88 (approved):

PR #2276, opened 2026-07-23
[DX-4553] Binary Caching for `loopinstall`

kalverra:
## Big Changes

* Add cache checks for plugins before building from source.
  * Supply cache dir through `cache-dir` flag or `CL_LOOPINSTALL_CACHE_DIR` env var. Defaults to `os.UserCacheDir()`
  * Builds hash key:
    ```go
    h := sha256.New()
    _, _ = fmt.Fprintf(
        h, 
        "%s|%s|%s|%s|%s|%s|%s", 
        pluginType, plugin.ModuleURI, plugin.GitRef, plugin.Flags, defaults.GoFlags, goos, goarch,
    )
    base := filepath.Base(filepath.Clean(plugin.ModuleURI))
    return fmt.Sprintf("%s-%x", base, h.Sum(nil)[:12])
    ```

## Small Changes

* Update standard workflow versions for CI
* Linting
* Update Go `1.26.2` -> `1.26.5`


jmank88 (dismissed):

jmank88 (dismissed):

jmank88 (dismissed):

.github/workflows/build_external.yml
```diff
@@ -148,7 +148,7 @@ jobs:
           go mod tidy
 
       - name: Install Nix
-        uses: cachix/install-nix-action@8887e596b4ee1134dae06b98d573bd674693f47c # v26
+        uses: cachix/install-nix-action@630ae543ea3a38a9a4166f03376c02c50f408342 # v31.11.0
```
chainchad: Calling this out as this is a multi-major version bump and wondering if this breaks compatibilty.

pkg/loop/cmd/loopinstall/cache.go
```diff
@@ -0,0 +1,322 @@
...
+	if plugin.BinaryURL != "" {
+		log.Printf("%s - cache miss, attempting binaryURL download (%s)", pluginKey, plugin.BinaryURL)
+
+		client := &http.Client{Timeout: 15 * time.Second}
+		req, err := http.NewRequest("GET", plugin.BinaryURL, nil)
+		if err != nil {
+			log.Printf("%s - binaryURL request creation failed (%v)", pluginKey, err)
+		} else {
+			if token := getAuthToken(); token != "" {
+				if strings.Contains(plugin.BinaryURL, "github.com") || strings.Contains(plugin.BinaryURL, "githubusercontent.com") || strings.Contains(plugin.BinaryURL, "amazonaws.com") {
```
chainchad: Instead of `Contains()`, we could be more strict/safe if we used [HasPrefix](https://pkg.go.dev/strings#HasPrefix).

pkg/loop/cmd/loopinstall/cache.go
```diff
@@ -0,0 +1,322 @@
...
+	Name               string `json:"name"`
+	BrowserDownloadURL string `json:"browser_download_url"`
+}
+
+type githubReleaseResponse struct {
+	Assets []githubReleaseAsset `json:"assets"`
+}
+
+// tryDownloadGitHubRelease attempts to download a pre-compiled binary from GitHub Releases.
+func tryDownloadGitHubRelease(plugin PluginDef, binaryName, goos, goarch, githubAPIBaseURL string) ([]byte, error) {
```
chainchad, line 121: Should we assume each binary release has a sha256sum file that we should check with it?

pkg/loop/cmd/loopinstall/cache.go
chainchad, line 1: Is there enough value to add this over just improving how we cache LOOPPs in the /chainlink core Dockerfile?

pkg/loop/cmd/loopinstall/cache.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+	userCache, err := os.UserCacheDir()
+	if err != nil || userCache == "" {
+		return filepath.Join(os.TempDir(), "loopinstall-cache"), nil
+	}
+	return filepath.Join(userCache, "loopinstall-cache"), nil
+}
+
+// computeCacheKey calculates a deterministic cache key for a plugin.
+func computeCacheKey(pluginType string, plugin PluginDef, defaults DefaultsConfig, goos, goarch string) string {
```
chainchad, line 38: Should we account for CL_PLUGIN_GOFLAGS and CL_PLUGIN_ENVVARS for the cache key?

chainchad (changes requested):

pkg/loop/cmd/loopinstall/cache.go
kalverra: Ya, makes sense

pkg/loop/cmd/loopinstall/cache.go
kalverra: We'd have to set up standard workflows for all of them that include a `.sha256` file.

I included this section as a fallback/future-proof if we did implement standard release pipelines for these down the line. If we want to consider that as part of the requirements for it, I'm down.

PR #2280, opened 2026-07-24
feat: add gofix Makefile target and CI check

pkcll:
- Add `gofix` Makefile target that runs `go fix ./...`
- Add `check-gofix` CI job that verifies `go fix ./...` produces no uncommitted changes
- Mark `check-gofix` with `continue-on-error: true` so it is non-blocking while evaluated
- Apply current `go fix` inlining changes to `pkg/beholder` tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,10 @@ modgraph: gomods
 	go install github.com/jmank88/modgraph@v0.1.0
 	./modgraph > go.md
 
+.PHONY: gofix
+gofix: ## Run go fix across all packages
+	go fix ./...
+
 .PHONY: dependabot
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -60,6 +60,10 @@ modgraph: gomods
 	go install github.com/jmank88/modgraph@v0.1.0
 	./modgraph > go.md
 
+.PHONY: gofix
+gofix: ## Run go fix across all packages
+	go fix ./...
```
jmank88: We can use gomods to cover all modules
```suggestion
	gomods -go fix ./...
```

jmank88 (approved):

PR #2288, opened 2026-07-28
chipingress: send whitelisted resource attributes as chainlink-* gRPC metadata

pkcll:
Follow-up to #2267, which added resource-attribute support to the ChipIngress client. Reworked 2026-09-14 from an open-keyspace `resource_` prefix to a **closed whitelist of fixed `chainlink-`* metadata headers**, per the gRPC metadata naming convention / header-design review: no operator-defined key ever becomes a header name.

## What this does

`WithResourceAttributeHeaders` now projects the configured resource attributes onto ten fixed, whitelisted gRPC metadata headers (the package-private `resourceAttributeHeaders` whitelist):

| attribute key (case-insensitive) | gRPC metadata header |
|---|---|
| `csa_public_key` | `chainlink-resource-csa-public-key` |
| `deployed_by` | `chainlink-resource-deployed-by` |
| `donID` | `chainlink-resource-don-id` |
| `host.name` | `chainlink-resource-host-name` |
| `internal_node_id` | `chainlink-resource-internal-node-id` |
| `node_id` | `chainlink-resource-node-id` |
| `platformEnv` | `chainlink-resource-platform-env` |
| `service.name` | `chainlink-resource-service-name` |
| `service.sha` | `chainlink-resource-service-sha` |
| `zone` | `chainlink-resource-zone` |

Attributes are attached once per request (they describe the producer, not any one event). Chip-ingress ([atlas#13201](https://github.com/smartcontractkit/atlas/pull/13201)) reads exactly these fixed header names and forwards them onto every Kafka record the request produces as `resource_<original attribute key>` (e.g. `chainlink-resource-service-name` → `resource_service.name`).

## Why a closed whitelist

- **The whitelist replaces the deny-list.** The header interceptor *appends* to outgoing metadata, so an attribute that could land on the CSA auth header's name would break authentication by sending a second value under it. With fixed, hand-named headers, `X-Beholder-Node-Auth-Token` is simply not in the whitelist and can never be emitted — no reserved-key set to maintain.
- **No open keyspace → no validation machinery.** The per-key charset validation, caps (32 attrs / 128B key / 512B value / 4KiB total), `DroppedAttribute` reporting, zap logger and otel drop counter from the previous revision are all deleted. What remains: values must be printable ASCII (grpc-go hard-fails the whole RPC otherwise) and are omitted, never rewritten.
- **Naming convention.** `chainlink-resource-` prefix (org-scoped `chainlink-` + category infix), lowercase hyphenated names; no `x-` (RFC 6648), no `_`/`.` in header names (nginx/Envoy/`textproto` hazards). The mapping is duplicated in chip-ingress (`chip-ingress/internal/constants`) and must stay in sync — same pattern as `authHeaderKey`.

## API changes vs the previous revision

- `SanitizeMetadataHeaders` returns to its released single-value signature (`map[string]string`); the `DroppedAttribute` return added earlier in this PR is removed.
- `SetResourceAttributeLogger` and the `chipingress.resource_attribute.dropped` counter are gone — the zap dependency they introduced is gone with them.
- The CloudEvent-extension path (`WithResourceAttributeExtensions`, `EventOpt`, `NewEventWithOpts`) stays deleted: resource attributes describe the producer, not events.

## Tests

- Whitelist translation pins all ten mappings (the mapping is the wire contract with chip-ingress).
- Case-insensitive matching; case-variant collisions resolve deterministically (sorted-first wins).
- Non-whitelisted keys — including the CSA auth header's name and gRPC-reserved keys — are omitted.
- Non-printable values are omitted; `TestSanitizeMetadataHeaders_AvoidsRPCFailure` proves an unsanitized value fails the RPC (`codes.Internal`) while the sanitized set passes.
- `TestClient_AuthHeaderCoexistsWithResourceAttributes`: auth token + whitelisted attributes + `x-include-nop-info` all arrive intact on one request.

Note: `pkg/chipingress/batch` `TestCallbacks` failures observed locally reproduce on `origin/main` — pre-existing, unrelated to this change.

Downstream: #2216 (beholder wiring, unchanged in shape — filtering happens here) → chainlink#23014 (pin bump).


4of9 (dismissed):

pkcll:
## Note on the `api-diff` failure (intentional, safe)

`api-diff / Analyze (pkg/chipingress)` fails with `enforce-compatible: true` because this PR removes 4 exported symbols that #2267 added. This is deliberate, and nothing downstream is affected:

**Removed symbols** (all belong to the per-event CloudEvent-extension path this PR deletes — resource attributes describe the producer, not events, and travel as whitelisted `chainlink-*` gRPC metadata instead):

| Symbol | Replacement |
|---|---|
| `WithResourceAttributeExtensions` | none — extension path deleted by design |
| `EventOpt` | none |
| `NewEventWithOpts` | `NewEvent` (signature unchanged) |
| `SanitizeMetadataValue` | validate-don't-rewrite: non-printable values are omitted, so no value-rewriter is needed |

**Unchanged signatures:** `WithResourceAttributeHeaders(map[string]string) Opt`, `SanitizeMetadataHeaders(map[string]string) map[string]string`, `NewStaticHeaderProvider`, `NewEvent` — which is why the downstream wiring in #2216 needed no code change.

**Why the break is safe at runtime:**

1. The resource-attribute transport has **no deployed consumers** — the only wiring (#2216) is unmerged, so no node sends these attributes today.
2. **No downstream repo pins a chipingress version containing the removed symbols.** Current default-branch pins: chainlink `v0.0.11-0.20260724142814`, chainlink-evm and chainlink-solana same (indirect), atlas chip-config `v0.0.11-0.20260610...` — all predate #2267 (merged 2026-07-27).
3. **Zero call sites outside this repo.** Org-wide code search for `WithResourceAttributeExtensions`, `NewEventWithOpts`, `SanitizeMetadataValue` returns hits only in `pkg/chipingress` itself (source + tests).

So the incompatibility is formal only — it exists relative to this repo's `main` exported surface, not in any consumer's module graph. Requesting maintainer sign-off on the api-diff gate for this PR; if strict compatibility is preferred instead, the alternative is restoring the 4 symbols as deprecated wrappers, which keeps the rejected extension path alive for zero users.

pkg/chipingress/types.go
```diff
@@ -13,33 +13,33 @@ import (
...
+// chainlink-resource-service-name becomes resource_service.name). The set is
+// closed on purpose: because no operator-defined key can ever become a header name, no attribute
+// can shadow a reserved gRPC metadata key such as the CSA auth token's, and the server needs no
+// deny-list to keep resource attributes away from "ce_" or identity headers it derives from the
+// verified auth token.
+//
+// The same mapping exists in chip-ingress (chip-ingress/internal/constants). Duplicating it across
+// repositories is deliberate, matching how authHeaderKey is already spelled in both pkg/beholder
+// and pkg/chipingress; the two must stay in sync or forwarding silently stops.
+var ResourceAttributeHeaders = map[string]string{
```
thomaska, line 32: I think we could lowercase this and make it package private? My assumption was different based on the comment "closed whitelist" :/ 
is this supposed to be accessed by callers outside the `chipingress` package?
in that case maybe we could expose an accessor?
```go
func ResourceAttributeHeader(key string) (string, bool) {
	header, ok := resourceAttributeHeaders[strings.ToLower(key)]
	return header, ok
}
```

thomaska (dismissed):
👍 🔢 It would be good to review the comment, but as far as I can tell there is no blocker here :) 
One side comment though: I'm not used to this amount of comments in the code! Is this a chainlink-common or a Go pattern!? Comments need maintenance and there is always the risk the code drifts 🤷

pkg/chipingress/types.go
pkcll: Good catch — done in 9e407e06f: the map is now package-private (`resourceAttributeHeaders`). Nothing outside the package referenced it (the one test assertion is now covered by the exact-output check in `TestSanitizeMetadataHeaders`, which still pins all ten translations). An exported mutable map contradicted the closed-whitelist contract, as you said — any importer could have added entries at runtime.

I did not add the accessor for now since no caller needs lookup today — `SanitizeMetadataHeaders` / `WithResourceAttributeHeaders` remain the public surface. Easy to add `ResourceAttributeHeader(key)` later if beholder logging or similar ever needs it.

thomaska (dismissed):

pkg/chipingress/header_provider.go
```diff
@@ -111,48 +113,67 @@ func newStaticHeaderProvider(headers map[string]string, requireTLS bool) HeaderP
...
+// sorted order of the original keys wins, deterministically.
 func SanitizeMetadataHeaders(in map[string]string) map[string]string {
-	out := make(map[string]string, len(in))
-	for _, pair := range sanitizeResourceAttributeKeys(in, reservedMetadataKeys) {
-		out[pair.name] = SanitizeMetadataValue(in[pair.key])
+	keys := make([]string, 0, len(in))
+	for k := range in {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
jmank88:
```suggestion
	keys := slices.Sorted(maps.Keys(in))
```

pkcll: Applied in f6651ffe — thanks.

pkcll: Applied in f6651ffe — thanks.

4of9 (approved):

jmank88 (approved):

PR #2299, opened 2026-07-31
PLEX-3289: pollOneoff fix -- prevent memory amplification

cedric-cordenier:


pkg/workflows/wasm/host/wasip1.go
```diff
@@ -137,96 +138,104 @@ const (
...
 // It doesn't actually sleep though, and will instead advance our fake clock by the sleep duration.
-func pollOneoff(caller *wasmtime.Caller, subscriptionptr int32, eventsptr int32, nsubscriptions int32, resultNevents int32) int32 {
-	if nsubscriptions <= 0 || nsubscriptions > max(math.MaxInt32/subscriptionLen, math.MaxInt32/eventsLen) {
-		return ErrnoInval
-	}
+func createPollOneoff(ctx context.Context, cfg *ModuleConfig) func(caller *wasmtime.Caller, subscriptionptr int32, eventsptr int32, nsubscriptions int32, resultNevents int32) int32 {
+	return func(caller *wasmtime.Caller, subscriptionptr int32, eventsptr int32, nsubscriptions int32, resultNevents int32) int32 {
+		maxSubscriptions, err := cfg.MaxSubscriptionsLimiter.Limit(ctx)
+		if err != nil {
+			return ErrnoInval
```
jmank88: Does it matter if the error was hitting a limit vs. failing to lookup/apply the limit? Or do we return `ErrnoInval` either way?

jmank88: Er nvm, this is just the lookup

jmank88: Ah, well if you are only doing lookup, then you could avoid creating the whole limiter, and just do the lookup directly yourself. Or you could call `Check` on the limiter instead of doing the math yourself, and then we will get metrics on the usage too.

cedric-cordenier: Usage metrics would be good -- will move to using check!

jmank88 (approved):

jmank88 (approved):

pkg/workflows/wasm/host/wasip1.go
```diff
@@ -137,96 +137,102 @@ const (
...
+
+				// Structure of event, per:
+				// https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md#-subscription_clock-struct
+				// - 0-8: clock id
+				// - 8-16: timeout
+				// - 16-24: precision
+				// - 24-32: flag
+				newTimeout := binary.LittleEndian.Uint64(argBuf[8:16])
+				flag := binary.LittleEndian.Uint16(argBuf[24:32])
+
```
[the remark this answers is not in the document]
cedric-cordenier: I'm not going to address this, this works as is, and matches the current docs which specify the flag as taking 8 bytes.

mchain0 (approved):

PR #2300, opened 2026-07-31
CRE-5747: Recover wasm panics in the host module

cedric-cordenier:


pkg/workflows/wasm/host/module.go
```diff
@@ -642,6 +651,21 @@ func (m *module) Run(ctx context.Context, request *wasmdagpb.Request) (*wasmdagp
 	return runWasm(ctx, m, request, setMaxResponseSize, linkLegacyDAG, nil, *m.cfg.Timeout)
 }
 
+// callStart invokes the wasm module's _start function, but recovers any
+// panic. It's sufficient to just recover here, wasmtime-go already recovers panics
+// that originated from guest calls to the host.
+// See https://pkg.go.dev/github.com/bytecodealliance/wasmtime-go/v47#Func.Call.
+func callStart(m *module, start *wasmtime.Func, store *wasmtime.Store) (result interface{}, err error) {
```
pavel-raykov: what happens if the same module always panics - will it be run indefinitely?

pkg/workflows/wasm/host/module.go
```diff
@@ -378,10 +380,15 @@ func NewModule(ctx context.Context, modCfg *ModuleConfig, binary []byte, opts ..
 		return nil, fmt.Errorf("failed to check decompressed binary size limit: %w", err)
 	}
 
-	return newModule(modCfg, binary)
+	metrics, err := newModuleMetrics()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create module metrics: %w", err)
```
mchain0, line 391: what if that fails? - no workflows can run.

maybe sth like in `pkg/capabilities/base_trigger.go:113-116` would be better?

```
metrics, err := NewBaseTriggerBeholderMetrics(capabilityId)
if err != nil {
    metrics = &noopBaseTriggerMetrics{}
}
```

pkg/workflows/wasm/host/module.go
```diff
@@ -642,6 +650,27 @@ func (m *module) Run(ctx context.Context, request *wasmdagpb.Request) (*wasmdagp
 	return runWasm(ctx, m, request, setMaxResponseSize, linkLegacyDAG, nil, *m.cfg.Timeout)
 }
 
+// callStart looks up and invokes the wasm module's _start function, but
+// recovers any panic. It's sufficient to just recover here, wasmtime-go
```
mchain0, line 661: nit. / not quite, suggestion: "wasmtime-go catches panics at the C/Go FFI boundary to avoid undefined behavior, then re-panics in Go. We recover here to convert that into an error instead of crashing the process"

pkg/workflows/wasm/host/module_metrics.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+type moduleMetricsImpl struct {
+	hostFnPanicRecoveredCount metric.Int64Counter
+}
+
+var _ moduleMetrics = &moduleMetricsImpl{}
+
+func newModuleMetrics() (moduleMetrics, error) {
+	hostFnPanicRecoveredCount, err := beholder.GetMeter().Int64Counter("platform_wasm_host_panic_recovered")
```
mchain0: nit / following the patterns we got elsewhere should be `platform_wasm_host_panic_recovered_total`

mchain0: nit / metric.WithDescription() could be helpful

pkg/workflows/wasm/host/module_test.go
```diff
@@ -685,3 +685,52 @@ func Test_CallAwaitRace(t *testing.T) {
...
+	  (func (export "_start")
+	    call $panic_me))`
+	wasmBytes, err := wasmtime.Wat2Wasm(wat)
+	require.NoError(t, err)
+
+	engine := wasmtime.NewEngine()
+	mod, err := wasmtime.NewModule(engine, wasmBytes)
+	require.NoError(t, err)
+
+	store := wasmtime.NewStore(engine)
```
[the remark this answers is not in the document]
mchain0: +1

pkg/workflows/wasm/host/module.go
```diff
@@ -642,6 +650,27 @@ func (m *module) Run(ctx context.Context, request *wasmdagpb.Request) (*wasmdagp
...
+// already recovers panics that originated from guest calls to the host.
+// See https://pkg.go.dev/github.com/bytecodealliance/wasmtime-go/v47#Func.Call.
+func callStart(m *module, instance *wasmtime.Instance, store *wasmtime.Store) (result interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			m.cfg.Logger.Errorw("panic during wasm execution", "panic", r)
```
mchain0, line 668: nil / very defensive - m.cfg and m.metrics being nil check. it's rather not a real risk since recovering does `newModule()`

pkg/workflows/wasm/host/module_test.go
```diff
@@ -685,3 +685,49 @@ func Test_CallAwaitRace(t *testing.T) {
 
 	wg.Wait()
 }
+
+// Test_callStart_RecoversHostFunctionPanic reproduces a panic raised by a host
+// function invoked from wasm (as would happen from a bug in e.g.
+// createCallCapFn, exec.log, etc.) and verifies that callStart recovers it
+// into a normal error instead of crashing the process.
+func Test_callStart_RecoversHostFunctionPanic(t *testing.T) {
```
mchain0: nit / no integration test through runWasm/Execute to confirm a host-fn panic during a real execution doesn't crash the process and returns an error to the caller

pkg/workflows/wasm/host/module.go
cedric-cordenier: That shouldn't happen tbh; we use this pattern of erroring if we fail to instantiate a metrics struct throughout chainlink. This is analogous to us returning an error in the constructor if we don't have a logger for instance, it's a function of incorrect setup + configuration; not something that will fail sporadically.

pkg/workflows/wasm/host/module.go
cedric-cordenier: Yes, I agree 👍

pkg/workflows/wasm/host/module_test.go
cedric-cordenier: Yeah, I didn't have a good idea how to do this tbh; if there was a way to trigger a panic via the module legitimately, we'd presumably fix the bug and make the test fail as a result. Let me think a bit what our options are here, since it would be great to have a proper test

pkg/workflows/wasm/host/module.go
cedric-cordenier: @pavel-raykov Yes, that's correct 👍

pkg/workflows/wasm/host/module_test.go
cedric-cordenier: OK, with some dependency injection this is sorted 👍

mchain0 (approved):

george-dorin (approved):

PR #2309, opened 2026-08-05
Updated protos

russell-stern:


nolag (commented):
I would just set to max uint32 and allow it to decrement, they are't making over 4 billion calls. Keep the code simpler.

nolag (approved):
I don't recall the audit issue with -ve numbers being unlimited but I know it was called out so I'll approve.

cedric-cordenier (approved):

PR #2316, opened 2026-08-10
Add a config flags to standardize how we do things

nolag:
TBD, should we have common structs for things like DB, OCR, etc? EVM and the like would then go in chainlink-evm.


pkg/config/configdoc/format.go
```diff
@@ -0,0 +1,53 @@
...
+// text with its marker stripped, the table's name, or the field's key.
+type Line struct {
+	Kind LineKind
+	Text string
+}
+
+// Format is a configuration file syntax. It both writes the pieces of a document (so callers
+// can assemble one from Go structs) and recognizes them when reading one back (so a
+// hand-written document can be turned into documentation). Implement it to document a format
+// other than TOML; see TOML for the reference implementation.
```
jmank88, line 29: What else did you have in mind? I thought we were leveraging TOMLs ease of parsing pretty directly here

jmank88: From call: I'm worried about this being premature abstraction, but only because it is public API. If this is useful for readability besides, then keeping it internal and evolving until it is ready to be public would be fine.

PR #2325, opened 2026-08-18
Remove obsolete Functions relayer methods

bolekk:


pavel-raykov (approved):

pkg/types/relayer.go
```diff
@@ -314,7 +314,6 @@ type Relayer interface {
 
 	NewMedianProvider(ctx context.Context, rargs RelayArgs, pargs PluginArgs) (MedianProvider, error)
 	NewMercuryProvider(ctx context.Context, rargs RelayArgs, pargs PluginArgs) (MercuryProvider, error)
-	NewFunctionsProvider(ctx context.Context, rargs RelayArgs, pargs PluginArgs) (FunctionsProvider, error)
```
jmank88, line 317: Let's avoid unnecessary breaking changes. These can still cause compilation problems in unrelated things, like patch releases when we try to backport.

jmank88 (changes requested):

bolekk:
ok makes sense, we don't have to clean this one up

PR #2326, opened 2026-08-18
fixes

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-5694


pkg/chipingress/batch/client.go
jmank88, line 1: Hmm now this doesn't look right to me...
- [x] double check

pkg/workflows/artifacts/artifacts_test.go
```diff
@@ -48,7 +48,7 @@ func (s *ArtifactsTestSuite) TestArtifacts() {
 	// Compare the keccak256 hash of the binary data against a value produced by
 	// the pinned Go toolchain (see GetBuildCmd in utils.go). Because Compile sets
 	// GOTOOLCHAIN from the nearest go.mod, this hash is stable across machines.
-	expKeccak256Hash, err := hex.DecodeString("fe4c48e42801b1425d771e4449c2b28551f4f433f33439a54ed0786f61085cf0")
```
pavel-raykov, line 51: why change the hash?

jmank88: Evidently it compiles differently now because of go version patch bump 🤷

pavel-raykov: I see, thanks!

pavel-raykov (approved):

pavel-raykov (dismissed):

pkcll (dismissed):

kirqz23 (dismissed):

pkg/loop/internal/relayer/aptos.go
```diff
@@ -112,6 +112,9 @@ func newAptosServer(impl types.AptosService, b *net.BrokerExt) *aptosServer {
 }
 
 func (s *aptosServer) AccountAPTBalance(ctx context.Context, req *aptospb.AccountAPTBalanceRequest) (*aptospb.AccountAPTBalanceReply, error) {
+	if len(req.Address) != 32 {
```
amit-momin: nit: Might be better to use the aptos [AccountAddressLength](https://github.com/smartcontractkit/chainlink-common/blob/2fb18e352bf84ae93ffc705e7e7df8a148549b27/pkg/types/chains/aptos/aptos.go#L6) here

pkg/loop/internal/relayerset/aptos.go
```diff
@@ -66,6 +66,10 @@ func (as *aptosServer) LedgerVersion(ctx context.Context, _ *aptospb.LedgerVersi
 }
 
 func (as *aptosServer) AccountAPTBalance(ctx context.Context, req *aptospb.AccountAPTBalanceRequest) (*aptospb.AccountAPTBalanceReply, error) {
+	if len(req.Address) != aptos.AccountAddressLength {
+		return nil, fmt.Errorf("invalid address length, must be 20 bytes but got: %d", len(req.Address))
```
amit-momin: I think this error message is inaccurate for Aptos. Think the expected length is 32 but might be better to just directly put `aptos.AccountAddressLength` into the error

amit-momin (dismissed):

nolag (dismissed):

pkg/billing/workflow_client.go
```diff
@@ -105,6 +105,10 @@ func NewWorkflowClient(lggr logger.Logger, address string, opts ...WorkflowClien
 		jwtGenerator: cfg.jwtGenerator,
 	}
 
+	if wc.jwtGenerator != nil && wc.tlsCert == "" {
```
patrickhuie19: i have to think about this, it may break the local-cre

pkg/storage/workflow_client.go
```diff
@@ -133,6 +134,10 @@ func NewWorkflowClient(lggr logger.Logger, address string, opts ...WorkflowClien
 		jwtGenerator: cfg.jwtGenerator,
 	}
 
+	if wc.jwtGenerator != nil && wc.tlsCert == "" && !testing.Testing() {
```
timothyF95: nit `!testing.Testing()` seems like a code smell?

timothyF95 (dismissed):

pkg/storage/workflow_client.go
jmank88: True this is a bit hacky, but the only risk is lack of test coverage 🤷 . Alternatively we can patch the tests that started failing from this stricter validation.

pkg/chipingress/batch/client.go
```diff
@@ -356,7 +356,8 @@ func (b *Client) sendBatch(ctx context.Context, messages []*messageWithCallback)
 			if err != nil {
 				b.log.Errorw("failed to publish batch", "error", err)
 				b.completeBatchCallbacks(batchMessages, err)
-			} else if !b.transactionEnabled && resp != nil && len(resp.Results) > 0 {
+			} else if !b.transactionEnabled {
```
patrickhuie19, line 359: was the previous issue here completeBatchCallbacksFromResults was not nil safe in resp.Results?

pkg/durableemitter/durable_emitter.go
```diff
@@ -262,7 +266,7 @@ func NewDurableEmitter(
 	}.NewServiceEngine(lggr)
 
 	if cfg.InsertBatchSize > 0 {
-		if bi, ok := store.(BatchInserter); ok {
+		if bi, ok := store.(BatchInserter); ok && isStoreBatch {
```
patrickhuie19: don't we do this now in line 239? should this just be 
`if isStoreBatch`?

pkg/jsonrpc2/types.go
```diff
@@ -76,7 +76,7 @@ func (r *Request[Params]) Digest() (string, error) {
 	}
 
 	canonicalJSONBytes := jsontext.Value(JSONBytes)
-	err = canonicalJSONBytes.Canonicalize()
+	err = canonicalJSONBytes.Canonicalize(jsontext.CanonicalizeRawInts(false), jsontext.CanonicalizeRawFloats(false))
```
patrickhuie19: what consumers use jsonrpc2? is there any hashing of these requests that would break?

pkg/workflows/artifacts/artifacts_test.go
patrickhuie19: are there implications here for verifiable workflowIDs?

pkg/jsonrpc2/types.go
jmank88: I think it would only break in cases where the hash was ambiguous before, because these numerical canonicalization funcs are lossy.

pkg/durableemitter/durable_emitter.go
jmank88: Yeah can grab the reference on 239 too.

pkg/chipingress/batch/client.go
jmank88: No nil bugs, just logically still needs to be called even with nil or empty results.

pkg/storage/workflow_client.go
jmank88: Reverted for now

pkg/billing/workflow_client.go
jmank88: Reverted

pkg/workflows/artifacts/artifacts_test.go
jmank88: I don't think we will have ID problems unless we have overly brittle tests that don't expect things to be recompiled 🤷

pkg/jsonrpc2/types.go
patrickhuie19: we do full payload hashing in many parts of the codebase, so caution here is warranted

jmank88: Reverted

amit-momin (approved):

patrickhuie19 (approved):

bolekk (approved):

pkcll (approved):

PR #2328, opened 2026-08-19
Added spend_value_in_gas_units to Metering report

russell-stern:


pavel-raykov (approved):

dhaidashenko (approved):

pkg/capabilities/capabilities.go
```diff
@@ -152,9 +152,10 @@ func ResponseToReportData(workflowExecutionID, referenceID string, responsePaylo
 }
 
 type MeteringNodeDetail struct {
-	Peer2PeerID string
-	SpendUnit   string
-	SpendValue  string
+	Peer2PeerID          string
+	SpendUnit            string
+	SpendValue           string
+	SpendValueInGasUnits string
```
patrickhuie19, line 158: is this struct used when serializing/deserializing a message from Don2DON? thinking about backwards compatability

russell-stern: It's not, it's in the metadata which is stripped before the hash is calculated

patrickhuie19: the struct is only used in `CapabilityResponseFromProto` so I think we should be good here.

patrickhuie19 (approved):

PR #2338, opened 2026-08-24
Add gcpkms keystore implementation

AmrMohamedRezk:
[CCIP-13120] https://smartcontract-it.atlassian.net/browse/CCIP-13120

**Add a Google Cloud KMS keystore backend**

Adds keystore/gcpkms, a keystore.Reader + keystore.Signer implementation backed by Google Cloud KMS, alongside the existing AWS backend in keystore/kms.

[CCIP-13120]: https://smartcontract-it.atlassian.net/browse/CCIP-13120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


keystore/reader.go
```diff
@@ -24,7 +24,7 @@ type GetKeyResponse struct {
 
 // Reader is the interface for reading keys from the keystore.
 // GetKeys returns all keys in the keystore if no names are provided, or the keys with the given names.
-// Keys are sorted by name in lexicographic order.
+// The order of returned keys is implementation-specific; callers must not rely on a particular ordering.
```
makramkd: Hm this seems like a breaking change, can we not satisfy this requirement (maybe sort client side)?

AmrMohamedRezk: No that shouldn't break any existing behavior. The old AWS behavior is as is. The new behavior doesn't, this is just to make sure clients don't always assume it's lexicographic order regardless of the backend.

makramkd: That's what I mean - if previous clients of this interface assume the sortation is lexicographical, and they upgrade and they don't get lexicographical, wouldn't their code silently break?

keystore/gcpkms/keystore.go
```diff
@@ -0,0 +1,418 @@
...
+var castagnoliTable = crc32.MakeTable(crc32.Castagnoli)
+
+// crc32c returns the CRC32C checksum of data, matching the value Google Cloud KMS uses in its
+// *_crc32c integrity fields.
+func crc32c(data []byte) int64 {
+	return int64(crc32.Checksum(data, castagnoliTable))
+}
+
+// checkCrc32c verifies that a received CRC32C checksum matches the computed value of the data.
+// This is Google's recommended integrity check for responses returned by Cloud KMS. A missing
```
makramkd: Can you link the docs where it mentions this? Would be useful for reviewers (and maybe auditors).

keystore/gcpkms/keystore.go
```diff
@@ -0,0 +1,418 @@
...
+// When no key names are given, the configured key ring is listed and keys that this keystore
+// cannot use — another purpose, an unsupported algorithm, or no enabled version — are skipped,
+// since a key ring may hold keys that have nothing to do with this keystore. Explicitly requested
+// keys always surface their errors.
+//
+// Listing requires a client that implements [KeyRingLister] and credentials holding a ring-wide
+// cloudkms.cryptoKeys.list. Deployments that follow least privilege grant neither and should pass an
+// explicit key allowlist instead.
+func (k *keystoreSignerReader) GetKeys(ctx context.Context, req keystore.GetKeysRequest) (keystore.GetKeysResponse, error) {
+	keyNames := append([]string(nil), req.KeyNames...)
```
makramkd: `req.KeyNames` is already a `[]string`, why do the re-append? You can do `len(s)` on a slice even though its `nil` (length would be zero).

keystore/gcpkms/keystore.go
```diff
@@ -0,0 +1,418 @@
...
+			return keystore.SignResponse{}, fmt.Errorf("signing response has name %q, expected %q", sig.Name, versionName)
+		}
+		if !sig.VerifiedDataCrc32C {
+			return keystore.SignResponse{}, errors.New("data CRC32C checksum was not verified by Cloud KMS")
+		}
+		if err = checkCrc32c(sig.Signature, sig.SignatureCrc32C); err != nil {
+			return keystore.SignResponse{}, fmt.Errorf("signature for key %s: %w", req.KeyName, err)
+		}
+		// Ed25519 signatures from Cloud KMS are already in the correct format (64 bytes).
+		if len(sig.Signature) != 64 {
```
makramkd: nit: can use `ed25519.SignatureSize`

keystore/gcpkms/client.go
```diff
@@ -0,0 +1,123 @@
...
+	KeyRingLister
+	Close() error
+}
+
+// ClientOptions contains options for creating a Cloud KMS client.
+type ClientOptions struct {
+	// CredentialsFile is the path to a GCP service account JSON key. Local development only —
+	// leave empty in production, where credentials come from the default credential chain
+	// (GKE Workload Identity, GCE instance/service accounts, or GOOGLE_APPLICATION_CREDENTIALS).
+	CredentialsFile string
```
makramkd: If one could use GOOGLE_APPLICATION_CREDENTIALS for local dev, can we get rid of this credentials file?

keystore/gcpkms/client.go
```diff
@@ -0,0 +1,123 @@
...
+// If CredentialsFile is specified, it uses service-account-key-based authentication (local dev).
+// Otherwise, it uses Application Default Credentials (Workload Identity in production, etc.).
+func NewClient(ctx context.Context, opts ClientOptions) (ClientWithClose, error) {
+	var clientOpts []option.ClientOption
+	if opts.CredentialsFile != "" {
+		// WithCredentialsFile is deprecated upstream because long-lived key files are a standing
+		// credential-leak risk. It is kept for the local-development path only; production leaves
+		// CredentialsFile empty and authenticates through Application Default Credentials.
+		//nolint:staticcheck // deliberate: local-development-only service-account key support
+		clientOpts = append(clientOpts, option.WithCredentialsFile(opts.CredentialsFile))
```
makramkd: Ideally we would remove this if its deprecated, and if an env var can be used instead. Those should be settable in devenv anyway via the testcontainers overrides

keystore/gcpkms/client.go
```diff
@@ -0,0 +1,115 @@
...
+// deployment that names its keys explicitly should never be forced to implement, or be granted, a
+// ring-wide list.
+type KeyRingLister interface {
+	ListCryptoKeys(ctx context.Context, keyRingName string) ([]*kmspb.CryptoKey, error)
+}
+
+// ClientWithClose is the client returned by NewClient: the full Cloud KMS surface, including key
+// ring listing, plus the underlying transport lifecycle. Whether listing actually succeeds is a
+// matter of the credentials' IAM bindings, not of the Go type.
+type ClientWithClose interface {
```
jmank88: Do the sub-interfaces need to be defined separately? If so, can we use `Client` here and name the other one differently?

keystore/gcpkms/keystore.go
```diff
@@ -0,0 +1,455 @@
...
+func signingKeyNames(listedKeys []*kmspb.CryptoKey) ([]string, error) {
+	keyNames := make([]string, 0, len(listedKeys))
+	for _, key := range listedKeys {
+		if key == nil || key.Name == "" {
+			return nil, errors.New("crypto key without a name returned by Cloud KMS")
+		}
+		if key.Purpose != kmspb.CryptoKey_ASYMMETRIC_SIGN {
+			continue
+		}
+		if template := key.VersionTemplate; template != nil {
```
pavel-raykov: what happens if the template is nil? Should the key also be discarded in this case?

keystore/gcpkms/keystore.go
```diff
@@ -0,0 +1,455 @@
...
+		return fmt.Errorf("CRC32C integrity check failed: computed %d, received %d", want, received.Value)
+	}
+	return nil
+}
+
+type keystoreSignerReader struct {
+	client      Client
+	keyRingName string
+
+	mu sync.RWMutex
```
pavel-raykov: why do you need all these caching mechanism? Can't you include the key version in its name when working with the keystore?

keystore/gcpkms/keystore.go
```diff
@@ -0,0 +1,455 @@
...
+	return publicKeyBytes, nil
+}
+
+// signingKeyNames filters a key ring listing down to the asymmetric signing keys this keystore can
+// use. Key rings are commonly shared, so keys with another purpose or an unsupported algorithm are
+// skipped rather than failing the whole listing.
+func signingKeyNames(listedKeys []*kmspb.CryptoKey) ([]string, error) {
+	keyNames := make([]string, 0, len(listedKeys))
+	for _, key := range listedKeys {
+		if key == nil || key.Name == "" {
```
pavel-raykov: Why return an error here? Can't you continue in the same way you do it below?

AmrMohamedRezk: Agree, no strong reason will drop.

keystore/gcpkms/keystore.go
AmrMohamedRezk: Good catch, VersionTemplate is optional in the API. I will address.

keystore/gcpkms/keystore.go
AmrMohamedRezk: We can, It doesn't remove the need for the pin, though. `ListCryptoKeys` returns `CryptoKeys`, not versions, so supporting ring listing means something must pick a version to report a public key for. And having `GetKeys` return the versioned name would break `keystore.Reader`, which specifies it returns "the keys with the given names". The pin keeps GetKeys and Sign agreeing without changing identities; it's one map under a mutex the public-key cache already needed, and a no-op when version names are configured.

The second cache is unrelated: public keys keyed by version name. Those are immutable, and it saves an RPC per signature.

Happy to drop bare-CryptoKey support entirely if you'd prefer it would delete the pin and most of latestEnabledVersion, at the cost of making every rotation a config change.

keystore/gcpkms/client.go
AmrMohamedRezk: Will drop the interface and replace it by an adapter.

keystore/gcpkms/keystore.go
pavel-raykov: > And having GetKeys return the versioned name would break keystore.Reader, which specifies it returns "the keys with the given names".

So far keystore has been implemented as not-version aware. Unless you would need it in your application, I don't see a reason to introduce this complextity. Hence, I would propose to just use "key/v1" and "key/v2" as fully qualified key names. It will be a client responsibility to choose a correct one.

> The second cache is unrelated: public keys keyed by version name. Those are immutable, and it saves an RPC per signature.

Sorry, I don't get it - can't I just call google keystore to sign with "key/v1" and simply fail if this key (or its version) is not available?

pavel-raykov: Should we also remove the caching layer then? (It only saves one lookup RPC, we can always reintroduce it later)

pavel-raykov (dismissed):

michelemin (approved):
LGTM

pavel-raykov (approved):

PR #2339, opened 2026-08-25
implement last-known-good fallback for dynamic limit settings

tarcisiozf:
This pull request improves the behavior and documentation of the settings limiters to ensure that, on a settings read failure, the limiters return the compiled default value alongside the error, rather than discarding the value. This makes the error advisory rather than fatal, allowing callers to proceed with a sensible default even when the settings service is unavailable. The changes also add comprehensive tests to verify this behavior across all limiter types.

**Behavioral improvements to limiter error handling:**

* All limiter `Limit` methods (`boundLimiter`, `rangeLimiter`, `gateLimiter`, `timeLimiter`) now return the resolved value (including the compiled default on error) alongside any error, rather than discarding the value on a settings read failure. This ensures callers can always use the returned value unless the tenant is missing.
* The `WithTimeout` method in `timeLimiter` now returns a usable context and the resolved default timeout even when a settings read fails, with the error being advisory. 

**Documentation updates:**

* The `Limiter` interface documentation is updated to clarify that returned values are always usable even when an error is present, except when the tenant is missing.


pkg/settings/limits/updater.go
```diff
@@ -90,18 +112,33 @@ func (u *updater[N]) updateLoop(ctx context.Context) {
...
+				u.setLast(limit)
 			}
 			u.recordLimit(ctx, limit)
 			if u.onLimitUpdate != nil {
 				u.onLimitUpdate(ctx)
 			}
 
 		case update := <-updates:
+			limit := update.Value
 			if update.Err != nil {
```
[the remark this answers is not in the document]
tarcisiozf: this is the original behavior and I'm going to keep it, as the change will make it worse by calling setLast with a empty value

pkg/settings/limits/updater.go
```diff
@@ -25,6 +26,9 @@ type updater[N any] struct {
 	stopCh    services.StopChan
 	done      chan struct{}
 	cancelSub func() // optional
+
+	// lastGoodValue is the fallback used on a read failure, instead of Setting.DefaultValue.
+	lastGoodValue atomic.Pointer[N]
```
jmank88: Why is this required? We should already be returning the correct default alongside any errors at a low level (`GetOrDefault`). Isn't returning that up the stack sufficient?

tarcisiozf: moving back to compiled default value

pkg/settings/limits/bound.go
```diff
@@ -213,14 +213,11 @@ func (b *boundLimiter[N]) Limit(ctx context.Context) (N, error) {
 	defer b.wg.Done()
 
 	tenant, bound, err := b.get(ctx)
-	if err != nil {
-		return zero, err
-	}
 	if tenant == "" && b.scope != settings.ScopeGlobal {
-		return zero, nil // fail open
+		return zero, err // no tenant: get() never resolved a value
```
jmank88: Should we return bound here? I suspect zero was a mistake before, from converting more complex enforcement code.

pkg/settings/limits/time.go
```diff
@@ -191,10 +191,10 @@ func (l *timeLimiter) WithTimeout(ctx context.Context) (context.Context, func(),
...
 
 	tenant, timeout, err := l.get(ctx)
-	if err != nil {
-		return nil, nil, err
-	}
 	if tenant == "" && l.scope != settings.ScopeGlobal {
+		if err != nil {
+			return nil, nil, err // no tenant: get() never resolved a value
+		}
 		return ctx, func() {}, nil // fail open
```
jmank88: Isn't there more to consider here? `timeout` may be set (or is even always set?), regardless of whether `err` is nil or tenant is empty, so shouldn't we be using it?

jmank88: Should this whole short circuit block be droppped?

pkg/settings/limits/bound.go
tarcisiozf: correct, this is redundant, in this branch of code get() will return a zero value anyways

pkg/settings/limits/time.go
tarcisiozf: The problem in this cases is that the "no-tenant path" leaves timeout at zero, but Limit() intentionally returns -1 there to distinguish from "a real 0s timeout." So fail-open branch would apply an already-expired deadline instead of leaving the context unbounded

pkg/settings/limits/limits.go
```diff
@@ -29,7 +29,10 @@ type Number interface {
 
 type Limiter[N any] interface {
 	io.Closer // Limiters spawn background goroutines and must be closed.
-	// Limit returns the current limit.
+	// Limit returns the current limit. The value is always usable, even when err is
+	// non-nil: on a read failure it's the compiled default (Setting.DefaultValue), so err
+	// is advisory, not a signal to discard the value. The exception is a required tenant
+	// missing from ctx, where no read is attempted and the zero value is returned instead.
```
jmank88: I think we should try to say less about the implementation details here. I'm not sure about the characterization of that exceptional case too. Zero is a usable value that achieves the intended effect of always limiting. Are there cases that need to understand whether tenant was missing? That seems like a different concern.
```suggestion
	// Limit returns the current limit, or an error along with a usable fallback value.
```

pkg/settings/limits/time.go
jmank88: Sorry, why is it important to have meaningful 0 vs. -1 cases?

tarcisiozf: I've checked again and it won't matter for Limit(), as the caller already handles it (like `limit <= 0`). Will clean up this

tarcisiozf: But for WithTimeout it's a different case, the value is passed to a `context.WithTimeoutCause`, so the returned context is already expired. That is a fail-closed (cancel immediatly) instead of current fail-open (run unbounded) behavior, which differs from what the original code comment said.

jmank88: Well why does Limit return -1? Should it be returning the `timeout` value alongside the error instead?

jmank88: And I believe part of the inconsistency in the "fail open" behavior is because some things are trivial to apply a default to (like this case) while others are complex data structures that must be indexed on the tenant value, meaning we don't really have a clean "default" fallback behavior. We should probably consider using a default rather than unbounded in this case 🤔

pkg/settings/limits/time.go
```diff
@@ -191,10 +191,10 @@ func (l *timeLimiter) WithTimeout(ctx context.Context) (context.Context, func(),
...
 	}
```
jmank88: Re: the other thread https://github.com/smartcontractkit/chainlink-common/pull/2339/changes#r3855088796 - I think we could drop this block and fall-through with an empty tenant value, because we only use it as metadata for the error anyways.
```suggestion
```
Missing tenant is only a blocking problem when there is some persistent state indexed on it.

tarcisiozf: I've changed the block but with a different path than just deleting it to avoid the case of immediately expired contexts (`WithTimeoutCause(ctx, 0)`)

tarcisiozf: But now it achieves what I believe was your goal, WithTimeout never returns nil context now, and still can handle missing tenant error individually.

jmank88 (approved):

pavel-raykov (approved):

PR #2342, opened 2026-08-25
[CRE] Refactor caching (fka fallback) OrgResolver

bolekk:
Change the behavior to always depend on the cached value (if available) to reduce the volume of calls. Only refresh periodically in the background (to recover from potential bad data).

Core counterpart: https://github.com/smartcontractkit/chainlink/pull/23485

Deployment validation: none until active in core: https://github.com/smartcontractkit/chainlink/pull/23485


jmank88:
Can we find a way to evolve without breaking the API?
> ⚠️ Breaking Changes (2)
> pkg/services/orgresolver (2)
> - NewOrgResolverWithFallback — 🗑️ Removed
>
> - OrgResolverFallback — 🗑️ Removed

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,379 @@
...
+const (
+	maxGetRetries       = 3
+	initialRetryBackoff = 100 * time.Millisecond
+	maxRetryBackoff     = 1 * time.Second
+
+	// DefaultRefreshInterval is how often a cached entry triggers a background
+	// revalidation call to the underlying resolver, used when
+	// CachingResolverConfig.RefreshInterval is unset.
+	DefaultRefreshInterval = 10 * time.Minute
+	DefaultRefreshWorkers  = 4
```
jmank88: Do these need to be exported?

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,379 @@
...
+	Set(ctx context.Context, owner string, entry CacheEntry) error
+}
+
+type InMemoryCache struct {
+	entries sync.Map // owner (string) -> CacheEntry
+}
+
+func NewInMemoryCache() *InMemoryCache {
+	return &InMemoryCache{}
+}
```
jmank88, line 62: Could exclude since it's not currently necessary and we may want to use a different signature in the future.
```suggestion
```

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,379 @@
...
+// entry is left in place and continues to be served.
+type CachingResolver struct {
+	inner           OrgResolver
+	cache           Cache
+	refreshInterval time.Duration
+	numWorkers      int
+	logger          log.SugaredLogger
+
+	refreshCh  chan refreshJob
+	refreshing sync.Map // owner (string) -> struct{}{} while queued or being refreshed
```
jmank88: Why use a `sync.Map` in this case?

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,379 @@
...
+// durable store, such as a DB table, depending on its deployment needs.
+type Cache interface {
+	// Get returns the cached entry for owner. ok is false if no entry exists.
+	Get(ctx context.Context, owner string) (entry CacheEntry, ok bool, err error)
+	// Set stores or updates the mapping for owner.
+	Set(ctx context.Context, owner string, entry CacheEntry) error
+}
+
+type InMemoryCache struct {
+	entries sync.Map // owner (string) -> CacheEntry
```
jmank88: Why a `sync.Map` instead of a regular map and mutex?

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
>
>The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

https://pkg.go.dev/sync#Map

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,379 @@
...
+	previous CacheEntry
+}
+
+// CachingResolver wraps an OrgResolver with a Cache of owner->orgID mappings.
+//
+// An owner's org mapping is not expected to change, so once an entry is
+// cached it is trusted indefinitely and returned immediately, without ever
+// blocking on a call to the underlying resolver. To guard against a bad
+// entry becoming permanent (e.g. from a resolver bug or a transient
+// corruption), an entry older than RefreshInterval also queues a
```
patrickhuie19: this will be improved by a batch API on the linking server

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,379 @@
...
+	c.storeInCache(ctx, owner, orgID)
+	return orgID, nil
+}
+
+// queueRefresh hands owner off to the background worker pool to revalidate
+// against the underlying resolver, so that Get never blocks on a cache hit.
+// At most one refresh per owner is queued/running at a time. If the queue is
+// full, the request is dropped; it will be retried on a later stale cache
+// hit.
+func (c *CachingResolver) queueRefresh(owner string, previous CacheEntry) {
```
patrickhuie19: I wonder if the additional complexity is worth it. If we just lazy load the values we can save on populating a queue, and once we get the batch API on the server side, we can warm the cache completely and basically drop TPS against the org resolver from this node to a call every few minutes (ofc we will need to paginate)

bolekk: Hmm with the batch API, are you suggesting that we could periodically fetch ALL entries in a single call?

patrickhuie19: with some reasonable pagination, a few calls. it should scale well for our needs for a long time, and if it ever becomes an issue we can consider upgrading to a websockets api to stream state changes

pkg/services/orgresolver/caching.go
bolekk: I'll keep it because now it also creates a map object.

pkg/services/orgresolver/caching.go
bolekk: Let me simplify this a bit.

bolekk:
@jmank88 I don't think it makes sense to keep old APIs since I renamed the file and types. It's not "fallback" any more.

jmank88:
> @jmank88 I don't think it makes sense to keep old APIs since I renamed the file and types. It's not "fallback" any more.

Compilation compatibility trumps undesirable naming though. It doesn't make sense to arbitrarily break unrelated things because the names don't make sense.

bolekk:
@jmank88 I disagree. It's only used in a single place in core and the old name is now confusing. NewOrgResolverWithFallback - there's no fallback, cached values are the new source of truth.

jmank88:
> @jmank88 I disagree. It's only used in a single place in core and the old name is now confusing. NewOrgResolverWithFallback - there's no fallback, cached values are the new source of truth.

Having one callsite from one module does not mean that there is only one thing to worry about. There are multiple versions of core that may need patch releases that could encounter this. There are also manys modules that import core and could be affected by this. Coincidentally, I am currently blocked from acting on basic dependabot security alerts in Aptos, because a trivial dependency bump is not possible, since we deleted a bunch of API that is superficially not related.

bolekk:
@jmank88 to me this tradeoff is not worth it. But to move this PR forward, I brought back the old fallback resolver and marked it as deprecated - is that OK?

pkg/services/orgresolver/fallback.go
```diff
@@ -3,58 +3,21 @@ package orgresolver
...
-// an org just before a WorkflowDeleted event is processed, causing the
-// resolver to return NotFound for an owner whose org was previously known.
+// @deprecated: Use CachingResolver
 type OrgResolverFallback struct {
 	inner  OrgResolver
 	cache  sync.Map // owner (string) -> orgID (string)
 	logger log.SugaredLogger
 }
 
+// @deprecated: Use NewCachingResolver
```
jmank88:
```suggestion
// Deprecated: Use NewCachingResolver
```

jmank88 (approved):

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,382 @@
...
+		owners = append(owners, owner)
+	}
+	c.ownersMu.Unlock()
+
+	for i, owner := range owners {
+		if i > 0 {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(refreshOwnerDelay):
```
patrickhuie19, line 207: should this be `i*refreshOwnerDelay`? if not, why have `refreshOwnerDelay`? the `refreshLoop` already waits for the `refreshInterval`.

pkg/services/orgresolver/caching.go
```diff
@@ -0,0 +1,382 @@
...
+
+// Start starts the inner resolver and the background refresh loop.
+func (c *CachingResolver) Start(ctx context.Context) error {
+	if err := c.inner.Start(ctx); err != nil {
+		return err
+	}
+
+	// The loop must outlive the Start call, so it gets its own context,
+	// cancelled explicitly by Close rather than inherited from ctx.
+	loopCtx, cancel := context.WithCancel(context.Background())
```
patrickhuie19, line 363: /nit i believe you could use the stopCh context helper here instead of a raw context.Background()

patrickhuie19 (approved):
mostly lgtm with one q on owner specific delay

pkg/services/orgresolver/caching.go
bolekk: Every 10 mins we cycle over all owners with 100ms gaps between them.

bolekk: This is to smooth out the spike in traffic on eery refreshInterval. It will go away with batching anyway.

PR #2345, opened 2026-08-27
Add outbound HTTP action concurrency limits

justinkaseman:


pkg/settings/cresettings/defaults.json
```diff
@@ -26,6 +26,8 @@
 	"GatewayConfidentialRelayPerNodeRate": "10rps:10",
 	"GatewayHTTPActionMtlsRequestRate": "every30s:0",
 	"GatewayHTTPActionMtlsConcurrencyLimit": "50",
+	"GatewayHTTPActionOutboundConcurrencyLimit": "256",
+	"GatewayHTTPActionOutboundPerNodeConcurrencyLimit": "32",
```
cedric-cordenier: We need to validate these against the existing rate limits that we set

I think we have RPS of 100 with a burst of 100 atm; so your per-node concurrency limit is too low I think

cedric-cordenier: (Best to validate against the traffic we see in prod though)

russell-stern (dismissed):

bolekk (dismissed):

bolekk (approved):

russell-stern (approved):

PR #2346, opened 2026-08-27
Add callback ID to the SecretsRequestParams type

vreff:
## Summary

Adds a `CallbackID int` field to `SecretsRequestParams` and binds it into the secrets response signature hash.

## Changes

- `pkg/capabilities/v2/actions/confidentialrelay/types.go`: add `CallbackID` to `SecretsRequestParams`; write it into the hash via a new `writeInt` helper, but only when non-zero (so older protocols that never set it hash identically).
- `types_test.go`: `TestSecretsResponseHash_BindsCallbackID` (different IDs → different hashes) and `TestSecretsResponseHash_CallbackIDZeroOmitted` (zero is omitted, never collides with non-zero).

## Why

The relay DON needs to correlate a secrets response back to the originating callback. Binding the callback ID into the signed hash prevents a signature over one callback from being replayed against another.

### DIFF
--- a/pkg/capabilities/v2/actions/confidentialrelay/types.go
+++ b/pkg/capabilities/v2/actions/confidentialrelay/types.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"hash"
 	"sort"
+	"strconv"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/teeattestation"
 	"github.com/smartcontractkit/libocr/ragep2p/peeridhelper"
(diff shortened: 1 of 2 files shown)


russell-stern (dismissed):

russell-stern (approved):

pavel-raykov:
> _ No description provided. _

It would be nice to have a jira/ticket for some explanation why this is needed.

pavel-raykov (approved):

PR #2353, opened 2026-09-02
Enable beholder metrics for sqlutil

dhaidashenko:


pkg/sqlutil/pg/beholder_metrics.go
```diff
@@ -0,0 +1,77 @@
+package pg
+
+import (
+	"context"
+	"database/sql"
+
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/beholder"
```
jmank88, line 10: I don't think this package should import beholder. We already provide a means of setting a custom hook via `StatsCustomReporterFn`. Why don't we use that?

pkg/sqlutil/beholder_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	m.histogram.Record(ctx, pct)
+}
+
+type noopSQLQueryTimeMetric struct{}
+
+func (noopSQLQueryTimeMetric) Record(context.Context, float64) {}
+
+var (
+	sqlQueryTimeMetricOnce   sync.Once
+	globalSQLQueryTimeMetric sqlQueryTimeMetric
```
jmank88, line 47: Do we strictly need a global instance? or could we inject something to use or construct from instead? We don't typically need to suppress errors for metric creation.

PR #2356, opened 2026-09-03
workflows/wasm/host: prevent limiter and Wasmtime resource leaks

cfal:
fixes resource leaks in the WASM host by explicitly closing module-owned limiters and Wasmtime resources.

confidential-workflows creates a new module per production execution without injecting limiters. the previous fallback limiters started 27 goroutines per module and were never closed.

this change:

- replaces constant fallback bounds with static limiters
- closes owned resources on constructor failure and module shutdown
- keeps fallback limiters module-local so async eviction cleanup can't affect a replacement module
- closes temporary modules, linkers, and Wasmtime resources deterministically

### DIFF
--- a/pkg/workflows/wasm/host/module.go
+++ b/pkg/workflows/wasm/host/module.go
@@ -27,7 +27,6 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/config"
 	"github.com/smartcontractkit/chainlink-common/pkg/custmsg"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
-	"github.com/smartcontractkit/chainlink-common/pkg/settings"
 	"github.com/smartcontractkit/chainlink-common/pkg/settings/cresettings"
 	"github.com/smartcontractkit/chainlink-common/pkg/settings/limits"
 	dagsdk "github.com/smartcontractkit/chainlink-common/pkg/workflows/sdk"
@@ -96,7 +95,8 @@ type ModuleConfig struct {
 	MaxResponseSizeLimiter       limits.BoundLimiter[config.Size] // supersedes MaxResponseSizeBytes if set
 
 	// MaxSubscriptionsLimiter bounds nsubscriptions in the WASI poll_oneoff host
-	// call. Defaults to cresettings.Default.WASMPollOneoffSubscriptionLimit.
+	// call. It uses the default value of cresettings.Default.WASMPollOneoffSubscriptionLimit;
+	// inject a limiter to provide dynamic settings.
 	MaxSubscriptionsLimiter limits.BoundLimiter[int]
 
 	MaxLogLenBytes      uint32
@@ -150,7 +150,8 @@ type module struct {
 	module  *wasmtime.Module
 	wconfig *wasmtime.Config
 
-	cfg *ModuleConfig
+	cfg             *ModuleConfig
+	defaultLimiters moduleLimiters
 
 	metrics moduleMetrics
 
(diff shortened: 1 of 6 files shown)


pkg/workflows/wasm/host/module.go
```diff
@@ -286,97 +333,64 @@ func NewModule(ctx context.Context, modCfg *ModuleConfig, binary []byte, opts ..
...
-		}
+		defaultLimiters.maxUserMetricLabelValueLength = limits.NewUpperBoundLimiter(int(modCfg.MaxUserMetricLabelValueLength))
 	}
 	if modCfg.MemoryLimiter == nil {
 		// Take the max of the min and the configured max memory mbs.
 		// We do this because Go requires a minimum of 16 megabytes to run,
 		// and local testing has shown that with less than the min, some
 		// binaries may error sporadically.
 		modCfg.MaxMemoryMBs = uint64(math.Max(float64(modCfg.MinMemoryMBs), float64(modCfg.MaxMemoryMBs)))
-		limit := settings.Size(config.Size(modCfg.MaxMemoryMBs) * config.MByte)
```
cfal, line 331: removed because settings.Size only wrapped the raw bound with a parser and unit:

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/settings/settings.go#L160-L163

this call site constructed a Factory with only Logger, so Settings and Meter were nil:

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/workflows/wasm/host/module.go#L287-L296

GetOrDefault returns DefaultValue immediately when Settings is nil:

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/settings/settings.go#L216-L229

despite that, MakeUpperBoundLimiter started a global updater:

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/settings/limits/bound.go#L65-L124

each updater also started cancellation and ticker goroutines:

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/settings/limits/updater.go#L68-L84

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/services/stop_!race.go#L9-L17

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/timeutil/ticker.go#L14-L20

so these nine constant defaults created 27 goroutines per module to repeatedly resolve the same value. NewUpperBoundLimiter enforces the same bound without that unused update machinery.

callers that need dynamic settings already inject configured limiters through ModuleConfig:

https://github.com/smartcontractkit/chainlink/blob/f05ecfe30b60d0be98aabbb1052657ef352b74d8/core/services/workflows/syncer/v2/handler.go#L836-L848

pkg/workflows/wasm/host/module.go
```diff
@@ -150,7 +150,8 @@ type module struct {
 	module  *wasmtime.Module
 	wconfig *wasmtime.Config
 
-	cfg *ModuleConfig
+	cfg             *ModuleConfig
+	defaultLimiters moduleLimiters
```
cfal, line 154: used to be written into the caller-owned ModuleConfig:

https://github.com/smartcontractkit/chainlink-common/blob/6b01cb8b44f2def95090582ce30a464bfe268837/pkg/workflows/wasm/host/module.go#L293-L368

that isn’t safe as EvictableModule retains the same config pointer, the old module closes asynchronously, and a replacement can be constructed while that cleanup is still queued:

https://github.com/smartcontractkit/chainlink/blob/f05ecfe30b60d0be98aabbb1052657ef352b74d8/core/services/workflows/syncer/v2/handler.go#L914

https://github.com/smartcontractkit/chainlink/blob/f05ecfe30b60d0be98aabbb1052657ef352b74d8/core/services/workflows/syncer/v2/evictable_module.go#L131-L150

https://github.com/smartcontractkit/chainlink/blob/f05ecfe30b60d0be98aabbb1052657ef352b74d8/core/services/workflows/syncer/v2/evictable_module.go#L291-L336

keeping defaults per-module means the old module can only close its own resources.

vreff (approved):

prashantkumar1982 (approved):

jmank88 (dismissed):
Can you please add a description? I'm having trouble understanding the goal of this PR.

pkg/workflows/wasm/host/module.go
```diff
@@ -286,97 +333,64 @@ func NewModule(ctx context.Context, modCfg *ModuleConfig, binary []byte, opts ..
...
 	}
 
 	if modCfg.MaxUserMetricPayloadLimiter == nil {
-		limit := settings.Size(config.Size(modCfg.MaxUserMetricPayloadBytes))
-		var err error
-		modCfg.MaxUserMetricPayloadLimiter, err = limits.MakeUpperBoundLimiter(lf, limit)
-		if err != nil {
-			return nil, fmt.Errorf("failed to make metric payload size limiter: %w", err)
-		}
+		defaultLimiters.maxUserMetricPayload = limits.NewUpperBoundLimiter(config.Size(modCfg.MaxUserMetricPayloadBytes))
```
jmank88, line 349: Why switch from the `*Make(Factory,*` variants? The only effect should be losing logging, which is especially important while we still have "fail open" cases which only log errors rather than return them.

pkg/workflows/wasm/host/module.go
jmank88: > callers that need dynamic settings already inject configured limiters through ModuleConfig:

IIRC we expect all production callers to do this. This other case was only meant to be a non-production fallback. Do we have callers in production that are not injecting limiters?

pkg/workflows/wasm/host/module.go
cfal: there is no effective logging lost. Factory.Settings is nil, so GetOrDefault always returns the default with a nil error. the fail-open log path only applies to missing tenants on non-global settings, while these defaults are all global. bound violations weren't logged before either.

so NewUpperBoundLimiter preserves the observable behavior; it only removes the unused updater machinery.

pkg/workflows/wasm/host/module.go
cfal: confidential-workflows creates a module per production execution without injecting any limiters:

https://github.com/smartcontractkit/chainlink-confidential-compute/blob/19cbd1cb7c4440fee21caddee229a8dba3a14f28/enclave/apps/confidential-workflows/app/wasm.go#L28-L44

WasmFileSpecFactory is another production path that supplies only Logger:

https://github.com/smartcontractkit/chainlink/blob/be67b7b77be4d515be111ddc935696b58edcec53/core/services/job/wasm_file_spec_factory.go#L24-L36

the standard workflow runtime also omits MaxSubscriptionsLimiter.

jmank88: Should we set them from confidential compute? Otherwise we cannot reconfigure and we don't get metrics.

The other one seems less important, given that it already has a null logger?

pkg/workflows/wasm/host/module.go
jmank88: I don't like the idea of depending on implementation details that could change but if this is only for the non-production cases then that is probably fine. Can we update the injection from confidential compute?
Otherwise, if we are stuck with using these for production, then something seems misaligned, because we should want overrides and metrics in that case, no?

jmank88 (approved):

PR #2357, opened 2026-09-03
Add vault don settings

russell-stern:


pkg/capabilities/actions/vault/messages.proto
```diff
@@ -136,6 +136,18 @@ message ObservationError {
...
+// DON-wide settings committed by per-field 2f+1 consensus.
+// Add fields freely; never reuse field numbers (use reserved instead).
+message NodeSettings {
+  uint64 max_identifier_key_length_bytes         = 1;
+  uint64 max_identifier_owner_length_bytes       = 2;
+  uint64 max_identifier_namespace_length_bytes   = 3;
+  uint64 max_share_length_bytes                  = 4;
+  uint64 max_blob_payload_bytes                  = 5;
+  uint64 max_pending_queue_write_size            = 6;
+  uint64 max_request_batch_size                  = 7;
```
jmank88, line 148: nit/ sometimes moving qualifiers like `max` towards the end of names makes things sort/group more naturally.

On the other hand, is this always just a collections of limits/maxes? If so, could we drop the qualifier from each field by communicating it another way? Might be confusing though

cedric-cordenier (approved):

PR #2361, opened 2026-09-08
pkg/workflows/dontime: introduce feature flagged fix for sequence number inconsistency

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-5838

Supports:
- https://github.com/smartcontractkit/chainlink/pull/23712

### DIFF
--- a/pkg/workflows/dontime/pb/dontime.go
+++ b/pkg/workflows/dontime/pb/dontime.go
@@ -1,0 +1,24 @@
+package pb
+
+import "math"
+
+// MaxSeqNum returns the max sequence number from TimestampsBySequence, or 0 if none exist.
+func (t *ObservedDonTimes) MaxSeqNum() (maxSeqNum int64) {
+	for seqNum := range t.TimestampsBySequence {
+		if seqNum > maxSeqNum {
+			maxSeqNum = seqNum
+		}
+	}
+	return
+}
+
+// EarliestTS returns the easliest timestamp value from TimestampsBySequence, or math.MaxInt64 if none exist.
+func (t *ObservedDonTimes) EarliestTS() (earliestTS int64) {
+	earliestTS = math.MaxInt64
+	for _, ts := range t.TimestampsBySequence {
+		if ts < earliestTS {
+			earliestTS = ts
+		}
+	}
+	return
+}
--- a/pkg/workflows/dontime/plugin.go
+++ b/pkg/workflows/dontime/plugin.go
@@ -293,22 +293,51 @@ func (p *Plugin) ObservationQuorum(_ context.Context, _ ocr3types.OutcomeContext
 }
 
 func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation) (ocr3types.Outcome, error) {
-	if p.sequencedTSEnabled.Check(ctx, config.NewTimestamp(time.Now())) == nil { //TODO inspect error
-		return p.sequencedOutcome(ctx, outctx, nil, aos)
-	}
-
-	return p.unsequencedOutcome(ctx, outctx, nil, aos)
-}
-
-// unsequencedOutcome executes the original outcome logic to produce an unsequenced slice of [pb.ObservedDonTimes.Timestamps].
-func (p *Plugin) unsequencedOutcome(ctx context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation) (ocr3types.Outcome, error) {
-	observationCounts := map[string]int64{} // counts how many nodes reported where a new DON timestamp might be needed
 	type timestampNodePair struct {
 		Timestamp        int64
 		NodeID           int
 		OffsetFromMedian int64
 	}
 	var timestampNodePairs []timestampNodePair
+	for idx, ao := range aos {
+		observation := &pb.Observation{}
+		if err := proto.Unmarshal(ao.Observation, observation); err != nil {
+			p.lggr.Errorf("failed to unmarshal observation in Outcome phase")
+			continue
+		}
+
+		timestampNodePairs = append(timestampNodePairs, timestampNodePair{Timestamp: observation.Timestamp, NodeID: idx})
+	}
+
+	if len(timestampNodePairs) == 0 {
+		return nil, errors.New("no observation contains a valid timestamp")
+	}
+
+	slices.SortFunc(timestampNodePairs, func(a, b timestampNodePair) int {
+		return cmp.Compare(a.Timestamp, b.Timestamp)
+	})
+	donTime := timestampNodePairs[len(timestampNodePairs)/2].Timestamp
+	for i := range timestampNodePairs {
+		timestampNodePairs[i].OffsetFromMedian = timestampNodePairs[i].Timestamp - donTime
+	}
+	p.lggr.Debugw("Observed Node Timestamps",
+		"timestampNodePairs", timestampNodePairs,
+		"median", donTime,
+		"collectedDataPoints", len(timestampNodePairs),
+		"minOffsetFromMedian", timestampNodePairs[0].OffsetFromMedian,
+		"maxOffsetFromMedian", timestampNodePairs[len(timestampNodePairs)-1].OffsetFromMedian,
+	)
+
+	if p.sequencedTSEnabled.Check(ctx, config.NewTimestamp(time.UnixMilli(donTime))) == nil { //TODO inspect error
+		return p.sequencedOutcome(ctx, outctx, nil, aos, donTime)
+	}
+
+	return p.unsequencedOutcome(ctx, outctx, nil, aos, donTime)
+}
+
+// unsequencedOutcome executes the original outcome logic to produce an unsequenced slice of [pb.ObservedDonTimes.Timestamps].
+func (p *Plugin) unsequencedOutcome(ctx context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation, donTime int64) (ocr3types.Outcome, error) {
+	observationCounts := map[string]int64{} // counts how many nodes reported where a new DON timestamp might be needed
 	limitByBatchSizeFlagEnabled := true
 
 	prevOutcome := &pb.Outcome{}
@@ -319,7 +348,7 @@ func (p *Plugin) unsequencedOutcome(ctx context.Context, outctx ocr3types.Outcom
 		prevOutcome.ObservedDonTimes = make(map[string]*pb.ObservedDonTimes)
 	}
 
-	for idx, ao := range aos {
+	for _, ao := range aos {
 		observation := &pb.Observation{}
 		if err := proto.Unmarshal(ao.Observation, observation); err != nil {
 			p.lggr.Errorf("failed to unmarshal observation in Outcome phase")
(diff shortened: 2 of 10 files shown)


pkg/workflows/dontime/plugin.go
```diff
@@ -213,6 +293,15 @@ func (p *Plugin) ObservationQuorum(_ context.Context, _ ocr3types.OutcomeContext
 }
 
 func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation) (ocr3types.Outcome, error) {
+	if p.sequencedTSEnabled.Check(ctx, config.NewTimestamp(time.Now())) == nil { //TODO inspect error
```
jmank88: Note: we discussed being more defensive against clock drift here, e.g. by "snapping" to a round sequence number after the timestamp.

pkg/workflows/dontime/pb/dontime.go
```diff
@@ -0,0 +1,24 @@
+package pb
```
bolekk, line 1: why in pb?

pkg/workflows/dontime/plugin.go
bolekk: Yeah, we can't use time.Now() but could we use DON Time? We can compute it first, before branching into un/sequenced logic.

pkg/workflows/dontime/plugin.go
```diff
@@ -341,6 +430,147 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
+	outcome.Timestamp = donTime
+
+	for id, numRequests := range observationCounts {
+		if numRequests > int64(p.config.F) {
+			observedDonTimes, ok := outcome.ObservedDonTimes[id]
+			if !ok {
+				observedDonTimes = &pb.ObservedDonTimes{TimestampsBySequence: make(map[int64]int64)}
+			}
+			currSeqNum := observedDonTimes.MaxSeqNum() + 1
+			observedDonTimes.TimestampsBySequence[currSeqNum] = donTime
```
bolekk, line 510: You also need to modify the transmitter: https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/workflows/dontime/transmitter.go#L53

pkg/workflows/dontime/pb/dontime.go
jmank88: Seemed fitting as methods, but could be a helper funcs instead 🤷

pkg/workflows/dontime/plugin.go
jmank88: Switched to donTime. That deduplicated some code too

pkg/workflows/dontime/plugin.go
jmank88: Added

pkg/workflows/dontime/plugin.go
```diff
@@ -204,6 +229,61 @@ func (p *Plugin) Observation(ctx context.Context, outctx ocr3types.OutcomeContex
...
+			maxSeqNum = times.MaxSeqNum()
+		}
+
+		if int64(req.SeqNum) > maxSeqNum+1 {
+			p.store.RemoveRequest(req.WorkflowExecutionID)
+			req.SendResponse(Response{
+				WorkflowExecutionID: req.WorkflowExecutionID,
+				SeqNum:              req.SeqNum,
+				Timestamp:           0,
+				Err: fmt.Errorf("requested seqNum %d for executionID %s is greater than expected based on the max seqNum observed so far %d",
```
bolekk, line 252: I don't think we should error here. What if the plugin didn't process all lower seq numbers in the previous round? I thought we would always add all local requests to observation.

pkg/workflows/dontime/plugin.go
```diff
@@ -255,27 +373,121 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
 					requestSeqNum, id, currSeqNum)
 			}
 		}
+	}
 
-		timestampNodePairs = append(timestampNodePairs, timestampNodePair{Timestamp: observation.Timestamp, NodeID: idx})
+	outcome := prevOutcome
+
+	// Compare with prior outcome to ensure DON time never goes backward.
```
bolekk, line 380: Shouldn't this check happen earlier, before we go into un/sequencedOutcome() ?

pkg/workflows/dontime/plugin.go
```diff
@@ -255,27 +373,121 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
+	p.metrics.donTime.Record(ctx, outcome.Timestamp)
+	p.metrics.donTimeEntries.Record(ctx, int64(len(outcome.ObservedDonTimes)))
+	p.metrics.outcomeBatchOverflow.Record(ctx, outcomeBatchOverflowCount)
+	p.metrics.outcomeSize.Record(ctx, int64(len(outcomeBytes)))
+	return outcomeBytes, err
+}
+
+// sequencedOutcome executed the updated outcome logic to produce a sequenced map of [pb.ObservedDonTimes.TimestampsBySequence].
+func (p *Plugin) sequencedOutcome(ctx context.Context, outctx ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation, donTime int64) (ocr3types.Outcome, error) {
+	observationCounts := map[string]int64{} // counts how many nodes reported where a new DON timestamp might be needed
```
bolekk, line 444: nit: comment what is the key in this map

pkg/workflows/dontime/plugin.go
```diff
@@ -255,27 +373,121 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
+	}
+
+	for _, ao := range aos {
+		observation := &pb.Observation{}
+		if err := proto.Unmarshal(ao.Observation, observation); err != nil {
+			p.lggr.Errorf("failed to unmarshal observation in Outcome phase")
+			continue
+		}
+
+		if !observation.GetLimitByBatchSizeFlag() {
```
bolekk, line 472: hmm I should've cleaned that up. We can assume this is enabled everywhere by now

pkg/workflows/dontime/plugin.go
```diff
@@ -255,27 +373,121 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
+			var currSeqNum int64
+			if times, ok := prevOutcome.ObservedDonTimes[id]; ok {
+				currSeqNum = times.MaxSeqNum() + 1
+			}
+			// We only count requests for the next sequence number and ignore all other ones.
+			if requestSeqNum == currSeqNum {
+				observationCounts[id]++
+			} else if requestSeqNum > currSeqNum {
+				// This should never happen since we don't include out of sequence requests in the Observation phase
+				p.lggr.Errorf("request seqNum %d for executionID %s is greater than the current seqNum %d",
```
bolekk, line 486: I don't think it's an error any more. Shouldn't we just process the highest number that we can reach consensus on and allow for gaps? cc @cedric-cordenier if I'm missing something here.

jmank88: Well what happens when consensus is not reached? Do we expect the next attempt to repeat the failed sequence number? Or skip to the next?

pkg/workflows/dontime/plugin.go
jmank88: Is it safe to drop this check completely? Or should it be relaxed to a sanity check?

pkg/workflows/dontime/plugin.go
```diff
@@ -255,27 +374,108 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
+			p.lggr.Errorf("failed to unmarshal observation in Outcome phase")
+			continue
+		}
+
+		for id, requestSeqNum := range observation.Requests {
+			var currSeqNum int64
+			if times, ok := prevOutcome.ObservedDonTimes[id]; ok {
+				currSeqNum = times.MaxSeqNum() + 1
+			}
+			// We only count requests for the next sequence number and ignore all other ones.
```
jmank88: @bolekk so does this need to change too? i.e Is there still a way to choose one "next" number? Or must we start handling all requests?

pkg/workflows/dontime/plugin.go
jmank88: Removed

pkg/workflows/dontime/plugin.go
jmank88: Comment shifted. Was on the line: `// We only count requests for the next sequence number and ignore all other ones.`

pkg/workflows/dontime/plugin.go
```diff
@@ -150,7 +157,22 @@ func (p *Plugin) Observation(ctx context.Context, outctx ocr3types.OutcomeContex
...
 		p.lggr.Errorf("failed to unmarshal previous outcome in Observation phase")
 	}
+	// inspect ObservedDonTimes to determine if the feature flag has taken effect
+	var observedDonTimes *pb.ObservedDonTimes
+	for _, observedDonTimes = range previousOutcome.ObservedDonTimes {
+		break
+	}
+	if observedDonTimes != nil && len(observedDonTimes.TimestampsBySequence) > 0 {
+		// feature flag for sequenced timestamps is in effect
+		// TODO can len be zero for both Timestamps and TimestampsBySequence?
```
jmank88: @bolekk Can I assume that any `Outcome` will have some timestamps populated? Or do we need to support empty `Outcome`s as well (currently ambiguous)?

cedric-cordenier: No I think you need to support no timestamps being populated

pkg/workflows/dontime/plugin.go
```diff
@@ -150,7 +157,22 @@ func (p *Plugin) Observation(ctx context.Context, outctx ocr3types.OutcomeContex
 	if err := proto.Unmarshal(outctx.PreviousOutcome, previousOutcome); err != nil {
 		p.lggr.Errorf("failed to unmarshal previous outcome in Observation phase")
 	}
+	// inspect ObservedDonTimes to determine if the feature flag has taken effect
+	var observedDonTimes *pb.ObservedDonTimes
```
cedric-cordenier: @jmank88 Note: you don't need to respect the feature flag here, you can just output the observation in both formats

pkg/workflows/dontime/plugin.go
jmank88: Hmmm well if we can ignore the feature flag per your other comment, then I think this ambiguity is irrelevant.

pkg/workflows/dontime/plugin.go
cedric-cordenier: @jmank88 You should choose the 1 next number by putting it through consensus; i.e. you should ask all nodes what their observed next sequence number is for a given execution ID, and then get consensus on that

pkg/workflows/dontime/plugin.go
cedric-cordenier: @jmank88 We should preserve this logic for the new flow: for a given sequence number we need to check a) whether we've already got a response for it (in which case we can skip it), b) whether we will never get a response for it (because we've skipped it) or c) if it's higher than the max sequence number and is therefore a candidate for consensus

pkg/workflows/dontime/plugin.go
jmank88: I think the answer here is to ensure that only the lowest requested seq num is chosen, and still error this way for any higher.

jmank88: Ah but the IDs are already unique at this point, so there are no conflicts to resolve.

pkg/workflows/dontime/plugin.go
jmank88: Updated. But this section just reduces to the new logic then, right? We tolerate all requests (up to batch limit) since we no longer make an assumption about what the next/current sequence number is.

pkg/workflows/dontime/plugin.go
jmank88: Updated

pkg/workflows/dontime/plugin.go
jmank88: Updated

pkg/workflows/dontime/plugin.go
```diff
@@ -272,15 +227,85 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
+		for id, requestSeqNum := range observation.Requests {
+			var currSeqNum int64
+			if times, ok := prevOutcome.ObservedDonTimes[id]; ok {
+				currSeqNum = int64(len(times.Timestamps))
+			}
+			// We only count requests for the next sequence number and ignore all other ones.
+			if requestSeqNum == currSeqNum {
+				observationCounts[id]++
+			} else if requestSeqNum > currSeqNum {
+				// This should never happen since we don't include out of sequence requests in the Observation phase
```
cedric-cordenier: Note: we now do so we should remove this comment

pkg/workflows/dontime/plugin.go
```diff
@@ -306,34 +331,73 @@ func (p *Plugin) Outcome(ctx context.Context, outctx ocr3types.OutcomeContext, _
...
-	)
-	p.metrics.donTime.Record(ctx, outcome.Timestamp)
-	p.metrics.donTimeEntries.Record(ctx, int64(len(outcome.ObservedDonTimes)))
-	p.metrics.outcomeBatchOverflow.Record(ctx, outcomeBatchOverflowCount)
-	p.metrics.outcomeSize.Record(ctx, int64(len(outcomeBytes)))
-	return outcomeBytes, err
+	outcome := prevOutcome
+	outcome.Timestamp = donTime
+
+	for key, numRequests := range observationCounts {
```
cedric-cordenier: You're iterating over a map here so this could be non-deterministic; you should sort the map's keys before iterating

cedric-cordenier: Eg. 

```
for key := range slices.Sorted(maps.Keys(observationCounts) {
    val := observationCounts[key]
    ...
```

PR #2369, opened 2026-09-09
Add Stellar get transaction status fields

Krish-vemula:
## Summary
- add explicit getTransaction status, tx hash, and result XDR fields
- make fee and ledger fields presence-aware
- reject unknown or status-inconsistent responses in converters

## Tests
- go test ./pkg/chains/stellar ./pkg/loop/internal/relayer ./pkg/loop/internal/relayerset

### DIFF
--- a/pkg/chains/stellar/proto_helpers.go
+++ b/pkg/chains/stellar/proto_helpers.go
@@ -781,23 +781,103 @@ func ConvertGetTransactionRequestFromProto(p *GetTransactionRequest) (stellar.Ge
 	return stellar.GetTransactionRequest{TxHash: p.GetTxHash()}, nil
 }
 
-func ConvertGetTransactionResponseToProto(resp stellar.GetTransactionResponse) *GetTransactionResponse {
+func ConvertGetTransactionResponseToProto(resp stellar.GetTransactionResponse) (*GetTransactionResponse, error) {
+	status, err := convertGetTransactionStatusToProto(resp.Status)
+	if err != nil {
+		return nil, fmt.Errorf("status: %w", err)
+	}
+	if err := validateGetTransactionResponse(resp); err != nil {
+		return nil, err
+	}
+
+	var resultXDR, resultMetaXDR []byte
+	if resp.ResultXDR != "" {
+		resultXDR, err = base64.StdEncoding.DecodeString(resp.ResultXDR)
+		if err != nil {
+			return nil, fmt.Errorf("invalid result xdr %q: %w", resp.ResultXDR, err)
+		}
+	}
+	if resp.ResultMetaXDR != "" {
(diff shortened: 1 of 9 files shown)


pkg/chains/stellar/proto_helpers.go
```diff
@@ -781,23 +781,103 @@ func ConvertGetTransactionRequestFromProto(p *GetTransactionRequest) (stellar.Ge
 	return stellar.GetTransactionRequest{TxHash: p.GetTxHash()}, nil
 }
 
-func ConvertGetTransactionResponseToProto(resp stellar.GetTransactionResponse) *GetTransactionResponse {
+func ConvertGetTransactionResponseToProto(resp stellar.GetTransactionResponse) (*GetTransactionResponse, error) {
+	status, err := convertGetTransactionStatusToProto(resp.Status)
+	if err != nil {
+		return nil, fmt.Errorf("status: %w", err)
```
amit-momin: nit: don't think this error wrapping helps much unless it's something like `convert to proto err` but think you can just skip it all together

pkg/chains/stellar/proto_helpers.go
```diff
@@ -781,23 +781,103 @@ func ConvertGetTransactionRequestFromProto(p *GetTransactionRequest) (stellar.Ge
...
+		return stellar.GetTransactionResponse{}, fmt.Errorf("status: %w", err)
+	}
+
+	resp := stellar.GetTransactionResponse{
+		Status:        status,
+		TxHash:        p.GetTxHash(),
+		ResultXDR:     base64.StdEncoding.EncodeToString(p.GetResultXdr()),
+		ResultMetaXDR: base64.StdEncoding.EncodeToString(p.GetResultMetaXdr()),
+	}
+	if p.FeeStroops != nil {
```
amit-momin: Should we use the getters for the nil check too? Also if we have getters, shouldn't we make the fields themselves private?

amit-momin, line 865: Should we use the getters for the nil check too? Also if we have getters, shouldn't we make the fields themselves private?

pkg/chains/stellar/proto_helpers.go
Krish-vemula: Good point on the extra wrapping. I removed the `status:` wrapper and return the converter error directly now.

pkg/chains/stellar/proto_helpers.go
Krish-vemula: For the optional proto fields, I kept the direct `nil` checks intentionally. The generated getters collapse unset and zero into the same value, so checking `p.FeeStroops != nil` / etc. is required to preserve presence semantics. These fields are generated protobuf fields, so making them private is not something we should do here.

amit-momin (approved):

ilija42 (approved):
Is there a chainlink-stellar PR follow up that populates these fields

PR #2371, opened 2026-09-09
[CRE] Add v2 SecretValue workflow config type

bolekk:
## Context
Part of CRE-5836 (Keystone cleanup: remove the legacy DAG workflow module and wasm utils). The legacy `pkg/workflows/sdk` root package is slated for deletion, but two of its symbols are still used by current v2 code in chainlink (PoR example + canary workflows):
- `sdk.SecretValue` — workflow config field marker type
- `sdk.BreakErr` — pure alias of `capabilities.ErrStopExecution`

### DIFF
--- a/pkg/workflows/sdk/v2/secrets.go
+++ b/pkg/workflows/sdk/v2/secrets.go
@@ -1,0 +1,5 @@
+// Package v2 provides shared types for v2 (NoDAG) workflows.
+package v2
+
+// SecretValue marks workflow config fields that hold secret values.
+type SecretValue string


pkg/workflows/sdk/v2/secrets.go
```diff
@@ -0,0 +1,5 @@
+// Package v2 provides shared types for v2 (NoDAG) workflows.
+package v2
+
+// SecretValue marks workflow config fields that hold secret values.
+type SecretValue string
```
bolekk, line 5: Do we need this at all? Can we just use "string" in places like this? [https://github.com/smartcontractkit/chainlink/blob/develop/system-tests/tests/canaries_sentinels/proof-of-reserve/cron-based/types/types.go#L27](https://www.google.com/url?q=https://github.com/smartcontractkit/chainlink/blob/develop/system-tests/tests/canaries_sentinels/proof-of-reserve/cron-based/types/types.go%23L27&sa=D&source=docs&ust=1788996005733355&usg=AOvVaw0nsUutosqz6y-ovSNWxNe6)

PR #2376, opened 2026-09-10
Require Stellar enum presence in proto responses

Krish-vemula:
## Summary
- make Stellar `event_type` and `tx_status` proto fields optional
- reject omitted enum fields on inbound conversion
- keep checked enum conversion for unknown values


ilija42 (commented):
Have you checked if this needs downstream api changes?

PR #2380, opened 2026-09-11
Introduce ErrMissingTenant for scoped limiters

tarcisiozf:
This pull request standardizes how missing tenant errors are reported across all scoped limiters in the `pkg/settings/limits` package. It introduces a new sentinel error, `ErrMissingTenant`, and updates all relevant limiters to wrap and return this error when a required tenant is missing from the context. Comprehensive tests are also added to ensure correct error handling and categorization.

**Error handling improvements:**

* Introduced a new sentinel error `ErrMissingTenant` in `errors.go` to represent the case where a scoped limiter is used without the required tenant in the context. This error is not a `LimitError` and signals a programming error rather than a system issue.
* Updated all scoped limiters (`bound.go`, `range.go`, `gate.go`, `queue.go`, `rate.go`, `resource.go`, `time.go`) to wrap and return `ErrMissingTenant` when the tenant is missing, ensuring consistent error reporting. 

**Testing and validation:**

* Added `missing_tenant_test.go` to assert that all scoped limiters return `ErrMissingTenant` when the tenant is missing, and to verify that this error is not misclassified as a `LimitError`.

### DIFF
--- a/pkg/settings/limits/errors.go
+++ b/pkg/settings/limits/errors.go
@@ -11,6 +11,14 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/settings"
 )
 
+// ErrMissingTenant is returned when a scoped limiter is used without the tenant that its
+// scope requires in ctx. Unlike a settings read failure, where the limiter still resolves a
+// usable value alongside the error, no lookup is attempted here, so callers must not treat
+// the returned value as a limit. It signals that the CRE context was never populated, which
+// is a programming error rather than a degraded system. Deliberately not a [LimitError]:
+// nothing was limited.
+var ErrMissingTenant = errors.New("missing tenant")
+
 // LimitError is implemented by errors returned when a limit is exceeded.
 // Use [errors.As] to identify limit errors, for example:
 //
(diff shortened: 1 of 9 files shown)


pkg/settings/limits/errors.go
```diff
@@ -11,6 +11,14 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/settings"
 )
 
+// ErrMissingTenant is returned when a scoped limiter is used without the tenant that its
+// scope requires in ctx. Unlike a settings read failure, where the limiter still resolves a
+// usable value alongside the error, no lookup is attempted here, so callers must not treat
+// the returned value as a limit. It signals that the CRE context was never populated, which
+// is a programming error rather than a degraded system. Deliberately not a [LimitError]:
+// nothing was limited.
+var ErrMissingTenant = errors.New("missing tenant")
```
jmank88, line 20: What do you think about using a custom error type, so that we can include the scope that was missing as a field, rather than just in the formatted error?

jmank88 (approved):

patrickhuie19 (approved):

PR #2381, opened 2026-09-11
Update configdoc to support more shapes of TOML

nolag:


pkg/config/configdoc/defaults.go
```diff
@@ -28,23 +28,32 @@ func DefaultsOnly(r io.Reader, cfg any, decode func(io.Reader, any) error) error
...
-		if skipUntil != nil {
-			if skipUntil(t) {
-				skipUntil = nil
+		// Indentation is ignored in TOML, and nested tables and their keys are conventionally indented.
+		trimmed := strings.TrimSpace(t)
+
+		if skipTable != "" {
+			if !strings.HasPrefix(trimmed, "[") {
+				// TOML does not end a table at a blank line, only at the next table header,
+				// so keys - and blank lines - still belong to the array of tables.
```
jmank88, line 40: Hmm this swallows comments? But I guess that is OK for this particular func 🤷

pkg/config/configdoc/configdoc.go
```diff
@@ -13,13 +13,16 @@ const (
...
 	TokenAdvanced = "**ADVANCED**"
+	TokenMap      = "**MAP**"
 )
 
 // Generate returns MarkDown documentation generated from the TOML string.
-// - Each field but include a trailing comment of either FieldDefault or FieldExample.
-// - If a description begins with TokenAdvanced, then a warning will be included.
-// - The markdown wil begin with the header, followed by the example
-// - Extended descriptions can be applied to top level tables
+//   - Each field but include a trailing comment of either FieldDefault or FieldExample.
```
jmank88: must?
```suggestion
//   - Each field must include a trailing comment of either FieldDefault or FieldExample.
```

pkg/config/configdoc/defaults.go
nolag: Claude says it's ok here.

> Correct observation, no impact. writeDefaults already drops every comment line unconditionally (strings.HasPrefix(trimmed, "#") → continue); its output is TOML fed straight to DecodeTOML, not documentation. Skipping comments inside an array-of-tables loses nothing that would have been emitted anyway.

jmank88 (approved):

DylanTinianov (approved):

PR #2382, opened 2026-09-11
[CRE] Remove legacy DAG paths from workflows/wasm/host

bolekk:
Strips the v1 legacy DAG machinery from the wasm host ahead of the DAG WASM removal (CRE-5836): ModuleV1/Run, linkLegacyDAG, IsLegacyDAG, GetWorkflowSpec, the legacy fetch/emit/log env imports and converters, the determinism random_get shim, the legacy poll_oneoff stub and fake clock, and the Fetch/MaxFetchRequests/Labeler/Determinism module config fields. IsLegacyDAG is also dropped from host.ModuleBase.

The sleep/oom test drivers are ported to v2 guests (rawsdk import) and driven via Execute, the two brotli/size tests are kept, and the 13 legacy guest test modules plus the dead test/requirements fixtures, mocks/module_v1 and its .mockery.yaml entry are deleted.

See full plan in  https://docs.google.com/document/d/1pRJ8zXt7Z_skqksG7-yRoa6v7pMEbZ7jmhu5DybaI6s/edit?usp=sharing

Deployment Validation: Existing Canaries.


pkg/workflows/host/module.go
```diff
@@ -14,7 +14,6 @@ import (
 type ModuleBase interface {
 	Start()
 	Close()
-	IsLegacyDAG() bool
```
nolag, line 17: Should we just return false and make it a non-breaking change?

pkg/workflows/wasm/host/module.go
```diff
@@ -641,32 +532,6 @@ func (m *module) Execute(ctx context.Context, req *sdkpb.ExecuteRequest, executo
 	return runWasm(ctx, m, req, setMaxResponseSize, m.linkV2, executor, timeout)
 }
 
-// Run is deprecated, use execute instead
-func (m *module) Run(ctx context.Context, request *wasmdagpb.Request) (*wasmdagpb.Response, error) {
```
nolag, line 645: If we want "non-breaking" for the interface, we can return an error saying we no longer run these.

nolag (dismissed):
Up to you to address my comments or not. I'm ok with the breaking change given we want to burn this bridge, but also get that we can leave a couple few and not worry about compliation.

pavel-raykov (dismissed):

pkg/workflows/host/module.go
bolekk: I'll keep it removed and clean up uses in core too.

cedric-cordenier (approved):

justinkaseman (approved):

PR #2387, opened 2026-09-15
Recover from panics in async capability calls

vreff:
Capability calls from the wasm host module spin up a goroutine to handle the call, then immediately return. Those goroutines seem to be unhandled by wasmtime if they panic.


cedric-cordenier:
@vreff Wasmtime will already be catching this panic for you, and then bubbling it back up to the call to module.Execute

This gets caught here: https://github.com/smartcontractkit/chainlink-common/blob/1d3a14a9b049680eb84225cb1c6954524a61ccde/pkg/workflows/wasm/host/module.go#L675 and should stop the panic

If this isn't working for you, I'd love to understand more

vreff:
> @vreff Wasmtime will already be catching this panic for you, and then bubbling it back up to the call to module.Execute
> 
> This gets caught here:
> 
> https://github.com/smartcontractkit/chainlink-common/blob/1d3a14a9b049680eb84225cb1c6954524a61ccde/pkg/workflows/wasm/host/module.go#L675
> and should stop the panic
> 
> If this isn't working for you, I'd love to understand more

The issue seems to be that the capability call spins up a new goroutine and then returns nil immediately. What happens if that instance shuts down for some other reason, like a timeout, and then the goroutine it started panics?

Problematic code:
```
	go func() {
		defer free()

		resp, err := e.executor.CallCapability(ctx, req)

		if err != nil {
			errString := err.Error()

			if caperror, ok := errors.AsType[caperrors.Error](err); ok {
				errString = caperror.SerializeToString()
			}
			resp = &sdkpb.CapabilityResponse{
				Response: &sdkpb.CapabilityResponse_Error{
					Error: errString,
				},
			}
		}

		select {
		case <-ctx.Done():
		case ch <- resp:
		}
	}()

	return nil
```

This just seems like a new goroutine, which is not guaranteed to be tracked by that wasm instance.

vreff:
@cedric-cordenier I've reworked this PR as simply a test to show the behavior I'm concerned about. LMK if some aspect of the test is missing context.

